PR #18, opened 2023-01-12
ci: add tests and lint to ci, fix lint errors

tychoish:
can haz CI checks


sharnoff (approved):
marking approved so it's green

Issue #27, opened 2023-01-24
Follow-up tasks from #25


sharnoff:
Items from review:

- [x] ~~Sentry integration - both in general and for `spawnBackgroundWorker` ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084733617))~~
  - No longer necessary; alerting on metrics is sufficient.
- [x] Replace unusual HTTP constants (e.g. 409 in `pkg/informant`) with their constants in `net/http` ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084734000))
  - Will be mostly handled once the informant is removed/deprecated
- [x] `pkg/agent` should be split into multiple packages ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084735705))
  - Implemented by https://github.com/neondatabase/autoscaling/pull/289, https://github.com/neondatabase/autoscaling/pull/309
- [x] agent<->informant protocol checks should be implemented in `pkg/api` ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084735906))
  - implemented by #30
- [x] Add tests to `pkg/agent` (particularly around `Runner`) ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084737232))
  - Replaced by #199
- [x] #45
- [x] `(*Runner).updateVMResources` loop body should be extracted into separate function ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084738213))
  - Obviated by https://github.com/neondatabase/autoscaling/pull/113
- [x] `(*Runner).doMetricsRequestIfEnabled` code between Lock() and Unlock() should be a separate method ([link](https://github.com/neondatabase/autoscaling/pull/25#discussion_r1084738683))
  - Replaced by #472

Item from the PR, before reviewing:

- [x] https://github.com/neondatabase/autoscaling/pull/51 ([comment](https://github.com/neondatabase/autoscaling/blob/bc1f146d9291e8c1ccf97855af445f554cbba63e/pkg/agent/runner.go#L605-L607), [comment](https://github.com/neondatabase/autoscaling/blob/bc1f146d9291e8c1ccf97855af445f554cbba63e/pkg/agent/runner.go#L1681))
[...]

Issue #23, opened 2023-01-20
agent vs NeonVM state is inconsistent when NeonVM fails


sharnoff:
During the all-hands demo 2023-01-20, the state of the VM as reported by NeonVM didn't match what the autoscaler-agent (and scheduler) thought it was. To be specific:

* Under high load, the autoscaler-agent decided to scale the VM up from 16 CPU + 64 GB to its maximum of 19 CPU + 76 GB, which was approved by the scheduler.
* Sometime after that, I looked at the state from NeonVM, and it didn't match — NeonVM reported 18 CPU + 66 GB, with the autoscaler-agent remaining constant.
* The autoscaler-agent's runner for the VM did not emit any logs relating to the request to NeonVM (so: presumably it did not error, unless the request handling silently ignored a failure).

Didn't have a chance to check from inside the VM (so: I don't know which is correct). Haven't checked whether this is easy to replicate.

---

Extra details:
* VM name: `compute-node-ep-throbbing-wave`
* Time of attempted scaling 16 -> 19 CU: `2023-01-20 17:09:32`

A helpful command to filter downloaded logs for the autoscaler-agent's runner for the VM:

```sh
grep '^{' log.txt \
    | jq -r '.log' \
    | grep -E '^(I|W|E|F).*Runner default:compute-node-ep-throbbing-wave-612027-7zz9t' \
    | sort > log
```

sharnoff:
Looks like NeonVM got stuck trying & failing to reconcile the state. Found some errors, all along the lines of `"a used vhost backend has no free memory slots left"`. Here's the logs from the attempted increase from 16 CPU + 64 GB to 19 CPU + 76 GB:

<details><summary>Expand for: <code>neonvm-controller</code> logs from <code>17:09:32</code> to <code>17:10:30</code></summary>
<pre>
1.6742345720475886e+09	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/mutate-vm-neon-tech-v1-virtualmachine", "UID": "c97a6ec3-4cd8-442b-9ac4-a4a533b44561", "kind": "vm.neon.tech/v1, Kind=VirtualMachine", "resource": {"group":"vm.neon.tech","version":"v1","resource":"virtualmachines"}}
1.6742345720478792e+09	INFO	virtualmachine-resource	default	{"name": "compute-node-ep-throbbing-wave-612027"}
[...]

Issue #12, opened 2023-01-11
Clean shutdown on exit


sharnoff:
See https://github.com/neondatabase/autoscaling/pull/8#discussion_r1065988708.

There's a few pieces to this (beyond what's in #8). Probably best to split them into separate PRs.

- [ ] Cleanup on VM informant exit
  - note: Currently (2023-01-10), NeonVM doesn't support clean shutdown, so this is not _immediately_ applicable
- [ ] Cleanup on autoscaler-agent exit (e.g., using pod lifecycle hooks + cancelling top-level context)
- [ ] Cleanup on scheduler exit
  - Some investigation is required here. The scheduler _framework_ knows when it's being replaced, even without any pod lifecycle hooks. However, the pod stays alive until the eviction timeout is reached.

sharnoff:
This was marked as done on 2023-07-12, but not actually closed.

PR #30, opened 2023-01-30
Add "memory high" upscaling via cgroups

sharnoff:
High-level features added:

* agent: `/try-upscale` endpoint, taking [`api.MoreResources`](https://github.com/neondatabase/autoscaling/blob/246367af592737e526ba43fe5b85191cc62b10a0/pkg/api/types.go#L267-L272) — VM informant can request upscaling. Refer to the changes to ARCHITECTURE.md for a brief overview.
* informant: off-by-default cgroup memory.high event tracking — triggers `/try-upscale`.

Prior issues fixed:

* VM informant will deny downscaling if it's too close to triggering a memory.high event (fixes the "downscaling below current usage" issue).

Remaining tasks:

- [x] Add a little "allocate a bunch of memory" test program for the test VM
- [x] Actually test that it works (and fix the bugs) — with and without cgroup handling enabled.

Follow-up tasks:

- [ ] Integrate the cgroups stuff into `compute_ctl`


pkg/util/arith.go
```diff
@@ -33,3 +33,31 @@ func Min[T constraints.Ordered](x, y T) T {
...
+	Swap(new I) (old I) //nolint:predeclared // same var names as methods
+}
+
+// AtomicMax atomically sets a to the maximum of *a and i, returning the old value at a.
+//
+// Eventually, one would *hope* that go could add atomic maximum (and minimum), so we don't need
+// this function (there are some ISAs that support such instructions). In the meantime though, this
+// tends to be pretty useful - event though it's not wait-free.
+func AtomicMax[A AtomicInt[I], I constraints.Integer](a A, i I) I {
+	for {
```
tychoish, line 54: I _think_ that this is how `atomic.Value.CompareAndSwap()` is implemented, that not withstanding, I have this feeling that if you get into this situation you probably want a different sort of operation (e.g. just a mutex protecting an int, then you can let the runtime be smart (potentially) about letting the goroutine calling this function sleep until it can be woken up logically.)

vm_image/init
```diff
@@ -5,6 +5,10 @@
 
 set -x
 
+# Setup the 'neon-test' cgroup:
+whoami
+mkdir /sys/fs/cgroup/neon-test
```
tychoish, line 10: does this leak into a production environment? does the whoami need to stay?

pkg/informant/endpoints.go
```diff
@@ -46,13 +113,64 @@ func (s *State) RegisterAgent(info *api.AgentDesc) (*api.InformantDesc, int, err
...
+		}
+
+		// Downscale:
+		//
+		// TODO: see similar note above. We shouldn't call methods on s.cgroup.mgr from here.
+		if err := s.cgroup.mgr.SetHighMem(newMemHigh); err != nil {
+			return false, "", fmt.Errorf("Error setting cgroup memory.high: %w", err)
+		}
+
+		status := fmt.Sprintf("Set cgroup memory.high down to %v MiB", float64(newMemHigh)/mib)
```
tychoish, line 157: %f for floats?

pkg/informant/endpoints.go
```diff
@@ -3,22 +3,89 @@ package informant
...
+			//  1. Respond "ok" to a downscale request
+			//  2. Restart
+			//  3. Read system memory
+			//  4. Get downscaled (as approved earlier)
+			// A potential way to fix this would be writing to a file to record approved downscale
+			// operations.
+			if err := s.cgroup.setMemoryHigh(); err != nil {
+				return fmt.Errorf("Error setting initial cgroup memory.high: %w", err)
+			}
+			go s.cgroup.handleCgroupSignalsLoop(config)
```
tychoish, line 86: one of my "rules" is "if you start a go routine without passing it a context, you should be really sure this is right, and then probably not do it," there are exceptions, but I've always regretted making them, for anything non-trivial.

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+	memEvents, eventErrCh := manager.EventChan()
+	sendEvent, recvEvent := util.NewCondChannelPair()
+
+	highEventCount := &atomic.Uint64{}
+	errCh := make(chan error, 1)
+
+	// Long-running handler task for memory events
+	go func() {
+		for {
+			select {
```
tychoish, line 61: how does this daemon thread handle shutdown

pkg/informant/cgroup.go
```diff
@@ -0,0 +1,228 @@
...
+// handleMemoryHighEvent performs the "freeze cgroup, request upscale, thaw cgroup" operation, in
+// response to a "memory high" event for the cgroup
+//
+// This method waits on s.agents.UpscaleEvents(), so incorrect behavior will occur if it's called at
+// the same time as anything else that waits on the upscale events. For that reason, both this
+// function and s.setMemoryHigh() are dispatched from within s.handleCgroupSignalsLoop().
+//
+// This method MUST be called while holding s.updateLock.
+func (s *CgroupState) handleMemoryHighEvent(config CgroupConfig) error {
+	klog.Infof("Received memory high event. Freezing cgroup")
```
tychoish, line 160: is it meaingful to tell us anything about that event?

pkg/informant/cgroup.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Immediately freeze the cgroup before doing anything else.
+	if err := s.mgr.Freeze(); err != nil {
+		return fmt.Errorf("Error freezing cgroup: %w", err)
+	}
+
+	startTime := time.Now()
+
+	// Start a timer for the maximum time we'll leave the cgroup frozen for:
+	maxWaitBeforeThaw := time.Millisecond * time.Duration(config.MaxUpscaleWaitMillis)
```
tychoish, line 170: I always prefer "num * <unit>" also because duration is just an int64 if your config variable is in int64 you *might* not need to convert. 

also I'd probably use `MS` over `Millis`

pkg/informant/cgroup.go
```diff
@@ -0,0 +1,228 @@
...
+			select {
+			case <-s.upscaleEventsRecvr.Recv():
+				klog.Warningf("Received memory high event, but also upscale event. Handling upscale and ignoring memory high")
+				func() {
+					s.updateLock.Lock()
+					defer s.updateLock.Unlock()
+
+					if err := s.setMemoryHigh(); err != nil {
+						panic(fmt.Errorf("Error handling upscale signal: %w", err))
+					}
```
tychoish, line 134: maybe make these methods? 

all of the panics worry me? I think if you expect that all of these errors are unrecoverable by their nature, I'd just panic in the functions that produce them rather than go through the effort of returning errors that are just going to convert them to panics? 

(unless these errors represent invariant violations, panic'ing this much seems to be indicative of some sort of other design flaw)

pkg/informant/agent.go
```diff
@@ -702,3 +764,45 @@ func (a *Agent) Resume(timeout time.Duration) error {
...
+// SpawnRequestUpscale requests that the Agent increase the resource allocation to this VM
+func (a *Agent) SpawnRequestUpscale(timeout time.Duration, handleError func(error)) {
+	// Quick unregistered check
+	select {
+	case <-a.unregistered.Recv():
+		klog.Warningf(
+			"RequestUpscale called for Agent %s/%s that is already unregistered (probably *not* a race?)",
+			a.serverAddr, a.id,
+		)
+		handleError(context.Canceled)
```
tychoish, line 777: this seems like a weird sentinel error to pass around, but sure.

pkg/informant/agent.go
```diff
@@ -321,6 +344,45 @@ func (s *AgentSet) RegisterNewAgent(info *api.AgentDesc) (uint32, int, error) {
...
+	defer s.lock.Unlock()
+
+	s.wantsMemoryUpscale = true
+
+	// explicitly copy s.current so that our error handling points to the right Agent
+	if agent := s.current; agent != nil {
+		agent.SpawnRequestUpscale(AgentUpscaleTimeout, func(err error) {
+			if errors.Is(err, context.Canceled) {
+				return
+			}
```
tychoish, line 366: what about deadline exceeded?

pkg/api/types.go
```diff
@@ -262,15 +270,52 @@ type UnregisterAgent struct {
...
+// MoreResources holds the data associated with a MoreResourcesRequest
 type MoreResources struct {
 	// Cpu is true if the VM informant is requesting more CPU
 	Cpu bool `json:"cpu"`
 	// Memory is true if the VM informant is requesting more memory
 	Memory bool `json:"memory"`
 }
 
+// Not returns the field-wise logical "not" of m
+func (m MoreResources) Not() MoreResources {
```
tychoish, line 291: none of these methods are actually called, I think?

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func NewCgroupManager(groupName string) (*CgroupManager, error) {
+	mode := cgroups.Mode()
+	if mode != cgroups.Unified && mode != cgroups.Legacy {
+		var modeString string
+		switch mode {
+		case cgroups.Unavailable:
+			modeString = "Unavailable"
+		case cgroups.Legacy:
```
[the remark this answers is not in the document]
sharnoff: Ah yep, good catch. The condition above should be `... && mode != cgroups.Hybrid`, not `cgroups.Legacy`.

pkg/informant/cgroupmanager.go
sharnoff: Refer to the comment 10 lines up:

https://github.com/neondatabase/autoscaling/blob/23e7b4c5ff02496946521d7540695e8d09b562c1/pkg/informant/cgroupmanager.go#L49-L51

And

#12:

> - [ ] Cleanup on VM informant exit
>     - note: Currently (2023-01-10), NeonVM doesn't support clean shutdown, so this is not immediately applicable

tl;dr: Currently nothing else _can_ shut down, so there isn't anything it _can_ do.

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Long-running handler task for memory events
+	go func() {
+		for {
+			select {
+			case event := <-memEvents:
+				highCount := event.High
+				oldHighCount := util.AtomicMax(highEventCount, highCount)
+				if highCount > oldHighCount {
+					sendEvent.Send()
```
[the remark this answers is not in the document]
sharnoff: `util.AtomicMax` isn't wait-free, but there's only two threads that can access `highEventCount` at that point.

`(*util.CondChannelSender).Send()` is wait-free.

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+				}
+			case err, ok := <-eventErrCh:
+				if err == nil && !ok {
+					errCh <- errors.New("Memory event channel closed without error")
+				} else {
+					errCh <- fmt.Errorf("Error while waiting for memory events: %w", err)
+				}
+				return
```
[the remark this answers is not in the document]
sharnoff: There aren't any other threads. `CgroupManager` stores it as a `<-chan error`.

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+	}, nil
+}
+
+// TODO: no way to do this with github.com/containerd/cgroups ? Seems like that should be
+// exposed to the user... We *can* just parse it directly, but it's a bit annoying.
+func parseMemoryEvents(groupName string) (*cgroup2.Event, error) {
+	path := cgroupPath(groupName, "memory.events")
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading file at %q: %w", path, err)
```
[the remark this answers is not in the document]
sharnoff: Can you elaborate?

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	event := cgroup2.Event{}
+	valueMap := map[string]struct {
+		v   *uint64
```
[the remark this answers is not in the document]
sharnoff: Not sure what you're saying

pkg/informant/endpoints.go
sharnoff: %v because it doesn't include decimals if it doesn't need to, which is _almost_ guaranteed to be the case.

For floats, %v is equivalent to %g - might make sense to use that instead tbh.

pkg/util/arith.go
```diff
@@ -33,3 +33,31 @@ func Min[T constraints.Ordered](x, y T) T {
 		return y
 	}
 }
+
+// AtomicInt represents the shared interface provided by various atomic.<NAME> integers
+//
+// This interface type is primarily used by AtomicMax.
+type AtomicInt[I any] interface {
```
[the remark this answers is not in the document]
sharnoff: What do you mean?

pkg/util/arith.go
sharnoff: > I think that this is how atomic.Value.CompareAndSwap() is implemented

_Kind of_, but not quite imo. It's [mostly pretty readable](https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/sync/atomic/value.go;l=135), though!

vm_image/init
sharnoff: good catch - the `whoami` is just for debugging. I'll make a note to remove it once I'm done testing stuff.

cmd/vm-informant/main.go
```diff
@@ -15,8 +16,32 @@ func main() {
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
+	nullString := "\x00"
+	var cgroupName string
+	flag.StringVar(&cgroupName, "cgroup", nullString, "Sets the cgroup to monitor (optional)")
```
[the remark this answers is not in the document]
tychoish: but _should_ it be a cgroup? would we want people to do this? seems like we can be slightly more restrictive than the kernel, to reduce confusion in general.

pkg/agent/runner.go
```diff
@@ -1476,6 +1500,28 @@ func (s *atomicUpdateState) computeUnitsBounds() (uint16, uint16) {
...
+// This method does not respect any bounds on Compute Units placed by the VM's maximum or minimum
+// resource allocation.
+func (s *atomicUpdateState) requiredCUForRequestedUpscaling() uint16 {
+	var required uint16
+
+	// note: floor(x / M) + 1 gives the minimum integer value greater than x / M.
+
+	if s.requestedUpscale.Cpu {
+		required = util.Max(required, s.vm.Cpu.Use/s.computeUnit.VCPU+1)
+	}
```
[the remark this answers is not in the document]
tychoish: maybe? I definitely had to read it more than times, and cross reference to the definition of `util.Max` which spins over an atomic. I'd feel a bit better about this if the util functions had unittests. 

I think I had to read it more than once because doing a max of two numbers when one is obviously a zero makes me think something fishy is up, or looks like it's unintentional.

pkg/informant/agent.go
```diff
@@ -321,6 +344,45 @@ func (s *AgentSet) RegisterNewAgent(info *api.AgentDesc) (uint32, int, error) {
...
+	// FIXME: we should assign a timeout to these upscale requests, so that we don't continue trying
+	// to upscale after the demand has gone away.
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.wantsMemoryUpscale = true
+
+	// explicitly copy s.current so that our error handling points to the right Agent
+	if agent := s.current; agent != nil {
```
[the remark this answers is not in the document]
tychoish: I've seen a convention in struct delcaration that says "if you're not going to mutex every field in the struct, put the mutex and the covered fields near the end in a group so it's clear what they protect." which I certainly haven't always done myself, but it definitely makes that more clear. 

One of the problems is that you only get immutability in go structs by copying the data (which all this being pointers doesn't do) or by pinky swearing that no code is going to modify a value.  

While I'm convinced that this is correct, this code *looks* like a bug, or is "valid form using the idiom of a common bug", so even though it's correct, the look and comment sort of makes it seem more suspect?

pkg/informant/cgroupmanager.go
```diff
@@ -0,0 +1,222 @@
...
+
+	cgroups "github.com/containerd/cgroups/v3"
+	"github.com/containerd/cgroups/v3/cgroup2"
+
+	"github.com/neondatabase/autoscaling/pkg/util"
+)
+
+type CgroupManager struct {
+	MemoryHighEvent util.CondChannelReceiver
+	ErrCh           <-chan error
```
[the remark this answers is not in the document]
tychoish: when I've used error channels in the past, I tend to make them buffered to the same extent as the number of threads/possible senders, or put the consume of the thread *really* close to the definition.

pkg/informant/cgroupmanager.go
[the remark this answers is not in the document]
tychoish: oh, just if you're writing code, you have an object of type `error` you know that it represents a conceptual error, the output is almost always `ERROR: <message>` so putting the word "error" in the text of the error, is a lot.

pkg/informant/cgroupmanager.go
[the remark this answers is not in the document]
tychoish: got it... 

I think this is, then, safely in the "this reads a bit suspect if you don't have the full context for how this works and a good mental model for the go routines in the system" category.

pkg/informant/cgroupmanager.go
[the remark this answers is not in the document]
tychoish: Let me attempt a different way: 

- why isn't the structure of the value type of this map named?
- why do you have a boolean field in the value of a map, for which accessing values already has an "ok" boolean value to let you now if its set?
- what is the meaning of:
  - `valueMap["foo"] = struct{v: nil, set: true}`

pkg/util/arith.go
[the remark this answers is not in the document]
tychoish: I believe that this is valid: `name := AtomicInt[string]`, because `I` can be any type, not just an integer, which is confusing.

pkg/informant/agent.go
tychoish: just declare a sentinel error in this named of `ErrRequestConflictTimeout` or something similar.

pkg/agent/runner.go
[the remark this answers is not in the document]
sharnoff: > `util.Max` which spins over an atomic

I think you're confusing it with `util.AtomicMax`? The definition of `util.Max` is:

https://github.com/neondatabase/autoscaling/blob/bfbf9873c8ea8400536b1b54999c0b421ac48687/pkg/util/arith.go#L19-L26

> doing a max of two numbers when one is obviously a zero makes me think something fishy is up

Fair enough. I think it's personal taste — for me, reading it without a maximum on the first means I have to double-check that the high-level logic on the first & second are still fundamentally the same, even though they're phrased differently.

sharnoff:
Pushed some updates. To get it working, opened https://github.com/neondatabase/neonvm/pull/33 — blocked on that before merging.

sharnoff:
Spent a while dealing with issues from an over-eager OOM killer after memory hotplug failure, but it seems like fixing some related VM informant bugs stopped that from happening - even though they couldn't possibly be the *cause*.

pkg/api/types.go
sharnoff: Removed `Or`.

pkg/informant/cgroup.go
sharnoff: Opened #35.

pkg/util/arith.go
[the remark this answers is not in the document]
sharnoff: `AtomicInt` is an interface, so you can't do that directly. You could 100% declare a type like that, but probably will run into trouble constructing it (because realistically nothing will have the appropriate methods returning `string`s)

vm_image/init
sharnoff: Done

pkg/informant/endpoints.go
sharnoff: Changed to %g.

pkg/informant/cgroupmanager.go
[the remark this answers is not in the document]
sharnoff: One of the recent commits expanded on a comment here, to provide some more context.

pkg/util/arith.go
```diff
@@ -33,3 +33,31 @@ func Min[T constraints.Ordered](x, y T) T {
...
+	Add(delta I) (new I)                      //nolint:predeclared // same var names as methods
+	CompareAndSwap(old, new I) (swapped bool) //nolint:predeclared // same var names as methods
+	Load() I
+	Store(val I)
+	Swap(new I) (old I) //nolint:predeclared // same var names as methods
+}
+
+// AtomicMax atomically sets a to the maximum of *a and i, returning the old value at a.
+//
+// Eventually, one would *hope* that go could add atomic maximum (and minimum), so we don't need
```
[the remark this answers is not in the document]
sharnoff: Fixed

sharnoff:
Ok, think this is good to merge. I'll give it a once-over before merging tomorrow, with a follow-up PR to https://github.com/neondatabase/neon to add cgroup handling there as well.

This feature has some unfortunate interactions with other open issues, particularly:

1. Our current metrics and scaling algorithm mean that it's hard to scale back down after increasing due to `/try-upscale` requests (if it can't go all the way back down to the level indicated by load average, it won't go down at all)
2. Some of the raciness around NeonVM means we have to "trust" what the autoscaler-agent says when it calls the `/upscale` endpoint (see: https://github.com/neondatabase/autoscaling/issues/23#issuecomment-1398805853, point (2))

I believe both of these require protocol changes, alongside some of the other changes from #27.

Issue #15, opened 2023-01-12
Use plugin's HTTP server for pod readiness check


sharnoff:
Per discussion with @tychoish; what the title says. Without this, it may be possible for a scheduler pod to be marked as ready before the plugin is sufficiently set up.

sharnoff:
I think we're _probably_ fine without this — the scheduler already exposes HTTP endpoints for readiness, and I'd imagine (though, haven't checked this) that those only respond that the scheduler is ready once the plugin is _also_ ready.

Even still, it's theoretically possible for the scheduler to be marked as ready before the plugin's HTTP server handling autoscaler-agent requests is bound & serving (if only because it's concurrent).

These are all mostly fine though because the autoscaler-agent will just retry if it can't connect to the scheduler yet.

PR #32, opened 2023-01-31
plugin: handle process start/shutdown

tychoish:


sharnoff:
Having only looked briefly, catching sigterm feels like overkill IMO - we should just use [container lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/). Kind of adjacent to #15.

tychoish:
I think container lifecycle hooks would require most of these changes (and maybe a bit more, to avoid having having services start unchecked?

Which is to say we need to have things listening for some kind of event to shut down, might as well be a signal, but the plumbing is the same regardless of the mechanism.

cmd/autoscale-scheduler/main.go
```diff
@@ -12,9 +15,30 @@ import (
...
+	// pauses for a *very* short period of time before exiting.
+	//
+	// eventually, the constructed application
+	sig := make(chan struct{})
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM)
+	go func() { defer close(sig); <-ctx.Done(); time.Sleep(500 * time.Millisecond) }()
+	defer func() { <-sig }()
+	defer cancel()
+
+	command := app.NewSchedulerCommand(app.WithPlugin(plugin.Name, plugin.NewAutoscaleEnforcerPlugin(ctx)))
```
tychoish, line 39: I actually think that by capturing the context here which wraps the other parts of the scheduler's shut down, we get our own orderly shutdown regardless of kube's 

(modulo actually waiting for things to shut down, I believe that we will have correctly passed a context through to all of the parts that could use one (the ones that don't get this one, will get one from the scheduler which has a similar trigger.) 

Also, unfortunately, kube doesn't use the context we pass to cobra, which is... unfortunate, but we tried.)

pkg/plugin/run.go
```diff
@@ -71,7 +72,39 @@ func (e *AutoscaleEnforcer) runPermitHandler() {
...
+	// in parallel) and wait for them to shut down.
+	go func() {
+		// this runs until something cancels it
+		if err := server.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
+			// this fatal will take down the entire process,
+			klog.Fatalf("[autoscale-enforcer] Resource request server failed: %s", err)
+		}
+	}()
+
+	go func() {
```
sharnoff, line 90: `runPermitHandler` is already launched in a new goroutine. I's say it's _probably_ (?) better to leave the contents of this function inside its caller (i.e. don't launch a new goroutine here).

cmd/autoscale-scheduler/main.go
sharnoff: That _might_ be the case. It's hard to say without looking into why the current shutdowns are stalling (e.g., maybe the scheduler waits for all plugin goroutines to exit before returning, so just waiting for `command.ExecuteContext` to finish might block).

sharnoff (approved):
Post-discussion verdict:

* K8s already uses SIGTERM to signal termination, so it's _reasonably_ clean to handle shutdown this way
* There's some weird preexisting interactions here because the K8s scheduler process will _already_ trap SIGTERM.

(note from later, about the second bullet-point) -- it looks like we're actually ok here. From the docs on os/signal.Notify:

> It is allowed to call Notify multiple times with different channels and the same signals: each channel receives copies of incoming signals independently.

I originally thought we might run into issues if the plugin shuts down first, but I think it's actually ok — none of the *scheduling* relies on the plugin's background tasks.

---

Approved, pending resolution of my comments (or: if you decide it's fine as-is, go for it)

cmd/autoscale-scheduler/main.go
tychoish: yeah, I definitely want to debug this. more, but I think resolving the long shutdowns should be debugged on its own.

pkg/plugin/run.go
```diff
@@ -71,7 +72,39 @@ func (e *AutoscaleEnforcer) runPermitHandler() {
...
+
+	// in general this isn't the right way to do this: we should
+	// have a group of functions which return service objects that
+	// have blocking close routines, and then start them at a
+	// higher level in the process, and then when we catch a
+	// shutdown signal, you call the shutdown methods (preferably
+	// in parallel) and wait for them to shut down.
+	go func() {
+		// this runs until something cancels it
+		if err := server.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
```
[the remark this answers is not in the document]
tychoish: yes this is just an error on my part. sorry, and thanks!

pkg/plugin/plugin.go
```diff
@@ -105,13 +113,14 @@ func NewAutoscaleEnforcerPlugin(obj runtime.Object, h framework.Handle) (framewo
...
-	go p.runPermitHandler()
+	go p.runPermitHandler(ctx)
 
 	// Periodically check that we're not deadlocked
 	go func() {
 		checkDelay := 5 * time.Second
 		timeout := 1 * time.Second
-
+		timer := time.NewTimer(checkDelay)
+		defer timer.Stop()
```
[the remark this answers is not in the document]
tychoish: yeah, I have a perhaps too aggressive anti-`time.After` prejudice.

pkg/plugin/run.go
tychoish: I think it doesn't really matter, but I turned it around.

PR #38, opened 2023-02-07
Require label for autoscaling to be enabled

sharnoff:
Adds the `autoscaling.neon.tech/enabled` label and requires that it's present for the autoscaler-agent to run.

### DIFF
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -6,20 +6,32 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	vmapi "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 )
 
-const LabelTestingOnlyAlwaysMigrate = "autoscaler/testing-only-always-migrate"
+const (
+	LabelTestingOnlyAlwaysMigrate = "autoscaler/testing-only-always-migrate"
+	LabelEnableAutoscaling        = "autoscaling.neon.tech/enabled"
+)
+
+// HasAutoscalingEnabled returns true iff the object has the label that enables autoscaling
+func HasAutoscalingEnabled[T metav1.ObjectMetaAccessor](obj T) bool {
+	labels := obj.GetObjectMeta().GetLabels()
+	_, enabled := labels[LabelEnableAutoscaling]
+	return enabled
+}
 
 // VmInfo is the subset of vmapi.VirtualMachineSpec that the scheduler plugin and autoscaler agent
 // care about
 type VmInfo struct {
-	Name          string    `json:"name"`
-	Namespace     string    `json:"namespace"`
-	Cpu           VmCpuInfo `json:"cpu"`
-	Mem           VmMemInfo `json:"mem"`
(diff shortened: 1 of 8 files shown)


sharnoff:
Current plan: wait until control plane is next released, then merge this, release & deploy. Because it's a breaking change, probably bump to v0.2

pkg/api/vminfo.go
```diff
@@ -6,20 +6,32 @@ import (
...
 
-const LabelTestingOnlyAlwaysMigrate = "autoscaler/testing-only-always-migrate"
+const (
+	LabelTestingOnlyAlwaysMigrate = "autoscaler/testing-only-always-migrate"
+	LabelEnableAutoscaling        = "autoscaling.neon.tech/enabled"
+)
+
+// HasAutoscalingEnabled returns true iff the object has the label that enables autoscaling
+func HasAutoscalingEnabled[T metav1.ObjectMetaAccessor](obj T) bool {
+	labels := obj.GetObjectMeta().GetLabels()
```
tychoish, line 21: is this a request or just producing a copy of the object. If it's a request, I'd be inclined to combine it with the pod.Ready op. 

also I think this doesn't need to be a generic and can just accept the interface as the type of the argument? I believe that go's calling convention allows conversions.

pkg/plugin/trans.go
```diff
@@ -214,6 +214,33 @@ func (r resourceTransition[T]) handleDeleted(currentlyMigrating bool) (verdict s
 	return verdict
 }
 
+// handleAutoscalingDisabled updates r.node with changes to clear any buffer and capacityPressure
+// from r.pod
+//
+// A pretty-formatted summary of the changes is returned as the verdict, for logging.
+func (r resourceTransition[T]) handleAutoscalingDisabled() (verdict string) {
```
tychoish, line 221: I don't think there's anything gained by declaring `verdict` here

pkg/plugin/trans.go
```diff
@@ -214,6 +214,33 @@ func (r resourceTransition[T]) handleDeleted(currentlyMigrating bool) (verdict s
...
+	for _, v := range valuesToReduce {
+		*v -= buffer
+	}
+
+	r.node.capacityPressure -= r.pod.capacityPressure
+	r.pod.capacityPressure = 0
+
+	fmtString := "pod had buffer %d, capacityPressure %d; " +
+		"node reserved %d -> %d, capacityPressure %d -> %d"
+	verdict = fmt.Sprintf(
```
tychoish, line 234: I'd avoid the concat in the fmtstring at least, and probably also avoid not just returning the result of the `fmt.Sprintf` directly.

pkg/api/vminfo.go
sharnoff: Just produces a copy of the object, should be ok :)

Removed the generic.

tychoish (approved):

pkg/plugin/trans.go
sharnoff: Leaving as-is, to match the rest of the file. May change in a follow-up PR, but undecided.

pkg/plugin/trans.go
sharnoff: Leaving as-is, to match the rest of the file. May change in a follow-up PR, but undecided.

PR #40, opened 2023-02-10
informant: Add auto-restarting capability

sharnoff:
There's some benefits to providing this ourselves — mostly that we'll do the cgroup cleanup operations that might be required on failure.

### DIFF
--- a/cmd/vm-informant/main.go
+++ b/cmd/vm-informant/main.go
@@ -3,7 +3,13 @@ package main
 import (
 	"errors"
 	"flag"
+	"fmt"
 	"net/http"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/containerd/cgroups/v3/cgroup2"
 
 	klog "k8s.io/klog/v2"
 
@@ -23,9 +29,34 @@ func main() {
 	// to distinguish between --cgroup=... and its absence, because null bytes aren't valid in paths.
 	noCgroup := "invalid\x00CgroupName"
 	var cgroupName string
+	var autoRestart bool
 	flag.StringVar(&cgroupName, "cgroup", noCgroup, "Sets the cgroup to monitor (optional)")
+	flag.BoolVar(&autoRestart, "auto-restart", false, "Automatically cleanup and restart on failure or exit")
 
 	flag.Parse()
+
+	// If we were asked to restart on failure, handle that separately:
+	if autoRestart {
+		var args []string
+		var cleanupHooks []func()
+		if cgroupName != noCgroup {
+			args = append(args, "-cgroup", cgroupName)
+			cleanupHooks = append(cleanupHooks, func() {
+				klog.Infof("vm-informant cleanup hook: making sure cgroup %q is thawed", cgroupName)
+				manager, err := cgroup2.Load(fmt.Sprint("/", cgroupName))
+				if err != nil {
+					klog.Warningf("Error making cgroup handler: %s", err)
+					return
+				}
+				if err := manager.Thaw(); err != nil {
+					klog.Warningf("Error thawing cgroup: %s", err)
+				}
+			})
+		}
+
+		runRestartOnFailure(args, cleanupHooks)
+		return
+	}
 
 	var stateOpts []informant.NewStateOpts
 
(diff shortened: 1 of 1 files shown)


tychoish (commented):
I think if you run under systemd you also get cgroup management/lifecycle. 

I guess I'm disinclined to do this outseleves, because we're running in an evironment that has other mechanisms...

cmd/vm-informant/main.go
```diff
@@ -64,3 +95,55 @@ func main() {
...
 	}
 }
+
+// runRestartOnFailure repeatedly calls this binary with the same flags, but with 'auto-restart'
+// removed.
+//
+// We execute ourselves as a subprocess so that it's possible to appropriately cleanup after
+// termination by various signals (or an unhandled panic!). This is worthwhile because we *really*
+// don't want to leave the cgroup frozen while waiting to restart.
+func runRestartOnFailure(args []string, cleanupHooks []func()) {
```
tychoish, line 105: what happens to signals and restarts?

cmd/vm-informant/main.go
```diff
@@ -64,3 +95,55 @@ func main() {
...
+	selfPath := os.Args[0]
+
+	minWaitDuration := time.Second * 5
+
+	for {
+		minWait := time.After(minWaitDuration)
+
+		cmd := exec.Command(selfPath, args...)
```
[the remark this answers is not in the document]
sharnoff: Done

cmd/vm-informant/main.go
sharnoff: What do you mean?

cmd/vm-informant/main.go
```diff
@@ -64,3 +95,55 @@ func main() {
...
+// We execute ourselves as a subprocess so that it's possible to appropriately cleanup after
+// termination by various signals (or an unhandled panic!). This is worthwhile because we *really*
+// don't want to leave the cgroup frozen while waiting to restart.
+func runRestartOnFailure(args []string, cleanupHooks []func()) {
+	selfPath := os.Args[0]
+
+	minWaitDuration := time.Second * 5
+
+	for {
+		minWait := time.After(minWaitDuration)
```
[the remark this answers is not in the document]
sharnoff: Don't think that gets us anything here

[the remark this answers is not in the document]
tychoish: I mean, ideally you'd have some kind of lifecycle (so that it would handle shut down/restart requests, upgrades, etc.) management, in which case we could handle signals more quickly? (which would involve having contexts, and not using this like a sleep, but it can all be changed later?

cmd/vm-informant/main.go
tychoish: what happens if the informat gets a signal or gets killed? do we spiral with restarts? do we propogate signals to the child processes? by what mechanism? 

(and sort of relatedly, though not in my original comment) how do we distinguish inner/outer logging messages?)

cmd/vm-informant/main.go
```diff
@@ -64,3 +96,55 @@ func main() {
...
+			h()
+		}
+
+		select {
+		case <-minWait:
+			klog.Infof("vm-informant restarting immediately")
+		default:
+			klog.Infof("vm-informant %s. respecting minimum wait of %s", exitMode, minWaitDuration)
+			<-minWait
+		}
```
tychoish, line 148: This is a really complicated way of (modulo logging): collecting the start time at the top of the loop and then: 

```go
time.Sleep(time.Since(startAt)-minWaitDuration)
```

The channel business only gets you something if you're processing multiple channels in a select or if you want to allow it to cancel.

sharnoff:
> I think if you run under systemd you also get cgroup management/lifecycle.
>
> I guess I'm disinclined to do this outseleves, because we're running in an evironment that has other mechanisms...

I think my original description should have been clearer. Basically the problem is this: If the VM informant exits without a chance to un-freeze the cgroup (either by SIGKILL, panic, or w/e), we _need_ to have some mechanism to thaw the cgroup on exit, so that a failure in the VM informant won't perma-freeze postgres.

(Un)fortunately, we aren't running under systemd.

cmd/vm-informant/main.go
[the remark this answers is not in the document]
sharnoff: Yeah, fair. Wrote half a comment re: systemd, forgot to finish it but now it's there - seems kind of related. Mostly this PR is just replicating 5 lines of bash, really just so we can keep the cgroups clean. Better to keep it simple for now I think.

tychoish:
> I think my original description should have been clearer. Basically the problem is this: If the VM informant exits without a chance to un-freeze the cgroup (either by SIGKILL, panic, or w/e), we _need_ to have some mechanism to thaw the cgroup on exit, so that a failure in the VM informant won't perma-freeze postgres.

I think the cases that are `sigkil` (as in it ooms and the oom killer gets us or we fail to shut down in a timely maner then we're *very likely* to get killed to, so it sort of doesn't help. I'm not sure that there are cases where this approach actually allows us to clean up in situations that won't end up killing the parent process too (or the bash script or whatever.) This is why I would tend to want to defer to systemd for this kind of thing. It's a bit of a monster, but it's not a monster we have to feed. 

For the panic cases defers at the top level should save us, which is frankly the 80% cause of an informant crashing. 

> (Un)fortunately, we aren't running under systemd.

That can be fixed :D

sharnoff:
My understanding is that switching to systemd isn't really an option right now. Would be good to discuss.

cmd/vm-informant/main.go
sharnoff: Fixed

sharnoff:
Merging, as previously discussed. Investigating whether using systemd would be worthwhile was added to the roadmap, as a post-launch item.

Also, ideally the outer loop should appropriately handle SIGTERM (and _maybe_ SIGINT?). Related to #12.

PR #43, opened 2023-02-10
billing: add client

tychoish:


sharnoff (approved):
Don't know if you wanted a review yet, but figured I'd have a look anyways.

Broadly looks good. I _was_ concerned about the ability to fetch information like tenant/timeline/endpoint IDs, but looks like that's available via pod labels :)

For example:

```shell
$ VM_NAME=compute-node-ep-calm-sky-909249
$ kubectl get neonvm $VM_NAME -o jsonpath='{.metadata.labels}' | jq                                                                                                                                                                    776ms  Sun 12 Feb 2023 10:58:05 PST
{
  "autoscaling.neon.tech/enabled": "true",
  "neon/branch-id": "br-fragrant-mouse-962162",
  "neon/component": "compute-node",
  "neon/endpoint-id": "ep-calm-sky-909249",
  "neon/project-id": "cool-cell-207703",
  "neon/tenant-id": "71b3dfc493552fc5b9f9972885e798ef",
  "neon/timeline-id": "f4f763f5d6930890669e7a09098ed9a8"
}
```

tychoish:
> Don't know if you wanted a review yet, but figured I'd have a look anyways.
> 
> Broadly looks good. I _was_ concerned about the ability to fetch information like tenant/timeline/endpoint IDs, but looks like that's available via pod labels :)

Cool, I think this is mergable. I think the problem is mostly that we may want to have a batch size so that we don't have too big payloads, but I don't want to optimize this too soon. 

There are two approaches to the batching solution:
- setting a limit on the batch and then calling Send lazily when the batch size is hit. (Would want contexts in the Add methods. 
- having a background thread that does sending, with adding being channel sends rather than appends to a slice. (this is a reasonable amount of machinery.

PR #49, opened 2023-02-17
agent: collect + send billing metrics

sharnoff:
Haven't tested it, still need to add more comments.

### DIFF
--- a/pkg/agent/billing.go
+++ b/pkg/agent/billing.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"math"
 	"net/http"
 	"time"
 
@@ -18,6 +19,7 @@ import (
 
 type BillingConfig struct {
 	URL                 string `json:"url"`
+	CPUMetricName       string `json:"cpuMetricName"`
 	CollectEverySeconds uint   `json:"collectEverySeconds"`
 	PushEverySeconds    uint   `json:"pushEverySeconds"`
 	PushTimeoutSeconds  uint   `json:"pushTimeoutSeconds"`
@@ -53,21 +55,27 @@ type vmMetricsInstant struct {
 
 // vmMetricsSeconds is like vmMetrics, but the values cover the allocation over time
 type vmMetricsSeconds struct {
-	// cpu stores the cpu seconds allocated to the VM, roughly equivalent to the integral of CPU
+	// cpu stores the CPU seconds allocated to the VM, roughly equivalent to the integral of CPU
 	// usage over time.
 	cpu uint32
 }
 
 const (
-	MetricNameCPU string = "cpu"
 	EndpointLabel string = "neon/endpoint-id"
 )
 
-func RunBillingMetricsColllector(backgroundCtx context.Context, conf *BillingConfig, store *util.WatchStore[vmapi.VirtualMachine]) {
+func RunBillingMetricsColllector(
+	backgroundCtx context.Context,
+	conf *BillingConfig,
+	targetNode string,
+	store *util.WatchStore[vmapi.VirtualMachine],
+) {
 	client := billing.NewClient(conf.URL, http.DefaultClient)
 
 	collectTicker := time.NewTicker(time.Second * time.Duration(conf.CollectEverySeconds))
 	defer collectTicker.Stop()
+	// Offset by half a second, so it's a bit more deterministic.
+	time.Sleep(500 * time.Millisecond)
 	pushTicker := time.NewTicker(time.Second * time.Duration(conf.PushEverySeconds))
 	defer pushTicker.Stop()
 
@@ -78,23 +86,29 @@ func RunBillingMetricsColllector(backgroundCtx context.Context, conf *BillingCon
 		pushWindowStart: time.Now(),
 	}
 
-	state.collect(conf, store)
+	state.collect(conf, targetNode, store)
+	batch := client.NewBatch()
 
 	for {
 		select {
 		case <-collectTicker.C:
-			state.collect(conf, store)
+			klog.Infof("Collecting billing state")
+			state.collect(conf, targetNode, store)
 		case <-pushTicker.C:
-			batch := state.makeBatch(&client)
+			klog.Infof("Creating billing batch")
+			state.drainAppendToBatch(conf, batch)
+			klog.Infof("Pushing billing events (count = %d)", batch.Count())
 			if err := pushBillingEvents(conf, batch); err != nil {
 				klog.Errorf("Error pushing billing events: %s", err)
 				continue
 			}
-			// Sending was successful; reset state
-			state.clear()
+			// Sending was successful; clear the batch.
+			batch = client.NewBatch()
 		case <-backgroundCtx.Done():
 			// If we're being shut down, push the latests events we have before returning.
-			batch := state.makeBatch(&client)
+			klog.Infof("Creating final billing batch")
+			state.drainAppendToBatch(conf, batch)
+			klog.Infof("Pushing final billing events (count = %d)", batch.Count())
 			if err := pushBillingEvents(conf, batch); err != nil {
 				klog.Errorf("Error pushing billing events: %s", err)
 			}
@@ -103,7 +117,7 @@ func RunBillingMetricsColllector(backgroundCtx context.Context, conf *BillingCon
 	}
 }
 
-func (s *billingMetricsState) collect(conf *BillingConfig, store *util.WatchStore[vmapi.VirtualMachine]) {
+func (s *billingMetricsState) collect(conf *BillingConfig, targetNode string, store *util.WatchStore[vmapi.VirtualMachine]) {
 	now := time.Now()
 
 	old := s.present
@@ -172,21 +186,16 @@ func (h *vmMetricsHistory) finalizeCurrentTimeSlice() {
 		return
 	}
 
-	duration := h.lastSlice.endTime.Sub(h.lastSlice.startTime).Nanoseconds()
+	duration := h.lastSlice.endTime.Sub(h.lastSlice.startTime)
 	if duration < 0 {
 		panic("negative duration")
 	}
 
-	// note: we could've used Duration.Seconds() above, but that gives us a float. We absolutely do
-	// not want to be doing billing math with floats.
-	//
-	// TODO: This approach here explicitly rounds us down - otherwise rounding is actually quite
-	// tricky, and it's safer to round down than up. In the future, we might internally store things
-	// as milliseconds or something, and round down to the nearest "CPU second" when sending the
-	// batch, in order to be more accurate. There's a variety of options.
-	seconds := duration / time.Second.Nanoseconds()
+	seconds := duration.Seconds()
+	// TODO: This approach is imperfect. Floating-point math is probably *fine*, but really not
+	// something we want to rely on. A "proper" solution is a lot of work, but long-term valuable.
 	metricsSeconds := vmMetricsSeconds{
-		cpu: uint32(h.lastSlice.metrics.cpu) * uint32(seconds),
+		cpu: uint32(math.Round(float64(h.lastSlice.metrics.cpu) * seconds)),
 	}
 	h.total.cpu += metricsSeconds.cpu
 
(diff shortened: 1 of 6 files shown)


pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+	state.collect(conf, store)
+
+	for {
+		select {
+		case <-collectTicker.C:
+			state.collect(conf, store)
+		case <-pushTicker.C:
+			batch := state.drainIntoBatch(&client)
+			if err := pushBillingEvents(conf, batch); err != nil {
+				return fmt.Errorf("Error pushing billing events: %w", err)
```
tychoish, line 79: does this mean that if we fail to push once we stop collecting consumption data? 

While I don't expect the console to be buggy, this seems aggressive.

pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+		present:    make(map[billingMetricsKey]vmMetrics),
+		lastTime:   nil,
+	}
+
+	state.collect(conf, store)
+
+	for {
+		select {
+		case <-collectTicker.C:
+			state.collect(conf, store)
```
tychoish, line 75: I was initially sort of concerned that by doing collection and pushing in the same thread, one could block the other, but I imagine that collection is always pretty fast, and if we don't collect for a couple of seconds that's probably net fine? 

The other option is to do them in different threads and use https://github.com/tychoish/fun/blob/main/pubsub/queue.go#L34 or similar to buffer/shed load.

pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if oldMetrics, ok := old[key]; ok {
+			// The VM was present from s.lastTime to now. Add a time slice to its metrics history.
+			timeSlice := metricsTimeSlice{
+				metrics: vmMetrics{
+					// strategically under-bill by assigning the minimum to the entire time slice.
+					cpu: util.Min(oldMetrics.cpu, presentMetrics.cpu),
+				},
+				// note: we know s.lastTime != nil because otherwise old would be empty.
+				startTime: *s.lastTime,
```
tychoish, line 118: I'd be inclined to avoid twiddling `time.Time` pointers and instead use the `time.Time.IsZero` to check the "is this set" status.

pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+	return batch
+}
+
+func (h vmMetricsHistory) forEachMergedSlice(process func(metricsTimeSlice)) {
+	if len(h.slices) == 0 {
+		return
+	}
+
+	last := h.slices[0]
+	for _, slice := range h.slices[1:] {
```
tychoish, line 163: range is kind of cute behind the scenes, so I think `slice` is always the same pointer object they just move what it points to? so I often (perhaps with an abundance of caution) use the index rather than the value output of range, a la: 

```go
for idx := range h.slices {
    cur := h.slices[idx] 
}
```

Though admittedly you're also doing someithing clever by starting at the second element, so maybe c-style rather than range is the way to go.

pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	process(last)
+}
+
+func pushBillingEvents(conf *BillingConfig, batch *billing.Batch) error {
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*time.Duration(conf.PushTimeoutSeconds))
+	defer cancel()
+
+	return batch.Send(ctx)
```
tychoish, line 180: short timeouts are probably good enough for most cases, but I think will definitely mean that we will miss events during console deploys, and it's worth calling that out, perhaps?

pkg/agent/billing.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (s *billingMetricsState) collect(conf *BillingConfig, store *util.WatchStore[vmapi.VirtualMachine]) {
+	now := time.Now()
+
+	old := s.present
+	s.present = make(map[billingMetricsKey]vmMetrics)
+	for _, vm := range store.Items() {
+		endpointID, ok := vm.Labels[EndpointLabel]
+		if !ok {
+			continue
```
tychoish, line 100: is this "something went away" or "we have a pod that we don't know about"

tychoish (approved):

pkg/agent/billing.go
sharnoff: "we have a pod that we don't know about" — added a comment.

pkg/agent/billing.go
sharnoff: in reverse order:

> admittedly you're also doing someithing clever by starting at the second element, so maybe c-style rather than range is the way to go.

I'll think about it for a bit. I like the readability of range syntax, but I see where you're coming from.

> I think `slice` is always the same pointer object they just move what it points to?

_kind of_ — IIRC it's more accurate to desugar this range as:

```go
{
    var slice metricsTimeSlice
    for i := 1; i < len(slice); i += 1 {
        slice = h.slices[i]
        // -- insert for loop body here --
    }
}
```

so you can run into issues when you create a reference to either `i` or `slice` that lives longer than the loop body, either explicitly with `&slice` or implicitly with a pointer method or anonymous function that captures it.

pkg/agent/billing.go
sharnoff: Yeah, that's what it'll currently do ­— originally wasn't sure about how best to resolve that and was planning on asking you. Although now I think the best way is probably just saving the batch for later if it fails, then re-sending with the merged batch on the next try? new code would look like:

```go
batch := client.NewBatch()
for {
    // -- skipped --
    case <-pushTicker.C:
        state.drainAppendToBatch(&batch)
        if err := pushBillingEvents(conf, batch); err != nil {
            klog.Errorf(...)
            continue
        }
        // reset batch, because sending was successful
        batch = client.NewBatch()
    // -- skipped --
}
```

pkg/agent/billing.go
sharnoff: > I imagine that collection is always pretty fast, and if we don't collect for a couple of seconds that's probably net fine?

that's basically my thoughts as well. Setting a short timeout on pushing means we should be _mostly_ ok — but probably worth improving this eventually.

pkg/agent/billing.go
sharnoff: I think we can miss during console deploys even with long timeouts — this is more a retry issue imo. We can discuss above.

pkg/agent/billing.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type metricsTimeSlice struct {
+	metrics   vmMetrics
+	startTime time.Time
+	endTime   time.Time
+}
+
+type vmMetrics struct {
+	cpu uint16
```
sharnoff, line 48: From discussion with @tychoish — this shouldn't be "cpu at a given time", instead should be "cpu seconds consumed over a timeslice". Will fix.

pkg/agent/billing.go
sharnoff: handled in latest commit, as previously discussed.

pkg/agent/billing.go
sharnoff: fixed in latest commit

pkg/agent/billing.go
```diff
@@ -93,84 +107,140 @@ func (s *billingMetricsState) collect(conf *BillingConfig, store *util.WatchStor
...
+func (s *billingMetricsState) makeBatch(client *billing.Client) *billing.Batch {
+	batch := client.NewBatch()
+
+	now := time.Now()
+
+	for key, history := range s.historical {
+		batch.AddIncrementalEvent(billing.IncrementalEvent{
+			IdempotencyKey: uuid.NewString(),
+			MetricName:     MetricNameCPU,
+			Type:           "", // set on JSON marshal
```
tychoish, line 223: I did the setting on json martial just to be cute, but we could delete that and set it here too?

pkg/agent/billing.go
```diff
@@ -93,84 +107,140 @@ func (s *billingMetricsState) collect(conf *BillingConfig, store *util.WatchStor
...
+
+// makeBatch creates a billing.Batch for the currently stored history, without modfiying any of it
+func (s *billingMetricsState) makeBatch(client *billing.Client) *billing.Batch {
+	batch := client.NewBatch()
+
+	now := time.Now()
+
+	for key, history := range s.historical {
+		batch.AddIncrementalEvent(billing.IncrementalEvent{
+			IdempotencyKey: uuid.NewString(),
```
tychoish, line 221: I feel like it's kind of wrong to use a random string (essentially) rather than the name of the object and some stored increment, there could be some weird http socket behavior that putting a random string here would cause us to think that we'd failed but have the request actually go through.

pkg/agent/billing.go
```diff
@@ -93,84 +107,140 @@ func (s *billingMetricsState) collect(conf *BillingConfig, store *util.WatchStor
...
+	}
+
+	// note: we could've used Duration.Seconds() above, but that gives us a float. We absolutely do
+	// not want to be doing billing math with floats.
+	//
+	// TODO: This approach here explicitly rounds us down - otherwise rounding is actually quite
+	// tricky, and it's safer to round down than up. In the future, we might internally store things
+	// as milliseconds or something, and round down to the nearest "CPU second" when sending the
+	// batch, in order to be more accurate. There's a variety of options.
+	seconds := duration / time.Second.Nanoseconds()
```
tychoish, line 187: durations are in nanoseconds already... so can't we just do duration/time.Second (?)

pkg/agent/billing.go
tychoish: yeah, as long as missed sending doesn't cause the payload to balloon, I think we're ok.

pkg/agent/billing.go
sharnoff: Made a change; no longer relevant.

sharnoff:
Did some manual testing, made some fixes, seems ok now. Also made some non-required but good to have changes - refer to the commit message on a46ca8b ("Some fixes, some changes").

Something I missed in the commit message: Decided to go back to an earlier approach w.r.t. batching, so that we preserve idempotency keys — it's essentially now equivalent to what I proposed here: https://github.com/neondatabase/autoscaling/pull/49#discussion_r1111314682. AFAICT, there's no way to respect idempotency that isn't equivalent to this approach.

pkg/agent/billing.go
sharnoff: See recent changes, should be handled.

pkg/agent/billing.go
sharnoff: I think there's better options; I'll open a follow-up PR.

pkg/agent/billing.go
```diff
@@ -0,0 +1,246 @@
...
+		case <-collectTicker.C:
+			klog.Infof("Collecting billing state")
+			state.collect(conf, targetNode, store)
+		case <-pushTicker.C:
+			klog.Infof("Creating billing batch")
+			state.drainAppendToBatch(conf, batch)
+			klog.Infof("Pushing billing events (count = %d)", batch.Count())
+			if err := pushBillingEvents(conf, batch); err != nil {
+				klog.Errorf("Error pushing billing events: %s", err)
+				continue
```
tychoish, line 103: if we get here, we have `n` things in the batch, the next time we get to line 98, we begin adding more things to the batch, correct, which means every time hit this condition, the batch will get bigger, thereby decreasing its chance of success. (particularly given the 500ms window and the chance that these could be running in different regions.)

pkg/agent/billing.go
tychoish: Also our little flap in cloud last week, makes me realize that we'll not be able to get to the control plane in 500ms from every agent. I think we should land this, and have an issue open to fiddle with this more, before GA

pkg/agent/billing.go
sharnoff: Yeah, it's not great. Realistically, size of JSON shouldn't cause us latency issues (as discussed, it'll still be quite small). Timeouts are worth considering, but also the configured duration is in seconds, so... :shrug:

When we add metrics, it'd be good to have something tracking the amount of time these requests take.

pkg/agent/billing.go
```diff
@@ -0,0 +1,246 @@
...
+
+// drainAppendToBatch clears the current history, adding it as events to the batch
+func (s *billingMetricsState) drainAppendToBatch(conf *BillingConfig, batch *billing.Batch) {
+	now := time.Now()
+
+	for key, history := range s.historical {
+		history.finalizeCurrentTimeSlice()
+
+		batch.AddIncrementalEvent(billing.IncrementalEvent{
+			IdempotencyKey: uuid.NewString(),
```
antonyc, line 225: Would be nice to store ISO time and hostname here for debug too.

PR #50, opened 2023-02-18
Add versioning to agent<->plugin protocol

sharnoff:
Mirrors the versioning for the agent<->informant protocol. Versioning here is required in some form to fix the "allow nil metrics" issue from #27.

Current plan is to release+deploy a new version _just_ with this change (requires autoscaler-agent first, then scheduler), then add the option for nil metrics in a follow-up PR + release.

### DIFF
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -64,6 +64,13 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
+// PluginProtocolVersion is the current version of the agent<->scheduler plugin in use by this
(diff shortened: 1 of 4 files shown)


sharnoff:
Realized that #53 is required to have a stable state between versions, which is generally something I want to maintain (i.e. for any two sequential versions X and Y, we _should_ have either plugin Y compatible with agent X and Y, or plugin X and Y compatible with agent Y — a side-effect of this is zero-ish downtime upgrades).

Anyways, #53 also means that a4d9d29 is not quite correct. I'll update this after it merges.

PR #51, opened 2023-02-19
Allow nil metrics in agent->plugin messages

sharnoff:
Builds on #50, fixes some annoyances around restart.

Will merge after #50 and its repo version bump.


sharnoff:
Resolved all comments; I believe we discussed them previously.

I've tested it, seems to be ok.

PR #52, opened 2023-02-19
Bump neonvm v0.4.5 -> v0.4.6

sharnoff:
Not strictly required, but worth doing this sooner rather than later.

### DIFF
--- a/scripts/download-deployments.sh
+++ b/scripts/download-deployments.sh
@@ -18,5 +18,5 @@ echo "downloading 'multus-daemonset.yaml'..."
 curl -sS https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset.yml \
     -o deploy/multus-daemonset.yaml
 echo "downloading 'neonvm.yaml'..."
-curl -sSL https://github.com/neondatabase/neonvm/releases/download/v0.4.5/neonvm.yaml \
+curl -sSL https://github.com/neondatabase/neonvm/releases/download/v0.4.6/neonvm.yaml \
     -o deploy/neonvm.yaml
--- a/vm_image/build.sh
+++ b/vm_image/build.sh
@@ -17,7 +17,7 @@ source '../scripts-common.sh'
 
 NEONVM_BUILDER_PATH='neonvm-builder'
 NEONVM_REPO='https://github.com/neondatabase/neonvm'
-NEONVM_VERSION='v0.4.5'
+NEONVM_VERSION='v0.4.6'
 
 if [ -e "$NEONVM_BUILDER_PATH" ]; then
     echo "Skipping downloading NeonVM vm-builder because '$NEONVM_BUILDER_PATH' already exists"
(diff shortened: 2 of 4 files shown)


vm_image/build.sh
```diff
@@ -57,6 +57,6 @@ echo " * IMAGE_SIZE = '$IMAGE_SIZE'"
 echo "Building 'Dockerfile.vmdata'..."
 docker buildx build -t "vmdata-temp:$TAG" -f Dockerfile.vmdata . | indent
 echo "Building NeonVM image..."
-"./$NEONVM_BUILDER_PATH" --src "vmdata-temp:$TAG" --dst "$REGISTRY/$NAME:$TAG" --size "$IMAGE_SIZE" | indent
+"./$NEONVM_BUILDER_PATH" --src "vmdata-temp:$TAG" --use-inittab --dst "$REGISTRY/$NAME:$TAG" --size "$IMAGE_SIZE" | indent
```
tychoish, line 60: is this orthogonal?

vm_image/build.sh
```diff
@@ -17,7 +17,7 @@ source '../scripts-common.sh'
 
 NEONVM_BUILDER_PATH='neonvm-builder'
 NEONVM_REPO='https://github.com/neondatabase/neonvm'
-NEONVM_VERSION='v0.4.5'
+NEONVM_VERSION='v0.4.6'
```
tychoish, line 20: is there a way that we can avoid the requirement of updating this value here and in the download deployments script so that upgrading a dependency can be done only in the go.mod file?

scripts/download-deployments.sh
```diff
@@ -18,5 +18,5 @@ echo "downloading 'multus-daemonset.yaml'..."
 curl -sS https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset.yml \
     -o deploy/multus-daemonset.yaml
 echo "downloading 'neonvm.yaml'..."
-curl -sSL https://github.com/neondatabase/neonvm/releases/download/v0.4.5/neonvm.yaml \
+curl -sSL https://github.com/neondatabase/neonvm/releases/download/v0.4.6/neonvm.yaml \
```
tychoish, line 21: is there a way we can get this from the go.mod to avoid hard coding it here?

tychoish (approved):

scripts/download-deployments.sh
bayandin:
```bash
$ go list -m -f '{{ .Version }}' github.com/neondatabase/neonvm
v0.4.5
```

vm_image/build.sh
sharnoff: Unfortunately no; `v0.4.5`'s inittab stuff was unintentionally a breaking change, so `v0.4.6` fixed the "breaking"-ness of it by putting the feature behind a flag.

vm_image/build.sh
sharnoff: duplicate, see below.

scripts/download-deployments.sh
sharnoff: Opened #62 to resolve this after merging.

Issue #55, opened 2023-02-20
scheduler takes a long time to shutdown


sharnoff:
Thought it'd be fixed by #32, but seems like it wasn't.

At first glance, seems like the context solution doesn't work either way, because K8s doesn't use the cobra command's context - see:

> ```go
> func runCommand(cmd *cobra.Command, opts *options.Options, registryOptions ...Option) error {
> 	verflag.PrintAndExitIfRequested()
> 
> 	// Activate logging as soon as possible, after that
> 	// show flags with the final logging configuration.
> 	if err := logsapi.ValidateAndApply(opts.Logs, utilfeature.DefaultFeatureGate); err != nil {
> 		fmt.Fprintf(os.Stderr, "%v\n", err)
> 		os.Exit(1)
> 	}
> 	cliflag.PrintFlags(cmd.Flags())
> 
> 	ctx, cancel := context.WithCancel(context.Background())
> 	//                                ^^^^^^^^^^^^^^^^^^^^ should be cmd.Context()
> 	defer cancel()
> 	go func() {
> 		stopCh := server.SetupSignalHandler()
> 		<-stopCh
> 		cancel()
> 	}()
> 
> 	cc, sched, err := Setup(ctx, opts, registryOptions...)
> 	if err != nil {
> 		return err
> 	}
> 	// add feature enablement metrics
> 	utilfeature.DefaultMutableFeatureGate.AddMetrics()
> 	return Run(ctx, cc, sched)
> 	//         ^^^ context used here.
> }
> ```

pointing comments added by me. source: https://github.com/kubernetes/kubernetes/blob/d6fe718e19881399fbb65f693c0c0f29f8f63861/cmd/kube-scheduler/app/server.go#L121-L147

---

HOWEVER, we're working with K8s 1.23 and I don't particularly want to fork kubernetes to backport a fix. AND this shouldn't matter — K8s is using `server.SetupSignalHandler()` to cancel the context, which _should_ work anyways.

sharnoff:
Checked, this issue still occurs after switching to 1.24 — will investigate further later this week.

sharnoff:
Looked into this, tried the fix in this issue, didn't fix it. Going to try the long way (start with base scheduler, gradually add each piece of the plugin until the behavior re-emerges).

sharnoff:
The issue was fixed in later K8s versions by this PR: https://github.com/kubernetes/kubernetes/pull/110207. Specifically, this diff:

```diff
[...]

Issue #37, opened 2023-02-07
Agent's informant server doesn't stop when context is cancelled


sharnoff:

tychoish:
Can you provide a bit more background here, and perhaps a link to the service that you'd expect to stop?

sharnoff:
Discovered that this is the cause of the goroutine leak we're seeing in prod. Lots and lots of goroutines with `net/http.(*Server).Serve(...)` in their callstack.

PR #59, opened 2023-02-21
agent: plumb contexts through

tychoish:
In persuit of #55 and #12 (but certianly not all of either.)


pkg/agent/informant.go
```diff
@@ -180,7 +180,7 @@ func NewInformantServer(
 	backgroundCtx, cancelBackground := context.WithCancel(ctx)
 
 	// note: docs for server.exit guarantee this function is called while holding runner.lock.
-	server.exit = func(status InformantServerExitStatus) {
+	server.exit = func(ctx context.Context, status InformantServerExitStatus) {
```
sharnoff, line 183: AFAICT this context is unused; we should avoid adding unused context arguments

pkg/agent/entrypoint.go
```diff
@@ -72,7 +70,7 @@ func (r MainRunner) Run() error {
 			globalState.Stop()
 			return nil
 		case event := <-podEvents:
-			globalState.handleEvent(event)
+			globalState.handleEvent(ctx, event)
```
sharnoff, line 73: this feels icky (`handleEvent` will spawn threads using the context long after `handleEvent` finishes, and we're always using the same context for `handleEvent`) - but the only better solution is storing the context in `globalstate` itself.

Thoughts?

pkg/agent/informant.go
```diff
@@ -199,16 +199,26 @@ func NewInformantServer(
...
 			unregisterName := fmt.Sprintf("InformantServer unregister (%s)", server.desc.AgentID)
-			runner.spawnBackgroundWorker(context.TODO(), unregisterName, func(c context.Context) {
-				if err := server.unregisterFromInformant(c); err != nil {
+			runner.spawnBackgroundWorker(ctx, unregisterName, func(context.Context) {
+				// we want shutdown to (potentially) live longer than the request which
+				// made it, but having a timeout is still good.
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
+
+				if err := server.unregisterFromInformant(ctx); err != nil {
```
sharnoff, line 221: `unregisterFromInformant` already uses a timeout on the request itself; do we need another one?

cmd/autoscaler-agent/main.go
```diff
@@ -48,8 +50,11 @@ func main() {
 		VMClient:   vmClient,
 	}
 
-	err = runner.Run()
-	if err != nil {
+	// TODO: add a signal handler here?
```
sharnoff, line 53:
```suggestion
	// TODO: add a signal handler here
```

pkg/agent/informant.go
```diff
@@ -199,16 +199,26 @@ func NewInformantServer(
 
 		// we need to spawn these in separate threads so the caller doesn't block while holding
 		// runner.lock
-		runner.spawnBackgroundWorker(context.TODO(), shutdownName, func(c context.Context) {
-			if err := httpServer.Shutdown(c); err != nil {
+		runner.spawnBackgroundWorker(ctx, shutdownName, func(context.Context) {
+			// we want shutdown to (potentially) live longer than the request which
+			// made it, but having a timeout is still good.
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
sharnoff, line 205: 20 seconds is a _super_ long timeout, compared to other things here (or: compared to our usual configuration of them). I'd either: make this shorter (e.g. 5s), add a config field for it, or calculate it from an existing config field

sharnoff (commented):
Had some thoughts, left comments.

Separately: It seems like this _should_ make it easy-ish to address #37 (basically, listen for context ending and trigger informant's `server.exit`).

pkg/agent/entrypoint.go
tychoish: I don't think there's any problem with passing a context to a function that returns before the goroutines it spawns (and indeed having this context means that shutting down the agents main loop will actually cause a shutdown. Eventually/soon the basecontext/shutdown stuff can/will help make some of this more manageable.

pkg/agent/informant.go
tychoish: I mean I think the question is more "how long could an HTTP request handler take to do a thing in normal operation and double(ish) it.

The cap on this is about 30s in my mind (which is probably just what the timeout handler was in System V init scripts between sigterm and sigkill if the process doesn't die, and which has definitely been carried further into the future.

pkg/agent/informant.go
tychoish: I think the fact that `doInformantRequest` takes a timeout *and* a context is a bit of a weird API, and I kind of planned to pull that apart in a later PR but I don't feel rushed about that.. 

it's plausible that we could just pass the the enclosing context to the unregister call, and not worry about it during shutdown, but.

tychoish:
> Separately: It seems like this _should_ make it easy-ish to address #37 (basically, listen for context ending and trigger informant's `server.exit`).

Yep! that's the hope.

pkg/agent/informant.go
```diff
@@ -199,16 +199,26 @@ func NewInformantServer(
 
 		// we need to spawn these in separate threads so the caller doesn't block while holding
 		// runner.lock
-		runner.spawnBackgroundWorker(context.TODO(), shutdownName, func(c context.Context) {
-			if err := httpServer.Shutdown(c); err != nil {
+		runner.spawnBackgroundWorker(ctx, shutdownName, func(context.Context) {
```
sharnoff, line 202:
```suggestion
		runner.spawnBackgroundWorker(context.TODO(), shutdownName, func(context.Context) {
```

pkg/agent/informant.go
```diff
@@ -199,16 +199,26 @@ func NewInformantServer(
...
+			if err := httpServer.Shutdown(ctx); err != nil {
 				runner.logger.Warningf("Error shutting down InformantServer: %s", err)
 			}
 		})
 		if server.madeContact {
 			// only unregister the server if we could have plausibly contacted the informant
 			unregisterName := fmt.Sprintf("InformantServer unregister (%s)", server.desc.AgentID)
-			runner.spawnBackgroundWorker(context.TODO(), unregisterName, func(c context.Context) {
-				if err := server.unregisterFromInformant(c); err != nil {
+			runner.spawnBackgroundWorker(ctx, unregisterName, func(context.Context) {
```
sharnoff, line 215:
```suggestion
			runner.spawnBackgroundWorker(context.TODO(), unregisterName, func(context.Context) {
```

sharnoff (approved):
I think the unresolved comments are mostly bikeshedding. Worst-case it's a little bit of tech debt that we already have plans to resolve.

Two suggested changes, mostly I think it'd be good to have `context.TODO()` so that it's more immediately obvious that the context for the background workers are disconnected.

sharnoff:
This appears to be causing the autoscaler-agent to immediately crash on startup. Logs:

```
I0304 00:36:48.696058       1 main.go:31] Got environment args: {ConfigPath:/etc/autoscaler-agent-config/config.json K8sNodeName:autoscale-sched-worker K8sPodIP:10.244.1.11}
I0304 00:36:48.696600       1 entrypoint.go:29] buildInfo.GitInfo:   bd9595e (2023-03-01 16:33:16 +0000) - agent: plumb contexts through (#59)
I0304 00:36:48.696615       1 entrypoint.go:30] buildInfo.NeonVM:    v0.4.6
I0304 00:36:48.696619       1 entrypoint.go:31] buildInfo.GoVersion: go1.19.6
I0304 00:36:48.696624       1 entrypoint.go:33] Starting pod watcher
I0304 00:36:48.796260       1 entrypoint.go:38] Pod watcher started
I0304 00:36:48.796273       1 entrypoint.go:40] Starting VM watcher
I0304 00:36:48.798530       1 entrypoint.go:45] VM watcher started
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x139c6ed]

goroutine 86 [running]:
github.com/tychoish/fun/seq.(*elemIter[...]).Next(0xc000652ff0?, {0x1940458?, 0xc0000ed080?})
	/go/pkg/mod/github.com/tychoish/fun@v0.7.0/seq/list.go:438 +0x2d
github.com/tychoish/fun/set.syncIterImpl[...].Next({{0x19339d8?, 0xc0005993b8?}, {0x1934d10?, 0xc0003c4b30?}}, {0x1940458?, 0xc0000ed080})
	/go/pkg/mod/github.com/tychoish/fun@v0.7.0/set/set.go:149 +0xdb
github.com/tychoish/fun/pubsub.(*Broker[...]).dispatchMessage(0x1940458, {0x1940458?, 0xc0000ed080?}, {0x193fc30, 0xc000456640}, {{{{0xc000652ff0, 0x24}, {0xc000564e40, 0xb}}, {0xc000653020, ...}, ...}, ...})
	/go/pkg/mod/github.com/tychoish/fun@v0.7.0/pubsub/broker.go:202 +0x2f1
github.com/tychoish/fun/pubsub.(*Broker[...]).startQueueWorkers.func2()
	/go/pkg/mod/github.com/tychoish/fun@v0.7.0/pubsub/broker.go:182 +0x145
created by github.com/tychoish/fun/pubsub.(*Broker[...]).startQueueWorkers
	/go/pkg/mod/github.com/tychoish/fun@v0.7.0/pubsub/broker.go:175 +0x250
```

AFAICT this is exclusively caused by the dependency bump.

If this is indeed an issue with `github.com/tychoish/fun`, this PR should be reverted before merging anything else if a fix is not yet available.

sharnoff:
Upgrading to v0.7.1 fixes the issue. Opened #73 to do so.

PR #60, opened 2023-02-21
informant: service orchestration

tychoish:


cmd/vm-informant/main.go
```diff
@@ -23,6 +23,18 @@ import (
...
+	ctx = srv.SetShutdown(ctx)      // allows workers to cause a shutdown
+	ctx = srv.WithOrchestrator(ctx) // creates and starts an orchestrator
+	ctx = srv.SetBaseContext(ctx)   // sets a context for starting async work in request scopes
+
+	orca := srv.GetOrchestrator(ctx)
+	svc := orca.Service()
+
+	defer func() {
+		if err := orca.Service().Wait(); err != nil {
+			klog.Fatal("shutting down service", err)
```
sharnoff, line 35:
```suggestion
			klog.Fatal("failed to shut down service", err)
```

cmd/vm-informant/main.go
```diff
@@ -92,14 +108,15 @@ func main() {
...
-	if errors.Is(err, http.ErrServerClosed) {
-		klog.Infof("Server ended.")
-	} else {
-		klog.Fatalf("Server failed: %s", err)
-	}
+	addr := "0.0.0.0:10301"
+	klog.Infof("Starting server at %s", addr)
+
+	// we create an http service and add it to the orchestrator,
+	// which will start it and manage it's lifecycle.
```
sharnoff, line 115:
```suggestion
	// which will start it and manage its lifecycle.
```

sharnoff (commented):
Ideally I'd like to keep the codepath for auto restarting as simple as possible — literally just "parse flags, run in a loop, exit".

I don't see the utility in involving a service orchestrator for that kind of thing, but I'm interested in your reasoning.

_Kind of_ a similar thing for the "normal" implementation - for that, I'm mostly curious what your longer-term plan is. Basically: this seems like it does nothing; why should we include it?

tychoish:
> Ideally I'd like to keep the codepath for auto restarting as simple as possible — literally just "parse flags, run in a loop, exit". I don't see the utility in involving a service orchestrator for that kind of thing, but I'm interested in your reasoning.

I think mostly, in order to productionize something like this, you have to try it out in smaller/simpler/easier services and then expanding into more complex systems. So this is really as much a proof of concept as anything else. I didn't move the restart/subprocess loop around much in this one, but it is... morally a service, so it was a thought. I'm not commited to it. 

> _Kind of_ a similar thing for the "normal" implementation - for that, I'm mostly curious what your longer-term plan is. Basically: this seems like it does nothing; why should we include it?

Again, mostly proving it out in the informant before applying the toolkit to bigger problems.

sharnoff:
Don't know why it CI didn't catch it, but this doesn't compile. `svc` on line 31 is not used, and error value from `orca.Add(...)` is unchecked.

tychoish:
> Don't know why it CI didn't catch it, but this doesn't compile. `svc` on line 31 is not used, and error value from `orca.Add(...)` is unchecked.

Super weird. Showed up locally, *shrug* Fixed now.

sharnoff (approved):
Looks good. I think ideally I'd put the context & orchestrator stuff after `if autoRestart { ... }`, because it's not used in auto-restart mode, but ultimately it doesn't really matter.

PR #61, opened 2023-02-21
informant: handle signals

tychoish:
as discussed


cmd/vm-informant/main.go
```diff
@@ -97,54 +102,103 @@ func main() {
...
+				// drain the timer before resetting it:
+				if !timer.Stop() {
+					<-timer.C
+				}
+				timer.Reset(minWaitDuration - dur)
+
+				klog.Infof("vm-informant %s. respecting minimum wait of %s", exitMode, minWaitDuration)
+				select {
+				case <-ctx.Done():
+					klog.Infof("vm0informant: received signal")
```
sharnoff, line 193:
```suggestion
					klog.Infof("vm-informant restart loop: received termination signal")
```

cmd/vm-informant/main.go
```diff
@@ -97,54 +102,103 @@ func main() {
...
+		}()
+
+		select {
+		case <-ctx.Done():
+			klog.Infof("vm0informant: received signal")
+			return
+		case <-sig:
+			dur := time.Since(startTime)
+			if dur < minWaitDuration {
+				// drain the timer before resetting it:
```
sharnoff, line 184:
```suggestion
				// drain the timer before resetting it, required by Timer.Reset:
```

cmd/vm-informant/main.go
```diff
@@ -97,54 +102,103 @@ func main() {
...
+			cmd.Stderr = os.Stderr
+			go func() {
+				defer close(sig)
+				select {
+				case <-pctx.Done():
+					return
+				case <-ctx.Done():
+					if cmd.Process == nil || cmd.ProcessState == nil {
+						return
+					}
```
sharnoff, line 138: This is racy. Aside from data races around `cmd.Process` and `cmd.ProcessState` (which shouldn't have a material impact beyond annoying the race detector), it's possible for the following sequence of events to occur:

1. Spawn this function
2. Parent vm-informant receives SIGTERM, this function sees `cmd.Process == nil` and returns
3. `cmd.Run()` starts
4. Child vm-informant never receives any signal.

We could partly fix this by putting `defer pcancel()` inside this goroutine. However:

AFAICT any solution using `cmd.Run()` will be similarly racy. I'd rewrite with `Start` + `Wait` instead.

cmd/vm-informant/main.go
```diff
@@ -97,54 +102,103 @@ func main() {
...
 
 	for {
 		startTime := time.Now()
-
-		cmd := exec.CommandContext(context.TODO(), selfPath, args...)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		err := cmd.Run()
-
+		sig := make(chan struct{})
```
sharnoff, line 120: maybe not relevant given the comment about raciness, but in future these should have more descriptive names ­— instead of answering "what is this" (a signalling channel), I find it more useful to answer "what does it mean" (letting us know when the process is done!). Here, I'd personally call it `cmdDone` or something.

(although afaict it isn't _really_ indicating anything concrete, other than "the goroutine that was going to send it a SIGTERM exited" - even though it _seems_ to be used approximately as "command finished"?)

cmd/vm-informant/main.go
```diff
@@ -97,54 +102,103 @@ func main() {
 	}
 }
 
+const minWaitDuration = 5 * time.Second
```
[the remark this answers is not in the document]
tychoish: I wanted it to be a constant and I think constants inside of functions are... odd, but can definitely fix up the name

cmd/vm-informant/main.go
tychoish: Yeah, I think this is a big part of why I didn't want us to have to be in the business of doing this kind of process management: there are a lot of corners. I don't know that Start+Wait definitely gets you out of the potential of 4 happening, (because what I think you really want is to run the subprocess in a cgroup that will terminate on the parent process exiting.) This us (unfortunately) a thing that windows has done better than Linux (with job objects), and you might be able to get it with cgroups/systemd managed processes, but it's dodgy to do this "correctly" without help from the operating system.

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,106 @@ func main() {
...
+
+			if err == nil {
+				go func() {
+					select {
+					case <-pctx.Done():
+						return
+					case <-ctx.Done():
+						if cmd.Process == nil || cmd.ProcessState == nil {
+							return
+						}
```
sharnoff, line 142: This condition is never true, I think?

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,106 @@ func main() {
...
+						if cmd.ProcessState.ExitCode() >= 0 {
+							return
+						}
+						var pid int
+						if cmd.Process != nil {
+							pid = cmd.Process.Pid
+						}
```
sharnoff, line 149: Same deal here, I think - this condition is probably never false?

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,106 @@ func main() {
...
+					select {
+					case <-pctx.Done():
+						return
+					case <-ctx.Done():
+						if cmd.Process == nil || cmd.ProcessState == nil {
+							return
+						}
+						if cmd.ProcessState.ExitCode() >= 0 {
+							return
+						}
```
sharnoff, line 145: Best I can tell, checking `ProcessState` is not thread-safe. Seems like we're not actually using `ProcessState` other than to check that the command exited — a safer solution could be just setting some atomic boolean when `Wait()` returns.

There's always a possibility of kill failures from TOCTOU issues, but those are unavoidable I think.

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,106 @@ func main() {
...
-			exitMode = ""
-			klog.Warningf("vm-informant exited without error. This should not happen.")
-		}
+		}()
+
+		select {
+		case <-ctx.Done():
+			klog.Infof("vm-informant restart loop: received termination signal")
+			return
+		case <-processedSignal:
```
sharnoff, line 192: I think this always triggers?

sharnoff (commented):
Ok one last set, then should be good I think.

Mainly unnecessary conditionals with a hint of thread-safety. For the conditionals, I'm not sure my assessment is right — please also check?

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,106 @@ func main() {
...
+			if err != nil {
+				// lint note: the linter's worried about wrapped errors being incorrect with switch, but
+				// this is cleaner than the alternative (via errors.As) and it's still correct because
+				// exec.Command.Wait() explicitly mentions ExitError.
+				switch err.(type) { //nolint:errorlint // see above.
+				case *exec.ExitError:
+					exitMode = "failed"
+					klog.Errorf("vm-informant exited with: %v", err)
+				default:
+					exitMode = "failed to start"
```
tychoish, line 168: I believe that if there are problems writing to the output we could have an error from wait that isn't "failed to start", having said that, I'd be inclined to drop the `exitMode` entirely and hope that the error doesn't need any more annotation.

cmd/vm-informant/main.go
tychoish: I think we can use the `pctx` for this purpose.

cmd/vm-informant/main.go
sharnoff: Fair enough, go for it

sharnoff (approved):
Looks good. Had a couple nits, made the changes myself. lmk if those seem right.

For ease of review (and not clobbering commit history), this PR should be merged after #60.

cmd/vm-informant/main.go
```diff
@@ -103,48 +110,94 @@ func main() {
...
-func runRestartOnFailure(args []string, cleanupHooks []func()) {
+func runRestartOnFailure(ctx context.Context, args []string, cleanupHooks []func()) {
 	selfPath := os.Args[0]
-
-	minWaitDuration := time.Second * 5
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+
+	// CancelFunc for each process's context, for internal signaling.
+	// The defer is out here so we don't leak deferred funcs on restart.
```
tychoish, line 119: I don't think this matters, particularly: 
- the previous defer was in a function that started on line 129, which meant that it always had run by at least line 174, which is before before the next iteration of the loop, under any branch. Though I think the previous code was a bit safer to edit, on the case that someone adds an early return to the 129 function that isn't the cause of the outer context being canceled? 
- the only place where this could matter is if the goroutine on 139 panics (which isn't handled right now, probably correctly) or if if something in the function on line 129 panics, (which, except for the cleanup hook(s) is just `os/exec`) so I think we're firmly in "moving the deck chairs on the titanic" territory.

sharnoff: Shoot yeah, right. Missed that it was in an inner function already.

sharnoff:
> I think the pcancel change is a bit dubious, more for ergonomics than for correctness, but I don't think we should block on it..

Yep, now see it's definitely dubious. Best to revert.

> I think we want both PRs to be separate commits into main, right? Because https://github.com/neondatabase/autoscaling/pull/60 is a PR into this branch, I think the way to do that is to merge this first, and then https://github.com/neondatabase/autoscaling/pull/60 will become a PR into main, but yeah.

Yup, I got mixed up. Better to merge this first.

I have no more comments/issues, feel free to merge.

PR #63, opened 2023-02-24
scheduler: service orchestration

tychoish:
In persuit of #55, a better solution I hope. I think it's better to
roll these out slowly, more of the goroutines can be serviceified later.


sharnoff (approved):
Overall, looks fine. Made a change to error messages, lmk if that's good.

One nit about unrelated changes. I noticed we've (both) been doing a bunch of those, want to be a little bit less sloppy about it.

Issue #34, opened 2023-02-01
deploy consistent locking discipline. 


tychoish:
In reading through the agent code, it's not particularly straight forward (unless you read the comments) which locks you need to be holding to call a specific function. There are some changes to the concurrency model that we might want to deploy in the future (e.g. fewer locks, locks only protect fields in their own structs, etc,) but I think we shouldn't let the perfect be the enemy of the good, and in that spirit... 

I would propose the following idiom:
> No exported (public) function or method requires that the caller hold a specific lock and we should assume that they are safe to call from any go routine. Un-exported/private functions and methods should generally be responsible for holding/managing the scope of their locks (or not needing locks). If it's imperative that the caller of a private function hold a lock the word `unsafe` should appear in the function or method name.

sharnoff:
I think I'm _generally_ in favor of this. There's some kind of complex untangling that'd have to take place for that to work, and there's some existing behaviors (in autoscaler-agent, in particular) that are hard to replicate while adhering to this — stuff to be mindful of.

It seems like this is closely related to the "pkg/agent should be split into multiple packages" item from #27 — It'd be good to keep this in mind while doing that. Probably doesn't make sense to try and do this before splitting stuff up more :)

tychoish:
I think the "reogranize/split the package" might actually depend on doing this because right now the splits would have to be centered on specific mutexes rather than on stronger abstractions, or you'd end up doing a lot of mutex rearranging in order to split the packages.

sharnoff:
Oooh, that's a good point. Thinking some more, it seems like these are heavily intertwined. Seems like realistically the cleanest option is handling both at the same time (probably as a series of smaller changes with a clear "end state" in mind). Thoughts?

tychoish:
I'd definitely do the lock untangling first and then split things out but you could interleave them into different commits/PRs. 

I think having some kind of base integration test for guard rails would be good, as a prerequisite for both.

sharnoff:
This is (presumably) no longer applicable -- there's been a lot of changes in the last year and a half :)

PR #65, opened 2023-02-27
testing: provide handcrank test watcher

tychoish:
This just pulls apart some of the util.Watcher parts to make it
possible to have a `testutil.Watch` that can produce a `WatchStore`
populated by mock events from a channel.

### DIFF
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -375,11 +375,11 @@ func ProcessEvent[T any, P WatchObject[T]](
 	handlers WatchHandlerFuncs[P],
 	args ProcessEventArgs[T, P],
 ) error {
+	store.mutex.Lock()
+	defer store.mutex.Unlock()
+
 	switch args.Event.Type {
 	case watch.Added:
-		store.mutex.Lock()
-		defer store.mutex.Unlock()
-
 		if _, ok := store.objects[args.UID]; ok {
 			return fmt.Errorf(
 				"watch %s: received add event for object %s that we already have",
@@ -389,13 +389,11 @@ func ProcessEvent[T any, P WatchObject[T]](
 		store.objects[args.UID] = (*T)(args.Object)
 		handlers.AddFunc((*T)(args.Object), false)
 	case watch.Bookmark:
-		// Nothing to do, just serves to give us a new ResourceVersion.
+		// no action necessary, but noop has to remain here to
+		// avoid panic.
 	case watch.Deleted:
 		// We're given the state of the object immediately before deletion, which
 		// *may* be different to what we currently have stored.
-		store.mutex.Lock()
-		defer store.mutex.Unlock()
-
 		old, ok := store.objects[args.UID]
 		if !ok {
 			return fmt.Errorf(
@@ -417,7 +415,7 @@ func ProcessEvent[T any, P WatchObject[T]](
 		store.objects[args.UID] = (*T)(args.Object)
 		handlers.UpdateFunc(old, (*T)(args.Object))
 	case watch.Error:
-		panic(errors.New("unreachable code reached")) // handled above
+		panic(errors.New("unreachable code reached"))
 	default:
 		panic(errors.New("unknown watch event"))
 	}
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -263,7 +263,6 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 					}
 				}
 			}
-
 		relist:
 			klog.Infof("watch %s: re-listing", config.LogName)
 			for {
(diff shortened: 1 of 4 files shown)


pkg/util/watch.go
```diff
@@ -409,10 +359,72 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
+		if !ok {
+			return fmt.Errorf(
+				"watch %s: received update event for object %s that's not present",
+				args.Config.LogName, args.Name,
+			)
+		}
+		store.objects[args.UID] = (*T)(args.Object)
+		handlers.UpdateFunc(old, (*T)(args.Object))
+	case watch.Error:
+		panic(errors.New("unreachable code reached")) // handled above
```
sharnoff, line 420: note `// handled above` is still handled in the original function. It's worth adding a note that `ProcessEvent` must not be called with an event with type `watch.Error`. Should also change this panic message.

pkg/util/watch.go
```diff
@@ -409,10 +359,72 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
+	return store, nil
+}
+
+type ProcessEventArgs[T any, P WatchObject[T]] struct {
+	Name   string
+	UID    types.UID
+	Config WatchConfig
+	Event  watch.Event
+	Object P
 }
```
sharnoff, line 371: I think it's probably cleaner to move the rest of the post-error-check handling into `ProcessEvent`, which allows `Name`, `UID`, and `Object` to be derived inside. With that, we probably don't need `ProcessEventArgs`?

pkg/util/watch.go
```diff
@@ -409,10 +359,72 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
+	case watch.Modified:
+		old, ok := store.objects[args.UID]
+		if !ok {
+			return fmt.Errorf(
+				"watch %s: received update event for object %s that's not present",
+				args.Config.LogName, args.Name,
+			)
+		}
+		store.objects[args.UID] = (*T)(args.Object)
+		handlers.UpdateFunc(old, (*T)(args.Object))
```
sharnoff, line 418: Just noticed we're not locking `store.mutex` here — oops. I'll open a PR to fix that. `ProcessEvent` should probably just lock `store.mutex` outside the switch statement; it'll be cleaner.

pkg/util/watch.go
```diff
@@ -409,10 +359,72 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
+		if _, ok := store.objects[args.UID]; ok {
+			return fmt.Errorf(
+				"watch %s: received add event for object %s that we already have",
+				args.Config.LogName, args.Name,
+			)
+		}
+		store.objects[args.UID] = (*T)(args.Object)
+		handlers.AddFunc((*T)(args.Object), false)
+	case watch.Bookmark:
+		// Nothing to do, just serves to give us a new ResourceVersion.
```
sharnoff, line 392: Maybe worth having a similar "don't call this function with `watch.Error`" thing for `watch.Bookmark`; this is kind of just noise here, and we can filter it out at the callsite.

sharnoff (commented):
As discussed, broadly I think this is a good direction.

Some specifics: I don't like that the hand-crank-ness is available on `WatchStore` directly. Instead of `NewWatchStore`, I'd make a separate function (e.g., `NewTestWatcher`, idk) that takes `Config` and `WatchHandlerFuncs` and returns `WatchStore` and `WatchStoreTestFeeder` (name's eh, do whatever), with methods on `WatchStoreTestFeeder` that instantiate and process each type of event through the underlying `WatchStore`. Basically, our internal testing code shouldn't need to know how to construct kubernetes watch events, and our non-testing code shouldn't be able to.

It'd also be good to hook up `WatchStore.Stop()` to work with testing `WatchStore` creation. `Stop` is also currently kind of a mess (not for any reason, it's just non-idiomatic and poorly written).

Might have some more comments the second time around; I left off comments on some of the specifics that I thought might be obviated by structural changes.

pkg/util/watch.go
tychoish: I think there's merit in being exhaustive in switch statements like these. It's also the case that if we filter things out in the main loop, then this function becomes less useless as the stateless filter through which events flow.

pkg/util/watch.go
tychoish: I moved the lock to the top. I think this a decent demonstration of why #34 would be good. Amusing the handlers don't cause a deadlock, that is.

pkg/util/watch.go
tychoish: I don't understand what you mean by `post-error-checking`. 

The `goto` and loops make it hard to move more around without making a larger (and therefore riskier) cut. I think as an intermediate state.

sharnoff:
Closing as we have no current plans to continue with this strategy for testing (and because #371 has been merged, which does finally have tests).

Can always reopen in the future.

PR #69, opened 2023-03-03
billing: change idempotency key

tychoish:
As a follow up to: https://github.com/neondatabase/autoscaling/pull/49#discussion_r1123623721

### DIFF
--- a/pkg/billing/client.go
+++ b/pkg/billing/client.go
@@ -46,7 +46,6 @@ func (b *Batch) idempotenize(key string) string {
 	}
 
 	return fmt.Sprintf("Host<%s>:ID<%s>:T<%s>", b.c.hostname, uuid.NewString(), time.Now().Format(time.RFC3339))
-
 }
 
 func (b *Batch) AddAbsoluteEvent(e AbsoluteEvent) {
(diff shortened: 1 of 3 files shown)


pkg/billing/client.go
```diff
@@ -5,34 +5,59 @@ import (
...
 func (b *Batch) Count() int {
 	return len(b.events)
 }
 
+func (b *Batch) idempotenize(key string) string {
+	if key != "" {
+		return key
+	}
+
+	return fmt.Sprintf("Host<%s>:ID<%s>:T<%s>", b.c.hostname, uuid.NewString(), time.Now().Format(time.RFC3339))
```
antonyc, line 48: NIT: having a space as a separator for RFC3339 time is slightly easier to parse.

antonyc (approved):

pkg/billing/client.go
tychoish: are we going to be parsing these or just reading them?

antonyc: We're not going to parse them.

sharnoff (approved):

PR #70, opened 2023-03-03
agent: add pprof to agent

tychoish:


sharnoff (approved):
Seems fine. Changed "problem adding pprof service" to a hard error

PR #76, opened 2023-03-06
agent, plugin: Add "dump state" endpoints

sharnoff:
There was some bulk renaming required in the plugin, so I extracted that into a separate PR (#75) to make reviewing / blame nicer.

I've done a little bit of manual testing for it, seems fine. The JSON output isn't small, so... thorough testing will probably be easier with a dedicated tool to analyze the output.

For now, due to "quirks" of `util.AddHandler`, the body of the request is expected to be "{}". So, example usage would be something like:
```
curl -X GET -d '{}' 'http://localhost:10300/"
```
(for the current default autoscaler-agent port)

### DIFF
--- a/pkg/agent/dumpstate.go
+++ b/pkg/agent/dumpstate.go
@@ -73,7 +73,7 @@ func (s *agentState) DumpState(ctx context.Context, stopped bool) (*StateDump, e
 		}
 		defer s.lock.Unlock()
 
-		var list []*podState
+		list := make([]*podState, 0, len(s.pods))
 		for name := range s.pods {
 			list = append(list, s.pods[name])
 		}
--- a/pkg/agent/dumpstate.go
+++ b/pkg/agent/dumpstate.go
@@ -1,0 +1,115 @@
+package agent
+
+// Utilities for dumping internal state
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/neondatabase/autoscaling/pkg/util"
+)
+
+type StateDump struct {
+	Stopped   bool           `json:"stopped"`
+	BuildInfo util.BuildInfo `json:"buildInfo"`
+	Pods      []podStateDump `json:"pods"`
+}
+
+func (s *agentState) StartDumpStateServer(shutdownCtx context.Context, config *DumpStateConfig) error {
+	// Manually start the TCP listener so we can minimize errors in the background thread.
+	addr := net.TCPAddr{IP: net.IPv4zero, Port: int(config.Port)}
+	listener, err := net.ListenTCP("tcp", &addr)
+	if err != nil {
+		return fmt.Errorf("Error binding to %v", addr)
+	}
+
+	go func() {
+		mux := http.NewServeMux()
+		util.AddHandler("dump-state: ", mux, "/", http.MethodGet, "<empty>", func(ctx context.Context, body *struct{}) (*StateDump, int, error) {
+			timeout := time.Duration(config.TimeoutSeconds) * time.Second
+
+			startTime := time.Now()
+			ctx, cancel := context.WithDeadline(ctx, startTime.Add(timeout))
+			defer cancel()
+
+			state, err := s.DumpState(ctx, shutdownCtx.Err() != nil)
+			if err != nil {
+				totalDuration := time.Since(startTime)
+				if totalDuration >= timeout {
+					return nil, 500, fmt.Errorf("timed out after %s while getting state", totalDuration)
+				} else {
+					// some other type of cancel; 400 is a little weird, but there isn't a great
+					// option here.
+					return nil, 400, fmt.Errorf("error while getting state: %w", err)
+				}
+			}
+
+			return state, 200, nil
+		})
+		// note: we don't shut down this server. It should be possible to continue fetching the
+		// internal state after shutdown has started.
+		server := &http.Server{Handler: mux}
+		if err := server.Serve(listener); err != nil {
+			klog.Errorf("dump-state server exited: %w", err)
+		}
+	}()
+
+	return nil
+}
+
+func (s *agentState) DumpState(ctx context.Context, stopped bool) (*StateDump, error) {
+	// Copy the high-level state, then process it
+	podList, err := func() ([]*podState, error) {
+		if err := s.lock.TryLock(ctx); err != nil {
+			return nil, err
+		}
+		defer s.lock.Unlock()
+
+		var list []*podState
+		for name := range s.pods {
+			list = append(list, s.pods[name])
+		}
+		return list, nil
+	}()
+	if err != nil {
+		return nil, err
+	}
+
+	results := make(chan podStateDump)
+
+	// TODO: We should have some concurrency limit on the number of active goroutines while fetching
+	// state.
+	for _, pod := range podList {
+		// pass pod through explicitly, so we avoid variable reuse issues
+		go func(p *podState) {
+			results <- p.Dump(ctx)
+		}(pod)
+	}
+
+	state := StateDump{
+		Stopped:   stopped,
+		BuildInfo: util.GetBuildInfo(),
+		Pods:      nil,
+	}
+
+	cleanupCourtesy := time.NewTimer(0)
+	cleanupCourtesy.Stop()
+
+	for i := 0; i < len(podList); i += 1 {
+		select {
+		case <-ctx.Done():
+			cleanupCourtesy.Reset(100 * time.Millisecond)
+		case <-cleanupCourtesy.C:
+			break
+		case pod := <-results:
+			state.Pods = append(state.Pods, pod)
+		}
+	}
+
+	return &state, nil
+}
--- a/pkg/agent/dumpstate.go
+++ b/pkg/agent/dumpstate.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	"k8s.io/klog/v2"
 
 	"github.com/neondatabase/autoscaling/pkg/util"
@@ -112,6 +114,11 @@ func (s *agentState) DumpState(ctx context.Context, stopped bool) (*StateDump, e
 	// that to make it back to state.Pods when the context expires, instead of proactively aborting
 	// in *this* thread.
 	wg.Wait()
+
+	// Sort the pods by name, so that we produce a deterministic ordering
+	slices.SortFunc(state.Pods, func(a, b podStateDump) (less bool) {
+		return a.PodName.Namespace < b.PodName.Namespace && a.PodName.Name < b.PodName.Name
+	})
 
 	return &state, nil
 }
--- a/pkg/agent/dumpstate.go
+++ b/pkg/agent/dumpstate.go
@@ -103,7 +103,7 @@ func (s *agentState) DumpState(ctx context.Context, stopped bool) (*StateDump, e
 				wg.Done()
 			}()
 
-			state.Pods[i] = pod.Dump(ctx)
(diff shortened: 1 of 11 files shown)


tychoish:
this PR didn't run all of the checks/e2e stuff, probably because it's behind main? I think merging main into this will help.

sharnoff:
Yep :) Rebased, should be good now

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+		// pass pod through explicitly, so we avoid variable reuse issues
+		go func(p *podState) {
+			results <- p.Dump(ctx)
+		}(pod)
+	}
+
+	state := StateDump{
+		Stopped:   stopped,
+		BuildInfo: util.GetBuildInfo(),
+		Pods:      nil,
```
tychoish, line 97: allocate a capacity on this rather than setting it to nil.

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	cleanupCourtesy := time.NewTimer(0)
+	cleanupCourtesy.Stop()
+
+	for i := 0; i < len(podList); i += 1 {
```
tychoish, line 103: It took me reading this code 2-3 times to get what's going here, because this is pretty clever and not consistent  with most idioms for doing these kinds of workerpool situations: typically doing a range over a closed channel, or a select over a channel and a cancel (checking the ok value from the channel reads) is a bit less unexpected.

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+	// Manually start the TCP listener so we can minimize errors in the background thread.
+	addr := net.TCPAddr{IP: net.IPv4zero, Port: int(p.state.conf.DumpState.Port)}
+	listener, err := net.ListenTCP("tcp", &addr)
+	if err != nil {
+		return fmt.Errorf("Error binding to %v", addr)
+	}
+
+	go func() {
+		mux := http.NewServeMux()
+		util.AddHandler("dump-state: ", mux, "/", http.MethodGet, "<empty>", func(ctx context.Context, body *struct{}) (*stateDump, int, error) {
```
tychoish, line 48: I believe body isn't used, I might not declare the variable at all (use `_`), also I'm pretty sure that all empty structs are actually just a singleton in memory, so the pointer is probably more expensive than using a value,) if that's your intent

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+	return "", nil
+}
+
+type stateDump struct {
+	Stopped   bool            `json:"stopped"`
+	BuildInfo util.BuildInfo  `json:"buildInfo"`
+	State     pluginStateDump `json:"state"`
+}
+
+func (p *AutoscaleEnforcer) startDumpStateServer(shutdownCtx context.Context) error {
```
tychoish, line 38: I think we don't actually use the shutdown context at all? is this an oversight?

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+	// Manually start the TCP listener so we can minimize errors in the background thread.
+	addr := net.TCPAddr{IP: net.IPv4zero, Port: int(p.state.conf.DumpState.Port)}
+	listener, err := net.ListenTCP("tcp", &addr)
+	if err != nil {
+		return fmt.Errorf("Error binding to %v", addr)
+	}
+
+	go func() {
```
tychoish, line 46: I read this function a few times to make sure that it couldn't panic... I think we're safe but the unchecked panic here makes me nervous for people to edit this in the future.

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (s *agentState) StartDumpStateServer(shutdownCtx context.Context, config *DumpStateConfig) error {
+	// Manually start the TCP listener so we can minimize errors in the background thread.
+	addr := net.TCPAddr{IP: net.IPv4zero, Port: int(config.Port)}
+	listener, err := net.ListenTCP("tcp", &addr)
+	if err != nil {
+		return fmt.Errorf("Error binding to %v", addr)
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
sharnoff: `StartDumpStateServer` is intended to be run from the main goroutine directly; spawning to run the server means we can fail faster with an address binding error in the main thread. Basically, error handling is a little nicer.

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+		mux := http.NewServeMux()
+		util.AddHandler("dump-state: ", mux, "/", http.MethodGet, "<empty>", func(ctx context.Context, body *struct{}) (*StateDump, int, error) {
+			timeout := time.Duration(config.TimeoutSeconds) * time.Second
+
+			startTime := time.Now()
+			ctx, cancel := context.WithDeadline(ctx, startTime.Add(timeout))
```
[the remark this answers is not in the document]
sharnoff: Fixed

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+		return list, nil
+	}()
+	if err != nil {
+		return nil, err
+	}
+
+	results := make(chan podStateDump)
+
+	// TODO: We should have some concurrency limit on the number of active goroutines while fetching
+	// state.
```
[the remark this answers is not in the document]
sharnoff: Done, kind of. I'd appreciate your thoughts on the new version.

pkg/agent/dumpstate.go
sharnoff: done

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	state := StateDump{
+		Stopped:   stopped,
+		BuildInfo: util.GetBuildInfo(),
+		Pods:      nil,
+	}
+
+	cleanupCourtesy := time.NewTimer(0)
+	cleanupCourtesy.Stop()
```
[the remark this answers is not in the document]
sharnoff: no longer applies

pkg/agent/dumpstate.go
```diff
@@ -0,0 +1,115 @@
...
+
+	for i := 0; i < len(podList); i += 1 {
+		select {
+		case <-ctx.Done():
+			cleanupCourtesy.Reset(100 * time.Millisecond)
```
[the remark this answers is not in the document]
sharnoff: no longer applies

pkg/agent/globalstate.go
```diff
@@ -132,9 +145,47 @@ type podState struct {
...
+	Runner          *RunnerState  `json:"runner,omitempty"`
+	CollectionError error         `json:"collectionError,omitempty"`
+}
+
+func (p *podState) Dump(ctx context.Context) podStateDump {
+	status := p.status.Dump()
+	runner, collectErr := p.runner.State(ctx)
+	if collectErr != nil {
+		collectErr = fmt.Errorf("error reading runner state: %w", collectErr)
+	}
```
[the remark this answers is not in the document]
sharnoff: > are these blocking on something other than mutexes or is there network requests? (looks like just mutexes?)

just mutexes

> are there multiple runners in one agent process? or is the runner multiplexed among many pods?

multiple runners in one agent process

> Seems like just presence/absence of runner would be easier to handle as a caller.

Fundamentally, the "caller" is meant to be a network request. So we're guaranteeing that there's exactly one of `Runner` or `CollectionError` present, which means it _should_ be easy enough to e.g. filter for runners where collection didn't error (`Runner` field is present) or find all collection errors.

pkg/plugin/dumpstate.go
sharnoff: It's used in the call to `p.dumpState` below, `shutdownCtx.Err() != nil`

pkg/plugin/dumpstate.go
sharnoff: Not sure what the suggestion is. OTOH there's plenty of other places we aren't catching panics, either

pkg/plugin/dumpstate.go
sharnoff: Yeah, this is due to that quirk of `AddHandler` that I mentioned in the PR description. Changing it is on the todo list, but for now `body` has to be a pointer to something. Naming it makes it a little clearer what's going on.

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	return &stateDump{
+		Stopped:   stopped,
+		BuildInfo: util.GetBuildInfo(),
+		State:     *state,
+	}, nil
+}
+
+type keyed[K any, V any] struct {
```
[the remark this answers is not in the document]
sharnoff: This solves the problem of "namespaced names are two fields, which makes them illegal JSON keys". We could equally solve this with e.g. generating map keys from `${namespace}:${name}`, but I want this to be more machine-readable (including `jq`) than human-readable, so keeping namespace and name separate provides a little more flexibility on the reading side.

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+	OtherPods []keyed[api.PodName, otherPodStateDump] `json:"otherPods"`
+	Mq        []*podNameAndPointer                    `json:"mq"`
+}
+
+type podStateDump struct {
+	Obj                      pointerString            `json:"obj"`
+	Name                     api.PodName              `json:"name"`
+	VMName                   string                   `json:"vmName"`
+	Node                     pointerString            `json:"node"`
+	TestingOnlyAlwaysMigrate bool                     `json:"testingOnlyAlwaysMigrate"`
```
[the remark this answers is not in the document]
sharnoff: it's a k8s thing :shrug:

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, err
+	}
+	defer s.lock.Unlock()
+
+	var vmPods []podNameAndPointer
+	for _, p := range s.podMap {
+		vmPods = append(vmPods, podNameAndPointer{Obj: makePointerString(p), PodName: p.name})
+	}
+
+	var otherPods []podNameAndPointer
```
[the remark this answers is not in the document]
sharnoff: done

pkg/agent/globalstate.go
```diff
@@ -132,9 +145,47 @@ type podState struct {
...
 }
 
+type podStateDump struct {
+	PodName         api.PodName   `json:"podName"`
+	Status          podStatusDump `json:"status"`
+	Runner          *RunnerState  `json:"runner,omitempty"`
+	CollectionError error         `json:"collectionError,omitempty"`
+}
+
+func (p *podState) Dump(ctx context.Context) podStateDump {
```
tychoish, line 155: make private, maybe.

pkg/plugin/dumpstate.go
```diff
@@ -0,0 +1,254 @@
...
+
+	mq := make([]*podNameAndPointer, 0, len(s.mq))
+	for _, p := range s.mq {
+		if p == nil {
+			mq = append(mq, nil)
+		} else {
+			v := podNameAndPointer{Obj: makePointerString(p), PodName: p.name}
+			mq = append(mq, &v)
+		}
+	}
```
tychoish, line 203: you probably want to sort these?

tychoish (approved):

pkg/agent/globalstate.go
sharnoff: done

pkg/plugin/dumpstate.go
sharnoff: done

PR #78, opened 2023-03-06
Add e2e tests 

bayandin:
This PR adds e2e tests for autoscaling. Some scripts from `neonvm` are used for environment preparation, so I've added `neonvm` as a submodule (for the time being, until we properly merge these repos).

Test `tests/e2e/autoscaling`:
- Create VM (1cpu, 1Gi)
- Start workload
- Check VM upscaled (to max 4cpu 4Gi)
- Stop workload
- Check VM downscaled (back to 1cpu 1Gi)


.github/workflows/test.yaml
```diff
@@ -28,4 +28,88 @@ jobs:
       - name: Run Go Tests
         if: env.GIT_DIFF
         run: |
-          go list -f "{{ if (or .TestGoFiles .XTestGoFiles) }}{{ .ImportPath }}{{ end }}" ./... | xargs -t go test -v 
+          go list -f "{{ if (or .TestGoFiles .XTestGoFiles) }}{{ .ImportPath }}{{ end }}" ./... | xargs -t go test -v
+
+  e2e:
```
tychoish, line 33: my inclination, and this is somewhat counter to the example in the cloud repo, is to keep different suites/classes of tests in different workflow files just to make things easier to find/trace, particularly if there aren't dependencies between them.

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "1.19"
+
+      - name: Install dependencies
+        run: |
```
tychoish, line 24: I'd love it if (more) of these were runnable outside of CI just to be able to verify changes locally before pushing to CI, or to use in development. (it'd be fine if this were linux only,)

Makefile
```diff
@@ -0,0 +1,53 @@
+.PHONE: vm-kernel
+vm-kernel:
+	cd neonvm && make kernel
```
tychoish, line 3: ```make -C neonvm kernel``` does this without the `cd`

similarly it'd be fun if `neonvm` were a target here that did submodule init or something related.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+.PHONY: kind-load
+kind-load: docker-build
+	kind load docker-image $(AUTOSCALER_SCHEDULER_IMG)
+	kind load docker-image $(AUTOSCALER_AGENT_IMG)
+	kind load docker-image $(VM_INFORMANT_IMG)
+
+.PHONE: deploy
+deploy: kind-load
+
+EXAMPLE_VM_IMG ?= vm-example:dev
```
tychoish, line 44: I'd put all the variables near the top, just for readability (make always has  to parse/run global stuff anyway, so you don't get much by putting things at the bottom.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+AUTOSCALER_SCHEDULER_IMG ?= kube-autoscale-scheduler:dev
+AUTOSCALER_AGENT_IMG ?= autoscaler-agent:dev
+VM_INFORMANT_IMG ?= vm-informant:dev
+
+.PHONY: docker-build
+docker-build:
+	docker buildx build \
+		-t $(AUTOSCALER_SCHEDULER_IMG) \
+		--build-arg "GIT_INFO=n/a" \
+		-f build/autoscale-scheduler/Dockerfile \
```
tychoish, line 22: nit: I'd maybe make all the arguments long form ...

tychoish (commented):
this is great! I'm psyched.

.github/workflows/e2e-test.yaml
bayandin: I've added a rough section to the README with instructions on how to run tests (and listing dependencies).

I plan to work more on Makefile (like conversing script to make targets for deduplication) and then will have a command to install/check dependencies.

Makefile
bayandin: > `make -C neonvm kernel` does this without the `cd`

Nice! Reusing this.

> similarly it'd be fun if neonvm were a target here that did submodule init or something related.

Will do something like this in the following work with Makefile

bayandin:
Ok, the PR is more or less ready for review.
It also includes a removal of local docker registry. I went through the setup mentioned in README, and it still works.

.github/workflows/e2e-test.yaml
tychoish: perfect!

tychoish (approved):

Makefile
```diff
@@ -0,0 +1,44 @@
...
+		--tag $(VM_INFORMANT_IMG) \
+		--build-arg "GIT_INFO=n/a" \
+		--file build/vm-informant/Dockerfile \
+		.
+
+.PHONY: kind-load
+kind-load: docker-build
+	kind load docker-image $(AUTOSCALER_SCHEDULER_IMG)
+	kind load docker-image $(AUTOSCALER_AGENT_IMG)
+	kind load docker-image $(VM_INFORMANT_IMG)
```
sharnoff, line 28:
```suggestion
```

Don't need to load the VM informant image. The image is only created to store the VM informant binary before building the test VM.

Makefile
```diff
@@ -0,0 +1,44 @@
...
+	docker buildx build \
+		--tag $(AUTOSCALER_AGENT_IMG) \
+		--build-arg "GIT_INFO=n/a" \
+		--file build/autoscaler-agent/Dockerfile \
+		.
+	docker buildx build  \
+		--tag $(VM_INFORMANT_IMG) \
+		--build-arg "GIT_INFO=n/a" \
+		--file build/vm-informant/Dockerfile \
+		.
```
sharnoff, line 22: not a blocker, but it would be good to actually include the git info here. If that seems like a pain to sort out, I'm happy to take care of it after merging

scripts/cluster-init.sh
```diff
@@ -27,6 +27,10 @@ kubectl apply -f deploy/neonvm.yaml | indent
 
 kubectl wait deployment -n neonvm-system neonvm-controller --for=condition=Available=True | indent
 
+for img in pg14-disk-test:dev kube-autoscale-scheduler:dev autoscaler-agent:dev; do
```
sharnoff, line 30:
```suggestion
for img in kube-autoscale-scheduler:dev autoscaler-agent:dev; do
```

(note: this requires changing the README to do the appropriate `kind load` there)

The test VM image isn't required for cluster startup, so I'd rather not require it to exist if we don't need to.

That said, [see note about switching away from these scripts].

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -0,0 +1,146 @@
...
+  - name: host-metrics
+    port: 9100
+    protocol: TCP
+    targetPort: host-metrics
+    nodePort: 30100
+  - name: metrics
+    port: 9187
+    protocol: TCP
+    targetPort: metrics
+    nodePort: 30187
```
sharnoff, line 33: Are `pooler` and/or `metrics` actually used anywhere?

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -0,0 +1,146 @@
...
+kind: ConfigMap
+metadata:
+  name: example-config
+data:
+  postgresql.conf: |
+    listen_addresses = '*'
+    shared_preload_libraries = 'pg_stat_statements'
+
+    max_connections = 64
+    shared_buffers = 512MB
```
sharnoff, line 117: 512MB here is actually quite big. We might run into issues with VMs with 1GB of memory. Thoughts?

tests/e2e/autoscaling/01-upscale.yaml
```diff
@@ -0,0 +1,48 @@
...
+      name: my-config
+  restartPolicy: Never
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-config
+data:
+  my-file.txt: |
+    select length(factorial(length(factorial(1223)::text)/2)::text);
```
sharnoff, line 48: filename should be something other than `my-file.txt`

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -0,0 +1,146 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- ../../../deploy/autoscale-scheduler.yaml
+- ../../../deploy/autoscaler-agent.yaml
```
sharnoff, line 5: At first glance, I feel like this should be handled in `make deploy`, not here. Thoughts? Maybe there's something about kuttl that means this makes more sense?

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -0,0 +1,146 @@
...
+        port: 9187
+      - name: informant
+        port: 10301
+  extraNetwork:
+      enable: true
+  disks:
+    - name: pgdata
+      mountPath: /var/lib/postgresql
+      emptyDisk:
+        size: 16Gi
```
sharnoff, line 86: Does this work? I'd be concerned about issues around disk migration

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -0,0 +1,146 @@
...
+kind: TestStep
+apply:
+- ../../../deploy/autoscale-scheduler.yaml
+- ../../../deploy/autoscaler-agent.yaml
+unitTest: false
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: example
```
sharnoff, line 11: Why do we need a service here?

sharnoff (commented):
Looks awesome. Some small items, some questions.

Eventually I'd love to switch away from the scripts directory and only use the makefile — there's some remaining changes required before that happens, though, and I don't think merging this should be blocked by them.

tests/e2e/autoscaling/00-create-vm.yaml
bayandin: I've copied this example from `neonvm`. 
Without service it doesn't work: `example:5432 - no response`, I've removed extra stuff that we don't really need (like a bunch of ports and vector), we could add it back if needed (or for other tests)

tests/e2e/autoscaling/00-create-vm.yaml
bayandin: It works (as in tests for neonvm). Do you think we need to change it to let's wait until something breaks? 😄

tests/e2e/autoscaling/00-create-vm.yaml
bayandin: Reduced to `256MB` just in case.

tests/e2e/autoscaling/00-create-vm.yaml
sharnoff: Ahhh I see - didn't realize it was used for providing a stable address. makes sense

tests/e2e/autoscaling/00-create-vm.yaml
sharnoff: Ah ok, yeah should be fine to leave it - plus easy enough to remove later if things break :sweat_smile:

sharnoff (approved):

PR #79, opened 2023-03-07
testing: add smoke test script

tychoish:
This is clearly rough, and I'll see how much is blows up on a github
runner (probably a lot).

Having said that, if you make a component (e.g. the agent) panic on
boot, it fails, so that's pretty good.


tychoish:
I think #78 is a cleaner solution, but it would probably still be good to have something that was easier to run locally.

PR #83, opened 2023-03-09
deploy: add default billing settings

tychoish:


sharnoff:
as discussed: `BillingConfig` is invalid if provided without `url`. It'd be good to have a test billing server to receive events, but until then I think we should leave billing config out of the default deployment files.

tychoish:
> as discussed: `BillingConfig` is invalid if provided without `url`. It'd be good to have a test billing server to receive events, but until then I think we should leave billing config out of the default deployment files.

I have a variant of this that just has an enabled/disabled flag (which will default to disabled.) I think making things harder/more error prone for ourselves when doing deploys just because we have unscheduled work in the future to make things better, just creates more risk.

PR #84, opened 2023-03-09
billing: collect active time metrics

tychoish:


pkg/agent/billing.go
```diff
@@ -46,6 +47,8 @@ type metricsTimeSlice struct {
 	endTime   time.Time
 }
 
+func (m *metricsTimeSlice) Duration() time.Duration { return m.endTime.Sub(m.startTime) }
```
sharnoff, line 50: nit: Is collapsing functions into a single line like this typical go style? Don't think I've seen it used, and it's a little harder to read, but happy to merge anyways if it's relatively common.

In other languages, I've tended only to do that when it's a simple expression (like field accessing, or forwarding to a different method).

sharnoff (approved):
Looks good, one nit.

pkg/agent/billing.go
tychoish: It's permissable as long as it doesn't run over the line length comment, and I see/use this sometimes? 

If there's a linter that disallows this and you feel strongly we should go that, but otherwise, I don't think it's good to have code formatting convention/rules that aren't automatically enforced.

PR #89, opened 2023-03-15
Custom task management

sharnoff:
## Motivation

Briefly:

1. Move task management away from business logic
2. Better stack traces on panic
3. Allow introspection of which tasks are running
4. Allow mock tests to check we aren't leaking goroutines

Currently `pkg/agent/runner.go` has the `spawnBackgroundWorker` method, which is _ok_, but is very specific to the `Runner` type, and is still somewhat limited in scope — for example, it appropriately triggers `Runner` shutdown on panic, but only captures the stack trace for the single goroutine.

On top of that, I'd like to eventually have some "e2e-ish" mock tests that connect the components outside of kubernetes — e.g. by feeding in fake k8s data + metrics and connecting the components' HTTP servers/clients to each other. Points (3) and (4) above would allow these tests to test more, in addition to providing better debugging capabilities when things go wrong.

## Implementation

The general idea is that we add a new package (`pkg/task`) that primarily exports a `Manager` type, which gets passed around everywhere. Every instance of `go f(...)` is instead replaced by `Manager.Spawn(...)` or `Manager.SpawnAsSubgroup(...)`.

`task.Manager` also does (approximately) the right thing w.r.t. signal handling and shutdown.

The implementation of `pkg/task` is backed by https://github.com/sharnoff/chord, which was designed with this project in mind. We may want to move that into this repo, or under `neondatabase`.

## Remaining items

- [x] Documentation in `pkg/task`
- [ ] Rebasing on `main`
- [x] Testing locally
- [ ] Self review

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -83,6 +83,7 @@ discussed more in the [high-level consequences] section below.
         protocol-relevant types independently used by both.
     * `pkg/informant/` — implementation of the VM informant
     * `pkg/plugin/` — implementation of the scheduler plugin
+    * `pkg/task/` — tools for managing goroutine lifecycle and service shutdown
     * `pkg/util/` — miscellaneous utilities that are too general to be included in `agent` or
       `plugin`.
 * `scripts/` — a collection of scripts for common tasks. Items of note:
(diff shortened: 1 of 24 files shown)


sharnoff:
Closing, this is not planned. _Maybe_ we'd like to resurrect this at some point, because it provides some useful abstractions, but for now there's little need.

PR #91, opened 2023-03-16
Integrate NeonVM into Autoscaling

bayandin:
General comment:
I've cut some corners while merging these projects; the main goal was to get all the parts to play well in the single repo without introducing unnecessary refactoring. Thus some parts look a bit rough and could be improved (for example, targets in Makefile); a difference in the project structure and directory layout is still visible. Nevertheless, I still see the benefits of getting this merged and continuing to work on both NeonVM and Autoscaling in the same repo.

ToDo list mostly for myself:

- [x] Merge repos https://github.com/neondatabase/autoscaling/pull/96
- [x] Fix make targets
- [x] Fix tests
- [x] Fix e2e tests
- [x] Unify .github/workflows
- [x] Fix scripts (for autoscaling)
- [x] sort out docs / READMEs
- [x] restore git history of neonvm https://github.com/neondatabase/autoscaling/pull/96
- [ ] Fix `golangci-lint` for neonvm (can be done in a separate PR)
- [ ] Test release workflow (create `0.5.0-pre` tag?)
- [x] Move tuning `fs.inotify.*` settings from GitHub job to ansible config https://github.com/neondatabase/runner/pull/20


.github/workflows/release.yaml
```diff
@@ -25,10 +34,24 @@ jobs:
...
+      #
+      # This step is analogue to the `make generate` and `make manifests` steps in the Makefile.
+      # Since we commit all generated files, we don't need to run these steps again.
+      #
+      # - name: generate code and k8s CRDs
+      #   run: |
+      #     git clone --branch=v${{ env.CODE_GENERATOR_VERSION }} --depth=1 https://github.com/kubernetes/code-generator.git ${RUNNER_TEMP}/code-generator
+      #     go install sigs.k8s.io/controller-tools/cmd/controller-gen@v${{ env.CONTROLLER_TOOLS_VERSION }}
+      #     ${RUNNER_TEMP}/code-generator/generate-groups.sh "deepcopy,client,informer,lister" github.com/neondatabase/neonvm/client github.com/neondatabase/neonvm/apis neonvm:v1 --go-header-file hack/boilerplate.go.txt
+      #     controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/common/crd/bases output:rbac:artifacts:config=config/common/rbac output:webhook:artifacts:config=config/common/webhook
```
bayandin, line 51: In NeonVM we have this code in [`release.yaml`](https://github.com/neondatabase/neonvm/blob/d9ce98c5d7e949ff25049d9b6dab7f2142184290/.github/workflows/release.yaml#L34-L39), but I far as I understand we commit all autogenerated code to the repo, so this step shouldn't be necessary, am I missing something?
/cc @cicdteam

Makefile
```diff
@@ -1,7 +1,32 @@
...
+IMG_VXLAN ?= vxlan-controller:dev
+VM_EXAMPLE_SOURCE ?= postgres:15-alpine
+VM_EXAMPLE_IMAGE ?= vm-postgres:15-alpine
+
+# Autoscaler related images
 AUTOSCALER_SCHEDULER_IMG ?= kube-autoscale-scheduler:dev
 AUTOSCALER_AGENT_IMG ?= autoscaler-agent:dev
 VM_INFORMANT_IMG ?= vm-informant:dev
 EXAMPLE_VM_IMG ?= vm-example:dev
+PG14_DIST_TEST_IMG ?= pg14-disk-test:dev
```
sharnoff, line 13: Is `DIST` instead of `DISK` a typo?

kind/config.yaml
```diff
@@ -1,34 +1,12 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
-# Using flannel for networking instead of `kindnet`
 networking:
   disableDefaultCNI: true
+  #podSubnet: "192.168.0.0/16"
```
sharnoff, line 5:
```suggestion
```

neonvm/hack/Dockerfile.generate
```diff
@@ -1,5 +1,12 @@
 FROM golang:1.19
 
+ARG USER_ID
+ARG GROUP_ID
+
+RUN addgroup --gid $GROUP_ID user
+RUN adduser --disabled-password --gecos '' --uid $USER_ID --gid $GROUP_ID user
+USER user
```
sharnoff, line 8: This seems funky - why is this needed? (or: can we have a comment explaining the purpose of this?)

I'm particularly wary of this because I have to run docker stuff as root, so this _looks_ like it might mess things up, but I can't really tell either way.

tests/e2e/autoscaling/00-create-vm.yaml
```diff
@@ -12,7 +12,6 @@ spec:
     port: 5432
     protocol: TCP
     targetPort: postgres
-    nodePort: 30432
```
sharnoff, line 15: Is this related?

sharnoff (approved):
Broadly looks good. The comments are mostly little things.

A couple notes:

1. A few of the changes don't seem strictly related to the PR — it'd be good to separate those out, to avoid giving the impression to future `git blame`s that they were related/required here.
2. I noticed a couple references to `vm_image` still around. Aside from the one I commented on, there's also references in `README.md` and `vm-examples/pg14-disk-test/README.md`

I'm a little concerned about some of the `kubectl` usage in the makefile, but AFAICT it's not meaningfully different from what we had before, so I don't think that should block merging. I might open up a PR to improve it sometime down the road.

---

Lastly: It seems like you're working on macos — have you already tested that things work as expected on linux as well? would it be useful for me to do that?

neonvm/hack/Dockerfile.generate
bayandin: I've cloned the repo as a local non-root user. Then if I run `make generate` (which uses this file), it generates files in a container from the root user. And I have issues deleting these files (during `git rebase`, for example, because they don't belong to the current user), so I've changed it to use the current user.

There's a comment regarding this in `Makefile` (for `generate` target), will add it to the Dockerfile as well.
```
# Use uid and gid of current user to avoid mismatched permissions
```

> I'm particularly wary of this because I have to run docker stuff as root, so this looks like it might mess things up, but I can't really tell either way.

I guess it just wouldn't work for your case 😄 (because the root user already exists in a container). I will add a fix for it.

bayandin (commented):
Thanks for the review @sharnoff!


> A few of the changes don't seem strictly related to the PR — it'd be good to separate those out, to avoid giving the impression to future git blames that they were related/required here.

Before merging PR, I'm going to squash commits into a couple of more meaningful ones (like `fix e2e tests`) and rebase on top of the `main` branch (without squashing everything into one commit). Hope this will help to make `git blame` more meaningful.

> I noticed a couple references to vm_image still around. Aside from the one I commented on, there's also references in README.md and vm-examples/pg14-disk-test/README.md

My bad, I will check and update these.

> Also, I think it's best to fix golangci-lint in a separate PR.

Sure! 

> Lastly: It seems like you're working on macos — have you already tested that things work as expected on linux as well? would it be useful for me to do that?

I'm using an ec2 Linux instance for this, `neonvm` can't be compiled yet on macOS because of a difference in netlink API (@cicdteam have some fixes for this in his [neonvm `arm64` branch](https://github.com/neondatabase/neonvm/compare/main...arm64))
But yes, it works on Linux for me

neonvm/controllers/virtualmachine_controller_test.go
```diff
@@ -81,7 +81,7 @@ var _ = Describe("VirtualMachine controller", func() {
 						Name:      VirtualMachineName,
 						Namespace: namespace.Name,
 					},
-					Spec: vmv1.VirtualMachineSpec{},
+					Spec: vmv1.VirtualMachineSpec{QMP: 1, RestartPolicy: "Never"},
```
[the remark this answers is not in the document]
sharnoff: Makes sense, sounds good

neonvm/hack/Dockerfile.generate
sharnoff: Makes sense - validated the fix works :)

tests/e2e/autoscaling/00-create-vm.yaml
sharnoff: Makes sense

sharnoff:
> > A few of the changes don't seem strictly related to the PR — it'd be good to separate those out, to avoid giving the impression to future git blames that they were related/required here.
> 
> Before merging PR, I'm going to squash commits into a couple of more meaningful ones (like fix e2e tests) and rebase on top of the main branch (without squashing everything into one commit). Hope this will help to make git blame more meaningful.

Sounds good! Yeah, seems like most of what I thought was unrelated is more ambiguous - my bad.

> I'm using an ec2 Linux instance for this, neonvm can't be compiled yet on macOS because of a difference in netlink API

Ah shoot, interesting. Out of curiosity, would compiling neonvm in containers (like what we do for the scheduler & autoscaler-agent) also fix that?

.github/workflows/release.yaml
cicdteam: seems yes, we can skip code generating in GA

cicdteam (approved):

bayandin:
> Ah shoot, interesting. Out of curiosity, would compiling neonvm in containers (like what we do for the scheduler & autoscaler-agent) also fix that?

Yep, it compiles (but we don't have arm64 kernel for `neonvm` yet, so can't say if it really works). I'd like to move the building of everything to containers, then we can improve caches (probably) and not require golang on host machine

cicdteam:
> Yep, it compiles (but we don't have arm64 kernel for `neonvm` yet, so can't say if it really works). I'd like to move the building of everything to containers, then we can improve caches (probably) and not require golang on host machine

Yes, netlink has no issue when cimplining in container (because there us Linux, not Darwin). I able run neonvm on my Apple M1 macbook (with `hvm` acceleration ever). I will add arm64 support after we finish moving neonvm into autiscaling repo.

PR #92, opened 2023-03-20
agent, plugin: Add prometheus metrics

sharnoff:
Some of the HTTP-related metrics are kind of funky right now, because adding the appropriate middleware is just a little bit trickier than I'd like.

I've checked it briefly, seems to be working as expected. I'm not expecting this to be anywhere close to the desired state of affairs — just wanted to put this here as a starting place.

## Implementation

Exposes prometheus metrics endpoints at `http://0.0.0.0:9100/metrics` in both the autoscaler-agent and scheduler plugin, with some custom metrics, plus go runtime and process metrics.

The collectors are set up in the `pkg/*/prommetrics.go` files; refer there for details.


pkg/plugin/run.go
```diff
@@ -55,7 +63,11 @@ func (e *AutoscaleEnforcer) startPermitHandler(ctx context.Context) error {
 			"[autoscale-enforcer] Received autoscaler-agent request (client = %s) %+v",
 			r.RemoteAddr, req,
 		)
-		resp, statusCode, err := e.handleAgentRequest(req)
+		nodeName := "<none>"
+		resp, statusCode, err := e.handleAgentRequest(req, &nodeName)
+		finalStatus = statusCode
+		e.metrics.validResourceRequests.WithLabelValues(nodeName, req.ProtoVersion.String(), strconv.FormatBool(req.Metrics != nil)).Inc()
```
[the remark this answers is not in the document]
sharnoff: That's fair. I don't know enough to know how concerned we should be about that, but node names seems like it would be really useful for figuring out what's going wrong.

pkg/agent/informant.go
```diff
@@ -625,7 +633,11 @@ func (s *InformantServer) informantURL(path string) string {
 // of that context.
 //
 // Returns: response body (if successful), status code, error (if unsuccessful)
-func (s *InformantServer) handleID(ctx context.Context, body *struct{}) (*api.AgentIdentificationMessage, int, error) {
+func (s *InformantServer) handleID(ctx context.Context, body *struct{}) (_ *api.AgentIdentificationMessage, code int, _ error) {
+	defer func() {
+		s.runner.global.metrics.informantRequestsInbound.WithLabelValues("/id", strconv.Itoa(code)).Inc()
```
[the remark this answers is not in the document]
sharnoff: I thought the control plane had leading slashes, so I put those here. I may be wrong though?

pkg/plugin/prommetrics.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/klog/v2"
+)
+
+type PromMetrics struct {
+	pluginCalls           *prometheus.CounterVec
+	resourceRequests      *prometheus.CounterVec
+	validResourceRequests *prometheus.CounterVec
+}
+
+func (p *AutoscaleEnforcer) startPrometheusServer(ctx context.Context) error {
```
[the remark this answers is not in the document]
sharnoff: Fixed

pkg/agent/globalstate.go
```diff
@@ -52,6 +65,9 @@ func podIsOurResponsibility(pod *corev1.Pod, config *Config, nodeName string) bo
 }
 
 func (s *agentState) Stop() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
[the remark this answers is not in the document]
sharnoff: this got added in `main`; rebasing removed this from the diff

pkg/agent/informant.go
```diff
@@ -555,6 +556,11 @@ func doInformantRequest[Q any, R any](
 	path string,
 	reqData *Q,
 ) (_ *R, statusCode int, _ error) {
+	result := "<internal error>"
```
tychoish, line 559: I think I might just key them on status code (potentially just rounded down to the nearest 100?) rather than these strings, but it's fine

pkg/agent/prommetrics.go
```diff
@@ -0,0 +1,104 @@
...
+	"k8s.io/klog/v2"
+)
+
+type PromMetrics struct {
+	schedulerRequests         *prometheus.CounterVec
+	informantRequestsOutbound *prometheus.CounterVec
+	informantRequestsInbound  *prometheus.CounterVec
+}
+
+func startPrometheusServer(ctx context.Context, globalstate *agentState) (PromMetrics, error) {
```
tychoish, line 23: I kind of tend to prefer returning pointers for things like this, even though the double indirection is admittedly annoying.

pkg/agent/prommetrics.go
```diff
@@ -0,0 +1,104 @@
...
+	// Separate binding from serving, so that we can catch any error in this thread, rather than the
+	// server's.
+	addr := net.TCPAddr{IP: net.IPv4zero, Port: 9100}
+	listener, err := net.ListenTCP("tcp", &addr)
```
tychoish, line 27: nit: combine these into one lines

pkg/agent/prommetrics.go
```diff
@@ -0,0 +1,104 @@
...
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+		schedulerRequests,
+		informantRequestsOutbound,
+		informantRequestsInbound,
+		totalVMs,
+	)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+	srv := &http.Server{Handler: mux}
```
tychoish, line 78: I've recently learned that without setting BaseContext, the handlers don't get a "good" context. Probably doesn't matter much, but might be nice.

pkg/agent/prommetrics.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}()
+
+	go func() {
+		// shutdown the shutdown watcher if we exit before it
+		defer shutdown()
+		if err := srv.Serve(listener); !errors.Is(err, http.ErrServerClosed) {
+			klog.Errorf("Prometheus server exited with unexpected error: %s", err)
+		}
+	}()
```
tychoish, line 97: I believe we're only ever shutting down the server after it's already closed, (the `shutdown()` call) or in the case of the parent context being canceled there's nothing that will block the process on waiting for everything to shut down, which means the process is almost certainly going to have exited before the http service cleanly shuts down. (Having thought too much about this in the past/for the fun/srv.Orchestrator stuff, might be more trouble than it's worth, happy to fuss with it later too. )

pkg/plugin/prommetrics.go
```diff
@@ -0,0 +1,91 @@
...
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(
+		collectors.NewGoCollector(),
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+		p.metrics.pluginCalls,
+		p.metrics.resourceRequests,
+		p.metrics.validResourceRequests,
+	)
+
+	mux := http.NewServeMux()
```
tychoish, line 67: can we refactor this so that we just hand a prom registry and then do the server on its own?

pkg/plugin/run.go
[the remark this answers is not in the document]
tychoish: I think I'd lean toward pulling the protocol version (it'll be disvoverable in a pinch,) and it seems like most of the time it'd be fine (but also useless) *except* if we had to deploy a few times to fix bugs, (e.g. in some kind of fire-fighting situation) where we'd run the risk of DoS'ing the metrics db, just when we need it the most?

pkg/plugin/run.go
```diff
@@ -97,7 +109,7 @@ func (e *AutoscaleEnforcer) startPermitHandler(ctx context.Context) error {
 }
 
 // Returns body (if successful), status code, error (if unsuccessful)
-func (e *AutoscaleEnforcer) handleAgentRequest(req api.AgentRequest) (*api.PluginResponse, int, error) {
+func (e *AutoscaleEnforcer) handleAgentRequest(req api.AgentRequest, nodeName *string) (*api.PluginResponse, int, error) {
```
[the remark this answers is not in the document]
tychoish: Either of these options seem preferable (and fine) to me.

tychoish (approved):

pkg/plugin/run.go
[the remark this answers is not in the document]
sharnoff: Makes sense. Went with the second.

pkg/plugin/run.go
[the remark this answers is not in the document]
sharnoff: Makes sense, done.

pkg/agent/prommetrics.go
sharnoff: euhhhh, yeah, I think this is probably too much of a pain to resolve now, because I do want to have binding vs listening in separate threads. (and realistically, it's _probably fine_ that we don't wait).

I do agree this is a problem, I think a "best of both worlds" solution is probably _easy enough_ in #89?

pkg/agent/prommetrics.go
sharnoff: Done

pkg/plugin/prommetrics.go
sharnoff: Done, moved the server part to pkg/util, made the plugin/agent versions of this function return the registry.

pkg/agent/prommetrics.go
```diff
@@ -20,15 +11,7 @@ type PromMetrics struct {
...
-func startPrometheusServer(ctx context.Context, globalstate *agentState) (PromMetrics, error) {
-	// Separate binding from serving, so that we can catch any error in this thread, rather than the
-	// server's.
-	addr := net.TCPAddr{IP: net.IPv4zero, Port: 9100}
-	listener, err := net.ListenTCP("tcp", &addr)
-	if err != nil {
-		return PromMetrics{}, fmt.Errorf("Error listening on TCP: %w", err)
-	}
-
+func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Registry) {
```
tychoish, line 14: why not put the registry in the `PromMetrics`

pkg/util/promserver.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"k8s.io/klog/v2"
+)
+
+// Starts the prometheus server in a background thread. Returns error if binding on the port fails.
+func StartPrometheusMetricsServer(ctx context.Context, port uint16, reg *prometheus.Registry) error {
```
tychoish, line 19: should something block until this server returns?

pkg/plugin/run.go
```diff
@@ -109,7 +107,17 @@ func (e *AutoscaleEnforcer) startPermitHandler(ctx context.Context) error {
...
 
 // Returns body (if successful), status code, error (if unsuccessful)
-func (e *AutoscaleEnforcer) handleAgentRequest(req api.AgentRequest, nodeName *string) (*api.PluginResponse, int, error) {
+func (e *AutoscaleEnforcer) handleAgentRequest(
+	req api.AgentRequest,
+) (_ *api.PluginResponse, status int, _ error) {
+	nodeName := "<none>" // override this later if we have a node name
+	defer func() {
+		hasMetrics := req.Metrics != nil
+		e.metrics.validResourceRequests.
```
tychoish, line 116: shouldn't we not log if we don't have metrics? 

What, operationally, do we learn from being able to distinguish metrics-ey requests from not-metricsy requests?

(I believe that the size of the metrics payload is (roughly) constant with regards to the number of timeseries (metrics*labels) even if storage is delta and zero run length encoded.)

pkg/agent/globalstate.go
```diff
@@ -38,11 +39,10 @@ type agentState struct {
 }
 
 func (r MainRunner) newAgentState(
-	ctx context.Context,
 	podIP string,
 	broker *pubsub.Broker[watchEvent],
 	schedulerStore *util.WatchStore[corev1.Pod],
-) (*agentState, error) {
+) (*agentState, *prometheus.Registry) {
```
tychoish, line 45: I might pass the registry in rather than creating it here.

sharnoff:
There's some unresolved comments, but I'm going to merge anyways. The gist is that the intent behind certain files/structs is more cleanly handled by things as they are now, but I don't currently have the words to explain that clearly.

PR #110, opened 2023-03-31
agent/billing: panic if VM store unexpectedly stopped

sharnoff:
This should act as a reasonable-ish fail-safe against the VM store ending early, which would otherwise cause us to continually send the same consumption metrics (and would be incorrect!)

This is an added concern because of #111, which actually does something in the VM store's handlers.

### DIFF
--- a/pkg/agent/billing.go
+++ b/pkg/agent/billing.go
@@ -97,7 +97,7 @@ func RunBillingMetricsCollector(
 		select {
 		case <-collectTicker.C:
 			klog.Infof("Collecting billing state")
-			if store.Stopped() && backgroundCtx.Err() != nil {
+			if store.Stopped() && backgroundCtx.Err() == nil {
 				panic(errors.New("VM store stopped but background context is still live"))
 			}
 			state.collect(conf, targetNode, store)
--- a/pkg/agent/billing.go
+++ b/pkg/agent/billing.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"math"
 	"net/http"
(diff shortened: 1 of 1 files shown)


pkg/agent/billing.go
```diff
@@ -96,6 +97,9 @@ func RunBillingMetricsCollector(
 		select {
 		case <-collectTicker.C:
 			klog.Infof("Collecting billing state")
+			if store.Stopped() && backgroundCtx.Err() != nil {
+				panic(errors.New("VM store stopped but background context is still live"))
+			}
```
tychoish, line 102: if the background context is canceled then everything's trying to shut down anyway, I think logging an error and returning is fine (and panic if the background context isn't canceled?) This seems like a recipe for bringing the process down in the middle of an orderly shutdown.

tychoish:
Is this shutdown bug something that we've seen?

sharnoff:
Not to my knowledge. I figured it'd be a good fail-safe against pushing stale consumption data if the VM watcher fails.

Although I don't know if that's worth worrying about, especially considering that a handler taking a long time to run would also affect it and we don't have a way to detect that. I don't know. Thoughts?

pkg/agent/billing.go
sharnoff: Oh, this was actually meant to be `backgroundCtx.Err() == nil`, so exactly what you said with "panic if the background context isn't canceled".

sharnoff: fixed

sharnoff:
Merging, as discussed

PR #111, opened 2023-03-31
agent: watch VMs for events, not pods

sharnoff:
This change mostly does nothing, but it has a few important impliciations that are worth listing:

1. vmIsOurResponsibility checks vm.Status.Phase == VmRunning, which means that we won't scale VMs that are currently migrating.
2. NeonVM can't do a FieldSelector on Status.NodeName, so we have to do the filtering ourselves. That was already the case with the billing-related code though, so :shrug:
3. A handful of functions deleted, because they aren't necessary if we're watching VMs directly :D

Required by #112.

### DIFF
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -84,8 +84,7 @@ type Runner struct {
 	// logger is the shared logger for this Runner, giving all log lines a unique, relevant prefix
 	logger RunnerLogger
 
-	// vm is non-nil for the duration of all background tasks, but is expected to be nil before
-	// r.getInitialVMInfo is called by (*Runner).Run
+	// vm stores some common information about the VM.
 	//
 	// FIXME: the value of vm.Using() is sometimes inaccurate - we need to prematurely mark
 	// resources as being used so that a new scheduler won't send a Register request at the same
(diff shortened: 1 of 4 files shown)


pkg/agent/watch.go
```diff
@@ -129,6 +51,41 @@ func startVMWatcher(
...
+				}
+			},
+		},
 	)
 }
+
+func makeVMEvent(vm *vmapi.VirtualMachine, kind vmEventKind) vmEvent {
+	info, err := api.ExtractVmInfo(vm)
+	if err != nil {
+		panic(fmt.Errorf("unexpected failure while extracting VM info from %s:%s: %w", vm.Namespace, vm.Name, err))
```
tychoish, line 82: `s/unexpected failure//`

pkg/agent/runner.go
```diff
@@ -1360,7 +1316,7 @@ func (r *Runner) getStateForVMUpdate(updateReason VMUpdateReason) *atomicUpdateS
 	return &atomicUpdateState{
 		computeUnit:      *r.computeUnit,
 		metrics:          *r.lastMetrics,
-		vm:               *r.vm,
+		vm:               r.vm,
```
tychoish, line 1319: 🥳

tychoish (approved):

pkg/agent/watch.go
sharnoff: Resolved by making this not a panic; it shouldn't have been one.

Issue #104, opened 2023-02-21
Overlay network


cicdteam:
## Access to test EKS cluster with NeonVM Overlay Network

```console
aws --profile dev sso login
aws --profile dev  eks --region us-east-2 update-kubeconfig --name dev-us-east-2-andrey --alias dev-us-east-2-andrey
```

### vm-1

```yaml
apiVersion: vm.neon.tech/v1
kind: VirtualMachine
metadata:
  name: vm-1
spec:
  terminationGracePeriodSeconds: 1
  nodeSelector:
    node: neon1
  guest:
    cpus:
      min: 1
    memorySlotSize: 1Gi
    memorySlots:
      min: 1
    rootDisk:
      image: andreyneon/vm-nginx:alpine
    ports:
      - name: http
        port: 80
  extraNetwork:
      enable: true
```

### vm-2

```yaml
apiVersion: vm.neon.tech/v1
kind: VirtualMachine
metadata:
  name: vm-2
spec:
  terminationGracePeriodSeconds: 1
  nodeSelector:
    node: neon2
  guest:
    cpus:
      min: 1
    memorySlotSize: 1Gi
    memorySlots:
      min: 1
    rootDisk:
      image: andreyneon/vm-nginx:alpine
    ports:
      - name: http
        port: 80
  extraNetwork:
      enable: true
```

### pod (just pod)

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: pod
spec:
  terminationGracePeriodSeconds: 1
  nodeSelector:
    node: neon3
  containers:
  - name: pod
    image: nginx:alpine
    imagePullPolicy: IfNotPresent
    ports:
      - name: http
        containerPort: 80
    securityContext:
      capabilities:
        add:
        - NET_RAW
        - NET_ADMIN
```

## Access to EKS nodes

node where `vm-1` running

```console
aws --profile dev --region us-east-2 ssm start-session --target i-04c441ec6d78200dc 
```

node where `vm-2` running

```console
aws --profile dev --region us-east-2 ssm start-session --target i-0d532401c1ede5ac5
```

node where test `pod` running

```console
aws --profile dev --region us-east-2 ssm start-session --target i-053b0d713ebec6aed
```

## Network structure

See attached picture

- each node has VXLAN configured:
  - linux bridge interface `neon-br0` with IP from Overlay CIDR `10.100.0.0/20`, linked to `neon-vxlan0` interface
[...]

PR #112, opened 2023-03-31
agent, plugin: Re-enable migration under load

sharnoff:
In order to properly handle migration, agent.Runner now has a way to shut itself down to prevent future requests.

### DIFF
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -84,6 +84,9 @@ type Runner struct {
 	// logger is the shared logger for this Runner, giving all log lines a unique, relevant prefix
 	logger RunnerLogger
 
+	// shutdown provides a clean way to trigger all background Runner threads to shut down
+	shutdown context.CancelFunc
+
 	// vm stores some common information about the VM.
 	//
 	// FIXME: the value of vm.Using() is sometimes inaccurate - we need to prematurely mark
@@ -98,6 +101,14 @@ type Runner struct {
 	vm      api.VmInfo
 	podName api.PodName
 	podIP   string
+
+	// schedulerRespondedWithMigration is true iff the scheduler has returned an api.PluginResponse
+	// indicating that it was prompted to start migrating the VM.
+	//
+	// This field MUST NOT be updated without holding BOTH lock and scheduler.requestLock.
+	//
+	// This field MAY be read while holding EITHER lock or scheduler.requestLock.
+	schedulerRespondedWithMigration bool
 
 	// lock guards the values of all mutable fields. The immutable fields are:
 	// - global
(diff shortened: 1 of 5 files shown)


pkg/agent/runner.go
```diff
@@ -356,8 +371,7 @@ func (r *Runner) setStatus(with func(*podStatus)) {
 
 // Run is the main entrypoint to the long-running per-VM pod tasks
 func (r *Runner) Run(ctx context.Context) error {
-	ctx, cancelCtx := context.WithCancel(ctx)
-	defer cancelCtx()
+	defer r.shutdown()
```
tychoish, line 374: I would just do `ctx, r.shutdown = context.WithCancel(ctx); defer r.shutdown()`  so that the context passed to run is definitely canceled.

pkg/plugin/run.go
```diff
@@ -130,11 +130,12 @@ func (e *AutoscaleEnforcer) handleAgentRequest(req api.AgentRequest) (*api.Plugi
...
 
-	// Check whether the pod *will* migrate, then update its resources, and THEN start its
-	// migration, using the possibly-changed resources.
-	mustMigrate := e.updateMetricsAndCheckMustMigrate(pod, node, req.Metrics)
-	// Don't migrate if it's disabled
-	mustMigrate = mustMigrate && e.state.conf.migrationEnabled()
+	mustMigrate := pod.migrationState == nil &&
+		// Check whether the pod *will* migrate, then update its resources, and THEN start its
+		// migration, using the possibly-changed resources.
+		e.updateMetricsAndCheckMustMigrate(pod, node, req.Metrics) &&
```
tychoish, line 136: I was going to comment "it would be cool collect this into a method with a more descriptive name, but this method looks like it has side effects (based on the name) which almost raises more questions, and probably makes this more fragile (or fragile seeming). 

Anyway, I think it's fine,

pkg/plugin/state.go
```diff
@@ -677,8 +678,28 @@ func (s *pluginState) startMigration(ctx context.Context, pod *podState, vmClien
...
+		Spec: vmapi.VirtualMachineMigrationSpec{
+			VmName: pod.vmName,
+		},
+	}
+	klog.Infof("[autoscale-enforcer] VM Migration create request for VM %s:%s", pod.name.Namespace, pod.vmName)
+	_, err := vmClient.NeonvmV1().
+		VirtualMachineMigrations(pod.name.Namespace).
+		Create(ctx, vmm, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("Error executing VM Migration create request: %w", err)
```
tychoish, line 699: I have a personal grudge against putting the word "error" in error messages.

tychoish (approved):

pkg/agent/runner.go

pkg/plugin/run.go
sharnoff: :shrug:

pkg/plugin/state.go
sharnoff: Yeah... Style difference. This fits with existing norms of the codebase, but those norms can be changed.

sharnoff:
Merging, but changing the title first. Technically this PR doesn't actually _enable_ migration; it just reimplements support for it. Actually turning it back on it will require https://github.com/neondatabase/autoscaling/issues/104; opening a separate PR to handle that.

PR #113, opened 2023-03-31
agent: Use single request lock for plugin & NeonVM

sharnoff:
This PR also does all the implied simplification from this, mainly:
    
- Remove vmStateLock
- Merge schedulerInternal into Scheduler (was only used for deadlock
  checker, which is no longer needed)
- Hold requestLock for all of updateVMResources
- Simplify internals of updateVMResources in cases where holding
  requestLock guarantees the state hasn't changed.
  - Remove retries from updateVMResources (no longer used)
- Don't pre-add to Runner's vm.{Cpu,Mem}.Use - no longer required,
  because scheduler requests can no longer co-occur with NeonVM requests

**Note:** For ease of review (and maybe future blame?), this PR is split into two commits. The first one _only touches whitespace._ Not yet sure if I should merge without squashing.

### DIFF
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -29,8 +29,9 @@ package agent
 //     * ... and a few more.
 //  * Each thread makes *synchronous* HTTP requests while holding the necessary lock to prevent any other
 //    thread from making HTTP requests to the same entity. For example:
-//    * Runner has Runner.vmStateLock that guards the abstract flow of "updating VM resources if we
-//      need to, according to current metrics".
+//    * All requests to NeonVM and the scheduler plugin are guarded by Runner.requestLock, which
+//      guarantees that we aren't simultaneously telling the scheduler one thing and changing it at
+//      the same time.
 //  * Each "background" thread is spawned by (*Runner).spawnBackgroundWorker(), which appropriately
 //    catches panics and signals the Runner so that the main thread from (*Runner).Run() cleanly
 //    shuts everything down.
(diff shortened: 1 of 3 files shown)


pkg/agent/runner.go
```diff
@@ -89,50 +89,36 @@ type Runner struct {
...
-	// FIXME: the value of vm.Using() is sometimes inaccurate - we need to prematurely mark
-	// resources as being used so that a new scheduler won't send a Register request at the same
-	// time as we're updating the Informant/NeonVM with an increase that was already approved. This
-	// is also tied to handling migration; there's a comment specifically about migration that
-	// should handle this.
-	//
-	// This field MUST NOT be updated without holding BOTH lock AND vmStateLock.
-	//
-	// This field MAY be read while holding EITHER lock OR vmStateLock.
+	// This field MUST NOT be read or updated without holding lock.
```
tychoish, line 92: which lock? 

There's a pattern I got off of someone last year, where you put the mutex directly above all the fields that it's supposed to protect to indicate this.

tychoish (approved):

pkg/agent/runner.go
sharnoff: > which lock?

There's a field named `lock`... perhaps that name should be changed :sweat_smile:

> There's a pattern I got off of someone last year, where you put the mutex directly above all the fields that it's supposed to protect to indicate this.

Yeah, I'm definitely a fan of that approach as well. The `Runner` struct _in general_ is kind of a lot, and I don't think reorganizing the fields would help much.

sharnoff:
Rebased & will merge once CI's happy. Want to keep the intermediate commit so we ignore it in blames.

PR #114, opened 2023-03-31
agent/informant: Don't return from /suspend until NeonVM requests finished

sharnoff:
Without this, it's possible for the informant to allow multiple active agents at the same time - something we generally want to avoid

Note: this builds on #113.

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -751,16 +751,18 @@ func (s *InformantServer) handleSuspend(ctx context.Context, body *api.SuspendAg
 		return nil, 400, errors.New("Cannot suspend agent that is not yet registered")
 	}
 
-	// Acquire s.runner.requestLock so that when we return, we can guarantee that
 	locked = false
 	s.runner.lock.Unlock()
 
+	// Acquire s.runner.requestLock so that when we return, we can guarantee that any future
+	// requests to NeonVM or the scheduler will first observe that the informant is suspended and
+	// exit early, before actually making the request.
 	if err := s.runner.requestLock.TryLock(ctx); err != nil {
 		err = fmt.Errorf("Context expired while trying to acquire requestLock: %w", err)
 		s.runner.logger.Errorf("%s", err)
 		return nil, 500, err
 	}
-	s.runner.requestLock.Unlock() // dont' actually hold the lock, we just wanted to wait
+	s.runner.requestLock.Unlock() // don't actually hold the lock, we're just using it as a barrier.
 
 	return &api.AgentIdentificationMessage{
 		Data:           api.AgentIdentification{AgentID: s.desc.AgentID},
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -710,7 +710,12 @@ func (s *InformantServer) handleSuspend(ctx context.Context, body *api.SuspendAg
 	}
 
 	s.runner.lock.Lock()
-	defer s.runner.lock.Unlock()
+	locked := true
+	defer func() {
+		if locked {
+			s.runner.lock.Unlock()
+		}
+	}()
 
 	if s.exitStatus != nil {
 		return nil, 404, errors.New("Server has already exited")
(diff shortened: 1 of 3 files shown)


pkg/agent/informant.go
```diff
@@ -746,6 +751,17 @@ func (s *InformantServer) handleSuspend(ctx context.Context, body *api.SuspendAg
...
+	// Acquire s.runner.requestLock so that when we return, we can guarantee that
+	locked = false
+	s.runner.lock.Unlock()
+
+	if err := s.runner.requestLock.TryLock(ctx); err != nil {
+		err = fmt.Errorf("Context expired while trying to acquire requestLock: %w", err)
+		s.runner.logger.Errorf("%s", err)
+		return nil, 500, err
+	}
+	s.runner.requestLock.Unlock() // dont' actually hold the lock, we just wanted to wait
```
tychoish, line 763: for what?

pkg/agent/informant.go
```diff
@@ -746,6 +751,17 @@ func (s *InformantServer) handleSuspend(ctx context.Context, body *api.SuspendAg
 		return nil, 400, errors.New("Cannot suspend agent that is not yet registered")
 	}
 
+	// Acquire s.runner.requestLock so that when we return, we can guarantee that
+	locked = false
+	s.runner.lock.Unlock()
```
tychoish, line 756: having two different locks within one runner struct is definitely a bit suspicious and difficult to get my head around. (and I think is one of the patterns that we would want to get rid of in #34. 

The comment doesn't help as much as it might?

pkg/agent/runner.go
```diff
@@ -1314,6 +1309,21 @@ func (r *Runner) doVMUpdate(
...
 	}
 
+	invalidInformantErr := func() error {
+		r.lock.Lock()
+		defer r.lock.Unlock()
+
+		if r.server == nil {
+			return errors.New("no informant server set")
+		}
+		return r.server.Valid()
```
tychoish, line 1319: nit: I'd make a method on `r` that did all of this rather than defining it in a function here

pkg/informant/consts.go
```diff
@@ -16,7 +16,7 @@ const (
 	AgentBackgroundCheckTimeout time.Duration = 250 * time.Millisecond
 
 	AgentResumeTimeout  time.Duration = 100 * time.Millisecond
-	AgentSuspendTimeout time.Duration = 200 * time.Millisecond
+	AgentSuspendTimeout time.Duration = 5 * time.Second        // may take a while; it /suspend intentionally waits
```
tychoish, line 19: why 5 seconds? 

[I read this code a few times before I realized what's happening is you make a timeout longer (because a thing takes longer in practice than we initially thought) and then fiddle with the locks so that you (maybe?) get a method to block until another operation returns. It might be clearer (and actually reflect your intent) to have there be a wait-group-ish barrier rather than a mutex barrier for the operation in question. Having said that, this is probably fine.]

pkg/agent/informant.go
```diff
@@ -746,6 +751,17 @@ func (s *InformantServer) handleSuspend(ctx context.Context, body *api.SuspendAg
...
+
+	if err := s.runner.requestLock.TryLock(ctx); err != nil {
+		err = fmt.Errorf("Context expired while trying to acquire requestLock: %w", err)
+		s.runner.logger.Errorf("%s", err)
+		return nil, 500, err
```
tychoish, line 761: this is effectively a timeout (or maybe with a bit of extra formatting it could be?) is 500 the right code?

pkg/agent/runner.go

sharnoff:
I updated some comments to be more clear. I _think_ that should resolve most of the questions?

pkg/agent/informant.go
sharnoff: > this is effectively a timeout

yep

> is 500 the right code?

I don't know. My motivating thought was that we have a lot of request timeouts for things that happen _while_ holding `requestLock`, so waiting for longer than expected probably means that something went wrong in the autoscaler-agent.

And if the request gets cancelled early and it's not the autoscaler-agent's fault, then the VM informant won't receive the 500 anyways.

tychoish: I think it's fine. This is more of a server-side timeout, so probably clients will get it. 

The thing, perhaps unnecessarily that I'm questioning is "if this is a 500 and we start keeping metrics on number of 500s" does this become noise?

tychoish (approved):

pkg/agent/informant.go
sharnoff: That's a fair concern. My thinking is that _generally_ we're expecting not to violate client-side timeouts, and if we do, that's primarily on the autoscaler-agent, not the vm-informant. So even if 500 isn't _technically_ the correct code from a metrics perspective, the _vibe_ of "which component messed up" is correct.

That said, my reasoning may be faulty.

PR #122, opened 2023-03-31
build: Use `git describe` instead of custom git_info

### LINKED ISSUE #16: build: commit messages are hard to "properly" escape for ldflags
## Background

Currently our dockerfiles use ldflags to embed git information into the built images — e.g.:

https://github.com/neondatabase/autoscaling/blob/59fce39dc15ec1ab2d3a4ac387aa1e6a4a5cf334/build/autoscaler-agent/Dockerfile#L15-L21

However, there's quoting issues with this approach:

https://github.com/neondatabase/autoscaling/blob/59fce39dc15ec1ab2d3a4ac387aa1e6a4a5cf334/scripts-common.sh#L38-L41

It can't be solved by adding extra backslashes, unfortunately.

This actually caused a release failure on 30c8bc9755e7caa284bbdd5714bf708098d71471, fixed in 3387ae4a0ce5cc2345000484a9a5e43a60f2548c.

## Desired outcome

There's a couple options here. Either we switch away from ldflags and use go:embed (e.g., as in https://levelup.gitconnected.com/a-better-way-than-ldflags-to-add-a-build-version-to-your-go-binaries-2258ce419d2d) or we use cleaner escaping (e.g., base64).

I think the base64 approach _may_ ultimately be a nicer solution here, but it's a little funky to deal with (and may run into cross-platform issues w.r.t. behavior of the `base64` command). However, the big upside is that it's easier to respect git worktrees with ldflags vs. go:generate (or whatever we'd use to make the embedded files for go:embed).

sharnoff:
The number of issues we ran into with the previous approach made it not worthwhile. We may eventually provide more git information at build time, but it's best to keep it simple for the time being.

Resolves #16.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-GIT_INFO := $(shell source ./scripts-common.sh && git_info)
+GIT_INFO := $(git describe --long --dirty)
 
 .PHONY: all
 all: build
--- a/pkg/util/buildinfo.go
+++ b/pkg/util/buildinfo.go
@@ -7,8 +7,9 @@ import (
 )
 
 // BuildGitInfo stores some pretty-formatted information about the repository and working tree at
-// build time. It's set by the GIT_INFO argument in the Dockerfiles and generated with the git_info
-// function in 'scripts-common.sh'.
+// build time. It's set by the GIT_INFO argument in the Dockerfiles and set to the output of:
+//
+//	git describe --long --dirty
 //
 // While public, this value is not expected to be used externally. You should use GetBuildInfo
 // instead.
--- a/scripts-common.sh
+++ b/scripts-common.sh
@@ -17,30 +17,6 @@ require_root () {
     fi
 }
 
-# Usage: GIT_INFO="$(git_info)"
-#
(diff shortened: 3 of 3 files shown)


tychoish (approved):
is the e2e failure known?

sharnoff:
> is the e2e failure known?

No, but hopefully a transient failure:

```
#12 [builder  6/11] RUN go mod download
#12 361.8 go: cloud.google.com/go/compute@v1.14.0: Get "https://proxy.golang.org/cloud.google.com/go/compute/@v/v1.14.0.mod": dial tcp 142.250.184.241:443: i/o timeout
#12 ERROR: process "/bin/sh -c go mod download" did not complete successfully: exit code: 1
------
 > [builder  6/11] RUN go mod download:
#12 361.8 go: cloud.google.com/go/compute@v1.14.0: Get "https://proxy.golang.org/cloud.google.com/go/compute/@v/v1.14.0.mod": dial tcp 142.250.184.241:443: i/o timeout
------
Dockerfile:8
--------------------
   6 |     COPY go.mod go.mod
   7 |     COPY go.sum go.sum
   8 | >>> RUN go mod download
   9 |     
  10 |     COPY pkg/ pkg/
--------------------
ERROR: failed to solve: process "/bin/sh -c go mod download" did not complete successfully: exit code: 1
make: *** [Makefile:161: vm-informant] Error 1
Error: Process completed with exit code 2.
```

Issue #120, opened 2023-03-31
e2e migration test is flaky


sharnoff:
Had [a CI failure] on #110 that didn't make sense, and re-running didn't fail. Not sure if this was due to external conditions or a bug. AFAICT this was the responsible log line (in the NeonVM controller):

> {"level":"error","ts":"2023-03-31T06:40:22.445Z","logger":"controller.virtualmachinemigration","msg":"Failed stop hypervisor in source runner pod","reconciler group":"vm.neon.tech","reconciler kind":"VirtualMachineMigration","name":"example","namespace":"kuttl-test-expert-weasel","error":"read tcp 10.244.162.130:45076->10.244.162.135:20183: read: connection reset by peer","stacktrace":"github.com/neondatabase/autoscaling/neonvm/controllers.(*VirtualMachineMigrationReconciler).Reconcile\n\t/workspace/neonvm/controllers/virtualmachinemigration_controller.go:186\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:114\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:311\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227"}

https://github.com/neondatabase/autoscaling/actions/runs/4572248349/jobs/8071305058#step:12:7103

(this failure seems to be repeated until the test timed out)

[a CI failure]: https://github.com/neondatabase/autoscaling/actions/runs/4572248349/jobs/8071305058

cc @cicdteam / @bayandin any ideas?

sharnoff:
Another, completely different case, from kuttl logs:

[...]

PR #125, opened 2023-04-02
e2e: Build docker images before cluster init

sharnoff:
AFAICT, some of the e2e flakiness might be partially resulting from system load after cluster start up, due to running vm-builder. I don't _actually_ know, though.

See also:
https://github.com/neondatabase/autoscaling/issues/120#issuecomment-1493405844

**Edit:** At first glance, it seems like this may allow the e2e tests to run faster. More investigation required, probably.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -64,11 +64,16 @@ jobs:
           VM_KERNEL_IMAGE: "neondatabase/vm-kernel"
           VM_KERNEL_VERSION: "5.15.80"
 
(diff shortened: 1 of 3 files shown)


bayandin (approved):
Makes sense!
Also we can try to use a bigger runner: `runs-on: [ self-hosted, gen3, large ]` in the future

sharnoff:
I've run into the "running vm-builder bricks a local kind cluster" issue on my computer (16 vCPU + 64 GB memory), so I'm not sure that it's something that a bigger runner would fix. But we can see - I won't say no to faster compile times :sweat_smile:

PR #109, opened 2023-03-31
util/watch fixup

sharnoff:
A couple convenience changes. `Stopped()` is required by #110.

PR #127, opened 2023-04-03
util/watch: add Relist() method

sharnoff:
Builds on #109, required by #128.

### DIFF
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -148,7 +148,7 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 
 	store := WatchStore[T]{
 		objects:       make(map[types.UID]*T),
-		triggerRelist: make(chan struct{}),
+		triggerRelist: make(chan struct{}, 1), // ensure sends are non-blocking
 		relisted:      make(chan struct{}),
 		stopSignal:    sendStop,
 	}
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -463,7 +463,6 @@ func (w *WatchStore[T]) Relist() <-chan struct{} {
 	w.mutex.Lock()
 	defer w.mutex.Unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/util/watch.go
```diff
@@ -407,10 +450,27 @@ type WatchStore[T any] struct {
...
+	triggerRelist chan struct{}
+	relisted      chan struct{}
+
 	stopSignal SignalSender
 	stopped    atomic.Bool
 }
 
+// Relist triggers re-listing the WatchStore, returning a channel that will be closed once the
+// re-list is complete
+func (w *WatchStore[T]) Relist() <-chan struct{} {
```
tychoish, line 462: this is ok, but you could also avoid the lock, and send (what is a reference, in effect) the response channel through the trigger channel, so `ch := make(chan struct{}); w.triggerRelist <- ch`, and then you wouldn't have to store the rellisted channel in the struct. 

Also the non-blocking send looks dodgy to my eyes, as I read the code, but I realized that it wasn't because you're mutating the w.relisted channel in the other thread.

tychoish (approved):

pkg/util/watch.go
sharnoff: > this is ok, but you could also avoid the lock, and send (what is a reference, in effect) the response channel through the trigger channel, so ch := make(chan struct{}); w.triggerRelist <- ch, and then you wouldn't have to store the relisted channel in the struct.

I decided not to create a channel for each call to `Relist` because subjectively, it felt like there's more cases that could lead to an unreasonable number of channels getting created (part of the challenge with having `Relist` return a channel).

> Also the non-blocking send looks dodgy to my eyes, as I read the code, but I realized that it wasn't

Well, because of this I double-checked and realized the send _was_ dodgy :upside_down_face: Needed to have `w.triggerRelist` be non-blocking, fixed that now.

> mutating the w.relisted channel in the other thread

Yep. It's more complicated than I'd like, but getting the semantics I want is actually kind of hard.

Issue #115, opened 2023-03-31
Epic: Create the pool of pre-allocated VMs


vadim2404:
### Motivation

To reduce start-up time for VMs (currently, it's ~4 seconds), we need to have several pre-warmed VMs and be able to change min/max boundaries at runtime.

### DoD
Start-up time for 99% of cases is < 1 sec

### Tasks:
* [x] https://github.com/neondatabase/autoscaling/pull/128 (includes handling runtime changes)
* [ ] https://github.com/neondatabase/cloud/pull/4741
* [x] https://github.com/neondatabase/autoscaling/issues/77

### Related epics/tasks/PRs: 
- neondatabase/neon#3610
- neondatabase/cloud#4433

ololobus:
I don't think that we have to do it specifically for VMs only. There are at least two things that prevent doing it only on the autoscaler / neonvm side.

1. Compute versions, they are picked based on storage node versions, so only control-plane knows which version of the pre-created compute is needed. After storage release there always will be pre-created computes left with stale version, so control-plane needs to tear them down and spawn fresh ones.

2. `compute_ctl` needs JWT and id (`compute_id`) to be able to get spec from the control-plane. These two are also known to the control-plane only, and future compute_id<->endpoint_id binding too.

Thus, as I imagined and discussed it with @tychoish and @kelvich primarily, pool of pre-created 'whatever' (pods, VMs or any other custom resource in the future) is maintained by the control-plane. From compute we only need an interface to 'wake it up' and notify that it now serves some particular timeline. I added some PoC `compute_ctl` code in this PR https://github.com/neondatabase/neon/pull/3923

So the only required part from the autoscaler perspective is:

> be able to change min/max boundaries at runtime

And as we discussed with @sharnoff, we will likely need to switch to using labels / annotations for that, since this is usually the only thing one can change on k8s objects without re-creation / restart.

vadim2404:
we discussed this item with you and it requires some on our side as well, therefore the epic is created.

> be able to change min/max boundaries at runtime
If it's enough, then who will be against it? :) 

> That said, I don't think that this target is realistic for a first iteration. Something like p90 or even p80 is more realistic, as any really huge spike from Hasura will likely hit k8s node autoscaler and p99 will ~20s-1min

I don't think that we need to stop after the first iteration. The ultimate goal is to have a really quick start-up for computes. Then, I don't want to discount the ultimate goal from the very beginning.

sharnoff:
[...]

PR #128, opened 2023-04-03
Allow annotation to set/override scaling bounds

sharnoff:
Part of #115. In progress. Builds on #38 additionally requires requires:

- [x] #127
- [x] #135

### DIFF
--- a/pkg/agent/globalstate.go
+++ b/pkg/agent/globalstate.go
@@ -78,15 +78,21 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
 	podName := util.NamespacedName{Namespace: event.vmInfo.Namespace, Name: event.podName}
 	state, hasPod := s.pods[podName]
 
+	if event.kind != vmEventAdded && !hasPod {
+		klog.Errorf("Received %s event for pod %v that isn't present", event.kind, event.podName)
+		return
+	}
+
 	switch event.kind {
 	case vmEventDeleted:
-		if !hasPod {
-			klog.Errorf("Received delete event for pod %v that isn't present", event.podName)
-			return
-		}
-
 		state.stop()
 		delete(s.pods, podName)
+	case vmEventUpdated:
+		state.status.lock.Lock()
+		defer state.status.lock.Unlock()
+
+		state.status.vmInfo = event.vmInfo
+		state.vmInfoUpdated.Send()
 	case vmEventAdded:
 		if hasPod {
 			klog.Errorf("Received add event for pod %v while already present", event.podName)
@@ -100,6 +106,7 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
 			done:     false,
 			errored:  nil,
 			panicked: false,
+			vmInfo:   event.vmInfo,
 		}
 
 		runner := &Runner{
@@ -129,14 +136,17 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
 			backgroundPanic:       make(chan error),
 		}
 
+		txVMUpdate, rxVMUpdate := util.NewCondChannelPair()
+
 		state = &podState{
-			podName: podName,
-			stop:    cancelRunnerContext,
-			runner:  runner,
-			status:  status,
+			podName:       podName,
+			stop:          cancelRunnerContext,
+			runner:        runner,
+			status:        status,
+			vmInfoUpdated: txVMUpdate,
 		}
 		s.pods[podName] = state
-		runner.Spawn(runnerCtx)
+		runner.Spawn(runnerCtx, rxVMUpdate)
 	default:
 		panic(errors.New("bad event: unexpected event kind"))
 	}
@@ -148,6 +158,8 @@ type podState struct {
 	stop   context.CancelFunc
 	runner *Runner
 	status *podStatus
+
+	vmInfoUpdated util.CondChannelSender
 }
 
 type podStateDump struct {
@@ -176,12 +188,20 @@ type podStatus struct {
 	done     bool // if true, the runner finished
 	errored  error
 	panicked bool // if true, errored will be non-nil
+
+	// vmInfo stores the latest information about the VM, as given by the global VM watcher.
+	//
+	// There is also a similar field inside the Runner itself, but it's better to store this out
+	// here, where we don't have to rely on the Runner being well-behaved w.r.t. locking.
+	vmInfo api.VmInfo
 }
 
 type podStatusDump struct {
 	Done     bool  `json:"done"`
 	Errored  error `json:"errored"`
 	Panicked bool  `json:"panicked"`
+
+	VMInfo api.VmInfo `json:"vmInfo"`
 }
 
 func (s *podStatus) dump() podStatusDump {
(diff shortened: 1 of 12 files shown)


pkg/agent/runner.go
```diff
@@ -511,6 +512,89 @@ func (r *Runner) handleVMResources(
...
+				panic("explicit VM update given but scaling is disabled")
+			}
+
+			// Update r.vm and r.lastApproved (see comment explaining why)
+			if changed := func() (changed bool) {
+				r.lock.Lock()
+				defer r.lock.Unlock()
+
+				if r.vm.Mem.SlotSize.Cmp(newVMInfo.Mem.SlotSize) != 0 {
+					panic("VM changed memory slot size")
```
tychoish, line 536: is this really a problem? seems like we could roll with it.

pkg/api/vminfo.go
```diff
@@ -129,3 +151,63 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
...
 	return &info, nil
 }
+
+func (vm VmInfo) EqualScalingBounds(cmp VmInfo) bool {
+	return vm.Min() != cmp.Min() || vm.Max() != cmp.Max()
+}
+
+type scalingBounds struct {
+	Min *resourceBound `json:"min,omitempty"`
+	Max *resourceBound `json:"max,omitempty"`
```
tychoish, line 161: why need these be pointers?

pkg/api/vminfo.go
```diff
@@ -129,3 +151,63 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
...
+	if b.Mem.IsZero() || b.Mem.Value() < 0 {
+		return ".mem", errors.New("value must be greater than zero")
+	} else if b.Mem.Value()%memSlotSize.Value() != 0 {
+		return ".mem", fmt.Errorf("value must be divisible by VM memory slot size %s", memSlotSize)
+	}
+
+	return "", nil
+}
+
+func (b scalingBounds) apply(info *VmInfo) {
```
tychoish, line 205: might it make sense to have this be a method on `VmInfo` rather than the other way around

pkg/plugin/watch.go
```diff
@@ -25,6 +25,48 @@ import (
...
+				if _, ok := pod.Labels[LabelVM]; ok {
+					vmDeletions <- name
+				} else {
+					podDeletions <- name
+				}
+			},
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("Error watching pod deletions: %w", err)
```
tychoish, line 62: I tend to get irritable about the word `Error` appearing in errors.

pkg/plugin/watch.go
```diff
@@ -51,52 +93,76 @@ import (
...
 // handled by cancelled contexts in *almost all* cases.
-func (e *AutoscaleEnforcer) watchPodEvents(
+func (e *AutoscaleEnforcer) watchVMEvents(
 	ctx context.Context,
-	vmDeletions chan<- util.NamespacedName,
 	vmDisabledScaling chan<- util.NamespacedName,
-	podDeletions chan<- util.NamespacedName,
-) error {
-	_, err := util.Watch(
+	updatedScalingBounds chan<- struct {
```
tychoish, line 99: can we declare the type?

pkg/plugin/watch.go
```diff
@@ -51,52 +93,76 @@ import (
...
-				// note: it doesn't matter whether we use newPod or oldPod for the name here; the
-				// pod name and namespace aren't allowed to change.
-				name := util.NamespacedName{Name: newPod.Name, Namespace: newPod.Namespace}
-				_, isVM := oldPod.Labels[LabelVM]
-				if isVM && api.HasAutoscalingEnabled(oldPod) && !api.HasAutoscalingEnabled(newPod) {
+		util.WatchHandlerFuncs[*vmapi.VirtualMachine]{
+			UpdateFunc: func(oldVM, newVM *vmapi.VirtualMachine) {
+				oldInfo, err := api.ExtractVmInfo(oldVM)
+				if err != nil {
+					klog.Errorf("[autoscale-enforcer] Error extracting VM info for %s:%s: %w", oldVM.Namespace, oldVM.Name, err)
```
tychoish, line 122: `s/Error//` etc.

pkg/agent/globalstate.go
```diff
@@ -78,15 +78,21 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
...
-		}
-
 		state.stop()
 		delete(s.pods, podName)
+	case vmEventUpdated:
+		state.status.lock.Lock()
+		defer state.status.lock.Unlock()
+
+		state.status.vmInfo = event.vmInfo
+		state.vmInfoUpdated.Send()
```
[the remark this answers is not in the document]
sharnoff: +1 to renaming the field; opened #153 as a follow-up.

The area you've highlighted is actually the end of the function (the outer switch statement is the last thing), but it's not great that this isn't clear. Also opened #154 to extract `vmEventAdded` into its own method.

pkg/agent/runner.go
sharnoff: Currently it would be a problem, yeah. But also it's not possible for VMs to change memory slot size at runtime, so if this happens something's probably gone horribly wrong somewhere else. Added a comment explaining that.

pkg/api/vminfo.go
sharnoff: They're pointers so that we can distinguish between "value set to zero" and "value wasn't set"

pkg/api/vminfo.go
```diff
@@ -129,3 +151,63 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
...
+	} else if b.Mem == nil {
+		return "", errors.New("missing field 'mem'")
+	}
+
+	if *b.CPU == 0 {
+		return ".cpu", errors.New("value cannot be zero")
+	}
+
+	if b.Mem.IsZero() || b.Mem.Value() < 0 {
+		return ".mem", errors.New("value must be greater than zero")
```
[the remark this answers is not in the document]
sharnoff: These are from user-supplied data (i.e. the control plane), so I'd rather not panic a long-running thread if it isn't our problem.

pkg/plugin/watch.go
sharnoff: Done

pkg/plugin/watch.go
```diff
@@ -158,10 +160,7 @@ func (e *AutoscaleEnforcer) watchVMEvents(
 					return
 				}
 
-				updatedScalingBounds <- struct {
-					vm      *api.VmInfo
-					podName string
-				}{newInfo, newVM.Status.PodName}
+				updatedScalingBounds <- vmPodInfo{vm: newInfo, podName: newVM.Status.PodName}
```
tychoish, line 163: are we worried about backpressure here? should we use the context?

tychoish (approved):

pkg/plugin/watch.go
sharnoff: Opened #160 for this

PR #130, opened 2023-04-04
Use go 1.20

bayandin:
Use latest go 1.20

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           # Required: the version of golangci-lint is required and
           # should be specified with patch version.
-          version: v1.50.1
+          version: v1.52.2
           args: --timeout 5m --fix
           github-token: ${{ secrets.github_token }}
         if: env.GIT_DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ ENVTEST_K8S_VERSION = 1.23.1
 # Get the currently used golang base path
 GOPATH=$(shell go env GOPATH)
 
+# Go 1.20 changed the handling of git worktrees:
+# https://github.com/neondatabase/autoscaling/pull/130#issuecomment-1496276620
+export GOFLAGS=-buildvcs=false
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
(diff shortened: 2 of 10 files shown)


sharnoff:
Broadly in favor of doing this. Can we set `-buildvcs=false`? Go 1.20 changed (read: [broke](https://github.com/golang/go/issues/59068)) the handling of git worktrees. (although, git worktrees also wouldn't work inside a container even if they implemented it correctly (because the referenced .gitdir wouldn't be in the container); they just didn't recognize .git _files_ prior to 1.20, so `-buildvcs=auto` didn't see a git repo)

tychoish:
Can we hold off on this until cloud moves to 1.20 (which I think is quasi blocked on the fact that some things make the linter oom with 1.20?) just so that it's almost possible to develop both of them on the same system/etc

tychoish:
> I _think_ that should be fixed by installing the right version of golangci-lint, unless there's a different OOM.

I'm giving this a second chance now, so we'll see. Still think we shouldn't get ahead of cloud but I don't think it matters.

bayandin:
> > I _think_ that should be fixed by installing the right version of golangci-lint, unless there's a different OOM.
> 
> I'm giving this a second chance now, so we'll see. Still think we shouldn't get ahead of cloud but I don't think it matters.

I don't mind waiting until we do it in cloud repo first. Is there a PR for this?

tychoish:
> I don't mind waiting until we do it in cloud repo first. Is there a PR for this?

Just made it!

bayandin:
> Can we set `-buildvcs=false`?

Added `GOFLAGS=-buildvcs=false` to Makefile.
I also passed the variable to all docker build/buildx commands. But can we skip this part or it's required as well?

tychoish (approved):
is the e2e test failure related? I didn't look, but seems like having a green checkmark would be nice :) 

Also, I think for the cloud docker images we pin to specific point-releases and here we're letting docker/etc (I think?) roll forward on point releases. I'm pretty confident in the stability of Go's release process, but I wanted to make sure we were ok with that difference, and its implications.

sharnoff (approved):
Sorry for such a late reply. I think we should only track the minor release, but if there's issues with a particular version we can pin to a patch until those are resolved.

Issue #124, opened 2023-04-02
Upgrade to k8s 1.24


sharnoff:
All our clusters are using k8s 1.24 now, so we should upgrade to that.

This also should allow filing https://github.com/neondatabase/autoscaling/issues/55 as a bug against k8s because 1.24 is not at end-of-life (need to check the behavior is still the same AND the fix in the issue description does resolve the issue).

Note: this will be a breaking change, because `neondatabase/autoscaling` will no longer work with k8s 1.23 (due to scheduler version skew requirements).

PR #132, opened 2023-04-04
Kubernetes 1.24

bayandin:
Resolves https://github.com/neondatabase/autoscaling/issues/124

I haven't tested https://github.com/neondatabase/autoscaling/issues/55


sharnoff:
I'd love to get this merged soon. Testing #55 can be done after merging, I think. Other than reviewing, is there anything else this PR needs?

cicdteam:
We need update versions for generators as well (in Makefile, scripts and  GitHub actions)

bayandin:
> We need update versions for generators as well (in Makefile, scripts and GitHub actions)

Are you about `CONTROLLER_TOOLS_VERSION`? I kept it on the 0.9 release because 0.10 has switched to k8s 1.25 (see https://github.com/kubernetes-sigs/controller-tools/releases). 
Should we bump it to 0.10 @cicdteam? (they should be compatible across +/-1 minor version, I think)

> Other than reviewing, is there anything else this PR needs?

Nothing else comes to my mind. You can play with it locally
If tests pass, it generally works 😄

cicdteam:
@bayandin 

Yes, you right, controller-gen was bumped to use k8s 1.24 in version 0.9.0, is current 0.9.2 is Ok

Nothing to change at this part.

cicdteam (approved):

sharnoff (approved):
Thought there were maybe some changes required in NeonVM controllers (because of the earlier downgrade 1.25 -> 1.23), but those pieces are strictly 1.25+, so we're good :)

PR #133, opened 2023-04-04
WIP: Vm builder separate

cicdteam:


Makefile
```diff
@@ -114,26 +115,32 @@ test: fmt vet envtest ## Run tests.
...
+bin/vm-builder-generic: ## Build vm-builder-generic binary.
+	go build -o bin/vm-builder-generic  neonvm/tools/vm-builder-generic/main.go
 
 .PHONY: run
 run: fmt vet ## Run a controller from your host.
 	go run ./neonvm/main.go
 
 .PHONY: vm-informant
 vm-informant: ## Build vm-informant image
+	echo "---$(GIT_INFO)---"
```
sharnoff, line 138:
```suggestion
	echo "---$(GIT_INFO)---"
```

Makefile
```diff
@@ -114,26 +115,32 @@ test: fmt vet envtest ## Run tests.
...
 	docker buildx build \
+		--quiet \
```
sharnoff, line 140:
```suggestion
		--quiet \
```

We don't use `--quiet` for other images

neonvm/runner/main.go
```diff
@@ -527,11 +531,46 @@ func main() {
 		qemuCmd = append(qemuCmd, "-incoming", fmt.Sprintf("tcp:0:%d", vmv1.MigrationPort))
 	}
 
+	go terminateQemu(vmSpec.QMP)
 	if err := execFg(QEMU_BIN, qemuCmd...); err != nil {
 		log.Fatal(err)
 	}
 }
 
+func terminateQemu(qmpPort int32) {
```
sharnoff, line 540:
```suggestion
func terminateQemuOnSigterm(qmpPort int32) {
```

neonvm/runner/main.go
```diff
@@ -527,11 +531,46 @@ func main() {
...
+		qmpcmd := []byte(`{"execute": "system_powerdown"}`)
+		_, err = mon.Run(qmpcmd)
+		if err != nil {
+			log.Println(err)
+			continue
+		}
+
+		log.Println("system_powerdown command sent to QEMU")
+		break
+	}
```
sharnoff, line 569: We should have some kind of minimum time to wait between loop iterations here so this loop can't _completely_ flood the logs if it's continually failing.

neonvm/tools/vm-builder-generic/main.go
```diff
@@ -0,0 +1,485 @@
...
+	"os"
+	"path/filepath"
+	"text/template"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+)
+
+// vm-builder --src alpine:3.16 --dst vm-alpine:dev --file vm-alpine.qcow2
```
sharnoff, line 21:
```suggestion
// vm-builder-generic --src alpine:3.16 --dst vm-alpine:dev --file vm-alpine.qcow2
```

Makefile
```diff
@@ -114,26 +115,32 @@ test: fmt vet envtest ## Run tests.
...
+build: fmt vet bin/vm-builder bin/vm-builder-generic ## Build all neonvm binaries.
 	go build -o bin/controller       neonvm/main.go
 	go build -o bin/vxlan-controller neonvm/tools/vxlan/controller/main.go
 	go build -o bin/vxlan-ipam       neonvm/tools/vxlan/ipam/main.go
 	go build -o bin/runner           neonvm/runner/main.go
 
 .PHONY: bin/vm-builder
 bin/vm-builder: ## Build vm-builder binary.
-	go build -o bin/vm-builder neonvm/tools/vm-builder/main.go
+	go build -o bin/vm-builder -ldflags "-X main.Version=${VERSION} -X main.VmInformant=${VM_INFORMANT_IMG}" neonvm/tools/vm-builder/main.go
```
sharnoff, line 126:
```suggestion
	go build -o bin/vm-builder -ldflags "-X main.Version=${GIT_INFO} -X main.VmInformant=${VM_INFORMANT_IMG}" neonvm/tools/vm-builder/main.go
```

I think we can use `GIT_INFO` for this, like some of the other images. That way, we don't need to add `VERISON` to the makefile, and we can get a little more info :)

neonvm/tools/vm-builder/main.go
```diff
@@ -191,38 +285,50 @@ for i in ${ETH_LIST}; do
...
 
 var (
-	srcImage   = flag.String("src", "", `Docker image used as source for virtual machine disk image: --src=alpine:3.16`)
-	dstImage   = flag.String("dst", "", `Docker image with resulting disk image: --dst=vm-alpine:3.16`)
-	size       = flag.String("size", "1G", `Size for disk image: --size=1G`)
-	outFile    = flag.String("file", "", `Save disk image as file: --file=vm-alpine.qcow2`)
-	useInittab = flag.Bool("use-inittab", false, `Use guest container's inittab, appending it to the default one`)
-	forcePull  = flag.Bool("pull", false, `Pull src image even if already present locally`)
+	Version     string
+	VmInformant string
```
sharnoff, line 323:
```suggestion
	VMInformant string
```

neonvm/tools/vm-builder/main.go
```diff
@@ -406,33 +532,37 @@ func main() {
...
-	buildArgs["USE_INITTAB"] = &inittabArg
-
 	opt := types.ImageBuildOptions{
 		Tags: []string{
 			dstIm,
 		},
 		BuildArgs:      buildArgs,
 		SuppressOutput: true,
-		NoCache:        false,
+		NoCache:        true,
```
sharnoff, line 565: Is this necessary? I'd be concerned that this causes a lot of unnecessary re-builds.

neonvm/runner/main.go
```diff
@@ -527,11 +531,46 @@ func main() {
 		qemuCmd = append(qemuCmd, "-incoming", fmt.Sprintf("tcp:0:%d", vmv1.MigrationPort))
 	}
 
+	go terminateQemu(vmSpec.QMP)
```
sharnoff, line 534: Is this required? If not, it'd be good to have this as a separate PR

sharnoff (approved):
Broadly looks good. My comments are _mostly_ nits.

neonvm/runner/main.go
cicdteam: @sharnoff it's just graceful shutdown for VM. What objections do not add it in this PR ?

sharnoff: The main problem I have is that, without reading the content of this PR, I had no idea that it implemented the graceful shutdown feature. I tend to split features into their own PRs (sometimes too much), but a reasonable alternative IMO is to just make sure that the features are included in the PR/commit name.

As an example, this was the motivation for #146. Here is a more significant case because there's actually a whole separate feature.

Anyways - do what you think is right. I'd _prefer_ to either have graceful shutdown in the commit message or as a separate PR, but it's up to you.

sharnoff: Fair enough, go for it

PR #134, opened 2023-04-05
deploy/sched: Re-enable migration

sharnoff:
Follow-up to #112 that turns the feature on. Requires #104.


Omrigan:
Can reopen later when needed

sharnoff:
This was actually superseded by #781, so no longer necessary.

PR #135, opened 2023-04-05
util/watch: Add user-defined indexing

sharnoff:
From the commit message:

> This allows making certain queries more efficient, but leaves the work of doing that for the user of WatchStore.
>
> One such example is implemented in this PR, by providing agent/billing with a WatchStore with an index allowing quick lookup of the VMs for its node.

This feature was implemented for #128 in order to allow efficient "get this VM" lookups in the scheduler plugin. It's necessary there in order to not have to deal with inconsistent state in the plugin extension methods vs the resource request handler.


pkg/agent/indexedstore.go
```diff
@@ -0,0 +1,56 @@
...
+	// note: delete is a no-op if the key isn't present.
+	delete(i.forNode, vm.UID)
+}
+
+func (i *VMNodeIndex) List() []*vmapi.VirtualMachine {
+	items := make([]*vmapi.VirtualMachine, len(i.forNode))
+	for _, vm := range i.forNode {
+		items = append(items, vm)
+	}
+	return items
```
tychoish, line 55: you allocate a slice of a specfied length, and then append, items to the end of it: so you have a lot of nil items and then you add more. adding a second 0 to the `make()` will resolve this issue. 

Given random iteration order, it wouldn't be unreasonable to sort these in some way, if there is any meaning or ordering of the list. 

It's also the case, that this does the right thing, but also it sort of looks/reminds me of an easy to make bug that I'm perhaps overly cautious to avoid: 

https://go.dev/play/p/oAOaGC-QrE1

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+// to efficiently answer queries.
+type IndexedWatchStore[T any, I WatchIndex[T]] struct {
+	*WatchStore[T]
+
+	index I
+
+	// id stores the id of this index in the WatchStore
+	id uint64
+	// collector allows us to
+	collector *struct{}
```
tychoish, line 593: allows us to...

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+	w.WithIndex(func(i I) {
+		list = f(i)
+	})
+	return
+}
+
+func NewNameIndex[T any]() *NameIndex[T] {
+	// check that *T implements metav1.ObjectMetaAccessor
+	var zero T
+	ptrToZero := any(&zero)
```
tychoish, line 623: I think `any(new(T))` does this?

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+	if _, ok := ptrToZero.(metav1.ObjectMetaAccessor); !ok {
+		panic("type *T must implement metav1.ObjectMetaAccessor")
+	}
+
+	// FIXME: does this need to be a pointer?
```
tychoish, line 628: probably not, but I don't think it hurts much.

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+	// collector allows us to
+	collector *struct{}
+}
+
+// WithIndex calls a function with the current state of the index, locking the WatchStore around it.
+//
+// It is almost guaranteed to be an error to indirectly return the index with this function.
+func (w IndexedWatchStore[T, I]) WithIndex(f func(I)) {
+	w.WatchStore.mutex.Lock()
+	defer w.WatchStore.mutex.Unlock()
```
tychoish, line 601: I had little visceral reaction to this from a locking hygiene perspective, but the inevitable deadlock would probably surface pretty quick?

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+		name      string
+	}]*T
+}
+
+// note: requires that *T implements metav1.ObjectMetaAccessor
+func keyForObj[T any](obj *T) struct {
+	namespace string
+	name      string
+} {
+	meta := any(obj).(metav1.ObjectMetaAccessor).GetObjectMeta()
```
tychoish, line 650: I really want there to be a way to specialize/type using interface types rather than generics, given that we have to do this runtime cast anyway, but I suspect you've had the same feeling and have figured out that there isn't

pkg/util/watch.go
```diff
@@ -511,3 +549,127 @@ func (w *WatchStore[T]) Items() []*T {
...
+// NameIndex is a WatchIndex that provides efficient lookup for a value with a particular name
+type NameIndex[T any] struct {
+	namespacedNames map[struct {
+		namespace string
+		name      string
+	}]*T
+}
+
+// note: requires that *T implements metav1.ObjectMetaAccessor
+func keyForObj[T any](obj *T) struct {
```
tychoish, line 646: I might just declare a key type rather than use the same anonymous function declaration everywhere.)

pkg/agent/indexedstore.go
```diff
@@ -0,0 +1,56 @@
...
+		node:    node,
+	}
+}
+
+func (i *VMNodeIndex) Add(vm *vmapi.VirtualMachine) {
+	if vm.Status.Node == i.node {
+		i.forNode[vm.UID] = vm
+	}
+}
+func (i *VMNodeIndex) Update(oldVM, newVM *vmapi.VirtualMachine) {
```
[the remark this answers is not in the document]
sharnoff: Nope; the watcher tracks things based on UIDs, which k8s guarantees are unique for the lifetime of the cluster.

pkg/agent/indexedstore.go
sharnoff: Oops, yep, good catch :sweat_smile:

pkg/util/watch.go
sharnoff: Yup, fair enough. done

pkg/util/watch.go
sharnoff: oops, fixed

pkg/util/watch.go
sharnoff: Yeah, basically what's going on here is that the `WatchStore` is parameterized by the type `T` rather than `*T`, so its indexes must be as well. The issue is that `metav1.ObjectMetaAccessor` is generally implemented for `*T`, not `T`, so we can't directly require that for this index's type without adding a second generic parameter (which would be too much of a hassle to deal with).

pkg/util/watch.go
sharnoff: Yeah, generally indexes are expected to be accessed only through their `WatchStore`, so instead of exposing the lock to the user, the pattern is "give me a function, I'll call it with the thing you want, instead of directly exposing control flow to you".

It's a bit of a rust-ism, but I think it's worth not exposing the locks here, especially considering that indexes will generally be _fairly_ simple.

pkg/util/watch.go
sharnoff: resolved

sharnoff:
@tychoish re-requested review. note that I merged #145 into here, so there's currently some other commits you might want to ignore for the purposes of reviewing this PR :)

pkg/util/watch.go
tychoish: Ah looks cool, I think when I was trying to do this, I saw some stack overflow that indicated that the `*new(T)` option was preferable in some way, but it looks like they're equivalent and despite forming a habit for `new` (which I was sort of glad to have a use for, for the first time,) to do this, I think your option is better.

tychoish (approved):

PR #136, opened 2023-04-05
util/watch: Delete old objects before adding new ones

sharnoff:
This commit fixes a problem with the relisting code, namely: Because we delete old objects *after* adding new ones, properties that are normally considered unique (e.g. a Pod's name) may be temporarily violated because they are not unique *across time*.

Required for the correctness of the API added in #135.

### DIFF
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -374,19 +374,40 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 				relistItems := accessors.Items(relistList)
 
 				func() {
-					// First, copy the contents of objects into oldObjects. We do this so that we can
-					// uphold some guarantees about the contents of the store.
-					oldObjects := make(map[types.UID]*T)
-					for uid, obj := range store.objects {
-						oldObjects[uid] = obj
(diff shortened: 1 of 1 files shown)


pkg/util/watch.go
```diff
@@ -374,20 +374,41 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
 				func() {
-					// First, copy the contents of objects into oldObjects. We do this so that we can
-					// uphold some guarantees about the contents of the store.
+					store.mutex.Lock()
+					defer store.mutex.Unlock()
+
+					// Copy the current contents of objects, and start tracking which ones have
+					// since been deleted.
 					oldObjects := make(map[types.UID]*T)
+					deleted := make(map[types.UID]bool)
```
tychoish, line 383: I tend to avoid maps to bools because it's often unclear what `deleted[uid] = false` is supposed to mean (and intent is hard to discern as a reader.

tychoish (approved):

pkg/util/watch.go
sharnoff: Makes sense. updated

PR #137, opened 2023-04-05
e2e: run k8s logs/events step if cluster exists

sharnoff:
From the commit message:

> This makes some of the handling around the "Get k8s logs and events" step a little more precise. We want to run it on failure-ish so long as the cluster was created.
>
> To do so, this commit makes fetching the logs conditional on cluster creation *not being skipped* (in case cluster creation failed, or was cancelled).

I'm not _actually_ sure that this does what I'm expecting; my knowledge of github workflows is limited :sweat_smile:

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -70,7 +70,9 @@ jobs:
       - run: make docker-build
       - run: make docker-build-examples
 
-      - run: make local-cluster
+      - name: make local-cluster
+        run: make local-cluster
+        id: make_local_cluster
 
       - run: make deploy
       - run: make example-vms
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -86,7 +88,7 @@ jobs:
           kubectl kuttl test --config tests/e2e/kuttl-test.yaml --skip-delete
 
       - name: Get k8s logs and events
-        if: failure()
+        if: ${{ (failure() || cancelled()) && steps.make_local_cluster.conclusion != 'skipped' }}
         run: |
           namespaces=$(kubectl get namespaces -o jsonpath='{.items[*].metadata.name}')
```
bayandin, line 93: Maybe we can just check if a cluster exists?
```sh
if [ $(kubectl config view -o jsonpath='{.contexts[*].name}') = "kind-kind" ]; then
  # Get logs
fi
```

bayandin (approved):

sharnoff:
Going to try running CI and cancelling it partway through, just to check it works as expected. Will merge afterwards.

Issue #101, opened 2023-03-29
Epic: support fractional CPU


sharnoff:
## Motivation

Free tier is currently defined as 0.25 CPU + 1 GB memory. We'd like to (a) move all users to VMs (which requires VMs with 0.25 CPU) and (b) _ideally_ allow scaling to/from the free tier limits of 0.25 CPU.

## DoD

Full support for fractional CPU allocations - in particular, allowing VMs to be created with 0.25 or 0.5 CPU, in addition to the typical integer scaling. The scheduler must appropriately handle bookkeeping for VMs with < 1 CPU.

As part of the implementation, we may also provide fractional CPUs above 1 - e.g. 11.5 CPU (instead of just integer number of CPUs when >= 1)

## Implementation ideas

There's a few possible ways of solving this. I think we've settled on doing this in NeonVM, not sure which of the 2nd or 3rd option.

1. Implement in the scheduler/autoscaler-agent via CPU overcommit.
2. Implement in NeonVM with K8s pod limits
3. Implement in NeonVM by running QEMU in a cgroup, inside the runner container

It's worth noting that option 2 would make it impossible for us to scale between CPU < 1 and CPU >= 1.

## Tasks

- [ ] Add support for fractional CPU in NeonVM
- [ ] Handle fractional CPU correctly in autoscaler-agent, scheduler (depends on NeonVM level of support)
- [ ] Enable fractional CPU for VMs in control plane/console

## Other related tasks and Epics

None

sharnoff:
Fully done, in prod, etc. Should have been closed by #172, I think.

PR #138, opened 2023-04-06
[NeonVM] use VM .spec.podResources in runner pod

cicdteam:
This PR allows set up resources for runner pod. Example of VM spec:

```yaml
spec:
  podResources:
    requests:
      cpu: 1
      memory: 1Gi
    limits:
      cpu: 2
      memory: 2Gi
```

>NOTE: `.spec.podResources` us immutable and controller doesn't allow change it after VM created.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_webhook.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_webhook.go
@@ -154,6 +154,9 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) error {
 	if !reflect.DeepEqual(r.Spec.Disks, before.Spec.Disks) {
 		return fmt.Errorf(".spec.disks is immutable")
 	}
+	if !reflect.DeepEqual(r.Spec.PodResources, before.Spec.PodResources) {
+		return fmt.Errorf(".spec.podResources is immutable")
+	}
 
 	// validate .spec.guest.cpu.use
 	if r.Spec.Guest.CPUs.Use != nil {
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -676,6 +676,7 @@ func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
 					Name:      "virtualmachineimages",
 					MountPath: "/vm/images",
 				}},
+				Resources: virtualmachine.Spec.PodResources,
 			}},
 			Volumes: []corev1.Volume{{
 				Name: "virtualmachineimages",


sharnoff:
Broadly, I think this is fine. However: The scheduler currently will not handle this well. Do you think it makes sense to say that VMs with `podResources` set cannot have autoscaling enabled?

Personally I'd like to see a more general solution to #101 than this, but I'm happy to implement that myself (especially considering that you probably have your hands full with #104). That said, I'm still happy to merge this after resolving the question above.

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -154,6 +154,9 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) error {
 	if !reflect.DeepEqual(r.Spec.Disks, before.Spec.Disks) {
 		return fmt.Errorf(".spec.disks is immutable")
 	}
+	if !reflect.DeepEqual(r.Spec.PodResources, before.Spec.PodResources) {
+		return fmt.Errorf(".spec.podResources is immutable")
```
tychoish, line 158: nit: I'd use `errors.New` for any error that doesn't have a formatted error.

tychoish (approved):

tychoish:
> The scheduler currently will not handle this well.

Do you mean "the scheduler will not handle this new error" or "the scheduler already has a problem handling changes to the podresources in the spec"? 

It was my understanding that these pod resources are (generally) immutable anyway, so if a caller is trying to change something that's actually immutable, then that would already be a problem, but perhaps I misunderstand a detail here.

sharnoff:
> Do you mean "the scheduler will not handle this new error" or "the scheduler already has a problem handling changes to the podresources in the spec"?

The scheduler currently distinguishes between VMs and not-VMs, with handling that is _very_ different for each. It'd be a not-insignificant amount of work to handle VMs that behave mostly like pods w.r.t. resource usage - especially considering that the meaning of "approving scaling to 4 CPUs" becomes muddled if the pod itself is limited to 3 CPU.

So _ideally_ I'd like to not have to deal with that :sweat_smile:

(of course, broadly the handling should be unified so that these kinds of things aren't _quite_ as big a deal, but such is the nature of tech debt)

tychoish:
> The scheduler currently distinguishes between VMs and not-VMs, with handling that is _very_ different for each. It'd be a not-insignificant amount of work to handle VMs that behave mostly like pods w.r.t. resource usage - especially considering that the meaning of "approving scaling to 4 CPUs" becomes muddled if the pod itself is limited to 3 CPU.

This makes sense. Does it use this field to decide pod/not-pod?

sharnoff:
> Does it use this field to decide pod/not-pod?

Currently it doesn't look at `podResources`. It'd be easy to say that VMs with `podResources` set should be treated like pods, but that also means that they'd have all scaling disabled. I'd like to be careful about that choice. (But also: if `podResources` can't be used with the rest of autoscaling, why include it?)

cicdteam:
Not necessary specify `podResources` but it could solve problem with free-tier where we need no autoscaling but need limit cpu usage at `250m`

sharnoff:
re: `use errors.New for unformatted errors` — there's a bunch more, elsewhere in `neonvm/`. So that this PR doesn't get off-topic, I opened https://github.com/neondatabase/autoscaling/pull/146 to handle that

cicdteam:
any objections to merge it ?

sharnoff:
Merged #146, conflicts are just duplicating work from this PR. No objections

sharnoff (approved):
Realized I never approved this. I'm not sure that podResources is useful long-term, but that's for later - this is still ok to merge for now.

PR #140, opened 2023-04-06
neonvm: Remove VirtualMachineSpec.ServiceAccountName

### LINKED ISSUE #103: ServiceAccountName-related mounts should be passed to VM
neondatabase/neonvm#12 added the ServiceAccountName field, but it's only useful with other commits from `sharnoff/dev`. The relevant /etc/kubernetes/... paths should be mounted in the VM as well (if possible; maybe this is not easy).

sharnoff:
This was a holdover from when we needed to run each autoscaler-agent in a sidecar. Supporting sidecar containers only ever existed in a non-main branch, and so ServiceAccountName currently doesn't do anything visible within the VM.

Closes #103.

See also: https://github.com/neondatabase/neonvm/pull/15.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -44,12 +44,11 @@ type VirtualMachineSpec struct {
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds"`
 
-	NodeSelector       map[string]string           `json:"nodeSelector,omitempty"`
-	Affinity           *corev1.Affinity            `json:"affinity,omitempty"`
-	Tolerations        []corev1.Toleration         `json:"tolerations,omitempty"`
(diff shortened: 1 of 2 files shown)


tychoish (approved):

cicdteam:
I'm not sure we need remove ability to set up `ServiceAccount` for runner pod. May be "tomorrow" we need it (for example to link with RBAC rules if VM need access to some k8s resources).

`ServiceAccount` always present. If you not set it up then `default` one used.

sharnoff:
As I understand it, the current issue with `ServiceAccount` is that it won't forward the added mounts from the service account to the VM, so the feature doesn't actually do anything right now (this is what #103 is about). IIRC there's some subtle tricky things there as well, so I think it's not as simple as "just" adding the config.

> May be "tomorrow" we need it (for example to link with RBAC rules if VM need access to some k8s resources).

I also think we'll probably need it in the future. However: I don't think it's worth keeping this feature that doesn't work just because we might need it later.

cicdteam (approved):

PR #141, opened 2023-04-06
Add .git-blame-ignore-revs

sharnoff:
This allows us to filter out commits from `git blame`, which should make it clearer after refactorings that are sufficiently broken up :)

To enable locally, run

    git config --add blame.ignoreRevsFile .git-blame-ignore-revs

For info on Github's handling, see:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

(this is a follow-up to #113)

**Edit:** The blame is a bit [funky in some places](https://github.com/neondatabase/autoscaling/blame/41cc47ff45dde4afba135db790be1c0ce4143c14/pkg/agent/runner.go#L1005-L1016), but I still think this is _probably_ a net win.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# "intermediate commit: de-indent some blocks (#113)" - factored out of #113 to make blame clear.
+480e9cc9534b49204d7e5bbd06fbc454c96e6c9e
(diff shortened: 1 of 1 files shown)


tychoish (approved):
this is super cool, do we need to filter out any of the merge-related commits for neonvm?

sharnoff:
> do we need to filter out any of the merge-related commits for neonvm?

I _think_ the blame for those is already working as it should. I did have a look through all prior commits to see if there were any other commits that can/should be ignored for blame, but didn't find any.

sharnoff:
Merging. Any other commits can be added later :)

PR #145, opened 2023-04-07
Add util.NamespacedName to replace api.PodName

sharnoff:
Having an `api.PodName`-equivalent available in util is required for #135. This PR bundles that with some cleanup of the name itself, which I'd been wanting to do for a while.

More justification from the 1st commit message:

> Defining this functionality in pkg/api means it can't be used in pkg/util, because api depends on util. https://github.com/neondatabase/autoscaling/pull/135 requires it in util.
>
> It's also worth making the name more generic, because it's not just pods that we want to handle namespaced names for.

Plan is to add the second commit to .git-blame-ignore-revs after merging.

---

**Suggestion for reviewing**: I'd probably just look at the first commit. The second _does_ have a couple changes that aren't just search-and-replace, but those are the actual deletion of `api.PodName` plus getting rid of unused imports.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -11,18 +11,7 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
-// PodName represents the namespaced name of a pod
-//
-// Some analogous types already exist elsewhere (e.g., in the kubernetes API packages), but they
-// don't provide satisfactory JSON marshal/unmarshalling.
-type PodName struct {
-	Name      string `json:"name"`
-	Namespace string `json:"namespace"`
-}
-
-func (n PodName) String() string {
-	return fmt.Sprintf("%s:%s", n.Namespace, n.Name)
-}
+type PodName = util.NamespacedName
 
 /////////////////////////////////
 // (Autoscaler) Agent Messages //
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -11,8 +11,6 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
-type PodName = util.NamespacedName
-
 /////////////////////////////////
 // (Autoscaler) Agent Messages //
 /////////////////////////////////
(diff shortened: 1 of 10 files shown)


pkg/api/types.go
```diff
@@ -11,18 +11,7 @@ import (
...
-// don't provide satisfactory JSON marshal/unmarshalling.
-type PodName struct {
-	Name      string `json:"name"`
-	Namespace string `json:"namespace"`
-}
-
-func (n PodName) String() string {
-	return fmt.Sprintf("%s:%s", n.Namespace, n.Name)
-}
+type PodName = util.NamespacedName
```
tychoish: canonically the type `=` (rather than conventional aliasing) exists for maintaining api compatibility during refactors to make it possible to move types from packages without needing to change lots of call sights, and sort of indicates something is still in a state of transition. 

Having said that, I think this is a silly convention, and I think this kind of type declaration can be useful more broadly, but because it's sort of against the convention, I think it bears a comment to explain the intent?

pkg/util/namespacedname.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// When printed with '%s' or '%v', NamespacedName is rendered as "<namespace>:<name>". Printing with
+// '%+v' or '%#v' renders as it would normally.
+type NamespacedName struct {
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+func (n NamespacedName) Format(state fmt.State, verb rune) {
+	if verb == 'v' {
```
tychoish, line 24: I've always seen these format methods implemented with switch statements, which I think is a little clearer to read, but thias preference.

pkg/util/namespacedname.go
```diff
@@ -0,0 +1,49 @@
...
+			_, _ = state.Write([]byte(n.Name))
+			_, _ = state.Write([]byte{'}'})
+			return
+		}
+		// Go syntax representation, e.g. `util.NamespacedName{Namespace:"foo", Name:"bar"}`
+		if state.Flag('#') {
+			_, _ = state.Write([]byte(string("util.NamespacedName{Namespace:")))
+			_, _ = state.Write([]byte(fmt.Sprintf("%q", n.Namespace)))
+			_, _ = state.Write([]byte(string(", Name:")))
+			_, _ = state.Write([]byte(fmt.Sprintf("%q", n.Name)))
```
tychoish, line 39: why not combine this into a single fmt.Sprintf?

tychoish (approved):
lgtm modulo a couple of take-it-or-leave-it style/comment nits

pkg/api/types.go
sharnoff: > sort of indicates something is still in a state of transition

that's exactly what's going on here :)

`api.PodName` is removed by the second commit.

pkg/util/namespacedname.go
sharnoff: Good point, done.

PR #148, opened 2023-04-08
fix GIT_INFO in Makefile

sharnoff:
Extracted from #133, because I needed it while working on #128.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-GIT_INFO=$(shell git describe --long --dirty)
+GIT_INFO := $(shell git describe --long --dirty)
 
 .PHONY: all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-GIT_INFO := $(git describe --long --dirty)
+GIT_INFO=$(shell git describe --long --dirty)
```
tychoish, line 36: The shift to the `shell` function makes a lot of sense. 

I think if I had done this myself, I wouldn't have swapped `:=` for `=` but that's an instinct thing not a purely rational thing, and I do much less makefile development than I used to.

tychoish (approved):

Makefile
sharnoff: Makes sense. `:=` seems better

PR #150, opened 2023-04-10
e2e tests: do not skip cleanup

bayandin:
I suspect `--skip-delete` flag could be the cause of `vm-migration` test flakiness (https://github.com/neondatabase/autoscaling/issues/120).
`--skip-delete` prevents namespaces (with pods) from being deleted after the test, so it means for the second test (`vm-migration`) we have a pod from the first test (`autoscaling`).

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -70,22 +70,12 @@ jobs:
       - run: make docker-build
       - run: make docker-build-examples
 
-      - name: make local-cluster
-        run: make local-cluster
-        id: make_local_cluster
+      - run: make local-cluster
 
       - run: make deploy
       - run: make example-vms
 
-      - name: Run make e2e
-        run: |
-          # Increase inotify limits for dnsmasq to fix
-          #   "dnsmasq: failed to create inotify: No file descriptors available"
-          sudo sysctl fs.inotify.max_user_instances=8192
-          sudo sysctl fs.inotify.max_user_watches=524288
-          sudo sysctl -p
-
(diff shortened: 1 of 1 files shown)


tychoish (approved):

bayandin:
On the sixth retry, it failed again https://github.com/neondatabase/autoscaling/actions/runs/4661965775/attempts/6 😞 
It looks like the cause of the flakiness is not this running pod from the previous test.

tychoish:
> On the sixth retry, it failed again https://github.com/neondatabase/autoscaling/actions/runs/4661965775/attempts/6 😞 It looks like the cause of the flakiness is not this running pod from the previous test.

Might still be good to clean things up better though...

.github/workflows/e2e-test.yaml
```diff
@@ -70,22 +70,12 @@ jobs:
...
       - run: make deploy
       - run: make example-vms
 
-      - name: Run make e2e
-        run: |
-          # Increase inotify limits for dnsmasq to fix
-          #   "dnsmasq: failed to create inotify: No file descriptors available"
-          sudo sysctl fs.inotify.max_user_instances=8192
-          sudo sysctl fs.inotify.max_user_watches=524288
-          sudo sysctl -p
```
sharnoff, line 86: This is no longer necessary?

sharnoff (approved):

.github/workflows/e2e-test.yaml
bayandin: Yep, this has been added to runners itself: https://github.com/neondatabase/runner/pull/20

PR #154, opened 2023-04-12
agent: move vmEventAdded handling into its own method

sharnoff:


pkg/agent/globalstate.go
```diff
@@ -94,64 +94,74 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
...
+		errored:  nil,
+		panicked: false,
+		vmInfo:   event.vmInfo,
+	}
+
+	runner := &Runner{
+		global: s,
+		status: status,
+		logger: RunnerLogger{
+			prefix: fmt.Sprintf("Runner %v: ", event.podName),
```
tychoish, line 129: what's the type of the podname and why is this a %v and not an %s

pkg/agent/globalstate.go
```diff
@@ -94,64 +94,74 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
...
+		return
+	}
+
+	runnerCtx, cancelRunnerContext := context.WithCancel(ctx)
+
+	status := &podStatus{
+		lock:     sync.Mutex{},
+		done:     false,
+		errored:  nil,
+		panicked: false,
```
tychoish, line 121: I still think that declaring all these zero values is a bit unidiomatic, and makes things harder to read.

pkg/agent/globalstate.go
```diff
@@ -94,64 +94,74 @@ func (s *agentState) handleEvent(ctx context.Context, event vmEvent) {
...
 }
 
+func (s *agentState) handleVMEventAdded(
+	ctx context.Context,
+	event vmEvent,
+	podName util.NamespacedName,
+) {
+	state, hasPod := s.pods[podName]
+
+	if hasPod {
```
tychoish, line 110: the `hasPod` vs `ok` for the name of the existence check threw me for a second.

tychoish (approved):

pkg/agent/globalstate.go
sharnoff: Oh good catch. This was previously a `util.NamespacedName`, but got changed to `string` by a recent PR. Opened #159

PR #155, opened 2023-04-12
neonvm: make spec.{cpus,memorySlots}.{min,max} required

sharnoff:
This change is mostly useful so that we don't have to have special-case handling for maybe-nil pointers that in practice are always non-nil.

Not aware of a good reason to keep them, but if there is, this change isn't necessary.

ref https://github.com/neondatabase/cloud/pull/4650#discussion_r1163938887


tychoish (approved):

cicdteam:
I think if we set `min`, `max` as required, then `use` also should be required.

Another solution - just set default values for `max` and `use` the same way as it already set for `min`. By other words - if you don't specify these fields in VM spec these fields will se to `1`.

But overall I have no objections for this PR.

sharnoff:
> I think if we set `min`, `max` as required, then `use` also should be required.

That makes sense - I'll do that.

> Another solution - just set default values for `max` and `use` the same way as it already set for `min`. By other words - if you don't specify these fields in VM spec these fields will se to `1`.

That would also make sense to me, somewhat. I like that the simplest spec stays clean, but it would be annoying if `use` is set to 1 when `min > 1`. I'll continue with the approach in this PR.

cicdteam:
@sharnoff 
>I like that the simplest spec stays clean, but it would be annoying if use is set to 1 when min > 1. I'll continue with the approach in this PR.

validation webhook doesn't allow it and you will get error when you try apply such spec

cicdteam (approved):

cicdteam:
I'm worried that this PR is too far off the `main` one right now and it's better to raise a new one (if it's still relevant) than to resolve conflicts.

sharnoff:
Yeah... tbh I think the costs outweigh the benefits on this one. Closing - maybe we change it in the future, idk.

PR #158, opened 2023-04-12
api: Use better formatting for VmInfo

sharnoff:
The reason this is useful is because `VmInfo` contains a k8s `resource.Quantity`, and the default formatting shows all the fields (instead of the more useful pretty-printing that it _could_ do). Basically: it makes the logs cleaner.

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -859,7 +859,7 @@ func (s *InformantServer) Downscale(ctx context.Context, to api.Resources) (*api
 	s.runner.logger.Infof("Sending downscale %+v", to)
 
 	timeout := time.Second * time.Duration(s.runner.global.config.Informant.DownscaleTimeoutSeconds)
-	rawResources := to.ConvertToRaw(&s.runner.vm.Mem.SlotSize)
+	rawResources := to.ConvertToRaw(s.runner.vm.Mem.SlotSize)
 
 	resp, statusCode, err := doInformantRequest[api.RawResources, api.DownscaleResult](
 		ctx, s, timeout, http.MethodPut, "/downscale", &rawResources,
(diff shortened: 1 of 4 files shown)


pkg/api/vminfo.go
```diff
@@ -129,3 +130,54 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
 
 	return &info, nil
 }
+
+// the reason we have custom formatting for VmInfo is because without it, the formatting of memory
+// slot size (which is a resource.Quanitty) has all the fields, meaning (a) it's hard to read, and
```
vadim2404, line 135: typo: `Quanitty` -> `Quantity`

sharnoff: good catch - done

pkg/api/vminfo.go
```diff
@@ -129,3 +130,54 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
...
+// (b) there's a lot of unnecessary information. But in order to enable "proper" formatting given a
+// VmInfo, we have to implement Format from the top down, so we do.
+func (vm VmInfo) Format(state fmt.State, verb rune) {
+	switch {
+	case verb == 'v' && state.Flag('#'):
+		state.Write([]byte(fmt.Sprintf("api.VmInfo{Name:%q, Namespace:%q, Cpu:", vm.Name, vm.Namespace)))
+		vm.Cpu.format(state, verb)
+		state.Write([]byte(", Mem:"))
+		vm.Mem.format(state, verb)
+		state.Write([]byte(fmt.Sprintf(", AlwaysMigrate:%t, ScalingEnabled:%t}", vm.AlwaysMigrate, vm.ScalingEnabled)))
```
tychoish, line 145: I'm 100% sure that this doesn't matter, but it seems like we could maybe combine the number of calls here into one(ish?) sprintf?

pkg/agent/informant.go
```diff
@@ -859,7 +859,7 @@ func (s *InformantServer) Downscale(ctx context.Context, to api.Resources) (*api
 	s.runner.logger.Infof("Sending downscale %+v", to)
 
 	timeout := time.Second * time.Duration(s.runner.global.config.Informant.DownscaleTimeoutSeconds)
-	rawResources := to.ConvertToRaw(&s.runner.vm.Mem.SlotSize)
+	rawResources := to.ConvertToRaw(s.runner.vm.Mem.SlotSize)
```
tychoish, line 862: yay!

tychoish (approved):

PR #164, opened 2023-04-13
add vector back to pg14-disk-test image

sharnoff:


vm-examples/pg14-disk-test/Dockerfile.vmdata
```diff
@@ -40,6 +40,13 @@ RUN adduser vm-informant --disabled-password --no-create-home
 # but we still have proper handling around cgroups, etc.
 RUN echo "::respawn:su vm-informant -c '/bin/vm-informant --auto-restart --cgroup=neon-test'" >> /etc/inittab
 
+# Install vector.dev binary
+RUN set -e \
+    && wget https://packages.timber.io/vector/0.26.0/vector-0.26.0-x86_64-unknown-linux-musl.tar.gz -O - \
+    | tar xzvf - --strip-components 3 -C /bin/ ./vector-x86_64-unknown-linux-musl/bin/vector
+ADD vector.yaml /etc/vector.yaml
+RUN echo "::respawn:/bin/vector -c /etc/vector.yaml --config-dir /etc/vector" >> /etc/inittab
```
nikitakalyanov, line 48: I had to remove the --config-dir /etc/vector thing (because we already pass -c and there is no /etc/vector dir anyway). But otherwise seems to work

sharnoff: ah ok, makes sense - thanks! fixed.

PR #167, opened 2023-04-13
neonvm: Simplify builder templating code

sharnoff:


neonvm/tools/vm-builder-generic/main.go
```diff
@@ -211,23 +211,33 @@ func printReader(reader io.ReadCloser) error {
...
 }
 
-func AddToTar(tw *tar.Writer, filename string, buf bytes.Buffer) error {
+func AddTemplatedFileToTar(tw *tar.Writer, buf *bytes.Buffer, tmplData any, filename string, tmplString string) error {
+	tmpl, err := template.New(filename).Parse(tmplString)
+	if err != nil {
+		return fmt.Errorf("failed to parse template for %q: %w", filename, err)
+	}
+
+	buf.Reset()
```
tychoish, line 220: why pass in a buffer that we're just going to reset and then write from?

neonvm/tools/vm-builder-generic/main.go
```diff
@@ -298,98 +308,40 @@ func main() {
 		log.Fatalln(err)
 	}
 
-	var SrcImageSpec ImageSpec
+	tmplCtx := TemplatesContext{
```
tychoish, line 311: s/Ctx/Args/ or s/Ctx/Opts/ (just because context.Context)

neonvm/tools/vm-builder-generic/main.go
```diff
@@ -298,98 +308,40 @@ func main() {
...
+		SrcImage:   *srcImage,
+		Entrypoint: imageSpec.Config.Entrypoint,
+		Cmd:        imageSpec.Config.Cmd,
+		Env:        imageSpec.Config.Env,
+	}
 
 	if len(imageSpec.Config.User) != 0 {
```
tychoish, line 318: if this is a string let's do != ""

neonvm/tools/vm-builder-generic/main.go
```diff
@@ -298,98 +308,40 @@ func main() {
...
-		log.Fatalln(err)
-	}
+	add("Dockerfile", dockerfileVmBuilder)
+	add("vmstart", scriptVmStart)
+	add("inittab", scriptInitTab)
+	add("vmacpi", scriptVmAcpi)
+	add("powerdown", scriptPowerDown)
+	add("vminit", scriptVmInit)
 
 	dockerFileTarReader := bytes.NewReader(buf.Bytes())
```
tychoish, line 346: isn't buf a reader too? also I thought that we reset the buf a lot so isn't this, in effect only the contents of the last file?

neonvm/tools/vm-builder-generic/main.go
```diff
@@ -298,98 +308,40 @@ func main() {
...
-	_, err = b.WriteString(scriptPowerDown)
-	if err != nil {
-		log.Fatalln(err)
-	}
-	if err = AddToTar(tw, "powerdown", b); err != nil {
-		log.Fatalln(err)
+	add := func(filename string, tmpl string) {
+		if err := AddTemplatedFileToTar(tw, buf, tmplCtx, filename, tmpl); err != nil {
+			log.Fatalln(err)
+		}
```
tychoish, line 336: I'd just write a loop rather than do this functional thing.

neonvm/tools/vm-builder/main.go
```diff
@@ -369,30 +369,36 @@ func printReader(reader io.ReadCloser) error {
 	return nil
 }
 
-func AddToTar(tw *tar.Writer, filename string, buf bytes.Buffer) error {
+func AddTemplatedFileToTar(tw *tar.Writer, buf *bytes.Buffer, tmplData any, filename string, tmplString string) error {
```
tychoish, line 372: I just had deja vu, but I suppose you did too. Is there any reason not to pull things into a util (perhaps not now, and I'm happy to use my fingers to do this.

tychoish (approved):

neonvm/tools/vm-builder-generic/main.go
sharnoff: The idea behind keeping the buffer around is that the backing slice's capacity is reused. It's.... mildly more efficient, and in-line with the previous implementation.

That said, I think (a) the performance benefit is minimal and (b) the weirdness is significant. Removed it.

neonvm/tools/vm-builder-generic/main.go
sharnoff: Makes sense. Changed to `tmplArgs`.

neonvm/tools/vm-builder-generic/main.go
sharnoff: ehhh, yes 100% agree. There's kind of a lot going on in this PR already; I'll open another PR to do this (and others) separately.

neonvm/tools/vm-builder-generic/main.go
sharnoff: Yeah good point. Fixed.

neonvm/tools/vm-builder-generic/main.go
sharnoff: Makes sense. Fixed

PR #168, opened 2023-04-13
neonvm/vm-builder: require --enable-file-cache for vm-informant integration

sharnoff:
Builds on #167


neonvm/tools/vm-builder/main.go
```diff
@@ -491,6 +493,11 @@ func main() {
 		tmplCtx.Cmd = []string{"/neonvm/bin/sleep", "3650d"}
 	}
 
+	// if file cache enabled, add the connection string to the vm-informant's args
+	if *enableFileCache {
+		tmplCtx.InformantFileCacheArgs = `--pgconnstr="dbname=neondb user=cloud_admin sslmode=disable"`
```
tychoish, line 498: just to confirm: we **only** use the `pgconstr` if the filecache is enabled?

tychoish (commented):
sgtm

PR #170, opened 2023-04-14
build: Run initial `go build ./...` to use docker cache

sharnoff:
Haven't tested this, not actually sure if it has the effect I think it does.


sharnoff:
Judging from the timing compared to other PRs / CI runs, I think this is strictly worse than what we have now. Closing.

PR #172, opened 2023-04-14
Support fractional CPUs with a cgroup

nikitakalyanov:
This adds a group to run qemu in. We could then scale by fractional CPU by changing the cgroup


neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -103,22 +109,13 @@ type Guest struct {
 }
 
 type CPUs struct {
-	// +kubebuilder:validation:Minimum=1
```
nikitakalyanov, line 106: I had to remove these validations because they only work for integers. Should probably add manual code for that to the controller

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -412,31 +415,36 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 					return err
 				}
 				// compare guest spec and count of plugged
-				if *virtualmachine.Spec.Guest.CPUs.Use > int32(len(cpusPlugged)) {
+				if int32(virtualmachine.Spec.Guest.CPUs.Use.Value()) > int32(len(cpusPlugged)) {
 					// going to plug one CPU
```
tychoish, line 419: what is the type of `.Value()`?  type casting both to `int32` seems weird and if I were writing this I'd cast them both to plain `int`

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -598,6 +606,29 @@ func affinityForVirtualMachine(virtualmachine *vmv1.VirtualMachine) *corev1.Affi
...
+		return err
+	}
+
+	resp, err := client.Post(url, "application/json", bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("unexpected status %d", resp.StatusCode)
```
tychoish, line 627: I'd use `resp.Status` which is already a string.

do we want to do `!= 200` or just `>= 400` are there non 200 responses that aren't errors. 

do we want to do any amount of retrying? on these requests

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -598,6 +606,29 @@ func affinityForVirtualMachine(virtualmachine *vmv1.VirtualMachine) *corev1.Affi
 	return a
 }
 
+func notifyRunner(vm *vmv1.VirtualMachine, r resource.Quantity) error {
+	client := http.Client{Timeout: 5 * time.Second}
```
tychoish, line 610: does neonvm have contexts? feels like we'd want contexts here.

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+	if err := createCgroup(cgroupPath); err != nil {
+		log.Fatal(err)
+	}
+
+	if err := setCgroupLimit(qemuCPUs.minFraction, cgroupPath); err != nil {
 		log.Fatal(err)
 	}
+
+	go terminateQemuOnSigterm(vmSpec.QMP)
+	go listenForCPUChanges(vmSpec.RunnerPort, cgroupPath)
```
tychoish, line 569: we should have contexts for these goroutines

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+
+	qemuString := shellescape.QuoteCommand(append([]string{QEMU_BIN}, qemuCmd...))
+	cgropedQemu := fmt.Sprintf("echo $$ >> %s && %s", path.Join(cgroupPath, "cgroup.procs"), qemuString)
+	// start a new shell, cgroup it and then start qemu. Qemu will get cgroup from the parent sh while runner is not cgrouped
+	if err := execFg("sh", "-c", cgropedQemu); err != nil {
+		log.Println(err)
```
tychoish, line 575: why `log` and not `klog`

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+	}
+
+	log.Println("cleaning up cgroup...")
+	err = os.Remove(cgroupPath)
```
tychoish, line 579: do we not want removeall?

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+}
+
+type cgroupContextKey struct{}
+
+func listenForCPUChanges(port int32, cgroupPath string) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/cpu_change", handleCPUChange)
+	baseCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	baseCtx = context.WithValue(baseCtx, cgroupContextKey{}, cgroupPath)
```
tychoish, line 635: we should do something to clean up the lifecycle of the server: 
- there's no way for the context to be canceled before the server exits
- there's no way to coordinate the server's shutdown with the process's shutdown (e.g. the process will exit before the service shuts down).

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+		ReadTimeout:       5 * time.Second,
+		ReadHeaderTimeout: 5 * time.Second,
+		WriteTimeout:      5 * time.Second,
+		BaseContext: func(_ net.Listener) context.Context {
+			return baseCtx
+		},
+	}
+	err := server.ListenAndServe()
+	if errors.Is(err, http.ErrServerClosed) {
+		log.Println("server closed")
```
tychoish, line 648: identify the name of the server?

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+	} else if err != nil {
+		log.Fatalf("error starting server: %s\n", err)
```
tychoish, line 650: I'm pretty sure that the log helpers add a `\n` at the end.

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+	}
+}
+
+// isSystemCgroup checks if we have systemd-managed-like cgroup layout
+func isSystemdCgroup() (bool, error) {
+	_, err := os.Stat("/sys/fs/cgroup/systemd")
+	if errors.Is(err, fs.ErrNotExist) {
```
tychoish, line 657: there is an `os.IsNotExist` that I that I might read.

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+}
+
+func setCgroupLimit(fraction float64, cgroupPath string) error {
+	isV2, err := isCgroupv2(cgroupPath)
+	if err != nil {
+		return err
+	}
+	if isV2 {
+		data := fmt.Sprintf("%d %d", int(fraction*float64(cgroupPeriod)), cgroupPeriod)
+		err := os.WriteFile(path.Join(cgroupPath, "cpu.max"), []byte(data), 0644)
```
tychoish, line 673: why do you use `path` and not `filepath`

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+	err = os.WriteFile(path.Join(cgroupPath, "cpu.cfs_period_us"), []byte(fmt.Sprintf("%d", cgroupPeriod)), 0644)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(path.Join(cgroupPath, "cpu.cfs_quota_us"), []byte(fmt.Sprintf("%d", int(fraction*float64(cgroupPeriod)))), 0644)
+}
+
+func createCgroup(cgroupPath string) error {
+	err := os.Mkdir(cgroupPath, os.ModeDir)
+	if os.IsExist(err) {
```
tychoish, line 689: I'm confused by the decision/motivation for your choices for `os.IsExist` vs `os.IsNotExist` and `errors.Is(err, fs.IsExist)`

neonvm/runner/main.go
```diff
@@ -532,10 +544,198 @@ func main() {
...
+		// unlikely but possible
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to create cgroup for qemu %w", err)
```
tychoish, line 693: I believe for error wrapping to work you have to put a colon before the `%w`

pkg/api/types.go
```diff
@@ -145,9 +145,20 @@ func (r Resources) ValidateNonZero() error {
 	return nil
 }
 
+func (r Resources) SanityCheck() error {
+	if r.VCPU.Cmp(*resource.NewMilliQuantity(250, resource.BinarySI)) == -1 {
```
sharnoff, line 149: IMO 0.25 is too big here. I think the big thing we're worried about here is the "accidentally multiplied or divided by 1000", and it's possible we want less than 0.25 CPU in the future. I'd use 0.05 instead (x1000 would be 50 CPUs, so this should have few false-positives and only has false-negatives on really big VMs)

cc @tychoish for second opinion here

pkg/api/types.go
```diff
@@ -145,9 +145,20 @@ func (r Resources) ValidateNonZero() error {
...
+		return errors.New("VCPU is too small")
+	}
+	if r.VCPU.Cmp(*resource.NewMilliQuantity(128*1000, resource.BinarySI)) == 1 {
```
sharnoff, line 152: Same comment here about the bounds. 128 is maybe too small (again, maybe we want something bigger in the future). I'd use e.g. 512.

cc @tychoish for second opinion here.

pkg/plugin/state.go
```diff
@@ -182,6 +182,16 @@ type podState struct {
...
 
+type milliCPU uint32
+
+func FromResourceQuantity(r resource.Quantity) milliCPU {
+	return milliCPU(r.MilliValue())
+}
+
+func (m *milliCPU) toResourceQuantity() resource.Quantity {
+	return *resource.NewMilliQuantity(int64(*m), resource.BinarySI)
+}
```
sharnoff, line 193: This should be public, in `pkg/api`, and used there (and the autoscaler-agent) for CPU instead of `resource.Quantity`. This requires implementing custom formatting and JSON (un)marshalling, but I think that's worth the effort.

In particular, using `MilliCPU` in `api.Resources` should make a lot of the plugin changes simpler, I think.

pkg/plugin/state.go
```diff
@@ -182,6 +182,16 @@ type podState struct {
 // podMigrationState tracks the information about an ongoing pod's migration
 type podMigrationState struct{}
 
+type milliCPU uint32
+
+func FromResourceQuantity(r resource.Quantity) milliCPU {
```
sharnoff, line 187: Also `FromResourceQuantity` should be `MilliCPUFromResourceQuantity`

tests/e2e/autoscaling/00-assert.yaml
```diff
@@ -11,5 +11,5 @@ status:
   conditions:
     - type: Available
       status: "True"
-  cpus: 1
+  cpus: "1"
```
sharnoff, line 14: We shouldn't need to use a string here, I think? Curious why that's required if we do.

vm-deploy.yaml
```diff
@@ -5,15 +5,15 @@ metadata:
   name: postgres14-disk-test
   annotations:
     # In this example, these bounds aren't necessary. So... here's what they look like :)
-    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 1, "mem": "1Gi" }, "max": { "cpu": 4, "mem": "4Gi" } }'
+    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 4, "mem": "4Gi" } }'
```
sharnoff, line 8:
```suggestion
    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.5, "mem": "4Gi" } }'
```

sharnoff (commented):
Broadly like this PR is going in the right direction. A high-level point to consider:

Note about doing the actual cgroup operations — basically, better to just use [`containerd/cgroups`](https://github.com/containerd/cgroups) instead of reimplementing it yourself. The vm-informant is using it as well. (its API is not great in some areas, but IIRC updating CPU limits should be pretty ok)

Going to give another review after changes, should be much fewer comments :) Happy to discuss offline as well

neonvm/runner/main.go
sharnoff: Don't think it's required:

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand

https://pkg.go.dev/fmt#Errorf

neonvm/runner/main.go
sharnoff: for context: we use klog (almost) everywhere else

neonvm/runner/main.go
sharnoff: removeall is not necessary because cgroups are ✨special✨

neonvm/controllers/virtualmachine_controller.go
sharnoff: `Value()` is int64, and it rounds up

pkg/plugin/state.go
sharnoff: Realized the architecture doc was out of date, opened https://github.com/neondatabase/autoscaling/pull/179 which should hopefully make my reasoning for "this should be in `pkg/api` clearer"

cicdteam:
@nikitakalyanov 

I didn't inspected code changes deep but I have short question.

What will be if I set CPU settings like
```yaml
spec:
  guest:
    cpu:
      min: 250m
      max: 8
      use: 500m
```

nikitakalyanov:
@cicdteam What's so special about this particular config?
I tested with
```
spec:
  guest:
    cpu:
      min: 250m
      max: 1500m
      use: 250m
```
And I observed that it could upscale 250m->500m->1000m and then back 1000m->500m->250m.
BTW, I feel that we may need to tweak some settings. When I start a VM, during boot it reports high LA and gets upscaled to 1000m and then back - i.e. I don't even have to provide any load at all to see this happening. Not sure it is a good thing, maybe autoscaling settings are too agressive.

Makefile
```diff
@@ -80,7 +80,7 @@ generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="neonvm/..." \
```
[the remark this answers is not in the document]
nikitakalyanov: It is, this generation is not working at all without this change for me (i.e. nothing gets generated, old content remains). Feel free to extract this change in a separate PR

tests/e2e/autoscaling/00-assert.yaml
nikitakalyanov: This is the "canonical form" for Kubernetes. Even if we pass 1, it will work but the created resource will have "1". This assert checks Kubernetes output (not input), so we need quotes

Makefile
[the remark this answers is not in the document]
sharnoff: Opened https://github.com/neondatabase/autoscaling/pull/187

neonvm/controllers/virtualmachine_controller.go
nikitakalyanov: The retryring is already there because this code runs inside the Kubernetes reconciliation loop

neonvm/controllers/virtualmachine_controller.go
nikitakalyanov: It does, I added the code to pass it, thanks!

neonvm/runner/main.go
nikitakalyanov: Right. It may also not work at all as rm -r does not work on a cgroup

neonvm/runner/main.go
nikitakalyanov: This code is now removed

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -332,7 +332,7 @@ func (r *VirtualMachineMigrationReconciler) doReconcile(ctx context.Context, vir
 					return err
 				}
 				// compare guest spec and count of plugged
-				if *vm.Spec.Guest.CPUs.Use > int32(len(cpusPlugged)) {
+				if int32(vm.Spec.Guest.CPUs.Use.Value()) > int32(len(cpusPlugged)) {
 					// going to plug one CPU
 					err := QmpPlugCpuToRunner(virtualmachinemigration.Status.TargetPodIP, vm.Spec.QMP)
```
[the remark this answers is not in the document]
nikitakalyanov: Thanks!

neonvm/runner/main.go
```diff
@@ -55,6 +61,10 @@ const (
 	resolveDefaultPath = "/etc/resolv.conf"
 	// alternatePath is a path different from defaultPath, that may be used to resolve DNS. See Path().
 	resolveAlternatePath = "/run/systemd/resolve/resolv.conf"
+
+	cgroupPeriod      = 100000
```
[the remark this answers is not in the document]
nikitakalyanov: Done

neonvm/runner/main.go
nikitakalyanov: I tried to not use them in new code but there may be some places left

nikitakalyanov:
I addressed most of the comments. Leaving some threads open if more discussion pops up

I feel that the recent changes merged to main make things better. I no longer observe too aggressive upscaling during boot - I have to provide some load to see it. Also, the upscaling now could do +500m/+750m/+1000m if there is a significant load and not always by 250m intervals. Looks really nice to me, thanks!

neonvm/runner/main.go
sharnoff: Makes sense, good to leave as-is for now. Opened #218 to track this.

pkg/api/vminfo_test.go
```diff
@@ -31,8 +31,8 @@ func TestFormatting(t *testing.T) {
 		AlwaysMigrate:  false,
 		ScalingEnabled: true,
 	})
-	defaultFormat := "{Name:foo Namespace:bar Cpu:{Min:1 Max:5 Use:3} Mem:{Min:2 Max:6 Use:4 SlotSize:1Gi} ScalingConfig:&{LoadAverageFractionTarget:0.7} AlwaysMigrate:false ScalingEnabled:true}"
-	goSyntaxRepr := `api.VmInfo{Name:"foo", Namespace:"bar", Cpu:api.VmCpuInfo{Min:1, Max:5, Use:3}, Mem:api.VmMemInfo{Min:2, Max:6, Use:4, SlotSize:&resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}, ScalingConfig:&api.ScalingConfig{LoadAverageFractionTarget:0.7}, AlwaysMigrate:false, ScalingEnabled:true}`
+	defaultFormat := "{Name:foo Namespace:bar Cpu:{Min:1000 Max:5000 Use:3000} Mem:{Min:2 Max:6 Use:4 SlotSize:1Gi} ScalingConfig:&{LoadAverageFractionTarget:0.7} AlwaysMigrate:false ScalingEnabled:true}"
```
sharnoff, line 34: _May_ want to instead output `Cpu:{Min:1 Max:5 Use:3}` here, with `100m` for e.g. `MilliCPU(100)`. Thoughts?

pkg/api/vminfo_test.go
```diff
@@ -15,9 +15,9 @@ func TestFormatting(t *testing.T) {
 		Name:      "foo",
 		Namespace: "bar",
 		Cpu: api.VmCpuInfo{
-			Min: 1,
-			Max: 5,
-			Use: 3,
+			Min: 1000,
+			Max: 5000,
+			Use: 3000,
```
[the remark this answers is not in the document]
sharnoff: Also one of these should be fractional, to test that formatting.

pkg/api/vminfo_test.go
```diff
@@ -31,8 +31,8 @@ func TestFormatting(t *testing.T) {
...
-	goSyntaxRepr := `api.VmInfo{Name:"foo", Namespace:"bar", Cpu:api.VmCpuInfo{Min:1, Max:5, Use:3}, Mem:api.VmMemInfo{Min:2, Max:6, Use:4, SlotSize:&resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}, ScalingConfig:&api.ScalingConfig{LoadAverageFractionTarget:0.7}, AlwaysMigrate:false, ScalingEnabled:true}`
+	defaultFormat := "{Name:foo Namespace:bar Cpu:{Min:1000 Max:5000 Use:3000} Mem:{Min:2 Max:6 Use:4 SlotSize:1Gi} ScalingConfig:&{LoadAverageFractionTarget:0.7} AlwaysMigrate:false ScalingEnabled:true}"
+	goSyntaxRepr := `api.VmInfo{Name:"foo", Namespace:"bar", Cpu:api.VmCpuInfo{Min:1000, Max:5000, Use:3000}, Mem:api.VmMemInfo{Min:2, Max:6, Use:4, SlotSize:&resource.Quantity{i:resource.int64Amount{value:1073741824, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1Gi", Format:"BinarySI"}}, ScalingConfig:&api.ScalingConfig{LoadAverageFractionTarget:0.7}, AlwaysMigrate:false, ScalingEnabled:true}`
```
sharnoff, line 35: IMO CPU field should be `api.VmCpuInfo{Min:api.MilliCPU(1000), Max:api.MilliCPU(5000), Use:api.MilliCPU(3000)}` - requires a change in the formatting code in `pkg/api/vminfo.go`.

Thoughts?

pkg/plugin/run.go
```diff
@@ -218,7 +218,8 @@ func (e *AutoscaleEnforcer) handleResources(
 	vCPUTransition := collectResourceTransition(&node.vCPU, &pod.vCPU)
 	memTransition := collectResourceTransition(&node.memSlots, &pod.memSlots)
 
-	vCPUVerdict := vCPUTransition.handleRequested(req.VCPU, startingMigration)
+	vCPUVerdict := vCPUTransition.handleRequested(
+		req.VCPU, startingMigration)
```
sharnoff, line 222:
```suggestion
	vCPUVerdict := vCPUTransition.handleRequested(req.VCPU, startingMigration)
```

unrelated formatting change?

pkg/plugin/config.go
```diff
@@ -226,10 +226,10 @@ func (c *nodeConfig) vCpuLimits(total *resource.Quantity) (_ nodeResourceState[u
...
 	margin.Sub(c.Cpu.System)
 
-	return nodeResourceState[uint16]{
-		Total:                uint16(totalRounded),
-		System:               uint16(systemCpus),
-		Watermark:            uint16(c.Cpu.Watermark * float32(reservableCpus)),
+	return nodeResourceState[api.MilliCPU]{
+		Total:                api.MilliCPU(totalRounded * 1000),
+		System:               api.MilliCPU(systemCpus * 1000),
+		Watermark:            api.MilliCPU(c.Cpu.Watermark * float32(reservableCpus) * 1000),
```
sharnoff, line 232: There's definitely a more thorough to make this better (i.e. don't count "full" CPUs), but realistically that's part of a broader change to the scheduler to move away from counting everything in terms of compute units.

So: this is going to add a bit of scheduling inefficiency, but that's ok for now I think. Just want to flag this for later.

pkg/plugin/state.go
```diff
@@ -327,7 +326,7 @@ func (r *nodeOtherResourceState) calculateReserved(memSlotSize *resource.Quantit
 
 // totalReservableCPU returns the amount of node CPU that may be allocated to VM pods -- i.e.,
 // excluding the CPU pre-reserved for system tasks.
-func (s *nodeState) totalReservableCPU() uint16 {
+func (s *nodeState) totalReservableCPU() api.MilliCPU {
```
sharnoff, line 329: note: this is called in `pkg/plugin/plugin.go`, used as: `totalCpu := ...`. That variable should be renamed to `totalMilliCpu`. (there's also `maxTotalCpu` below it, which should be renamed in the same way).

pkg/util/arith.go
```diff
@@ -25,7 +27,16 @@ func Max[T constraints.Ordered](x, y T) T {
 	}
 }
 
-// Max returns the minimum of the two values
+// Max returns the minimum of the two resource.Quantity values
+func MaxQuantity(x, y resource.Quantity) resource.Quantity {
```
sharnoff, line 31: Not used anymore

pkg/util/arith.go
```diff
@@ -34,6 +45,15 @@ func Min[T constraints.Ordered](x, y T) T {
 	}
 }
 
+// Min returns the minimum of the two resource.Quantity values
+func MinQuantity(x, y resource.Quantity) resource.Quantity {
```
sharnoff, line 49: Not used anymore

sharnoff (commented):
It's looking good! _Should_ just be one final round of review after this, I think.

There's a couple larger comments buried among the "make this small change" ones. Feel free to message me or comment here for more context. 

A final note: There's a bunch of resource formatting in the scheduler and autoscaler-agent. As currently written, this change will cause all the places where we mention CPU to be multiplied by 1000, so logs will go from looking like:

```
I0424 16:09:34.251442       1 run.go:227] [autoscale-enforcer] Handled resources from pod kuttl-test-sharing-cricket/example-9hzc9 AgentRequest.
	vCPU verdict: Register 1 -> 1 (pressure 0 -> 0); node reserved 1 -> 1 (of 11), node capacityPressure 0 -> 0 (0 -> 0 spoken for)
	 mem verdict: Register 1 -> 1 (pressure 0 -> 0); node reserved 1 -> 1 (of 62), node capacityPressure 0 -> 0 (0 -> 0 spoken for)
```

([source](https://github.com/neondatabase/autoscaling/actions/runs/4788499642/jobs/8515190304#step:16:314) - logs from most recent CI run on main)

to looking like:

```
I0424 16:09:34.251442       1 run.go:227] [autoscale-enforcer] Handled resources from pod kuttl-test-sharing-cricket/example-9hzc9 AgentRequest.
	vCPU verdict: Register 1000 -> 1000 (pressure 0 -> 0); node reserved 1000 -> 1000 (of 11000), node capacityPressure 0 -> 0 (0 -> 0 spoken for)
	 mem verdict: Register 1 -> 1 (pressure 0 -> 0); node reserved 1 -> 1 (of 62), node capacityPressure 0 -> 0 (0 -> 0 spoken for)
```

I kind of mentioned this briefly in one of the comments, but I _think_ it's probably best to have custom formatting for `MilliCPU` so that instead of multiplying all logged values by 1000, we see integer CPU represented the same, and e.g. 100m for 0.1 CPU — to align with formatting for `resource.Quantity`.

With that, there's some amount of finding all the places where CPU gets formatted that'd be required. So: If it'd be helpful, I'm happy to make a PR into this one to implement that.

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -264,7 +261,7 @@ type VirtualMachineStatus struct {
 	// +optional
 	Node string `json:"node,omitempty"`
 	// +optional
-	CPUs int `json:"cpus,omitempty"`
+	CPUs resource.Quantity `json:"cpus,omitempty"`
```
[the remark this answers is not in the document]
nikitakalyanov: Sounds right to me. Also the men field is a pointer too

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -412,33 +415,38 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 					return err
 				}
 				// compare guest spec and count of plugged
-				if *virtualmachine.Spec.Guest.CPUs.Use > int32(len(cpusPlugged)) {
+				if int32(virtualmachine.Spec.Guest.CPUs.Use.Value()) > int32(len(cpusPlugged)) {
```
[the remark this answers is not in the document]
nikitakalyanov: +1. BTW, is there some special reason why it was done this way initially (with int32 and not int that `Len` already returns)?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -412,33 +415,38 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
 				// update status by count of CPU cores used in VM
-				virtualmachine.Status.CPUs = len(cpusPlugged)
+				virtualmachine.Status.CPUs = *virtualmachine.Spec.Guest.CPUs.Use
 				// record event about cpus used in VM
 				r.Recorder.Event(virtualmachine, "Normal", "CpuInfo",
-					fmt.Sprintf("VirtualMachine %s uses %d cpu cores",
+					fmt.Sprintf("VirtualMachine %s uses %v cpu cores",
 						virtualmachine.Name,
-						virtualmachine.Status.CPUs))
+						&virtualmachine.Status.CPUs))
```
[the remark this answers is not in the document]
nikitakalyanov: I was thinking of adding a GET endpoint too, because otherwise it just feels wrong to update the cgroup every second (every reconciliation loop iteration)

neonvm/runner/main.go
```diff
@@ -427,7 +443,7 @@ func main() {
 
 	// create iso9660 disk with runtime options (command, args, envs, mounts)
 	if err = createISO9660runtime(runtimeDiskPath, vmSpec.Guest.Command, vmSpec.Guest.Args, vmSpec.Guest.Env, vmSpec.Disks); err != nil {
-		log.Fatalln(err)
+		log.Fatalf("Failed to createISO9660runtime %s", err)
```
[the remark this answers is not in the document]
nikitakalyanov: Ooops, forgot to do this

neonvm/runner/main.go
```diff
@@ -532,17 +548,199 @@ func main() {
...
+		return
+	}
+
+	parsed := api.VCPUChange{}
+	err = json.Unmarshal(body, &parsed)
+	if err != nil {
+		log.Printf("could not parse body: %s\n", err)
+		w.WriteHeader(500)
+		return
+	}
```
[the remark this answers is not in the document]
nikitakalyanov: Makes sense

neonvm/runner/main.go
```diff
@@ -532,17 +548,199 @@ func main() {
...
+
+type cgroupContextKey struct{}
+
+func listenForCPUChanges(ctx context.Context, port int32, cgroupPath string, wg *sync.WaitGroup) {
+	defer wg.Done()
+	mux := http.NewServeMux()
+	mux.HandleFunc("/cpu_change", handleCPUChange)
+	baseCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	baseCtx = context.WithValue(baseCtx, cgroupContextKey{}, cgroupPath)
```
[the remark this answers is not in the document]
nikitakalyanov: Thanks!

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: I'd imagine that `Spec.Guest.CPUs.Use` was `*int32` instead of `*int` so that the size of the integer wasn't platform-dependent (maybe kubebuilder complains?). But I don't know for sure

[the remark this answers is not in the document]
nikitakalyanov: Yeah, but isn't it better to upcast to not lose data? Like int(int32(1)) is better than int32(int(1))

neonvm/runner/main.go
```diff
@@ -55,6 +63,12 @@ const (
 	resolveDefaultPath = "/etc/resolv.conf"
 	// alternatePath is a path different from defaultPath, that may be used to resolve DNS. See Path().
 	resolveAlternatePath = "/run/systemd/resolve/resolv.conf"
+
+	// cgroupPeriod is the period for evaluating cgroup quota
+	// in microseconds. Min 1000 microseconds, max 1 second
+	// the actual quota is <= this period
+	cgroupPeriod     = uint64(100000)
```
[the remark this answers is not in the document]
nikitakalyanov: I googled a stack overflow answer that this quota works independent of number of cores. If you have a 25000 quota and 4 cores then it could be one core at 100% or 4 cores at 25%, or anything in between. Apparently, this is a lie, as I can confirm your results. Thanks for noticing!

pkg/api/vminfo_test.go
nikitakalyanov: I changed the formatting code but haven't checked all its usages yet. Plan to do that gradually as I read some logs

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
nikitakalyanov: BTW, how do we handle old runners that won't reply to this (and won't handle POST either)? We could have code that handles it gracefully, but I think we also have to work with full CPU for old runners in scheduler to avoid extra overcommit because of the absence of cgroup

pkg/plugin/config.go
nikitakalyanov: Could you elaborate what's the deficiency here (out of curiosity)?

neonvm/runner/main.go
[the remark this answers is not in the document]
nikitakalyanov: Should be ok now

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Ah, sorry yeah. I think _in general_ it's better, but realistically we're never going to have more than 2^32 CPUs plugged

So, I'm not @cicdteam, but I know that personally, I'd rather make the more unknown type clear (`len` is broadly known, but someone reading may not know that `CPUs.Use` is `int32`).

Either way, should be fine :smile:

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Oooh, yeah good point. I think a reasonable path forward is making "failed to connect" (or whatever equivalent network error) a warning, not hard error — for now. Then, once all old runners are gone, we change it to a hard error with a future update. HTTP error responses can still be a hard error though, because we should only be getting valid HTTP for new runners.

neonvm/runner/main.go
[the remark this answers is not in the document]
sharnoff: > Apparently, this is a lie

:grimacing:

well, glad we got there in the end :)

pkg/plugin/config.go
sharnoff: (note: this is defined by `vCpuLimits` ([pkg/plugin/config.go](https://github.com/neondatabase/autoscaling/blob/b206290f5898f263a79682fb08de9d1ec485b503/pkg/plugin/config.go#L206-L237)) and `buildInitialNodeState` ([pkg/plugin/state.go](https://github.com/neondatabase/autoscaling/blob/b206290f5898f263a79682fb08de9d1ec485b503/pkg/plugin/state.go#L476-L553)) which calls it)

Sure: As _currently_ written (again: it's bad, we should change it) - the scheduler determines the allocatable CPU as follows:

1. Set `total := Node.Status.Allocatable().Cpu()` (or `.Capacity().Cpu()` if not present) — provided by the K8s API
2. Subtract "system" CPUs from `total` (amount of CPU reserved for daemon pods, etc)
3. Round down `total` to the nearest integer

So basically we're rounding `total` down, which is now not needed. This is not _really_ clear in `vCpuLimits`, unfortunately, because there's some other things it's trying to do as well.

There may also be some issues with how `otherPodResources` interacts with this - going to double-check that during my next review.

neonvm/runner/main.go
```diff
@@ -532,17 +548,230 @@ func main() {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	wg := sync.WaitGroup{}
+
+	wg.Add(1)
+	go terminateQemuOnSigterm(ctx, vmSpec.QMP, &wg)
+	wg.Add(1)
+	go listenForCPUChanges(ctx, vmSpec.RunnerPort, cgroupPath, &wg)
+
+	args := append([]string{"-g", fmt.Sprintf("cpu:%s", cgroupPath), QEMU_BIN}, qemuCmd...)
+	log.Printf("build qemu args: %s", args)
```
sharnoff, line 567: Not sure what "build qemu args" is referring to here. Seems like this is actually the arguments to cgexec? +1 to logging this, but description should be more clear, I think.

pkg/agent/billing.go
```diff
@@ -52,14 +53,14 @@ func (m *metricsTimeSlice) Duration() time.Duration { return m.endTime.Sub(m.sta
 
 type vmMetricsInstant struct {
 	// cpu stores the cpu allocation at a particular instant.
-	cpu uint16
+	cpu api.MilliCPU
```
sharnoff, line 56: Thinking about this more, I think `cpu` would actually be better off as a float all the way through. The difference should be negligible though, so happy to leave that for a follow-up.

pkg/agent/billing.go
```diff
@@ -250,7 +252,7 @@ func (s *billingMetricsState) drainAppendToBatch(conf *BillingConfig, batch *bil
 			EndpointID:     key.endpointID,
 			StartTime:      s.pushWindowStart,
 			StopTime:       now,
-			Value:          int(math.Round(history.total.activeTime.Seconds())),
+			Value:          math.Round(history.total.activeTime.Seconds()),
```
sharnoff, line 255:
```suggestion
			Value:          history.total.activeTime.Seconds(),
```

We can drop the `math.Round` now that we don't need to convert to an integer :)

pkg/agent/runner.go
```diff
@@ -1373,16 +1373,16 @@ func (s *atomicUpdateState) computeUnitsBounds() (uint16, uint16) {
...
 
 	// note: floor(x / M) + 1 gives the minimum integer value greater than x / M.
 
 	if s.requestedUpscale.Cpu {
-		required = util.Max(required, s.vm.Cpu.Use/s.computeUnit.VCPU+1)
+		required = util.Max(required, uint32(s.vm.Cpu.Use/s.computeUnit.VCPU)+1)
 	}
 	if s.requestedUpscale.Memory {
-		required = util.Max(required, s.vm.Mem.Use/s.computeUnit.Mem+1)
+		required = util.Max(required, uint32(s.vm.Mem.Use/s.computeUnit.Mem+1))
```
sharnoff, line 1385: The difference in positioning of the parens at the end of these lines looks incorrect (I know it's actually fine, but best to keep them consistent).

I'd change the second one to match the first (i.e. `uint32(...)+1`)

pkg/api/types.go
```diff
@@ -488,3 +501,73 @@ type SuspendAgent struct {
...
+// it represents the vCPU usage as controlled by cgroup
+type VCPUCgroup struct {
+	VCPUs resource.Quantity
+}
+
+// MilliCPU is a special type to represent vCPUs * 1000
+// e.g. 2 vCPU is 2000, 0.25 is 250
+type MilliCPU uint32
+
+// MilliCPUFromResourceQuantity converts resource.Qunatity into MilliCPU
```
sharnoff, line 525:
```suggestion
// MilliCPUFromResourceQuantity converts resource.Quantity into MilliCPU
```

:D

pkg/api/types.go
```diff
@@ -488,3 +501,73 @@ type SuspendAgent struct {
...
+func MilliCPUFromResourceQuantity(r resource.Quantity) MilliCPU {
+	return MilliCPU(r.MilliValue())
+}
+
+// ToResourceQuantity converts a MilliCPU to resource.Qunatity
```
sharnoff, line 530:
```suggestion
// ToResourceQuantity converts a MilliCPU to resource.Quantity
```

pkg/api/types.go
```diff
@@ -488,3 +501,73 @@ type SuspendAgent struct {
...
+
+// this a similar version type for controller <-> runner communications
+// see PluginProtoVersion comment for details
+type RunnerProtoVersion uint32
+
+const (
+	RunnerProtoV1_0 RunnerProtoVersion = iota + 1
+)
+
+func (v RunnerProtoVersion) SupportsCgroup() bool {
```
sharnoff, line 571: Name here maybe could be improved? I'd probably say `SupportsCgroupFractionalCPU`, but I don't know. I'm thinking along the lines of: "Can this be more clear to someone who doesn't know the details runner's implementation?".

Thoughts?

pkg/plugin/config.go
sharnoff: Double-checked: Looks like this diff made it ok, because it's no longer rounding:

```diff
diff --git a/pkg/plugin/state.go b/pkg/plugin/state.go
index dc5d508..f2ef947 100644
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -301,8 +301,7 @@ func (r *nodeOtherResourceState) calculateReserved(memSlotSize *resource.Quantit
 		// set cupCopy := r.rawCpu - r.marginCpu
 		cpuCopy := r.RawCPU.DeepCopy()
 		cpuCopy.Sub(*r.MarginCPU)
-		// note: Value() rounds up, which is the behavior we want here.
-		r.ReservedCPU = uint16(cpuCopy.Value())
+		r.ReservedCPU = api.MilliCPUFromResourceQuantity(cpuCopy)
 	}
 
 	// If rawMemory doesn't exceed the margin ..., set reserved = 0
```

so no issues :)

tests/e2e/autoscaling/00-assert.yaml
sharnoff: Resolving, now that it's been added

pkg/util/arith.go
```diff
@@ -25,7 +25,7 @@ func Max[T constraints.Ordered](x, y T) T {
 	}
 }
 
-// Max returns the minimum of the two values
+// Min returns the minimum of the two values
```
sharnoff, line 28: Oops, good catch. Opened #226 so it's not bundled into this PR.

neonvm/runner/main.go
sharnoff: linking my review from a few days ago, re: this topic: https://github.com/neondatabase/autoscaling/pull/172/files#r1175697800

> This has the trailing `\n` but honestly I think it's mostly fine. There's existing code that has it, so a full cleanup can be done in a separate PR.

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: We talked about this elsewhere; a different approach is now implemented (explicit versioning).

pkg/api/vminfo_test.go
sharnoff: Resolving, now that it was implemented

sharnoff (commented):
Some final comments, plus another semi-significant change that I forgot we needed. Everything but that last bit is approved (pending resolving comments). Future review will only focus on new changes.

---

Very sorry I forgot until now, but: technically this should also come with an accompanying major version bump of the agent<->plugin protocol (and then the scheduler should work with integer CPU for older autoscaler-agents). Requires special support in `(resourceTransition[T]).handleRequested()` (or just a separate function implementing the old behavior). Also _ideally_ we'd handle that gracefully in `(*AutoscaleEnforcer).handleAgentRequest` as well, by converting CU to one with integer CPU (e.g. 0.25+1Gi -> 1+4Gi). Another option is to just do a two-phase deploy, only updating the config to use fractional CPU in the CU after everything's updated to support it.

For testing cross-version support, I'd recommend doing the whole "run tests locally" thing, but with the autoscaler-agent's image changed to `neondatabase/autoscaler-agent:v0.6.0`. A newer vm-informant should still be ok, because the change from #203 was backwards-compatible both ways (refer to [`pkg/api/VERSIONING.md`](https://github.com/neondatabase/autoscaling/blob/159a1f1eabe4332c6190b31867b0b8f6c2f0ec80/pkg/api/VERSIONING.md#agent-informant-protocol) for more)

---

Also, played around with this PR locally, figured this would be easier to suggest here: to make things line up with the updated default CU, I'd recommend the following:

```diff
diff --git a/tests/e2e/autoscaling/00-create-vm.yaml b/tests/e2e/autoscaling/00-create-vm.yaml
index e6fd326..b03ff4c 100644
--- a/tests/e2e/autoscaling/00-create-vm.yaml
+++ b/tests/e2e/autoscaling/00-create-vm.yaml
@@ -23,13 +23,13 @@ metadata:
   labels:
     autoscaling.neon.tech/enabled: "true"
   annotations:
-    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 1, "mem": "1Gi" }, "max": { "cpu": 4, "mem": "4Gi" } }'
+    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 250m, "mem": "1Gi" }, "max": { "cpu": 1, "mem": "4Gi" } }'
 spec:
   schedulerName: autoscale-scheduler
   guest:
     cpus:
-      min: 1
-      max: 5 # set value greater than bounds so our tests check we don't exceed the bounds.
+      min: 0.25 
+      max: 1.25 # set value greater than bounds so our tests check we don't exceed the bounds.
       use: 1
     memorySlotSize: 1Gi
     memorySlots:
diff --git a/vm-deploy.yaml b/vm-deploy.yaml
index f556f2e..d7a6940 100644
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -5,7 +5,7 @@ metadata:
   name: postgres14-disk-test
   annotations:
     # In this example, these bounds aren't necessary. So... here's what they look like :)
-    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.5, "mem": "4Gi" } }'
+    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.25, "mem": "5Gi" } }'
   labels:
     autoscaling.neon.tech/enabled: "true"
     # Set to "true" to continuously migrate the VM (TESTING ONLY)
@@ -13,9 +13,9 @@ metadata:
 spec:
   schedulerName: autoscale-scheduler
   guest:
-    cpus: { min: 0.25, use: 0.25, max: 1.5 }
+    cpus: { min: 0.25, use: 0.25, max: 1.25 }
     memorySlotSize: 1Gi
-    memorySlots: { min: 1, use: 1, max: 4 }
+    memorySlots: { min: 1, use: 1, max: 5 }
     rootDisk:
       image: pg14-disk-test:dev
       size: 8Gi
```

And also, because average CPU has been made smaller, I'd recommend reducing the load from `scripts/run-bench.sh`:

```diff
diff --git a/scripts/run-bench.sh b/scripts/run-bench.sh
index 8e09def..ef62b32 100755
--- a/scripts/run-bench.sh
+++ b/scripts/run-bench.sh
@@ -31,7 +31,7 @@ apk add postgresql-client
 echo '"'$QUERY'"' > factorial.sql
 echo "Running pgbench. Query:"
 echo '"'   $QUERY'"'
-pgbench -h '"$vm_ip"' -U postgres -c 20 -T 1000 -P 1 -f factorial.sql postgres
+pgbench -h '"$vm_ip"' -U postgres -c 10 -T 1000 -P 1 -f factorial.sql postgres
 '
 
 # FIXME: same multus stuff as above.
```

(reduces concurrent connections from 20 -> 10, which should help a lot)

pkg/plugin/run.go
```diff
@@ -175,17 +177,33 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
...
 }
 
+// getComputeUnitForResponse tries to return compute unit that the agent supports
+// if we have a fractional CPU but the agent does not support it we multiply the result
+func getComputeUnitForResponse(node *nodeState, supportsFractional bool) api.Resources {
+	computeUnit := *node.computeUnit
+	if !supportsFractional {
+		initialCU := computeUnit
+		i := uint16(2)
+		for ; computeUnit.VCPU < 1000; i++ {
```
sharnoff, line 193:
```suggestion
		for i := uint16(2); computeUnit.VCPU % 1000 != 0; i++ {
```

As currently written, this will be incorrect if `1000 % computeUnit.VCPU != 0` (e.g. if `VCPU` is 750m, we will see that `1500 >= 1000` and exit, which would still give the autoscaler-agent a non-integer value)

Another option, food for thought: Multiply `computeUnit` by the least common multiple of `computeUnit.VCPU` and 1000. Maybe overkill (LCM is annoying); "dumb" solution is ok.

pkg/plugin/run.go
```diff
@@ -175,17 +177,33 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
...
+
+	return computeUnit
+}
+
 func (e *AutoscaleEnforcer) handleResources(
 	pod *podState,
 	node *nodeState,
 	req api.Resources,
 	startingMigration bool,
+	suportsFractional bool,
```
sharnoff, line 206:
```suggestion
	suportsFractionalCPU bool,
```

pkg/plugin/trans.go
```diff
@@ -152,6 +153,10 @@ func (r resourceTransition[T]) handleRequested(requested T, startingMigration bo
 			// adjust node pressure accordingly. We can have old < new or new > old, so we shouldn't
 			// directly += or -= (implicitly relying on overflow).
 			r.node.CapacityPressure = r.node.CapacityPressure - r.oldPod.capacityPressure + r.pod.CapacityPressure
+			if onlyThousands {
+				// this branch is for the case when caller requests increase in thousands
+				maxIncrease = T(math.Floor(float64(maxIncrease) / 1000.0))
+			}
```
sharnoff, line 159: I believe this is incorrect: `maxIncrease` should just round down, not fully divide. Also, while it's probably not an issue here, float conversion is scary.

The placing here is also not _necessarily_ incorrect (provided `increase % 1000 == 0`, which caller must verify), but I feel it conveys intent better if it's outside the `if increase > maxIncrease`.

Here's what I'd recommend:

```diff
diff --git a/pkg/plugin/trans.go b/pkg/plugin/trans.go
index d745a6b..06f734d 100644
--- a/pkg/plugin/trans.go
+++ b/pkg/plugin/trans.go
@@ -148,6 +148,11 @@ func (r resourceTransition[T]) handleRequested(requested T, startingMigration bo
 		increase := requested - r.pod.Reserved
 		// Increases are bounded by what's left in the node
 		maxIncrease := remainingReservable
+		// if only in thousands, round down maxIncrease to nearest multiple of 1000
+		if onlyThousands {
+			thousand := T(100) * 10 // avoid compiler complaining about 1000 > maximum int8
+			maxIncrease = (maxIncrease / thousand) * thousand
+		}
 		if increase > maxIncrease /* increases are bound by what's left in the node */ {
 			r.pod.CapacityPressure = increase - maxIncrease
 			// adjust node pressure accordingly. We can have old < new or new > old, so we shouldn't
```

pkg/plugin/run.go
sharnoff, line 206: Also: should have some validation here, returning HTTP 400 if `!supportsFractionalCPU && req.VCPU % 1000 != 0`

pkg/api/types.go
```diff
@@ -41,6 +42,15 @@ const (
 	// Currently the latest version.
 	PluginProtoV1_1
 
+	// PluginProtoV2_0 represents v2.0 of the agent<->scheduler plugin protocol.
+	//
+	// Changes from v1.1:
+	//
+	// * Supports fractional CU
```
sharnoff, line 49:
```suggestion
	// * Supports fractional CPU
```

Technically, this isn't fractional CU, but CPU

pkg/api/types.go
```diff
@@ -41,6 +42,15 @@ const (
...
+	//
+	// Currently the latest version.
```
sharnoff, line 51: > Currently the latest version.

`PluginProtoV1_1` should be updated to say "Last used in release version v0.6.0."

pkg/api/types.go
```diff
@@ -41,6 +42,15 @@ const (
...
+	PluginProtoV2_0
```
sharnoff, line 52: Need to update `pkg/api/VERSIONING.md` to add this as current version

sharnoff (approved):
Final comments - great work! :)

PR #173, opened 2023-04-14
Switch to '/' for namespace/name separator, not ':'

sharnoff:
Quoting the comment that marked this as a todo:

> K8s uses '/' as its separator. We currently use this for backwards-compatibility, but we should change to '/'. That change will involve fixing the many places where we create strings with %s:%s instead of %s/%s as well; those should just use NamespacedName directly.

So that should hopefully explain why this change is good, and what's actually going on here.

Planning to merge both commits separately, not squash.

### DIFF
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -148,8 +148,8 @@ type podState struct {
 	// name will not change after initialization, so it can be accessed without holding a lock.
 	name util.NamespacedName
 
-	// vmName is the name of the VM, as given by the 'vm.neon.tech/name' label.
-	vmName string
+	// vmName is the name of the VM, as given by the 'vm.neon.tech/name' label (and name.Namespace)
+	vmName util.NamespacedName
 
 	// testingOnlyAlwaysMigrate is a test-only debugging flag that, if present in the pod's labels,
 	// will always prompt it to mgirate, regardless of whether the VM actually *needs* to.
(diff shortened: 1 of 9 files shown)


pkg/plugin/state.go
```diff
@@ -669,15 +669,17 @@ func (e *AutoscaleEnforcer) handlePodDeletion(podName util.NamespacedName) {
...
-	klog.Infof("Handling updated scaling bounds for VM pod %s:%s", vm.Namespace, podName)
+	podName := util.NamespacedName{Namespace: vm.Namespace, Name: unqualifiedPodName}
 
-	pod, ok := e.state.podMap[util.NamespacedName{Namespace: vm.Namespace, Name: podName}]
+	klog.Infof("Handling updated scaling bounds for VM pod %v", podName)
+
+	pod, ok := e.state.podMap[podName]
 	if !ok {
-		klog.Errorf("[autoscale-enforcer] cannot find VM pod %s:%s to update scaling bounds", vm.Namespace, podName)
+		klog.Errorf("[autoscale-enforcer] cannot find VM pod %v to update scaling bounds", podName)
```
tychoish, line 682: isn't this stil a string/stringer?

pkg/util/watch.go
```diff
@@ -318,8 +316,8 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 							old, ok := store.objects[uid]
 							if !ok {
 								return fmt.Errorf(
-									"watch %s: received update event for object %s that's not present",
-									config.LogName, name(meta),
+									"watch %s: received update event for object %v that's not present",
```
tychoish, line 319: is the name not a string/stringer?

tychoish (commented):
all good except for some questions about formatting spec.

tychoish (approved):
we had some conversations IRL and I think this is good.

pkg/plugin/state.go
sharnoff: discussed IRL - not a stringer because it's implemented with Format

pkg/util/watch.go
sharnoff: same as above

PR #174, opened 2023-04-14
agent: Add "unhealthy vm-informants" metric

sharnoff:
Builds on #92. Should allow things like this to be detected via metrics: https://neondb.slack.com/archives/C03H1K0PGKH/p1681461362196019?thread_ts=1681459030.472469&cid=C03H1K0PGKH

### DIFF
--- a/pkg/agent/globalstate.go
+++ b/pkg/agent/globalstate.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tychoish/fun/pubsub"
@@ -130,6 +131,9 @@ func (s *agentState) handleVMEventAdded(
 		errored:  nil,
 		panicked: false,
 		vmInfo:   event.vmInfo,
+
+		startTime:                   time.Now(),
+		lastSuccessfulInformantComm: nil,
 	}
 
 	runner := &Runner{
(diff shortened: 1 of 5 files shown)


pkg/agent/informant.go
```diff
@@ -646,6 +646,16 @@ func (s *InformantServer) handleID(ctx context.Context, body *struct{}) (_ *api.
 		return nil, 404, errors.New("Server has already exited")
 	}
 
+	// Update our record of the last successful time we heard from the informant, if the server is
+	// currently enabled. This allows us to detect cases where the informant is not currently
+	// communicating back to the agent - OR when the informant never /resume'd the agent.
+	if s.mode == InformantServerRunning {
+		s.runner.setStatus(func(s *podStatus) {
+			now := time.Now()
```
tychoish, line 654: my gut feeling is that we might want to trim/round this to millisecond, because the monotonic/nanosecond properties  of time.Now() might create some noise?

pkg/agent/globalstate.go
```diff
@@ -209,13 +216,30 @@ type podStatus struct {
...
+	defer s.mu.Unlock()
+
+	startupGracePeriod := time.Second * time.Duration(config.Informant.UnhealthyStartupGracePeriodSeconds)
+	unhealthySilencePeriod := time.Second * time.Duration(config.Informant.UnhealthyAfterSilenceDurationSeconds)
+
+	if s.lastSuccessfulInformantComm == nil {
+		return time.Since(s.startTime) >= startupGracePeriod
+	} else {
+		return time.Since(*s.lastSuccessfulInformantComm) >= unhealthySilencePeriod
+	}
```
tychoish, line 240: I'd read from the time source once.

tychoish (approved):
This is great. All my comments are mostly just "minor anxieties about nanosecond timekeeping", which we can take or leave.

pkg/agent/globalstate.go
sharnoff: Talked IRL, this is ok

pkg/agent/informant.go
sharnoff: I think it _should_ be fine (and not rounding is simpler/less weird than rounding, I think), but leaving this comment open for posterity.

PR #177, opened 2023-04-14
agent: tidy+improve config validation

tychoish:


pkg/agent/config.go
```diff
@@ -108,56 +110,34 @@ func ReadConfig(path string) (*Config, error) {
...
-		return cannotBeZero(".dumpState.port")
-	} else if c.DumpState != nil && c.DumpState.TimeoutSeconds == 0 {
-		return cannotBeZero(".dumpState.timeoutSeconds")
-	}
-
-	return nil
+	ec := &erc.Collector{}
+
+	const (
+		emptyTmpl = "filed %q cannot be empty"
```
sharnoff, line 116:
```suggestion
		emptyTmpl = "field %q cannot be empty"
```

sharnoff (approved):
Looks good. Still a bit unwieldy, but I think this is probably the cleanest "no magic" solution we can get, which I appreciate :)

I'd probably rename the PR to something more like "agent: tidy+improve config validation" (as it is, I feel like it implies that there wasn't config validation already) — up to you though

tychoish:
> Looks good. Still a bit unwieldy, but I think this is probably the cleanest "no magic" solution we can get, which I appreciate :)

Yeah, I'm generally not a fan of the magic. I think it'll end up being unwieldy if someone has a config with *many* errors in it, but with two or three, I think the error message will be pretty clear. There's an `erc.Unwind()` function that will turn the aggregated error into a slice of errors, if you want to (at some point) pull apart the output error. 

Also I just sorted all of the calls lexically, out of lack of a better order. I believe that the presentation of the errors is the reverse of the collection order (which kind of makes sense from an `errors.Is/errors.As` situation), so if you have better ideas about how to order these so more important things end up earlier in the message, we can do that later.

> I'd probably rename the PR to something more like "agent: tidy+improve config validation" (as it is, I feel like it implies that there wasn't config validation already) — up to you though

Ah, yeah, fair enough. your wording is clear. That implication wasn't my intent. 

I'm going to fix the merge conflict and try and get this landed soon, but there are some small tweaks that we can follow up with.

PR #178, opened 2023-04-14
bump: tychoish/fun

tychoish:
I think I'm mostly done (at least for now, probably most changes will be if only if we/cloud need more/different things).

### DIFF
--- a/cmd/vm-informant/main.go
+++ b/cmd/vm-informant/main.go
@@ -25,9 +25,9 @@ const minSubProcessRestartInterval = 5 * time.Second
 func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM)
 	defer cancel()
-	ctx = srv.SetShutdown(ctx)      // allows workers to cause a shutdown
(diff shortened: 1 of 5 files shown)


cmd/vm-informant/main.go
```diff
@@ -80,8 +80,9 @@ func main() {
 		}
 
 		runRestartOnFailure(ctx, args, cleanupHooks)
-		// this cancels the context
-		srv.GetShutdown(ctx)()
+		closer := srv.GetShutdownSignal(ctx)
+		// this cancels the process' underlying context
+		closer()
```
sharnoff, line 85:
```suggestion
		cancel := srv.GetShutdownSignal(ctx)
		// this cancels the process' underlying context
		cancel()
```

Not sure if this is better, but a thought. Take it or leave it

sharnoff (approved):

cmd/vm-informant/main.go
tychoish: I think maybe it should be `shutdown`, but I'm indifferent.

PR #180, opened 2023-04-15
agent: Add runner panics metrics

sharnoff:
Typically, panics would be visible in other ways, like K8s events. But because the autoscaler-agent isolates panics just to the threads handling a single VM, these can go unnoticed unless we do something about it.

### DIFF
--- a/pkg/agent/prommetrics.go
+++ b/pkg/agent/prommetrics.go
@@ -94,5 +94,6 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
 		schedulerRequests:         schedulerRequests,
 		informantRequestsOutbound: informantRequestsOutbound,
 		informantRequestsInbound:  informantRequestsInbound,
+		runnerThreadPanics:        runnerThreadPanics,
 	}, reg
 }
--- a/pkg/agent/prommetrics.go
+++ b/pkg/agent/prommetrics.go
@@ -9,6 +9,7 @@ type PromMetrics struct {
 	schedulerRequests         *prometheus.CounterVec
 	informantRequestsOutbound *prometheus.CounterVec
 	informantRequestsInbound  *prometheus.CounterVec
+	runnerThreadPanics        prometheus.Counter
 }
 
 func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Registry) {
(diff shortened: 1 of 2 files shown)


pkg/agent/prommetrics.go
```diff
@@ -33,6 +34,37 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
...
+		},
+	)
+	totalPanickedVMs := prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name: "autoscaling_panicked_vm_runners_current",
+			Help: "Number of VMs whose per-VM runner has panicked (and not restarted)",
+		},
+		func() float64 {
+			globalstate.lock.Lock()
+			defer globalstate.lock.Unlock()
```
tychoish, line 50: It looks like these functions are called pretty often (the frequency wasn't super clear to me but maybe once per metrics scrape? maybe there's another interval?) and I worry some about the lock contention (particularly given the global lock and the fact that we need to take and release a bunch of locks while holding other locks. 

In the short term this is probably fine, in the longer term:
- maybe panicked can be an atomic so we don't need the inner lock?
- maybe we can have metrics for: number started, number gracefully shutdown, and number currently running and use subtraction to infer this number.

sharnoff: Called once per metrics scrape, yep. Every 10s, currently (although maybe that's way too often).

I also considered the lock contention issue. I think realistically it's not a problem, although _it would be nice_ if it weren't.

I didn't want metrics for number started / number shutdown / etc because that creates a "multiple sources of truth" situation, which _realistically_ is fine, but felt kinda meh.

tychoish: I don't think that using subtraction really counts as multiple sources of truth: this only ends up being inaccurate if there are lots of updates during metrics collection _and_ the atomics in prom are slow (which they aren't really), or if we end up incorrectly recording data (which can always happen?)

Faced with "maybe we'll contend a global lock" vs "there's a small chance of transient off-by-one" errors, I'd pick the second one all the time.

sharnoff: Makes sense. I think I care most about getting visibility into this ASAP, but agree that this should be changed in the future. Opened #198 for this.

PR #181, opened 2023-04-15
api: Make scalingBounds public

sharnoff:
Allows the scaling bounds to be used directly in cloud, instead of duplicating the type, as in https://github.com/neondatabase/cloud/pull/4650.

### DIFF
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -137,12 +137,12 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
 	}
 
 	if boundsJSON, ok := vm.Annotations[AnnotationAutoscalingBounds]; ok {
-		var bounds scalingBounds
+		var bounds ScalingBounds
 		if err := json.Unmarshal([]byte(boundsJSON), &bounds); err != nil {
 			return nil, fmt.Errorf("Error unmarshaling annotation %q: %w", AnnotationAutoscalingBounds, err)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/api/vminfo.go
```diff
@@ -179,17 +179,24 @@ func (vm *VmInfo) applyBounds(b scalingBounds) {
...
 }
 
-type scalingBounds struct {
-	Min *resourceBound `json:"min,omitempty"`
-	Max *resourceBound `json:"max,omitempty"`
+// ScalingBounds is the type that we deserialize from the "autoscaling.neon.tech/bounds" annotation
+//
+// All fields (and sub-fields) are pointers so that our handling can distinguish between "field not
+// set" and "field equal to zero". Please note that all field are still required to be set and
+// non-zero, though.
```
tychoish, line 186: I feel like the error "<x> must be set to a non-zero value" is clear and actionable and it seems like the set/unset level means that a user might not set the value, set it to zero, retry and then get the next error (possibly multiplied (ish) for both fields,) when we can simplify the validation logic significantly and provide more actionable feedback to users more quickly.

sharnoff: That's a really good point; I'll change that before merging this PR.

tychoish: also if they're not pointers the logic gets simpler and you can do the `ec := &erc.Collector{}; erc.When(ec, cond, "message")` to get all the errors at once to avoid the extra retry cycle

sharnoff: Opened #190 to do this

PR #188, opened 2023-04-17
Update VM CRD for #140

sharnoff:
Realized I never ran `make manifests` during the changes from #140.


cicdteam:
may be revert this step (make generate) into CICD flow ?

PR #189, opened 2023-04-17
e2e: always get k8s logs + events

sharnoff:
Follow-up to a conversation with @nikitakalyanov yesterday. Currently we get logs/events on everything but success, which means that when you see errors in logs, there's no way to tell if those are "normal" or not.

cc @bayandin in case there's reasons not to do this.


bayandin (approved):
Yep, totally makes sense, thanks!

PR #191, opened 2023-04-17
lint: Check that `make manifests` has no changes

sharnoff:
ref https://github.com/neondatabase/autoscaling/pull/188#issuecomment-1510896020

Expecting this to fail for now, until #188 is merged.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -41,3 +41,14 @@ jobs:
             echo "ERROR: 'go mod tidy' modified the source tree."
             exit 1
           fi
+
+      - name: make manifests
+        run: |
+          make manifests
+          if ! (test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"); then
+            git ls-files --exclude-standard --others .
+            git diff .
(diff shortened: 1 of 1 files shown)


sharnoff:
Ok, seems to be failing as expected :) https://github.com/neondatabase/autoscaling/actions/runs/4719485521/jobs/8370358257

.github/workflows/lint.yml
```diff
@@ -41,3 +41,14 @@ jobs:
             echo "ERROR: 'go mod tidy' modified the source tree."
             exit 1
           fi
+
+      - name: make manifests
```
tychoish, line 45: does it make sense to combine this in some way with the previous task (e.g. is the working directory reset? would the state of the repo be dirty when this starts? if go mod tidy fails, does this task even run?)

sharnoff: IMO it's worth keeping separate so that we can tell which step modified the repo.

> is the working directory reset?

Nope

> would the state of the repo be dirty when this starts? if go mod tidy fails, does this task even run?

No to both

PR #195, opened 2023-04-18
agent: Rework scaling algorithm

### LINKED ISSUE #82: Rework scaling algorithm
The "scaling algorithm" is defined exclusively in the `(*atomicUpdateState).desiredVMState()` method, in `pkg/agent/runner.go`. There's some instructive comments:

```go
// FIXME: This should have *some* access to prior scaling decisions, so that we can e.g. use slower
// scaling to start, and accelerate it over time.
//
// FIXME: Even factoring in the above, this implementation is *pretty bad*.
func (s *atomicUpdateState) desiredVMState(allowDecrease bool) api.Resources {
	// There's some annoying edge cases that this function has to be able to handle properly. For
	// the sake of completeness, they are:
	//
	// 1. s.vm.Using() is not a multiple of s.computeUnit
	// 2. s.vm.Max() is less than s.computeUnit (or: has at least one resource that is)
	// 3. s.vm.Using() is a fractional multiple of s.computeUnit, but !allowDecrease and rounding up
	//    is greater than s.vm.Max()
	// 4. s.vm.Using() is much larger than s.vm.Min() and not a multiple of s.computeUnit, but load
	//    is low so we should just decrease *anyways*.
	//
	// ---
	// Now, it's worth noting that we only *barely* handle the edge cases above. We don't do it
	// well, but we do handle them in a protocol-compliant way, and that's what counts! Eventually,
	// this function will be rewritten, this note removed, and all will be well.

	...
}
```

## Current algorithm

The general idea of the current version is that we double CU when load is >0.9×CPU and halve CU when load is <0.4×CPU. The rest of the function is basically just edge-case handling (making sure we're within bounds, etc.)

Doubling is great for quick demos, but not suitable for long-term usage.

## Desired algorithm

Realistically, we need #81 for the best implementation. However, that might take a while, and we can do better with what we have.

A clean option IMO is that we aim for target CPU ≈ load average. We can make a configurable parameter that controls how close to cut it (e.g., load average = 0.8 × target CPU).

We should also have some kind of "maximum speed of change" (i.e. "don't scale by more than X CU at a time"), optionally with some kind of acceleration (i.e. "start with small changes, make bigger ones if load is still very different from allocation").

sharnoff:
The new algorithm essentially just tries to target the current compute units equal to load average (modulo some fixed under/overcommit), relying on the 1-minute load average to smooth things out.

Most of this PR is just passing the "fixed under/overcommit" config around, into the function itself.

Closes #82.

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -1,11 +1,15 @@
 package agent
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 
+	"github.com/tychoish/fun"
 	"github.com/tychoish/fun/erc"
+
+	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
 type Config struct {
@@ -30,6 +34,9 @@ type DumpStateConfig struct {
 type ScalingConfig struct {
 	// RequestTimeoutSeconds gives the timeout duration, in seconds, for VM patch requests
 	RequestTimeoutSeconds uint `json:"requestTimeoutSeconds"`
+	// DefaultConfig gives the default scaling config, to be used if there is no configuration
+	// supplied with the "autoscaling.neon.tech/config" annotation.
+	DefaultConfig api.ScalingConfig `json:"defaultConfig"`
 }
 
 type InformantConfig struct {
@@ -145,6 +152,10 @@ func (c *Config) validate() error {
 	erc.Whenf(ec, c.Metrics.LoadMetricPrefix == "", emptyTmpl, ".metrics.loadMetricPrefix")
 	erc.Whenf(ec, c.Metrics.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.secondsBetweenRequests")
 	erc.Whenf(ec, c.Scaling.RequestTimeoutSeconds == 0, zeroTmpl, ".scaling.requestTimeoutSeconds")
+	if err := c.Scaling.DefaultConfig.Validate(); err != nil {
+		stack := err.(*erc.Stack)
+		fun.Observe(context.TODO(), stack.Iterator(), ec.Add)
+	}
 	erc.Whenf(ec, c.Scheduler.RequestPort == 0, zeroTmpl, ".scheduler.requestPort")
 	erc.Whenf(ec, c.Scheduler.RequestTimeoutSeconds == 0, zeroTmpl, ".scheduler.requestTimeoutSeconds")
 	erc.Whenf(ec, c.Scheduler.SchedulerName == "", emptyTmpl, ".scheduler.schedulerName")
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -52,6 +52,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"runtime/debug"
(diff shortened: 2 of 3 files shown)


pkg/agent/config.go
```diff
@@ -145,6 +152,12 @@ func (c *Config) validate() error {
 	erc.Whenf(ec, c.Metrics.LoadMetricPrefix == "", emptyTmpl, ".metrics.loadMetricPrefix")
 	erc.Whenf(ec, c.Metrics.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.secondsBetweenRequests")
 	erc.Whenf(ec, c.Scaling.RequestTimeoutSeconds == 0, zeroTmpl, ".scaling.requestTimeoutSeconds")
+	if err := c.Scaling.DefaultConfig.Validate(); err != nil {
+		stack := err.(*erc.Stack) //nolint:errorlint // return ty guaranteed by Validate()
+		if err := fun.Observe(context.TODO(), stack.Iterator(), ec.Add); err != nil {
+			panic(fmt.Errorf("unexpected failure Observe-ing errors: %w", err))
+		}
+	}
```
tychoish, line 160: oh, you can just `ec.Add()` the error and [it'll do what you want](https://github.com/tychoish/fun/blob/main/erc/errors.go#L33) (append the input stack to the collector's stack.)

tychoish: also, there's `erc.Unwind` which returns an error slice which might be more usable in some situations.

also there's this which might help in  the future: https://github.com/tychoish/fun/blob/main/erc/helpers.go#L245

pkg/agent/runner.go
```diff
@@ -1300,29 +1304,26 @@ func (s *atomicUpdateState) desiredVMState(allowDecrease bool) api.Resources {
...
-	goalCU := currentCU
-	if s.metrics.LoadAverage1Min > 0.9*float32(s.vm.Cpu.Use) {
-		goalCU *= 2
-	} else if s.metrics.LoadAverage1Min < 0.4*float32(s.vm.Cpu.Use) && allowDecrease {
-		goalCU /= 2
-	}
+	// Goal compute unit is at the point where (CPUs) × (LoadAverageFractionTarget) == (load
+	// average),
+	// which we can get by dividing LA by LAFT.
+	goalCU := uint16(math.Round(float64(s.metrics.LoadAverage1Min) / s.config.LoadAverageFractionTarget))
```
tychoish, line 1316: it might be instructive/demonstrative to factor this (and some surrounding code) out into functions that you can write unittests: just so the inputs/expected outputs can be written down somewhere and validated.

pkg/agent/config.go
sharnoff: Done

pkg/agent/runner.go
sharnoff: Yes, 100%. I definitely want tests for this. Currently I feel like the "agent is 5 packages stacked on top of each other pretending to be 1 package" situation is making this harder, but we can _probably_ add some tests anyways.

This will conflict with #172, so I'd like to wait until after that's done to do so. Opened #199 to track this post-merge.

tychoish (approved):

PR #202, opened 2023-04-19
lint: Always run golangci-lint

sharnoff:
There are sometimes non-Go files that affect linting (e.g., changes to the linting workflow itself, as in #201) and realistically the longest CI job is the e2e tests, so being efficient about linting isn't as significant.


bayandin (commented):
We also need to delete `if: env.GIT_DIFF` (from `golangci/golangci-lint-action@v3` step itself).

In general, I'm ok with these changes, alternatively, we can add the workflow itself `.github/workflows/lint.yml` to `PATTERNS`

PR #204, opened 2023-04-19
e2e: show previous logs if restarted

sharnoff:
Currently there's no good way to debug a crash if the pod's logs on restart don't show what went wrong. Ran into this while working on #203.


sharnoff:
Intentionally cancelled the job, to trigger fetching logs faster. The logs step _seems_ ok, AFAICT

PR #209, opened 2023-04-20
agent, plugin: Allow restricted namespace

sharnoff:
cc @seymourisdead


tychoish (approved):
this seems reasonable.

PR #214, opened 2023-04-23
agent/informant: Stop informant server when context canceled

### LINKED ISSUE #37: Agent's informant server doesn't stop when context is cancelled

### LINKED ISSUE #196: autoscaler-agent leaks goroutines
Spotted on staging (v0.6.0):

Screenshot of grafana, graph titled "autoscaler-agent number of goroutines (total)", showing some spikes, but an overall increase

The individual blips up-and-down are from cluster autoscaling, I believe - the general trend line is primarily due to a single instance that's gradually increasing.

sharnoff:
Without this the autoscaler-agent slowly leaks goroutines each time a new VM is handled.

Fixes #37, fixes #196.

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -257,7 +257,7 @@ func NewInformantServer(
 
 	// Thread waiting for the context to be canceled so we can use it to shut down the server
 	shutdownWaiterName := fmt.Sprintf("InformantServer shutdown waiter (%s)", server.desc.AgentID)
-	runner.spawnBackgroundWorker(ctx, shutdownWaiterName, func(c context.Context) {
+	runner.spawnBackgroundWorker(ctx, shutdownWaiterName, func(context.Context) {
 		// Wait until parent context OR server's context is done.
 		<-backgroundCtx.Done()
 		server.exit(InformantServerExitStatus{Err: nil, RetryShouldFix: false})
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -114,7 +114,8 @@ type InformantServerState struct {
 }
 
 type InformantServerExitStatus struct {
-	// Err is the non-nil error that caused the server to exit
+	// Err is the error, if any, that caused the server to exit. This is only non-nil when context
+	// used to start the server becomes canceled (i.e. the Runner is exiting).
 	Err error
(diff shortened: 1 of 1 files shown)


pkg/agent/informant.go
```diff
@@ -254,6 +255,14 @@ func NewInformantServer(
 		}
 	})
 
+	// Thread waiting for the context to be canceled so we can use it to shut down the server
+	shutdownWaiterName := fmt.Sprintf("InformantServer shutdown waiter (%s)", server.desc.AgentID)
+	runner.spawnBackgroundWorker(ctx, shutdownWaiterName, func(c context.Context) {
+		// Wait until parent context OR server's context is done.
+		<-backgroundCtx.Done()
```
tychoish, line 262: it looks like the way this gets canceled is if `server.exit` runs, which this function then calls in the next line (am I reading this correctly?) I suppose other things call exit too.) 

If the parent context gets canceled (e.g. the one passed to spawnBackgroundWorker) will this run? I suppose it will because nothing is waiting on `c` (I might just delete that variable in the signature to make that more clear). 

it seems plausible that this could help, but it's not totally clear to me that this is *definitely* the right solution: moving L204-213 out of the exit function blocking on background/the input context being canceled seems _roughly_ equally functional. Thought I don't have a perfect mental model for all of the threads and where the informant monitoring parts in the agent can exit, so if this is the best way, then let's do it?

sharnoff: > it looks like the way this gets canceled is if server.exit runs, which this function then calls in the next line (am I reading this correctly?) I suppose other things call exit too.)
>
> If the parent context gets canceled (e.g. the one passed to spawnBackgroundWorker) will this run?

Yeah so `server.exit` is called from a handful of places, _usually_ due to some error. (before this PR, _only_ due to an error) `backgroundCtx` is derived from the parent context, so all should be good on that front.

> (I might just delete that variable in the signature to make that more clear)

done

> moving L204-213 out of the exit function blocking on background/the input context being canceled seems roughly equally functional

At first I was about to agree & implement this. However: other parts of the code rely on `server.exitStatus != nil` being equivalent to "the server is no longer running", and `server.exit` sets `exitStatus`, so avoid it would create weird behavior when the context is canceled.

That said, maybe the semantics there are wrong.

tychoish (approved):

PR #220, opened 2023-04-24
Add .gitattributes, mark CRDs as generated

sharnoff:
Further reading:

* https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github
* https://git-scm.com/docs/gitattributes

Noticed while reviewing #172. With #191, CRDs shouldn't generally need human review, I think. Curious to hear other opinions on this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+neonvm/config/common/crd/bases/* linguist-generated=true


tychoish (approved):
seems good to me. github typically seems to do a decent job at finding generated files without help... if anything it's a little over eager, and some diffs end up getting hidden more than the churn  in generated files ends up biting us.

PR #231, opened 2023-04-28
neonvm: Add current usage annotation to runner Pod

sharnoff:
The annotation is based on the value of Spec.Guest.*.Use, not the status, because that's what the rest of the autoscaling system uses to make decisions with.

To observe this annotation changing live, you can use:

    kubectl get pod -w -o jsonpath='{.metadata.name}{": "}{.metadata.annotations.vm\\.neon\\.tech/usage}{"\n"}'

(different shells may handle the backslashes differently; the argument passed to kubectl should have single backslashes, like vm\\.neon\\.tech)

This feature is probably the easiest path towards patching cluster-autoscaler to work with our VMs - it avoids needing to pass through an extra neonvm client everywhere.

~~This PR currently semantically conflicts with #172 and `VirtualMachineUsage.CPU` should be updated after that merges.~~

### DIFF
--- a/neonvm/Dockerfile
+++ b/neonvm/Dockerfile
@@ -12,6 +12,7 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
+COPY pkg/util            pkg/util
 COPY neonvm/main.go      neonvm/main.go
 COPY neonvm/apis/        neonvm/apis/
 COPY neonvm/controllers/ neonvm/controllers/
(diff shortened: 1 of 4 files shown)


sharnoff:
NB: The NeonVM images from this PR are currently built and pushed as `sharnoff/neonvm-*:v0.7.0-alpha1`, and deployed in `dev-eu-central-1-alpha`.

PR #232, opened 2023-04-28
Add cluster-autoscaler patch

### LINKED ISSUE #77: cluster-autoscaler does not work with VMs
The basic problem is that cluster-autoscaler uses pods' `resources.requests` and NeonVM doesn't set that (but also _can't_ correctly set that, unless it upgraded to K8s v1.27).

For now, the plan is to provision our metal nodes manually. That's not a good long-term solution though. As I see it, here's a few possible options:

1. Wait until K8s v1.27 is available on EKS, make NeonVM set `resources.requests`
2. Fork cluster-autoscaler, teach it about VMs
3. Build cluster scaling logic into the scheduler (or expose info for something else)

None of these options are _good_. The best scaling behavior requires knowledge that only the scheduler has (e.g. whether a pod was rejected during Filter or Reserve - the second should be retried, but the first means there's no room).

It's also worth considering that we can combine options: e.g. fork cluster-autoscaler for now, then change to setting `resources.requests` for VMs once K8s v1.27 is available.

sharnoff:
Refer to the added [`cluster-autoscaler/README.md`](./cluster-autoscaler/README.md) for the "what". The tl;dr is:

> This directory contains a patch for [cluster-autoscaler] (CA) that allows it to correctly interpret the resource usage of VMs.

Resolves #77. Requires #231.

## Implementation

The "how":

The basic idea here is based on a few observations:

1. cluster-autoscaler currently doesn't work with VMs, because;
    1. cluster-autoscaler uses resources.requests (because internally it calls a copy of the _default_ k8s scheduler)
    2. If CA thinks that there _should_ be room for a pod, it won't scale up
    3. NeonVM cannot set resources.requests because it needs to change (and we're using K8s 1.24)
    4. so: there's no easy fix on the VM/autoscaling side
2. Implementing our own version of cluster-autoscaler is _possible_, but annoying
3. cluster-autoscaler is _really big_ — the source code itself is 403 MiB, and that's not including the other things in the repo
4. cluster-autoscaler contains a _single file_ that all Pod `List` requests originate from (more below)

So, with all that, my hypothesis ~~(which has not been tested)~~ is that cluster-autoscaler only fetches pod information with `List` requests, and that by setting resources.requests to match the VM's resource usage at the point the pods are retrieved, this will fix the issue.

The patch introduces a wrapper around the `ListWatch` that's fed into the typical pod listers that modifies the resources.requests of any pod with the `vm.neon.tech/usage` annotation, before passing it on.

---

note: to find all the places Pods were listed, I used:

```sh
rg --pcre2 '\.List\b\(|func.*\bList\(' -g '!vendor' -g '!cloudprovider'
```

## Remaining items

- [ ] Review of general idea
- [x] Review of patched Go code
- [ ] Review of workflows
- [x] Modify release workflow to build CA
- [x] Testing this works
- [x] Add `cluster-autoscaler` to ARCHITECTURE.md


sharnoff:
Not sure exactly what's causing CI to fail here. "IO timeout" is a little vague :disappointed:

sharnoff:
I went with `gcr.io/distroless/static:latest-amd64` because that's what CA uses. However: it looks like distroless has deprecated a bunch of images, `static` included:

> Any other tags are considered deprecated and are no longer updated

Looks like `static-debian11` is probably our best bet?

rahulinux:
@sharnoff is this running somewhere in dev to inspect logs and behavior?

cluster-autoscaler/ca.patch
```diff
@@ -0,0 +1,145 @@
...
++			wrappedEvents := make(chan watch.Event)
++			proxyIface := watch.NewProxyWatcher(wrappedEvents)
++
++			go func() {
++				defer iface.Stop()
++
++				events := iface.ResultChan()
++
++				for {
++					var ev watch.Event
```
tychoish, line 102: I would declare on 107 and nest the remainder of the function in that arm of the select statment.

tychoish (approved):

sharnoff:
Finished a bunch of testing for this PR, in its current state.

Testing was done in `dev-eu-central-1-alpha` with the latest commit built as `sharnoff/cluster-autoscaler-neonvm:v0.7.3-alpha3`.

Full writeup is here: https://neondb.slack.com/archives/C03TN5G758R/p1683697723436409?thread_ts=1683697717.657759&cid=C03TN5G758R

sharnoff:
Got the ok from @rahulinux, merging to include in the pre-release. If there's issues, we can fix in post.

PR #233, opened 2023-04-28
e2e: use 5m timeout on actions/setup-go

sharnoff:
Sometimes it gets stuck, and won't get killed for a very long time.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -16,6 +16,8 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
+        # Sometimes setup-go gets stuck. Without this, it'll keep going until the job gets killed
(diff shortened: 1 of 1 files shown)


tychoish (approved):
do you know what this happens? 

what's the timeout that this overrides?

sharnoff:
> do you know what this happens?

Some issue pulling from cache in `actions/setup-go`. It hangs sometimes

> what's the timeout that this overrides?

Not sure where it's configured, but IIRC it takes 6 hours to get killed.

bayandin:
> Here's a great example of the timeout in action: https://github.com/neondatabase/autoscaling/actions/runs/4862960306/jobs/8670317530

This is a kinda known issue with `actions/cache`; with `neondatabase/neon` we hit it as well with build artifacts caches and were able to solve it only by switching to manually uploading/downloading cache to s3

PR #235, opened 2023-05-01
disable service links

nikitakalyanov:
if we have a lot of services in one namespace service links lead to pods failing to start due to args too long


sharnoff:
Didn't know about `enableServiceLinks` before, but: My understanding is that this has an effect inside the VM because the environment variables will get passed through.

If that's true, I think I'd prefer having this as a field on the VM spec, passing it through that way, instead of changing the default like here. Thoughts?

nikitakalyanov:
Yeah, sounds good. I've changed the code

sharnoff (approved):

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -637,6 +637,7 @@ func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
 			Annotations: virtualmachine.Annotations,
 		},
 		Spec: corev1.PodSpec{
+			EnableServiceLinks:            virtualmachine.Spec.ServiceLinks,
```
cicdteam, line 640: would it trigger errors when `virtualmachine.Spec.ServiceLinks` node defined (`null`)

nikitakalyanov: It shouldn't. If you don't pass this field, it is already null because this is the default value of the bool pointer. So, nothing should change

PR #244, opened 2023-05-08
billing: Revert value to int, not float

sharnoff:
This was part of #172 that I'd mistakenly suggested. It turns out the control plane actually cannot accept fractional values for billing values.


sharnoff:
Validated with a shim billing implementation, numbers look right. Merging w/o review so that staging isn't left broken.

PR #261, opened 2023-05-14
neonvm/api: Add (MilliCPU).AsFloat64()

sharnoff:
In general, resource.Quantity's AsApproximateFloat64() method is less precise than if we do the conversion ourselves. This PR adds a new method to do that, and switches to it in all our usage.

---

Originally spotted this in the logs. As a reproduction, take for example, this program:

```go
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/api/resource"
)

func main() {
	milli := int64(27278)

	q := resource.NewMilliQuantity(milli, resource.BinarySI)
	fmt.Println("resource:", q.AsApproximateFloat64())
	f := float64(milli) / 1000
	fmt.Println("float:", f)
}
```

With current K8s versions, outputs:

```
resource: 27.278000000000002
float: 27.278
```

(Go playground link: https://go.dev/play/p/S7gW9hxgE7C)

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -179,6 +179,14 @@ func (m MilliCPU) ToResourceQuantity() *resource.Quantity {
 	return resource.NewMilliQuantity(int64(m), resource.BinarySI)
 }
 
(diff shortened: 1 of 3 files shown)


nikitakalyanov (approved):
Nice!

PR #271, opened 2023-05-16
agent/watch: Add custom vmEvent formatting

sharnoff:
This matches the existing vmInfo formatting, so that logs that print vmEvent will be able to have better formatting. Without this, we get things like:

    Handling VM event {... vmInfo:{... SlotSize:0xc0006cce00} ...}

("..." added for brevity)

### DIFF
--- a/pkg/agent/watch.go
+++ b/pkg/agent/watch.go
@@ -123,3 +123,30 @@ func makeVMEvent(vm *vmapi.VirtualMachine, kind vmEventKind) (vmEvent, error) {
 		podIP:   vm.Status.PodIP,
 	}, nil
 }
+
+// custom formatting for vmEvent so that it prints in the same way as VmInfo
+func (e vmEvent) Format(state fmt.State, verb rune) {
+	switch {
+	case verb == 'v' && state.Flag('#'):
+		state.Write([]byte(fmt.Sprintf(
+			// note: intentionally order podName and podIP before vmInfo because vmInfo is large.
+			"agent.vmEvent{kind:%q, podName:%q, podIP:%q, vmInfo:%#v}",
+			e.kind, e.podName, e.podIP, e.vmInfo,
+		)))
+	default:
+		if verb != 'v' {
+			state.Write([]byte("%!"))
+			state.Write([]byte(string(verb)))
+			state.Write([]byte("(api.VmInfo="))
+		}
+
+		state.Write([]byte(fmt.Sprintf(
+			"{kind:%s podName:%s podIP:%s vmInfo:%s}",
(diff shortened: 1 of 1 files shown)


pkg/agent/watch.go
```diff
@@ -123,3 +123,30 @@ func makeVMEvent(vm *vmapi.VirtualMachine, kind vmEventKind) (vmEvent, error) {
...
+		state.Write([]byte(fmt.Sprintf(
+			// note: intentionally order podName and podIP before vmInfo because vmInfo is large.
+			"agent.vmEvent{kind:%q, podName:%q, podIP:%q, vmInfo:%#v}",
+			e.kind, e.podName, e.podIP, e.vmInfo,
+		)))
+	default:
+		if verb != 'v' {
+			state.Write([]byte("%!"))
+			state.Write([]byte(string(verb)))
+			state.Write([]byte("(api.VmInfo="))
```
tychoish, line 140: why do you sprintf and call Write once in other cases?

pkg/agent/watch.go
```diff
@@ -123,3 +123,30 @@ func makeVMEvent(vm *vmapi.VirtualMachine, kind vmEventKind) (vmEvent, error) {
...
 	}, nil
 }
+
+// custom formatting for vmEvent so that it prints in the same way as VmInfo
+func (e vmEvent) Format(state fmt.State, verb rune) {
+	switch {
+	case verb == 'v' && state.Flag('#'):
+		state.Write([]byte(fmt.Sprintf(
+			// note: intentionally order podName and podIP before vmInfo because vmInfo is large.
+			"agent.vmEvent{kind:%q, podName:%q, podIP:%q, vmInfo:%#v}",
```
tychoish, line 133: this is prefixed as `api.VmInfo=` below... seems like we could be consistent there?

also just to confirm this is for controlling how things are printed to logs, or is there another consumer of this format?

tychoish (approved):

pkg/agent/watch.go
tychoish: > also just to confirm this is for controlling how things are printed to logs, or is there another consumer of this format?

sorry, see the description of the PR now.

pkg/agent/watch.go
sharnoff: Honestly, only because `api.VmInfo` does it that way, and to match that. Happy to change if you think `sprintf` would be cleaner

https://github.com/neondatabase/autoscaling/blob/2f2b73eb8f8802c7a6be9012b23e3c171b1a8a60/pkg/api/vminfo.go#L298-L302

tychoish (approved):

PR #273, opened 2023-05-16
agent: Restart panicked/errored runners (+metrics)

sharnoff:
This doesn't directly fix any individual bug, but it should lessen the impact of some behaviors we're currently seeing.

The bulk of this PR is adding a new function, `(*agentState).TriggerRestartIfNecessary`. It is called each time the Runner exits, and queues restarting the Runner if it failed (from panic or other fatal error). If the Runner was long-running, it's immediately restarted. If it has not been up for long, it is restarted after a random delay of 5-10s.

This PR also:

* Changes `podStatusDump` to have a record of previous "end states" (time, error, why it exited)
* Adds a metric `autoscaling_agent_runner_starts`
* Adds a metric `autoscaling_agent_runner_restarts`

## Todo before merge

- [x] Split some of the metrics changes into a separate PR
- [x] Testing
- [x] Write a more descriptive commit message

### DIFF
--- a/pkg/agent/globalstate.go
+++ b/pkg/agent/globalstate.go
@@ -126,42 +126,16 @@ func (s *agentState) handleVMEventAdded(
 	runnerCtx, cancelRunnerContext := context.WithCancel(ctx)
 
 	status := &podStatus{
-		mu:       sync.Mutex{},
-		done:     false,
-		errored:  nil,
-		panicked: false,
-		vmInfo:   event.vmInfo,
+		mu:                sync.Mutex{},
+		endState:          nil,
+		previousEndStates: nil,
+		vmInfo:            event.vmInfo,
 
 		startTime:                   time.Now(),
 		lastSuccessfulInformantComm: nil,
 	}
 
-	runner := &Runner{
-		global: s,
-		status: status,
-		logger: RunnerLogger{
-			prefix: fmt.Sprintf("Runner %v: ", podName),
-		},
-		schedulerRespondedWithMigration: false,
-
-		shutdown:              nil, // set by (*Runner).Run
-		vm:                    event.vmInfo,
-		podName:               podName,
-		podIP:                 event.podIP,
-		lock:                  util.NewChanMutex(),
-		requestLock:           util.NewChanMutex(),
-		requestedUpscale:      api.MoreResources{Cpu: false, Memory: false},
-		lastMetrics:           nil,
-		scheduler:             nil,
-		server:                nil,
-		informant:             nil,
-		computeUnit:           nil,
-		lastApproved:          nil,
-		lastSchedulerError:    nil,
-		lastInformantError:    nil,
-		backgroundWorkerCount: atomic.Int64{},
-		backgroundPanic:       make(chan error),
-	}
+	runner := s.newRunner(status, event.vmInfo, podName, event.podIP)
 
 	txVMUpdate, rxVMUpdate := util.NewCondChannelPair()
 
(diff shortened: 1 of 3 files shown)


pkg/agent/globalstate.go
```diff
@@ -172,9 +146,170 @@ func (s *agentState) handleVMEventAdded(
...
 	runner.Spawn(runnerCtx, rxVMUpdate)
 }
 
+// FIXME: make these timings configurable.
+const (
+	RunnerRestartMinWaitSeconds = 5
+	RunnerRestartMaxWaitSeconds = 10
+)
+
+func (s *agentState) TriggerRestartIfNecessary(runnerCtx context.Context, podName util.NamespacedName, podIP string) {
```
tychoish, line 159: I think `context.Context` objects should be named `ctx`, particularly the ones that get passed to a function unless there's a good reason for it, which might exist but it's not obvious (and I read the function looking for it because I was suspicious.)

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,170 @@ func (s *agentState) handleVMEventAdded(
...
+	// Three steps:
+	//  1. Check if the Runner needs to restart. If no, we're done.
+	//  2. Wait for a random amount of time (between RunnerRestartMinWaitSeconds and RunnerRestartMaxWaitSeconds)
+	//  3. Restart the Runner (if it still should be restarted)
+
+	pod, ok := func() (*podState, bool) {
+		// note: intentionally release s.lock as soon as we're done with it, so that there's no
+		// possibility of holding onto it longer than we need.
```
tychoish, line 167: I think that comments like this really just duplicate what's going on in the code, and we should let the code speak for itself. Using a lambda for this kind of thing is fine, and it would be fine to write a specific method to do this operation (which would help with the documentation aspect, on the idea that one good method name is worth 10 comments). 

It's also the case that this comment made me think/notice that the podState is a pointer and that wile you use this lock to access the map (which is cool/good) we can still race accesses to `podState`, you solve this by using a lock on 178. If this lambda were a method (say), the code would just "look right" and your intent would be clear. Having said that, while I'm not in favor of the comment (and think that the open issue for improved lock hygine will take care of some of this), I think if you want to leave the code as it is, you should mention that the use of `podState` is governed by its own lock.

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,170 @@ func (s *agentState) handleVMEventAdded(
...
+			select {
+			case <-time.After(waitDuration):
+			case <-runnerCtx.Done():
+				klog.Infof("Canceling restart of Runner %v after %s: %s", podName, time.Since(endTime), runnerCtx.Err())
+				return
+			}
+		}
+
+		s.lock.Lock()
+		defer s.lock.Unlock()
```
tychoish, line 240: why don't we do the narrow scope lock situation here? 

also we still have a pointer to pod (which should be the same object) in scope because of like 165.

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,170 @@ func (s *agentState) handleVMEventAdded(
...
+		// Runner was already restarted
+		if pod.status.endState == nil {
+			addedInfo := "this generally shouldn't happen, but could if there's a new pod with the same name"
+			klog.Warningf(
+				"Canceling restart of Runner %v after %s: Runner was already restarted (%s)",
+				podName, time.Since(endTime), addedInfo,
+			)
+		}
+
+		klog.Infof("Restarting %s Runner %v after %s, was running for %s", exitKind, podName, time.Since(endTime), totalRuntime)
```
tychoish, line 260: do we still need/want to use exitKind from the calling function? would the value have changed by now?

tychoish, line 260: do we still need/want to use exitKind from the calling function? would the value have changed by now?

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,170 @@ func (s *agentState) handleVMEventAdded(
...
+		//                     ^^
+		// note: exitKind is one of "panicked" or "errored" - e.g. "Restarting panicked Runner ..."
+
+		runner := s.newRunner(pod.status, pod.status.vmInfo, podName, podIP)
```
tychoish, line 264: given that we pass pod.status while holding the pod.status lock, feels like a great way to accidentally cause a deadlock on this line.

pkg/agent/prommetrics.go
```diff
@@ -115,6 +163,9 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
 		schedulerRequests:         schedulerRequests,
 		informantRequestsOutbound: informantRequestsOutbound,
 		informantRequestsInbound:  informantRequestsInbound,
+		runnerFatalErrors:         runnerFatalErrors,
 		runnerThreadPanics:        runnerThreadPanics,
+		runnerStarts:              runnerStarts,
+		runnerRestarts:            runnerRestarts,
```
tychoish, line 134: I think having to edit 2-4 spots to add a metric, it might be cleaner to declare `PromMetrics` at the top of the file, set the fields on this when you construct the runner, and then call `MustRegister` for each metric.

pkg/agent/runner.go
```diff
@@ -289,22 +289,35 @@ func (r *Runner) State(ctx context.Context) (*RunnerState, error) {
 
 func (r *Runner) Spawn(ctx context.Context, vmInfoUpdated util.CondChannelReceiver) {
 	go func() {
+		// Trigger restart if necessary *after* we've handled any panics.
+		defer r.global.TriggerRestartIfNecessary(ctx, r.vm.NamespacedName(), r.podIP)
```
tychoish, line 293: I'd just put this on like 305 within the same defer statement: same effect and you don't have to think about LIFO defers (which are great, but also confusing.

pkg/agent/runner.go
```diff
@@ -289,22 +289,35 @@ func (r *Runner) State(ctx context.Context) (*RunnerState, error) {
...
 		// Gracefully handle panics:
 		defer func() {
 			if err := recover(); err != nil {
 				r.setStatus(func(stat *podStatus) {
-					stat.panicked = true
-					stat.done = true
-					stat.errored = fmt.Errorf("runner panicked: %v", err)
+					stat.endState = &podStatusEndState{
+						ExitKind: podStatusExitPanicked,
+						Error:    fmt.Errorf("Runner %v panicked: %v", r.vm.NamespacedName(), err),
```
tychoish, line 300: theoretically, it might be nice to use `%w` (but this requires making sure that the recover actually produced a panic) [I wrote some panic handling stuff in `fun` that handles this, but it means that you get yourself an error object that you can introspect usefully.

pkg/agent/runner.go
```diff
@@ -289,22 +289,35 @@ func (r *Runner) State(ctx context.Context) (*RunnerState, error) {
...
+						Time:     time.Now(),
```
tychoish, line 301: this `time.Now()` returns when the lambda passed to `setStatus` is called, which could happen at an arbitrary future time. It's probably picky to do this, but I might declare a `now` before line 297..

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,167 @@ func (s *agentState) handleVMEventAdded(
...
+
+		txVMUpdate, rxVMUpdate := util.NewCondChannelPair()
+		// note: pod is *podState, so we don't need to re-assign to the map.
+		pod.vmInfoUpdated = txVMUpdate
+		pod.runner = runner
+
+		pod.status.previousEndStates = append(pod.status.previousEndStates, *pod.status.endState)
```
tychoish, line 271: is there a cap on how long we're willing to let this slice grow?

tychoish, line 271: is there a cap on how long we're willing to let this slice grow?

pkg/agent/globalstate.go
```diff
@@ -172,9 +146,167 @@ func (s *agentState) handleVMEventAdded(
...
+		defer s.lock.Unlock()
+
+		pod, ok := s.pods[podName]
+		if !ok {
+			klog.Warningf("Canceling restart of Runner %v after %s: no longer present in pod map", podName, time.Since(endTime))
+			return
+		}
+
+		pod.status.mu.Lock()
+		defer pod.status.mu.Unlock()
```
tychoish, line 249: I understand why this has to and should be this way, and while I think it might make sense to break this function apart into different functions/methods, it's fine. 

Having said that, the "holding two locks in one function  that doesn't have unit tests (and may not have integration tests either?" is definitely dodgy

tychoish (commented):
I made a bunch of code style/"readers notes"  while going through this. 

I think a comment in front of `TriggerRestartIfneeded` along the lines of "this restarts the runner for a specific pod <definition of what needed means>. The type plus method name (make this slightly confusing, and as I haven't been looking at this code for a few weeks, it took me a while to remember "one-runner-thread-per-vm-in-the-agent, one-agent-per-node, one-scheduler-per-cluster, one-informant-in-the-vm"

tychoish (commented):
I made a bunch of code style/"readers notes"  while going through this. 

I think a comment in front of `TriggerRestartIfneeded` along the lines of "this restarts the runner for a specific pod <definition of what needed means>. The type plus method name (make this slightly confusing, and as I haven't been looking at this code for a few weeks, it took me a while to remember "one-runner-thread-per-vm-in-the-agent, one-agent-per-node, one-scheduler-per-cluster, one-informant-in-the-vm"

pkg/agent/globalstate.go
sharnoff: Added a comment above the function:

> NB: runnerCtx is the context *passed to the new Runner*. It is only used here to end our restart process early if it's already been canceled.

lmk if that's suitable

pkg/agent/globalstate.go
sharnoff: Added a comment:

> Need to update pod itself; can't release s.lock. Also, pod *theoretically* may been deleted + restarted since we started, so it's incorrect to hold on to the original podStatus.

lmk if that works

pkg/agent/globalstate.go
sharnoff: Fair. the `podStatus` was only passed to fill out a field. Changed it so that it's now the caller's responsibility to set that

pkg/agent/prommetrics.go
sharnoff: I think I agree this is an issue, but I don't think it can get much better than it is now (unless I'm missing something?). To me it seems like we'd kind of always have to have multiple places — the struct definition of `PromMetrics`, instantiating `PromMetrics`, and `MustRegister`-ing the metric. (plus whatever usage is needed throughout the codebase)

Given that, I'm inclined to leave it as is. Thoughst?

pkg/agent/runner.go
sharnoff: Yeah. Definitely could be nicer, going to leave it for now — I think currently there'd be relatively little benefit from changing it to use `%w`

pkg/agent/globalstate.go
sharnoff: Not currently, no. It should only increase every 5s, so IMO that's probably fine.

(if we're getting OOM-killed because things are restarting too much, then we probably have bigger issues)

pkg/agent/runner.go

pkg/agent/runner.go

pkg/agent/globalstate.go
tychoish: I mean, if runners are going to crash (and all of them might at the same time if there's a bug? possibly?) and that happens a lot... hitting memory limits seems quite possible, but I suppose there's not much to do other than panic, and the oom will do the thing.

Because I don't have a great mental model of this: do these states end up in the "dump state endpoint", so if this got big would that endpoint die?

pkg/agent/prommetrics.go
sharnoff: Yeah, this will be resolved post-merge by #276.

pkg/agent/runner.go
sharnoff: The downside is the implementation complexity — and panic handling in particular benefits from being simple (low code-coverage, things go very wrong if it messes up)

pkg/agent/globalstate.go
sharnoff: > suppose there's not much to do other than panic, and the oom will do the thing

:+1:

> do these states end up in the "dump state endpoint", so if this got big would that endpoint die?

Realistically I think this would have to get _really_ big for that to happen — the only processing we do on these is to copy it into a new slice, then convert to JSON and respond.

We have metrics for runner panics (and #274 adds for fatal errors), so we should be getting a consistent "SYSTEM IS ON FIRE" signal long before that happens.

sharnoff:
Ok, fixed some bugs, now validated that this is working as expected. I used this patch to do some manual restart testing:

```diff
diff --git a/pkg/agent/runner.go b/pkg/agent/runner.go
index 7b7f519..6a2fadf 100644
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -386,6 +386,13 @@ func (r *Runner) Run(ctx context.Context, vmInfoUpdated util.CondChannelReceiver
 		r.serveInformantLoop(c, sendInformantUpd, sendUpscaleRequested)
 	})
 
+	r.spawnBackgroundWorker(ctx, "DEBUG: test panicker", func(c context.Context) {
+		duration := util.NewTimeRange(time.Second, 1, 30).Random()
+		r.logger.Infof("DEBUG: going to panic in %s", duration)
+		time.Sleep(duration)
+		panic("test panic")
+	})
+
 	// Note: Run doesn't terminate unless the parent context is cancelled - either because the VM
 	// pod was deleted, or the autoscaler-agent is exiting.
 	select {
```

PR #276, opened 2023-05-18
metrics: change metrics registration management

tychoish:


pkg/agent/prommetrics.go
```diff
@@ -15,53 +15,76 @@ type PromMetrics struct {
...
+	//   (even though MustRegister is variadic, the function calls
+	//   are cheap and calling it more than once means that when
+	//   it panics, we know exactly which metric caused the error.)
+	reg.MustRegister(collectors.NewGoCollector())
+	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+
+	// the register() function registers the collector and returns
+	// it so we can set it directly on the output structure.
+	metrics.schedulerRequests = register(reg,
+		prometheus.NewCounterVec(
```
sharnoff, line 37:
```suggestion
	metrics.schedulerRequests = register(reg, prometheus.NewCounterVec(
```

so that formatting matches with the ones below

sharnoff (commented):
Looks good. Same thing should happen for `pkg/plugin/prommetrics.go` as well

pkg/plugin/prommetrics.go
```diff
@@ -14,38 +15,38 @@ type PromMetrics struct {
...
+	// register stock collectors directly:
+	//   (even though MustRegister is variadic, the function calls
+	//   are cheap and calling it more than once means that when
+	//   it panics, we know exactly which metric caused the error.)
+	reg.MustRegister(collectors.NewGoCollector())
+	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+
 	p.metrics = PromMetrics{
-		pluginCalls: prometheus.NewCounterVec(
+		pluginCalls: util.RegisterMetric(reg, prometheus.NewCounterVec(
```
sharnoff, line 28: replicate comment about `util.RegisterMetric()` here?

pkg/util/metrics.go
```diff
@@ -0,0 +1,8 @@
+package util
+
+import "github.com/prometheus/client_golang/prometheus"
+
+func RegisterMetric[P prometheus.Collector](reg *prometheus.Registry, collector P) P {
+	reg.MustRegister(collector)
+	return collector
+}
```
sharnoff, line 8: I think this should probably be merged with `pkg/util/promserver.go`. I like the name `metrics.go` better for the two combined.

pkg/agent/prommetrics.go
```diff
@@ -16,52 +17,67 @@ type PromMetrics struct {
...
+		runnerRestarts: util.RegisterMetric(reg, prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: "autoscaling_agent_runner_restarts",
+				Help: "Number of existing per-VM Runners restarted due to failure",
+			},
+		)),
+	}
+
+	// the remaining metrics are computed on timers run within
+	// prom: register them directly.
```
sharnoff, line 79: I was under the impression that these are actually in response to metrics requests, not on some regular timer. Do you know if that's incorrect?

sharnoff (approved):
Looks good! Left some comments. The only other (very minor) thing that stands out to me is the newline gaps between the gauge metrics in `pkg/agent` — the file is already pretty sparse without them. Not super important though, so feel free to leave as-is.

**NB: This should not be merged until after #274**

sharnoff:
Heads-up: the PRs this was building on have now been merged, so this PR requires rebase (or _maybe_ merge would work ok? not actually sure tbh - rebase tends to be easier for this case because the state underneath is still the same, just the commit hashes have changed)

pkg/agent/prommetrics.go
tychoish: huh, no idea where that idea came from on my part. you're right. I've modified my comment accordingly

PR #277, opened 2023-05-19
neonvm: Rename 'runner' container to 'neonvm-runner'

sharnoff:
Currently "runner" makes it hard to understand the log output of other systems (e.g. k8s events) interacting with NeonVM. IMO renaming should help make things more understandable to others.

NB: this may require updating some Grafana dashboards.


cicdteam:
not sure if a rename is needed, but if so then you also need to update the container name in that annotation.
https://github.com/neondatabase/autoscaling/blob/main/neonvm/controllers/virtualmachine_controller.go#L933

cicdteam (approved):

sharnoff:
> also need to update the container name in that annotation

Thanks for the heads up. I think the runner name _probably_ should be some constant somewhere, but ok to leave for now (realistically we won't change it that often)

PR #278, opened 2023-05-19
agent/billing: Don't count VMs that aren't running

sharnoff:
To do this, adds the `(VirtualMachineStatus).IsAlive()` method, so that the autoscaler-agent doesn't need to be kept up-to-date with changes in NeonVM.

Also adds a call to IsAlive() in the agent's `vmIsOurResponsibility` check, so that future changes to NeonVM don't require modification there.

This requires updating #256 after merging, but should mean that changes are localized to `IsAlive`, instead of updating the autoscaler-agent as well.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -390,6 +390,16 @@ const (
 	VmMigrating VmPhase = "Migrating"
 )
 
+// IsAlive returns whether the guest in the VM is expected to be running
+func (p VmPhase) IsAlive() bool {
+	switch p {
+	case VmRunning, VmMigrating:
+		return true
+	default:
+		return false
+	}
+}
+
 //+genclient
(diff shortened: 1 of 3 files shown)


neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -390,6 +390,16 @@ const (
 	VmMigrating VmPhase = "Migrating"
 )
 
+// IsAlive returns whether the guest in the VM is expected to be running
+func (p VmPhase) IsAlive() bool {
+	switch p {
+	case VmRunning, VmMigrating:
```
cicdteam, line 396: need add there `VmPreMigrating` as well I think. It's short time phase (about 1 second usually) but anyway.

sharnoff: That'll be required with #256, right? (`VmPreMigrating` isn't on main yet)

PR #279, opened 2023-05-19
neonvm/controller: Update runner pod labels/annotations when VM's change

### LINKED ISSUE #275: VM label/annotation updates do not propagate to pods
## Impact

I'm not sure what our plan is for keeping compute logs dashboards once the compute pool is merged (cc @tychoish, @nikitakalyanov), but if it relies on labels or annotations, it currently will not work with VMs.

Billing is based on the VM objects themselves (not the pods), so should be ok.

## Steps to reproduce

For testing, apply this patch to `vm-deploy.yaml`:

```diff
diff --git a/vm-deploy.yaml b/vm-deploy.yaml
index d7a6940..5d482c9 100644
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -7,6 +7,7 @@ metadata:
     # In this example, these bounds aren't necessary. So... here's what they look like :)
     autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.25, "mem": "5Gi" } }'
   labels:
+    my-test-label: "foobar"
     autoscaling.neon.tech/enabled: "true"
     # Set to "true" to continuously migrate the VM (TESTING ONLY)
     autoscaling.neon.tech/testing-only-always-migrate: "false"
```

And then setup cluster + perform operations as follows:

```sh-session
$ # Set up local cluster...
$ make local-cluster && make deploy && make pg-14-disk-test

$ # ... with test VM running
$ kubectl apply -f vm-deploy.yaml

$ # validate initial VM/pod label:
$ kubectl get neonvm -o json | jq '.items[] | { name: .metadata.name, my_label: .metadata.labels["my-test-label"] }'
{
  "name": "postgres14-disk-test",
  "my_label": "foobar"
}
$ kubectl get pod -o json | jq '.items[] | { name: .metadata.name, my_label: .metadata.labels["my-test-label"] }'
{
  "name": "postgres14-disk-test-dcrqd",
  "my_label": "foobar"
}

$ # Update label
$ kubectl patch neonvm postgres14-disk-test --type=json --patch '[{"op": "replace", "path": "/metadata/labels/my-test-label", "value": "baz" }]'
virtualmachine.vm.neon.tech/postgres14-disk-test patched

$ # validate VM label updated
$ kubectl get neonvm -o json | jq '.items[] | { name: .metadata.name, my_label: .metadata.labels["my-test-label"] }'
{
  "name": "postgres14-disk-test",
  "my_label": "baz"
}

$ # validate Pod label updated (uh oh!)
$ kubectl get pod -o json | jq '.items[] | { name: .metadata.name, my_label: .metadata.labels["my-test-label"] }'
{
  "name": "postgres14-disk-test-dcrqd",
  "my_label": "foobar"
}
```

## Expected result

VM pod label should be updated along with VM

## Actual result

VM pod label is not updated.

## Environment

Cluster running locally, latest commit in main.

sharnoff:
Fixes #275.

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -773,6 +773,8 @@ func updatePodMetadataIfNecessary(ctx context.Context, c client.Client, vm *vmv1
 		log.Info(msg, "VirtualMachine", vm.Name, "Pod", runnerPod.Name)
 	}
 
+	// NOTE: We don't need to update the data in runnerPod ourselves because c.Patch will update it
+	// with what we get back from the k8s API after the patch completes.
 	return c.Patch(ctx, runnerPod, client.RawPatch(types.JSONPatchType, patchData))
 }
 
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -391,6 +391,13 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 			log.Error(err, "Failed to get runner Pod")
 			return err
 		}
+
+		// Update the metadata (including "usage" annotation) before anything else, so that it
+		// will be correctly set even if the rest of the reconcile operation fails.
+		if err := updatePodMetadataIfNecessary(ctx, r.Client, virtualmachine, vmRunner); err != nil {
+			log.Error(err, "Failed to sync pod labels and annotations", "VirtualMachine", virtualmachine.Name)
+		}
+
 		// runner pod found, check/update phase now
 		switch vmRunner.Status.Phase {
 		case corev1.PodRunning:
@@ -400,13 +407,6 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 			virtualmachine.Status.Phase = vmv1.VmRunning
 			// update Node name where runner working
 			virtualmachine.Status.Node = vmRunner.Spec.NodeName
-
-			// update Pod "usage" annotation before anything else, so that it will be correctly set,
-			// even if the rest of the reconcile operation fails.
-			if err := updateRunnerUsageAnnotation(ctx, r.Client, virtualmachine, virtualmachine.Status.PodName); err != nil {
-				log.Error(err, "Failed to set Pod usage annotation", "VirtualMachine", virtualmachine.Name)
-				return err
-			}
 
 			// get CPU details from QEMU and update status
 			cpuSlotsPlugged, _, err := QmpGetCpus(virtualmachine)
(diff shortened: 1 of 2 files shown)


cicdteam:
@sharnoff this PR is quite old, should we re-look it?

sharnoff:
> should we revisit it?

Probably yes, I think. I've just rebased to fix after #256, so it should still be good.

I'm not 100% sure that this is a good idea. Here are my thoughts:

1. I think this would _generally_ be useful for logs and metrics, especially with the compute pools stuff.
2. In order to implement this properly, we have to propagate label _deletion_ (and the only way to do that correctly is to compare the sets of labels on the VM vs pod). 

For (2), I've listed some annotations that we shouldn't delete even if they're on the pod but not the VM:

https://github.com/neondatabase/autoscaling/blob/f8dac9cb8a10b4a285e9935f0514f438fb15d1dc/neonvm/controllers/virtualmachine_controller.go#L689-L694

However, with this approach (and AFAICT, with _any_ approach that handles label/annotation deletion), we're always going to run the risk of accidentally deleting labels that are supposed to be there, so I don't know. 

Anyways, what do you think?

cicdteam:
Looks good, but I want you to pay attention to [kubectl.kubernetes.io/last-applied-configuration](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-create-objects) annotation. When I create vm (`kubectl apply -f vm.yaml`) I see such annotation in VM (`kubectl get neonvm vm -oyaml`) and the same annotation in the runner pod. I think this annotation should not pass to the pod level.

Example:

vm
```console
k get neonvm vm -ojsonpath='{.metadata.annotations.kubectl\.kubernetes\.io/last-applied-configuration}'

{"apiVersion":"vm.neon.tech/v1","kind":"VirtualMachine","metadata":{"annotations":{},"name":"vm","namespace":"default"},"spec":{"extraNetwork":{"enable":true},"guest":{"args":["nginx","-g","daemon off;"],"cpus":{"min":1},"memorySlotSize":"1Gi","memorySlots":{"min":1},"ports":[{"name":"http","port":80}],"rootDisk":{"image":"andreyneon/vm-nginx:alpine","imagePullPolicy":"Always"}},"terminationGracePeriodSeconds":1}}
```

pod
```console
k get pod vm-r8t56 -ojsonpath='{.metadata.annotations.kubectl\.kubernetes\.io/last-applied-configuration}'

{"apiVersion":"vm.neon.tech/v1","kind":"VirtualMachine","metadata":{"annotations":{},"name":"vm","namespace":"default"},"spec":{"extraNetwork":{"enable":true},"guest":{"args":["nginx","-g","daemon off;"],"cpus":{"min":1},"memorySlotSize":"1Gi","memorySlots":{"min":1},"ports":[{"name":"http","port":80}],"rootDisk":{"image":"andreyneon/vm-nginx:alpine","imagePullPolicy":"Always"}},"terminationGracePeriodSeconds":1}}
```

cicdteam (approved):

PR #286, opened 2023-05-22
agent/informant: Fix informant server exit logs

sharnoff:
This commit contains both (a) a fix so that we're not logging %!s(\<nil\>) and (b) minor formatting improvements. The end result is that our logs should go from:

    Informant server exiting with retry: false, err: %!s(<nil>)

to:

    Informant server exiting with: retry=false, err=nil

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -194,7 +194,11 @@ func NewInformantServer(
 				logFunc = runner.logger.Infof
 			}
 
-			logFunc("Informant server exiting with retry: %v, err: %s", status.RetryShouldFix, status.Err)
+			errDesc := "nil"
+			if status.Err != nil {
+				errDesc = fmt.Sprintf("%s", status.Err)
+			}
+			logFunc("Informant server exiting with: retry=%v, err=%s", status.RetryShouldFix, errDesc)
 		}
 
 		shutdownName := fmt.Sprintf("InformantServer shutdown (%s)", server.desc.AgentID)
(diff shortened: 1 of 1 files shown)


pkg/agent/informant.go
```diff
@@ -194,7 +194,11 @@ func NewInformantServer(
 				logFunc = runner.logger.Infof
 			}
 
-			logFunc("Informant server exiting with retry: %v, err: %s", status.RetryShouldFix, status.Err)
+			errDesc := "nil"
+			if status.Err != nil {
+				errDesc = fmt.Sprintf("%s", status.Err)
+			}
+			logFunc("Informant server exiting with: retry=%v, err=%s", status.RetryShouldFix, errDesc)
```
tychoish, line 201: I'd just use %q and let the empty quote marks to give an indication that there's an issue 

I might also add a method to status that's like `ReportError` just for clarity.

sharnoff: double-checking I understand - you're suggesting that if `err == nil`, we should print `err=""`?

If so, I'd rather the logs be more explicit about that. Although on a related note, I think it'd make sense to set `errDesc = fmt.Sprintf("%q", status.Err.Error())` instead, so that it's unambiguous.

tychoish: I mean I don't think it super matters. I think this is a lot of code to get, essentially what you get from `fmt.Sprintf("'%v', err)`

https://go.dev/play/p/dp58wuSUGD5

sharnoff: Ended up going with just using `%v` for the error

PR #293, opened 2023-05-23
implement restart policy for VM

cicdteam:
Restarting VM runner pods.

- by default `.spec.restartPolicy` is `Always` now
- if the runner pod is gone (manually removed or evicted by k8s control plane) the VM will go into `VmFailed` state and then runner will start again
- in case `.spec.restartPolicy: OnFailure` runner pod **will not** restart if it completed successfully

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -500,7 +500,6 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		}
 
 	case vmv1.VmSucceeded, vmv1.VmFailed:
-		// TODO: implement RestartPolicy
 		switch virtualmachine.Spec.RestartPolicy {
 		case vmv1.RestartPolicyAlways:
 			log.Info("Restarting VM runner pod", "VM.Phase", virtualmachine.Status.Phase, "RestartPolicy", virtualmachine.Spec.RestartPolicy)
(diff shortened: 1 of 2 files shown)


sharnoff (commented):
There's maybe a couple nits, but given the controller rewrite I think it's fine.

The one thing I'd note is that IMO we should keep the RestartPolicy default as "Never", and instead update the control plane to explicitly set it to "Always". AFAICT it doesn't matter if the control plane sets it to Always before this is merged, because currently RestartPolicy does nothing, right?

Thoughts?

(I'll open a control plane PR to do that anyways, because it's worth being explicit)

**EDIT:** Cloud PR: https://github.com/neondatabase/cloud/pull/5146

sharnoff:
Oh actually, turns out cloud explicitly sets it to Never - so it needs updating anyways.

cicdteam:
> it doesn't matter if the control plane sets it to Always before this is merged, because currently RestartPolicy does nothing, right?

No, this is wrong. Currently, RestartPolicy is passed to the Pod spec and doesn't work correctly (if the pod is down, it restarts by k8s and goes to Error state at the end (because k8s just restarts, not recreates the pod))

sharnoff:
> Currently, RestartPolicy is passed to the Pod spec

Ah right, forgot. Good catch. Added a note to the control plane PR that it must not be merged until after the VM controller update is deployed.

sharnoff (approved):
Hm, actually - I just read through the pod spec, turns out "Always" is the default RestartPolicy there as well. Happy to change the default to Always.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -503,9 +503,68 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
 		case vmv1.RestartPolicyAlways:
-			// TODO: restart VirtualMachine
+			log.Info("Restarting VM runner pod", "VM.Phase", virtualmachine.Status.Phase, "RestartPolicy", virtualmachine.Spec.RestartPolicy)
+			// get runner to delete
+			vmRunner := &corev1.Pod{}
+			err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
+			if err == nil {
+				// delete current runner
+				if err = r.Delete(ctx, vmRunner); err != nil {
+					log.Error(err, "Failed to delete runner Pod", "Pod.Namespace", vmRunner.Namespace, "Pod.Name", vmRunner.Name)
```
tychoish, line 513: my instinct is that we should avoid logging errors and then returning them: if you do this every time you get an error, you end up with one error being logged potentially many times. I don't, from this diff, how much logging we end up doing, but it seems like it could be a lot.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -503,9 +503,68 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+					return err
+				}
+				r.Recorder.Event(virtualmachine, "Normal", "Deleted",
+					fmt.Sprintf("VM runner pod was deleted: %s", vmRunner.Name))
```
tychoish, line 517: why is this line split? we have a bunch of quite long lines, and I think if we're going to have long lines, we should avoid splits like this except when it improves clarity, and I'm not sure that it makes sense in this case.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -503,9 +503,68 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+				log.Info("VM runner pod was deleted", "Pod.Namespace", vmRunner.Namespace, "Pod.Name", vmRunner.Name)
+				// TODO: remove thos hack after redesign reconcile loop
+				time.Sleep(time.Second)
```
tychoish, line 520: are TODOs tracked somewhere (issues, centrally?) on what time scale are we expecting to resolve this? 

also a comment about why 1 second was chosen and what would happen (assuming the refactor happens after we merge this) if we changed this in either direction.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -503,9 +503,68 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+			// drop VM phase
+			virtualmachine.Status.Phase = ""
 		case vmv1.RestartPolicyOnFailure:
-			// TODO: restart VirtualMachine
+			log.Info("Restarting VM runner pod", "VM.Phase", virtualmachine.Status.Phase, "RestartPolicy", virtualmachine.Spec.RestartPolicy)
+			// get runner to delete
+			found := true
+			vmRunner := &corev1.Pod{}
+			err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
+			if err != nil && apierrors.IsNotFound(err) {
```
tychoish, line 539: I think/hope that `apierrors.IsNotFound` will return false if `err == nil` so it makes this easier to read if you delete the `err != nil` on this line.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -503,9 +503,68 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+				return err
+			}
+			// do cleanup
+			virtualmachine.Status.PodName = ""
+			virtualmachine.Status.PodIP = ""
+			virtualmachine.Status.Node = ""
+			virtualmachine.Status.CPUs = nil
+			virtualmachine.Status.MemorySize = nil
+			// drop VM phase
+			virtualmachine.Status.Phase = ""
```
tychoish, line 532: we have this code in two places. can we put it into a method somewhere?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -760,7 +819,7 @@ func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
 		},
 		Spec: corev1.PodSpec{
 			EnableServiceLinks:            virtualmachine.Spec.ServiceLinks,
-			RestartPolicy:                 corev1.RestartPolicy(virtualmachine.Spec.RestartPolicy),
+			RestartPolicy:                 corev1.RestartPolicyNever,
```
tychoish, line 822: I don't quite understand this one.

tychoish (commented):
Nothing here that should block merging, just style and clarity questions.

neonvm/controllers/virtualmachine_controller.go
cicdteam: Recorder Events have the same style over whole code. It was initially generated by kubebuilder and i think it's good.

neonvm/controllers/virtualmachine_controller.go
cicdteam: `time.Sleep` has been removed as it now works without delay. In my tests with the wrong vm phase cleanup, this resulted in duplicate runner pods, but now everything is fine.

neonvm/controllers/virtualmachine_controller.go
cicdteam: I originally decided not to add another method to the VirtualMachine type because cleanup only happens in those two places, but now I've added Cleanup().

neonvm/controllers/virtualmachine_controller.go
cicdteam: deleted

PR #295, opened 2023-05-25
util/watch: Move event handling into its own function

sharnoff:
This also fixes a possible race on update events, because we weren't locking the store.

The function basically already existed by itself in the body of Watch as an anonymous function. Other than moving the locks, no other changes were made.

I think something very similar was originally implemented in #65, but that PR seems mostly stalled.

This PR is mostly "cleaning things up a little as I'm figuring out where to add metrics".

### DIFF
--- a/pkg/util/watch.go
+++ b/pkg/util/watch.go
@@ -261,77 +261,12 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 					}
 
 					meta := obj.GetObjectMeta()
-					uid := meta.GetUID()
 					// Update ResourceVersion so subsequent calls to client.Watch won't include this
 					// event, which we're currently processing.
 					opts.ResourceVersion = meta.GetResourceVersion()
 
-					name := GetNamespacedName(obj)
-
 					// Wrap the remainder in a function, so we can have deferred unlocks.
-					err := func() error {
-						switch event.Type {
-						case watch.Added:
-							store.mutex.Lock()
-							defer store.mutex.Unlock()
-
-							if _, ok := store.objects[uid]; ok {
-								return fmt.Errorf(
-									"watch %s: received add event for object %v that we already have",
-									config.LogName, name,
-								)
-							}
-							store.objects[uid] = (*T)(obj)
-							for _, index := range store.indexes {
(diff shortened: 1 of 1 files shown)


pkg/util/watch.go
```diff
@@ -493,6 +428,82 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
...
+	eventType watch.EventType,
+	meta metav1.Object,
+	ptr P,
+) error {
+	uid := meta.GetUID()
+	name := NamespacedName{Namespace: meta.GetNamespace(), Name: meta.GetName()}
+	obj := (*T)(ptr)
+
+	// Some of the cases below don't actually require locking the store. Most of the events that we
+	// recieve *do* though, so we're better off doing it here for simplicity.
```
tychoish, line 445: could have each case in the switch statement be: 
```go
return func() error {
// lock handling
// current code
}
```
I think it'd be worth it, given how contended this lock may be.

pkg/util/watch.go
```diff
@@ -493,6 +428,82 @@ func Watch[C WatchClient[L], L metav1.ListMetaAccessor, T any, P WatchObject[T]]
 	return &store, nil
 }
 
+// helper for Watch. Error events are expected to already have been handled by the caller.
+func handleEvent[T any, P ~*T](
```
tychoish, line 432: wasn't this my handcrank change, more or less?

tychoish (approved):
love it. 

I'm worried enough about lock contention in this codebase that I'd move the lock around.

tychoish (approved):
love it. 

I'm worried enough about lock contention in this codebase that I'd move the lock around.

pkg/util/watch.go
sharnoff: yep!

> I think something very similar was originally implemented in https://github.com/neondatabase/autoscaling/pull/65, but that PR seems mostly stalled.

pkg/util/watch.go
sharnoff: Realistically we'll be fine. The only "normal" case that won't require locking is if we receive a "bookmark" event, which shouldn't be very often. (the other cases are things that we're not expecting, so will panic)

Issue #35, opened 2023-02-06
Tech debt around duration config


sharnoff:
A few things here:

1. Our config files have durations for various things, and they're largely defined as "integer number of seconds"
2. We have a bunch of ad-hoc duration constants with FIXMEs to make them configurable
3. We have a dedicated type for picking random durations in a range (`util.TimeRange`), but we're not using that outside of `util.Watch`.

It'd be good to resolve some of these, ideally
- [ ] Using `util.TimeRange` in the configs; and
- [ ] Determining which duration constants should be made configurable (and which should be left as-is)

sharnoff:
See also: https://github.com/neondatabase/autoscaling/pull/223#discussion_r1185110442

PR #296, opened 2023-05-26
util: Move TimeRange into its own time.go

sharnoff:
Precursor to #297. Should also set up (finally) resolving #35 nicely.

### DIFF
--- a/pkg/util/time.go
+++ b/pkg/util/time.go
@@ -1,0 +1,35 @@
+package util
+
+import (
+	"errors"
+	"math/rand"
+	"time"
+)
+
+type TimeRange struct {
+	min   int
+	max   int
+	units time.Duration
+}
+
+func NewTimeRange(units time.Duration, min, max int) *TimeRange {
+	if min < 0 {
+		panic(errors.New("bad time range: min < 0"))
+	} else if min == 0 && max == 0 {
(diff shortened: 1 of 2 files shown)


pkg/util/time.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+)
+
+type TimeRange struct {
+	min   int
+	max   int
+	units time.Duration
+}
+
+func NewTimeRange(units time.Duration, min, max int) *TimeRange {
```
tychoish, line 15: as an aside, while I think you should merge this as is, because more files probably make sense... 

I think it'd be more clear to just do min/max durations and cut out the third argument... and I'd probably just do `max(0, r.Random())` to avoid the (logical) underflow rather than panic (potentially I think rather a long time after startup?)

tychoish (approved):

pkg/util/time.go
sharnoff: > I think it'd be more clear to just do min/max durations and cut out the third argument

Yeah, there's definitely some improvements that can be made. Going to leave that for #35.

> rather than panic (potentially I think rather a long time after startup?)

Currently, `TimeRange` is _mostly_ used in `Watch` calls, which are all done at startup. So: should be ok, for now.

PR #303, opened 2023-05-30
ExtractVmInfo: change hard error to warn when min !< use or use !< max

### LINKED ISSUE #249: `ExtractVmInfo` doesn't allow updating bounds before resources
Ran into this while testing #232. tl;dr is that if you try to patch a VM to update the scaling bounds so min > use (to convince the autoscaler-agent to do it for you), then `ExtractVmInfo` will error, saying that it's illegal to have min > use.

fprasx:
Resolves #249

### DIFF
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -11,6 +11,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 
 	vmapi "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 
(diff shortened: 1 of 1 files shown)


pkg/api/vminfo.go
```diff
@@ -195,9 +196,9 @@ func ExtractVmInfo(vm *vmapi.VirtualMachine) (*VmInfo, error) {
 
 	// check: min <= using <= max
 	if using.HasFieldLessThan(min) {
-		return nil, fmt.Errorf("current usage %+v has field less than minimum %+v", using, min)
+		klog.Warningf("current usage %+v has field less than minimum %+v", using, min)
```
sharnoff, line 199: With this change, we lose some context about which VM triggered the warning (before, it was the caller's responsibility to include it in the logs). It'd be good to add that into the message here. Maybe something along the lines of:

```go
klog.Warningf("VM %v current usage ...", info.NamespacedName(), ...)
```

(`NamespacedName()` returns type `NamespacedName`, which provides some standardized formatting for VM names)

sharnoff (commented):
Looks good — one comment for more info in the logs. It'd also be good to retain the implicit check that `min <= max`, because that will always be invalid

sharnoff (approved):

PR #304, opened 2023-05-30
agent: Use queue for global vmEvents, not channel

sharnoff:
Pre-req of a later fix in util/watch (#305). This change is similar in spirit to #160 (using a queue in the plugin), arising from a similar need: It's hard to avoid deadlocks without using a queue.

### DIFF
--- a/pkg/agent/entrypoint.go
+++ b/pkg/agent/entrypoint.go
@@ -25,17 +25,24 @@ type MainRunner struct {
 }
 
 func (r MainRunner) Run(ctx context.Context) error {
-	vmEvents := make(chan vmEvent)
-
 	buildInfo := util.GetBuildInfo()
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
+	vmEventQueue := pubsub.NewUnlimitedQueue[vmEvent]()
+	defer vmEventQueue.Close()
+	pushToQueue := func(ev vmEvent) {
+		if err := vmEventQueue.Add(ev); err != nil {
+			klog.Warningf("error adding vmEvent %+v to queue: %s", ev, err)
+		}
+	}
+
 	klog.Info("Starting VM watcher")
-	vmWatchStore, err := startVMWatcher(ctx, r.Config, r.VMClient, r.EnvArgs.K8sNodeName, vmEvents)
+	vmWatchStore, err := startVMWatcher(ctx, r.Config, r.VMClient, r.EnvArgs.K8sNodeName, pushToQueue)
 	if err != nil {
 		return fmt.Errorf("Error starting VM watcher: %w", err)
 	}
+	defer vmWatchStore.Stop()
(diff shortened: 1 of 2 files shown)


pkg/agent/entrypoint.go
```diff
@@ -25,17 +25,24 @@ type MainRunner struct {
...
 
 func (r MainRunner) Run(ctx context.Context) error {
-	vmEvents := make(chan vmEvent)
-
 	buildInfo := util.GetBuildInfo()
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
+	vmEventQueue := pubsub.NewUnlimitedQueue[vmEvent]()
+	defer vmEventQueue.Close()
```
tychoish, line 33: the queue doesn't have any background threads: closing the queue just means that it'll error if you try and add new things to the queue. 

Add will only ever error in two cases: 
- the queue is closed (this is the condition where a channel would panic)
- you are using a limited queue and it's at capacity. (retryable. I also made some interfaces to queues that let you iterate over results/have "force" or blocking ops that may make more sense in other cases.)

(both of these conditions can be checked with `errors.Is` and are probably not worth logging)

pkg/agent/entrypoint.go
```diff
@@ -70,25 +78,15 @@ func (r MainRunner) Run(ctx context.Context) error {
...
-					break loop
-				}
+		event, err := vmEventQueue.Wait(ctx)
+		if err != nil {
+			if ctx.Err() != nil {
+				return nil
+			} else {
+				klog.Errorf("vmEventQueue returned error: %s", err)
+				return err
 			}
```
tychoish, line 88: I believe that you can (should? unclear.) do `errors.Is()` (I also have a `fun/erc.ContextExpired(err)` helper that checks both deadline/canceled cases).

I think, the same thing above: goes for here: Wait will block till the queue is closed or the context is canceled, in either case I think just returning the error (or nil? not sure how this error is processed,) would be _fine_

sharnoff:
Recording a key point from discussing:

> have the :grimacing: emotion about slapping an unbounded queue here: if the even processing gets behind or deadlocks, we end up with an OOM instead of backpressure

basically, it's preferable to OOM rather than either: (a) get behind on processing events [means billing operates on stale data!] or (b) drop some events [we have no recovery, so: autoscaling outage]

pkg/agent/entrypoint.go
sharnoff: Double-checked, and the caller does a `klog.Fatalf` if the error is non-nil — I think it's better here to explicitly return nil if the context is canceled. I'm going to stick with what's there (roughly) rather than checking via `errors.Is` because I think it's worthwhile to avoid dependence on the reason why the context is done.

pkg/agent/entrypoint.go
sharnoff: Going to keep this, so that if we still have lots of things being added to the queue after shutdown, we have logs that that's what's going on

PR #305, opened 2023-05-30
util/watch: Fix racy behavior with InitModeDefer

sharnoff:
Not 100% sure, but this *may* be the cause of an issue affecting prod that causes preexisting VMs to be ignored by pkg/agent/billing until there's an event for them. (the theory is that the index is added before the deferred adds are processed)

**Edit:** After thinking about it some more, realized that actually the issue is _guaranteed_ to happen because of this, without #304

**Builds on #304, must not be merged before that PR.**

We may also just want to get rid of `InitModeDefer` — it's nice in theory, but the autoscaler-agent's global VM event watching is the only place that uses it, and after #304 it's not _actually_ necessary. `InitModeDefer` sounds like good flexibility in theory, but in practice it's probably a code smell.

### DIFF
--- a/pkg/util/watch/watch.go
+++ b/pkg/util/watch/watch.go
@@ -172,8 +172,19 @@ func Watch[C Client[L], L metav1.ListMetaAccessor, T any, P Object[T]](
 		return nil, fmt.Errorf("Initial watch failed: %w", err)
 	}
 
+	// Lock the store to pass it into the goroutine, so that we don't have to worry about immediate
+	// operations on the store racing with any deferred additions.
+	store.mutex.Lock()
+
 	// With the successful Watch call underway, we hand off responsibility to a new goroutine.
 	go func() {
+		holdingInitialLock := true
+		defer func() {
+			if holdingInitialLock {
+				store.mutex.Unlock()
+			}
+		}()
+
 		// note: instead of deferring watcher.Stop() directly, wrapping it in an outer function
 		// means that we'll always Stop the most recent watcher.
 		defer func() {
@@ -184,6 +195,8 @@ func Watch[C Client[L], L metav1.ListMetaAccessor, T any, P Object[T]](
 		defer store.Stop()
 
 		// Handle any deferred calls to AddFunc
+		// NB: This is only sound because we're still holding store.mutex; otherwise we'd have to
+		// deal with possible racy operations (including adding an index).
 		for i := range deferredAdds {
 			obj := &deferredAdds[i]
 			uid := P(obj).GetObjectMeta().GetUID()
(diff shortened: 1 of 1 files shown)


pkg/util/watch/watch.go
```diff
@@ -172,8 +172,19 @@ func Watch[C Client[L], L metav1.ListMetaAccessor, T any, P Object[T]](
...
 	go func() {
+		holdingInitialLock := true
+		defer func() {
+			if holdingInitialLock {
+				store.mutex.Unlock()
+			}
+		}()
+
 		// note: instead of deferring watcher.Stop() directly, wrapping it in an outer function
 		// means that we'll always Stop the most recent watcher.
```
tychoish, line 189: I'd just combine the two defers (forcing folks to read the code backwards is just friction that is often best avoided.

pkg/util/watch/watch.go
```diff
@@ -172,8 +172,19 @@ func Watch[C Client[L], L metav1.ListMetaAccessor, T any, P Object[T]](
...
 
+	// Lock the store to pass it into the goroutine, so that we don't have to worry about immediate
+	// operations on the store racing with any deferred additions.
+	store.mutex.Lock()
+
 	// With the successful Watch call underway, we hand off responsibility to a new goroutine.
 	go func() {
+		holdingInitialLock := true
+		defer func() {
+			if holdingInitialLock {
```
tychoish, line 183: my assumption/understanding is that the only way that you might get here is: 
- it panics, so you'd release the lock, (though nothing catches the panic before this, and I'm not 100% sure what happens in this case, but it might not run). 
- there's a case where the go-routine might exit early: on (new) 207, you return early here, and you could break from the loop, release the lock as you do on 212, and then check the context before you start the loops right after that. (I'm not super worried about checking contexts in loops of things that are unlikely to be slow (e.g. stuff that's all in memory,) frankly it seems plausible that you could just let the context checks in the loop take care of the early return. [file this under the heading of: getting clever with locks and unlocking them outside of defers provides a great opportunity to make a deadlock.]

tychoish (approved):
seems fine. my comments are mostly just fussy pointers toward "it might be nice to make this more simple" but I think it's fine to merge this as is.

tychoish (approved):
seems fine. my comments are mostly just fussy pointers toward "it might be nice to make this more simple" but I think it's fine to merge this as is.

pkg/util/watch/watch.go
sharnoff: Going to leave it as-is for now, because I think it makes the high-level control flow easier (IMO: repeatedly seeing "defer" makes it clear that the things are deferred — order doesn't really matter. Whereas combining everything into one function it's easy for your eyes to skip over the leading "defer" at the top)

pkg/util/watch/watch.go
sharnoff: I don't entirely follow the second bullet point. But generally, yeah "getting clever with locks" is not ideal here, and really the correct thing to do is to get rid of `InitModeDefer` because I think it's fundamentally smelly code.

PR #306, opened 2023-05-30
util/watch: Add metrics

sharnoff:
Adds metrics to `util/watch.Watch` calls, for better observability on deeply internal systems. These are aggregated per _binary_, so the new metrics are:

* `$PREFIX_client_calls_total`
* `$PREFIX_relist_requests_total`
* `$PREFIX_events_total`
* `$PREFIX_errors_total`
* `$PREFIX_alive_current`
* `$PREFIX_failing_current`

where `$PREFIX` is either `autoscaling_agent_watchers` or `autoscaling_plugin_watchers`.

A fuller description of each metric can be found in `pkg/util/watch/metrics.go`.

All metrics have the `watcher_instance` label, which associates the metrics with an individual call to `watch.Watch` .

### DIFF
--- a/pkg/agent/entrypoint.go
+++ b/pkg/agent/entrypoint.go
@@ -31,8 +31,10 @@ func (r MainRunner) Run(ctx context.Context) error {
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
+	watchMetrics := watch.NewMetrics("autoscaling_agent_watchers_")
+
 	klog.Info("Starting VM watcher")
-	vmWatchStore, err := startVMWatcher(ctx, r.Config, r.VMClient, r.EnvArgs.K8sNodeName, vmEvents)
+	vmWatchStore, err := startVMWatcher(ctx, r.Config, r.VMClient, watchMetrics, r.EnvArgs.K8sNodeName, vmEvents)
 	if err != nil {
 		return fmt.Errorf("Error starting VM watcher: %w", err)
 	}
@@ -44,7 +46,7 @@ func (r MainRunner) Run(ctx context.Context) error {
 	}
 
 	watcherLogger := util.PrefixLogger{Prefix: "Scheduler Watcher: "}
-	schedulerStore, err := schedwatch.StartSchedulerWatcher(ctx, watcherLogger, r.KubeClient, broker, r.Config.Scheduler.SchedulerName)
+	schedulerStore, err := schedwatch.StartSchedulerWatcher(ctx, watcherLogger, r.KubeClient, watchMetrics, broker, r.Config.Scheduler.SchedulerName)
 	if err != nil {
 		return fmt.Errorf("starting scheduler watch server: %w", err)
 	}
@@ -57,6 +59,8 @@ func (r MainRunner) Run(ctx context.Context) error {
 	}
 
 	globalState, promReg := r.newAgentState(r.EnvArgs.K8sPodIP, broker, schedulerStore)
+	watchMetrics.MustRegister(promReg)
+
 	if err := util.StartPrometheusMetricsServer(ctx, 9100, promReg); err != nil {
 		return fmt.Errorf("Error starting prometheus metrics server: %w", err)
 	}
--- a/pkg/agent/entrypoint.go
+++ b/pkg/agent/entrypoint.go
@@ -31,7 +31,7 @@ func (r MainRunner) Run(ctx context.Context) error {
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
-	watchMetrics := watch.NewMetrics("autoscaling_agent_watchers_")
+	watchMetrics := watch.NewMetrics("autoscaling_agent_watchers")
 
 	klog.Info("Starting VM watcher")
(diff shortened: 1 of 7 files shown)


pkg/agent/entrypoint.go
```diff
@@ -31,8 +31,10 @@ func (r MainRunner) Run(ctx context.Context) error {
 	klog.Infof("buildInfo.GitInfo:   %s", buildInfo.GitInfo)
 	klog.Infof("buildInfo.GoVersion: %s", buildInfo.GoVersion)
 
+	watchMetrics := watch.NewMetrics("autoscaling_agent_watchers_")
```
tychoish, line 34: trailing underscore?

pkg/util/watch/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+				Name: fmt.Sprint(prefix, "failing_current"),
+				Help: "For each watcher, 1 iff the watcher's last request failed *and* it's waiting to retry, else 0",
+			},
+			[]string{metricInstanceLabel},
+		),
+	}
+}
+
+// MustRegister registers all the collectors in the Metrics
+func (m *Metrics) MustRegister(reg *prometheus.Registry) {
```
tychoish, line 99: why keep this seperate and not use `util.RegisterMetric` for these?

pkg/util/watch/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+	reg.MustRegister(m.aliveCurrent)
+	reg.MustRegister(m.failingCurrent)
+}
+
+///////////////////////////////////////////////
+// Internal helper methods for MetricsConfig //
+///////////////////////////////////////////////
+
+func (m *MetricsConfig) alive() {
+	if m != nil {
```
tychoish, line 113: I think it wouldn't be absurd to require that metrics be non-nil? what's your thinking on that?

pkg/util/watch/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+	wasFailing, *store = *store, true
+
+	if m != nil && !wasFailing {
+		m.failingCurrent.WithLabelValues(m.Instance).Inc()
+	}
+}
+
+func (m *MetricsConfig) unfailing(store *bool) {
+	var wasFailing bool
+	wasFailing, *store = *store, false
```
tychoish, line 135: I guess I'd probably store (an `atomic.Bool` (??)) in the `MetricsConfig` struct and track the state of failing/not-failing there rather than passing a pointer in/out and mutating here?

pkg/agent/entrypoint.go
sharnoff: Yeah... Something's gotta have the leading/trailing underscore. Although on second thought I'd prefer that to be internal to the `watch` package. Changed.

pkg/util/watch/metrics.go
sharnoff: Record of discussion: reasoning was that `watch.Metrics` wants to be purely internal to the `watch` package, so we can't really directly expose the collector(s).

pkg/util/watch/metrics.go
tychoish: this doesn't seem to have changed?

pkg/util/watch/metrics.go
tychoish: Is there a case where you would construct a `*Metrics` object without registering metrics? It doesn't look like that's the case, and you could just pass a registry to the `NewMetrics` constructor, am I wrong about this?

sharnoff: > Is there a case where you would construct a `*Metrics` object without registering metrics?

That's currently the case in the autoscaler-agent. Setting up the scheduler store:

https://github.com/neondatabase/autoscaling/blob/9ee46220f1c0dba14bf576f0f0bbde20311c0f88/pkg/agent/entrypoint.go#L56-L61

uses a `watch.Metrics`, and is required before calling `r.newAgentState`:

https://github.com/neondatabase/autoscaling/blob/9ee46220f1c0dba14bf576f0f0bbde20311c0f88/pkg/agent/entrypoint.go#L70

which produces the registry.

---

All that aside, I think it's generally preferable to allow constructing a `watch.Metrics` without registering it — if only for the possibility of future testing.

pkg/util/watch/metrics.go
sharnoff: Check again? That function now reads:

```go
func (m *MetricsConfig) alive() {
    m.aliveCurrent.WithLabelValues(m.Instance).Inc()
}
```

pkg/util/watch/metrics.go
tychoish: I think, in general, it's better to prevent a pretty easy to make mistake (creating a new metric without registering it) rather than making the code a bit (though, to be fair, only a bit) harder to maintain, to support theoretical future testing (particularly given the current state of the unit testing). 

Aside from all that, my earlier attempt to retrofit tests into the watcher have left me convinced that adding tests to the watcher would require a non-trival refactor, and the metrics registry wouldn't be the deal breaker if/when that is undertaken.

Though I haven't dug into the agent's metrics since the code review, It seems like the registry really ought to be passed into the agent state constructor rather than the other way around.

tychoish (approved):

pkg/util/watch/metrics.go
sharnoff: > It seems like the registry really ought to be passed into the agent state constructor rather than the other way around.

Yup, agree. Don't want to do that in this PR (or block on a separate PR to do that), but 100% should do that after merging.

PR #308, opened 2023-05-31
Suspend old agent then resume new one during switch

### LINKED ISSUE #24: Informant allows multiple active agents
Currently (3f9650e604cb0fb768cab2a574fd0832ca9e082e), when the VM informant switches to a newer autoscaler-agent, it sends a resume request to the new one *before* sending a suspend request to the old one. This means that there's a period of time between the "resume new" and "suspend old" in which there can be multiple active autoscaler-agents for a single VM.

Relevant code locations:

Resume new...

https://github.com/neondatabase/autoscaling/blob/3f9650e604cb0fb768cab2a574fd0832ca9e082e/pkg/informant/agent.go#L183-L206

... and then suspend old:

https://github.com/neondatabase/autoscaling/blob/3f9650e604cb0fb768cab2a574fd0832ca9e082e/pkg/informant/agent.go#L221-L235

`SpawnSuspend` is probably also not right, if we want to wait until the suspend completes.

fprasx:
Resolves #24.

### DIFF
--- a/pkg/informant/agent.go
+++ b/pkg/informant/agent.go
@@ -180,7 +180,7 @@ func (s *AgentSet) tryNewAgents(signal <-chan struct{}) {
 				break loopThroughAgents
 			}
 
-			// Resume() the Agent if we need to
+			// Do we need to resume the agent? We will use this later
 			shouldResume := func() bool {
 				candidate.lock.Lock()
 				defer candidate.lock.Unlock()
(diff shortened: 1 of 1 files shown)


pkg/informant/agent.go
```diff
@@ -180,7 +180,7 @@ func (s *AgentSet) tryNewAgents(signal <-chan struct{}) {
 				break loopThroughAgents
 			}
 
-			// Resume() the Agent if we need to
+			// Do we need to resume the agent? We will use this later
```
sharnoff, line 183: love the comments like this

sharnoff (approved):
Looks good! Tested, works as expected

PR #310, opened 2023-06-01
agent/billing: Add prometheus metrics

sharnoff:
This commit adds the following new metrics:

- `autoscaling_agent_billing_vms_processed_total`
- `autoscaling_agent_billing_vms_current`
- `autoscaling_agent_billing_batch_size`
- `autoscaling_agent_billing_send_errors_total`

These should resolve prior issues ("how do we see how many VMs the autoscaler-agent thinks there are?") as well as provide some introspection into the information we're using for billing - one of the more sensitive components.

Builds on #309.


pkg/agent/billing/prommetrics.go
```diff
@@ -0,0 +1,99 @@
...
+		sendErrorsTotal: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: "autoscaling_agent_billing_send_errors_total",
+				Help: "Total errors from attempting to send billing events",
+			},
+		),
+	}
+}
+
+func (m PromMetrics) MustRegister(reg *prometheus.Registry) {
```
tychoish, line 51: can you register metrics in the constructor?

pkg/agent/billing/prommetrics.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return batchMetrics{
+		total: make(map[batchMetricsLabels]int),
+
+		vmsProcessedTotal: m.vmsProcessedTotal,
+		vmsCurrent:        m.vmsCurrent,
+	}
+}
+
+func (b batchMetrics) inc(isEndpoint, autoscalingEnabled bool, phase vmapi.VmPhase) {
```
tychoish, line 82: I find having two successive boolean arguments to a function to be deeply confusing, I would either: 

make the arguments to this function a struct, write some chaining interface, so that it's clear what's happening, *or* split it into 4 methods.

pkg/agent/billing/prommetrics.go
sharnoff: Technically, yes. TBH though, I'd rather keep it similar to `util/watch` metrics (see also https://github.com/neondatabase/autoscaling/pull/306#discussion_r1214702650)

pkg/agent/billing/prommetrics.go
sharnoff: Changed from booleans to distinct types, I think that resolves this with relatively little boilerplate.

pkg/agent/billing/prommetrics.go
tychoish: I don't find this pattern/argument particularly compelling for either case, but I don't think this is important enough to reach consensus on.

tychoish (approved):

PR #312, opened 2023-06-02
Add another handler for usage updates in non-autoscaling VM's

### LINKED ISSUE #251: scheduler doesn't recognize usage updates from non-autoscaling VMs
Ran into this while testing #232.

## Steps to reproduce

1. Create VM _without_ `autoscaling.neon.tech/enabled=true`
2. Update the VM's spec.guest.{cpus,memorySlots}.use ~~_and/or_ its scaling bounds~~
3. Observe the plugin's internal state for the VM: it still only has the original value reserved

fprasx:
Resolves #251


pkg/plugin/plugin.go
```diff
@@ -121,14 +121,30 @@ func makeAutoscaleEnforcerPlugin(ctx context.Context, obj runtime.Object, h fram
...
+			memDiff := newVM.Using().Mem - oldVM.Using().Mem
+
+			p.state.lock.Lock()
+			defer p.state.lock.Unlock()
+
+			p.state.nodeMap[vmname].memSlots.Reserved += memDiff
+			p.state.nodeMap[vmname].vCPU.Reserved += cpuDiff
+
+			p.state.podMap[name].memSlots.Reserved += memDiff
+			p.state.podMap[name].vCPU.Reserved += cpuDiff
```
sharnoff, line 137: To match with the other callbacks, this should be a method on `(*AutoscaleEnforcer)` in state.go

pkg/plugin/watch.go
```diff
@@ -136,29 +137,29 @@ func (e *AutoscaleEnforcer) watchVMEvents(
...
 					return
 				}
 
 				if oldInfo.ScalingEnabled && !newInfo.ScalingEnabled {
 					name := util.NamespacedName{Namespace: newInfo.Namespace, Name: newVM.Status.PodName}
-					klog.Infof(
-						"[autoscale-enforcer] watch: Received update to disable autoscaling for pod %v",
-						name,
-					)
+					klog.Infof("[autoscale-enforcer] watch: Received update to disable autoscaling for pod %v", name)
```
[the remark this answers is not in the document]
fprasx: yeah :(

pkg/plugin/plugin.go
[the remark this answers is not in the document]
fprasx: I opted to just leaved everything the handler without making a shim because it's really only setting stuff - there's no other calculation. It seems like use cases like this are rare when looking at the other "actual" handlers, which all perform some calculation.

pkg/plugin/state.go
```diff
@@ -695,6 +695,33 @@ func (e *AutoscaleEnforcer) handleUpdatedScalingBounds(vm *api.VmInfo, unqualifi
...
+	node := pod.node
+
+	newMem, newCPU := vm.Using().Mem, vm.Using().VCPU
+	oldMem, oldCPU := pod.memSlots.Reserved, pod.vCPU.Reserved
+
+	node.memSlots.Reserved, node.vCPU.Reserved = newMem, newCPU
+	pod.memSlots.Reserved, pod.vCPU.Reserved = newMem, newCPU
+
+	klog.Infof(
+		"[autoscale-enforcer]: updated VM pod %v resources: vCPU %v -> %v, updated memory: %v -> %v",
```
sharnoff, line 718: It'd be good to include information for the node as well, e.g. a format string like:

```
updated VM pod %v resources in node %s: vCPU %v -> %v (node %v -> %v), memory %v -> %v (node %v -> %v)
```

**NB:** Having `node %s` in the format string is useful for log filtering (all other handlers do the same), at least until we make things a little more standard. The rest is just so that we can deduce the state of the node from the logs if the internal logic isn't correct.

pkg/plugin/state.go
```diff
@@ -695,6 +695,33 @@ func (e *AutoscaleEnforcer) handleUpdatedScalingBounds(vm *api.VmInfo, unqualifi
 	klog.Infof(fmtString, podName, pod.node.name, cpuVerdict, memVerdict)
 }
 
+func (e *AutoscaleEnforcer) handleNonAutoscalingUsageChange(vm *api.VmInfo, name util.NamespacedName) {
```
sharnoff, line 698:
```suggestion
func (e *AutoscaleEnforcer) handleNonAutoscalingUsageChange(vm *api.VmInfo, podName util.NamespacedName) {
```

It'd be good to make it clear that this refers to the name of the _pod_, not the _VM_

pkg/plugin/state.go
```diff
@@ -695,6 +695,33 @@ func (e *AutoscaleEnforcer) handleUpdatedScalingBounds(vm *api.VmInfo, unqualifi
...
+	if !ok {
+		klog.Errorf("[autoscale-enforcer] cannot find non-autoscaling VM pod %v to update usage", name)
+		return
+	}
+	node := pod.node
+
+	newMem, newCPU := vm.Using().Mem, vm.Using().VCPU
+	oldMem, oldCPU := pod.memSlots.Reserved, pod.vCPU.Reserved
+
+	node.memSlots.Reserved, node.vCPU.Reserved = newMem, newCPU
```
sharnoff, line 714: This line is incorrect — `node.{memSlots,vCPU}.Reserved` is the sum of all pods on the node; we need to update the node's resources based on the change in the pod (instead of e.g. setting it _equal_ to the pod's)

---

as an aside: This kind of error is easy to make. While it's not necessary here, this is partly why I've tended to make the handler functions generic over the resource. I find less repetition makes it easier to parse what's actually going on (of course, there's two sides to every coin — wanted to share some motivation behind why things are the way they are)

pkg/plugin/plugin.go
```diff
@@ -121,14 +121,17 @@ func makeAutoscaleEnforcerPlugin(ctx context.Context, obj runtime.Object, h fram
 	submitVMBoundsChanged := func(vm *api.VmInfo, podName string) {
 		pushToQueue(func() { p.handleUpdatedScalingBounds(vm, podName) })
 	}
+	submitNonAutoscalingVmUsageChanged := func(newVM *api.VmInfo, name util.NamespacedName) {
```
sharnoff, line 124: Another comment on this: The pod's namespace is guaranteed to be the same as the VM's (i.e. `newVM.Namespace == name.Namespace` always holds). For that reason, the other handler functions just take a `string` instead of a `util.NamespacedName`

(also `newVM` looks left over from earlier commits?)

pkg/plugin/plugin.go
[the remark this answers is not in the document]
sharnoff: Sure, makes sense

pkg/plugin/state.go
fprasx: Still wrapping my head around the pod-node-VM hierarchy but this makes sense.

pkg/plugin/state.go
sharnoff: Yup. For Kubernetes in general, Nodes have multiple Pods; Pods each have ≥ 1 Container (defined at creation of the Pod). For us, we only have one Container per-Pod, and it's running the VM.

pkg/plugin/trans.go
```diff
@@ -221,6 +221,17 @@ func (r resourceTransition[T]) handleDeleted(currentlyMigrating bool) (verdict s
 	return verdict
 }
 
+func (r resourceTransition[T]) handleNonAutoscalingUsageChange(newUsage T) (verdict string) {
+	diff := newUsage - r.node.Reserved
+	r.pod.Reserved += diff
+	r.node.Reserved = newUsage
```
sharnoff, line 227: hm, this is also incorrect, similar-ish to before. As written, `newUsage` corresponds to the _pod's_ usage.

pkg/plugin/plugin.go
```diff
@@ -130,7 +133,9 @@ func makeAutoscaleEnforcerPlugin(ctx context.Context, obj runtime.Object, h fram
 	}
 
 	klog.Infof("[autoscale-enforcer] Starting VM watcher")
-	vmStore, err := p.watchVMEvents(ctx, watchMetrics, submitVMDisabledScaling, submitVMBoundsChanged)
+
+	vmStore, err := p.watchVMEvents(ctx, watchMetrics, submitVMDisabledScaling, submitVMBoundsChanged, submitNonAutoscalingVmUsageChanged)
+
```
sharnoff, line 138:
```suggestion
	vmStore, err := p.watchVMEvents(ctx, watchMetrics, submitVMDisabledScaling, submitVMBoundsChanged, submitNonAutoscalingVmUsageChanged)
```

I'd ditch the extra lines. Merge artifact?

pkg/plugin/watch.go
```diff
@@ -169,6 +170,15 @@ func (e *AutoscaleEnforcer) watchVMEvents(
 					submitVMDisabledScaling(name)
 				}
 
+				if !oldInfo.ScalingEnabled && !newInfo.ScalingEnabled && oldInfo.Using() != newInfo.Using() {
```
sharnoff, line 173: I think I was mistaken when I suggested this condition. It should actually be:

```go
if (!oldInfo.ScalingEnabled || !newInfo.ScalingEnabled) && oldInfo.Using() != newInfo.Using() {
    ...
}
```

Because otherwise we won't handle usage changes if happens at the same time as scaling being enabled/disabled.

sharnoff (commented):
Generally looking good — once `handleNonAutoscalingUsageChange` is fixed, it'll be good to go :)

pkg/plugin/plugin.go
fprasx: Yup

pkg/plugin/watch.go
fprasx: Good catch.

PR #316, opened 2023-06-05
Include AgentID in /downscale and /upscale responses

fprasx:
Not sure if this is the exact right way to do it. It seems protocol version agnostic.

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -984,7 +984,7 @@ func (s *InformantServer) Downscale(ctx context.Context, to api.Resources) (*api
 	var statusCode int
 	var resp *api.DownscaleResult
 	if s.protoVersion.SignsResourceUpdates() {
-		signedRawResources := api.SignedRawResources{RawResources: rawResources, Id: id}
+		signedRawResources := api.ResourceMessage{RawResources: rawResources, Id: id}
 		reqData := api.AgentResourceMessage{Data: signedRawResources, SequenceNumber: s.incrementSequenceNumber()}
 		resp, statusCode, err = doInformantRequest[api.AgentResourceMessage, api.DownscaleResult](
 			ctx, s, timeout, http.MethodPut, "/downscale", &reqData,
@@ -1037,7 +1037,7 @@ func (s *InformantServer) Upscale(ctx context.Context, to api.Resources) error {
 
 	var statusCode int
 	if s.protoVersion.SignsResourceUpdates() {
-		signedRawResources := api.SignedRawResources{RawResources: rawResources, Id: id}
+		signedRawResources := api.ResourceMessage{RawResources: rawResources, Id: id}
 		reqData := api.AgentResourceMessage{Data: signedRawResources, SequenceNumber: s.incrementSequenceNumber()}
 		_, statusCode, err = doInformantRequest[api.AgentResourceMessage, struct{}](
 			ctx, s, timeout, http.MethodPut, "/upscale", &reqData,
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -20,7 +20,7 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
-// The autoscaler-agent currently supports v1.0 to v1.1 of the agent<->informant protocol.
+// The autoscaler-agent currently supports v1.0 to v2.0 of the agent<->informant protocol.
 //
 // If you update either of these values, make sure to also update VERSIONING.md.
 const (
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -978,10 +978,11 @@ func (s *InformantServer) Downscale(ctx context.Context, to api.Resources) (*api
 	s.runner.logger.Infof("Sending downscale %+v", to)
 
 	timeout := time.Second * time.Duration(s.runner.global.config.Informant.DownscaleTimeoutSeconds)
-	rawResources := to.ConvertToRaw(s.runner.vm.Mem.SlotSize)
+	id := api.AgentIdentification{AgentID: s.desc.AgentID}
+	signedRawResources := api.SignedRawResources{RawResources: to.ConvertToRaw(s.runner.vm.Mem.SlotSize), Id: id}
 
-	resp, statusCode, err := doInformantRequest[api.RawResources, api.DownscaleResult](
-		ctx, s, timeout, http.MethodPut, "/downscale", &rawResources,
+	resp, statusCode, err := doInformantRequest[api.SignedRawResources, api.DownscaleResult](
+		ctx, s, timeout, http.MethodPut, "/downscale", &signedRawResources,
 	)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/informant/endpoints.go
```diff
@@ -375,7 +393,22 @@ func (s *State) NotifyUpscale(ctx context.Context, newResources *api.RawResource
...
+	// 2: The agent that responded is using an old protocol, and thus did not respond with an AgentId.
+	//    In this case, due to Go's uninitialized variable semantics, the "AgentID" we read will be
+	//    all 0's, which will never match, since UUID's cannot be all 0's
+	if incomingId != currentId {
+		klog.Errorf(
+			"Got upscale response from agent %v, while current agent is %v",
+			incomingId, currentId,
+		)
+		return nil, 400, fmt.Errorf("Received upscale response from unknown agent %v", incomingId)
+	}
```
sharnoff, line 409: This check should be done before the "no action needed" check above — if the agent's bad, we _always_ want to return 400 (not: "sometimes, if the vm-informant is in a silly goofy mood")

pkg/api/types.go
```diff
@@ -495,6 +514,13 @@ type RawResources struct {
 	Memory *resource.Quantity `json:"memory"`
 }
 
+// Similar to RawResources, stores raw resource amounts. However, also stores the ID of the agent
+// notifying the VM of the granted resource request
+type SignedRawResources struct {
+	RawResources
+	Id AgentIdentification `json:"id"`
+}
```
sharnoff, line 522: There's a few things I'd note here:

1. _In general_ the type names here serve to describe the _function_ of the message, rather than the _structure_. (e.g. `MoreResourcesRequest` or `DownscaleResult`) I think `RawResources` was kind of a unique exception. I'd initially though that the struct embedding was weird here, but I think it's just the name.
2. Simiarly, the comment above this type should have additional information about the function (i.e. "what is this used for? who sends it? why?")
3. Another thing I wanted to put on your radar: instead of writing `api.AgentMessage[api.AgentIdentification]` everywhere, we have a type alias `AgentIdentificationMessage`. A similar approach _may_ be useful here

pkg/api/VERSIONING.md
```diff
@@ -13,7 +13,7 @@ commit in this repository, possibly unreleased.
 
 | Release | autoscaler-agent | VM informant |
 |---------|------------------|--------------|
-| _Current_ | v1.0 - v1.2 | v1.1 - v1.2 |
+| Forthcoming | **v1.0 - v1.3** | **v1.1 - v1.2** |
```
sharnoff, line 16:
```suggestion
| _Current_ | **v1.0 - v1.3** | **v1.1 - v1.2** |
```

In the past we've (read: I've) left this as "Current" even when it's different from the latest release. I'd just stick with that tbh.

---

Some mostly unrelated context: The release flow (build+push images, etc) is triggered by pushing a tag; I'll make a commit directly to main that updates these tables, tag it (with a bunch of info in the tag), and then push everything.

pkg/informant/endpoints.go
```diff
@@ -258,7 +258,22 @@ func (s *State) TryDownscale(ctx context.Context, target *api.RawResources) (*ap
...
+	// 2: The agent that responded is using an old protocol, and thus did not respond with an AgentId.
+	//    In this case, due to Go's uninitialized variable semantics, the "AgentID" we read will be
+	//    all 0's, which will never match, since UUID's cannot be all 0's
+	if incomingId != currentId {
+		klog.Errorf(
+			"Got downscale response from agent %v, while current agent is %v",
+			incomingId, currentId,
+		)
+		return nil, 400, fmt.Errorf("Received downscale response from unknown agent %v", incomingId)
+	}
```
sharnoff, line 274: Missed this in earlier comments, but similar "move this above the other check" for the same reason.

sharnoff (commented):
Left a handful of comments, varying levels of depth/significance.

Some additional places that need updating:

* ARCHITECTURE.md refers to "RawResources" in the "Agent-Informant protocol details" section
* The constants for the actual versions that the vm-informant uses should be updated as well (defined at the top of `pkg/informant/agent.go` — `{Min,Max}ProtocolVersion`)
  * OTOH, this (in combination with switching to sending an `AgentMessage` for older versions [see comment]) might be why the e2e tests failed.

pkg/agent/informant.go
```diff
@@ -978,11 +978,24 @@ func (s *InformantServer) Downscale(ctx context.Context, to api.Resources) (*api
...
+	var resp *api.DownscaleResult
+	if s.protoVersion.ResourceUpdatesSigned() {
+		signedRawResources := api.SignedRawResources{RawResources: rawResources, Id: id}
+		reqData := api.AgentMessage[api.SignedRawResources]{Data: signedRawResources, SequenceNumber: s.incrementSequenceNumber()}
+		resp, statusCode, err = doInformantRequest[api.AgentMessage[api.SignedRawResources], api.DownscaleResult](
+			ctx, s, timeout, http.MethodPut, "/downscale", &reqData,
+		)
+	} else {
+		reqData := api.AgentMessage[api.RawResources]{Data: rawResources, SequenceNumber: s.incrementSequenceNumber()}
+		resp, statusCode, err = doInformantRequest[api.AgentMessage[api.RawResources], api.DownscaleResult](
```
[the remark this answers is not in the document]
sharnoff: Before I forget: both the upscale and downscale functions are the same in this regard, so both should be fixed.

sharnoff (commented):
Broadly looks good. I think the remaining significant items are around comments documentation and error messages. On a related note:

In a number of places you refer to the /upscale and /downscale endpoints as the agent "responding" to a request. To clarify: the agent _always_ sends requests to the informant's /upscale and /downscale endpoints whenever there's a change in the VM resources. Sometimes that "change in the VM resources" is triggered by a request _from_ the informant (to the agent's /try-upscale endpoint), but often it's not.

---

Also ARCHITECTURE.md still needs updating.

pkg/informant/endpoints.go
```diff
@@ -242,7 +242,19 @@ func (s *State) HealthCheck(ctx context.Context, info *api.AgentIdentification)
 // amount is ok
 //
 // Returns: body (if successful), status code and error (if unsuccessful)
-func (s *State) TryDownscale(ctx context.Context, target *api.RawResources) (*api.DownscaleResult, int, error) {
+func (s *State) TryDownscale(ctx context.Context, target *api.AgentResourceMessage) (*api.DownscaleResult, int, error) {
+	currentId := s.agents.current.id
```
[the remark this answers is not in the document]
fprasx: Ahh good catch!!!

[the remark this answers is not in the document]
sharnoff: Hm, actually, thinking about this further, we also have to handle that `s.agents.current` may be nil. We currently have `(*AgentSet).Get()` — maybe a similar method `(*AgentSet).Current()` would be useful?

sharnoff (approved):

PR #321, opened 2023-06-06
deploy: Change metrics scrape interval 10s -> 60s

sharnoff:
10 seconds is too often, for little gain, at the expense of shoving a bunch more data through VictoriaMetrics.

Realistically it's not *that* bad, but still worth changing.


lassizci (approved):
We do mostly with 60s interval, so it’s good to reduce from 10s if there’s no real benefit, as currently there’s no support for multiple retention periods

sharnoff:
Makes sense. I'll switch it to 60s

PR #323, opened 2023-06-07
agent, plugin, informant: Start switch from klog -> zap

sharnoff:
The goal here is to switch _some_ components to _some_ kind of structured logging, so that we can have:

1. Uniformity among logged lines
2. Fewer isolated ad-hoc ways of adding context to log lines
3. Easier ways of retaining context in log lines
4. More machine-readable logs (this lets us have _much_ nicer dashboards)

I've also written up a general "here's how we want to do logging in this repo" document. [Rendered](https://github.com/neondatabase/autoscaling/blob/sharnoff/zap/LOGGING.md)

**This PR only touches the `autoscaler-agent`, `scheduler`, and `vm-informant`.** (I wanted a smaller initial PR, but it isn't possible to switch any of those fully away from klog without affecting the others, unfortunately.)

Notes for reviewers:

* Feel free to review as much or as little as you'd like. Feedback on LOGGING.md is just as useful as review of the implementation.
* Correct usage of `zap.Any` requires care because it just uses the standard library's JSON marshaller. I've manually gone through each usage and checked that it's ok.


cmd/autoscale-scheduler/main.go
```diff
@@ -53,7 +60,12 @@ func runProgram() (err error) {
 		return err
 	}
 
-	command := app.NewSchedulerCommand(app.WithPlugin(plugin.Name, plugin.NewAutoscaleEnforcerPlugin(ctx, conf)))
+	// The normal scheduler outputs to klog, and there isn't *really* a way to stop that. So to make
+	// everything fit nicely, we'll redirect it to zap as well.
+	redirectKlog(logger.Named("klog"))
```
sharnoff, line 65: Noticed that the scheduler's klog output still isn't being redirected... need to fix that before merging. Maybe I need to supply the klog instance

pkg/plugin/state.go
```diff
@@ -745,14 +804,15 @@ func (s *pluginState) startMigration(ctx context.Context, pod *podState, vmClien
...
 
-	klog.Infof(
-		"[autoscale-enforcer] Migrate pod %v from node %s; node.vCPU.capacityPressure %d -> %d (%d -> %d spoken for)",
-		pod.name, pod.node.name, oldNodeVCPUPressure, pod.node.vCPU.CapacityPressure, oldNodeVCPUPressureAccountedFor, pod.node.vCPU.PressureAccountedFor,
-	)
+	// FIXME: this is incorrect, needs fixing by another PR
+	// klog.Infof(
+	// 	"[autoscale-enforcer] Migrate pod %v from node %s; node.vCPU.capacityPressure %d -> %d (%d -> %d spoken for)",
+	// 	pod.name, pod.node.name, oldNodeVCPUPressure, pod.node.vCPU.CapacityPressure, oldNodeVCPUPressureAccountedFor, pod.node.vCPU.PressureAccountedFor,
+	// )
```
sharnoff, line 815: * #324

fprasx (approved):
Looks pretty solid to me. One thing: for the log messages that used to be prefixed with `[autoscale-enforcer]`, how is that information now communicated?

sharnoff:
> for the log messages that used to be prefixed with `[autoscale-enforcer]`, how is that information now communicated?

For context, that was previously used to distinguish between the "core" scheduler pieces outside our plugin, and the plugin itself. Now, we (should) have that core scheduler logged to zap as well, with the logger named `autoscale-scheduler.klog` name (and the plugin is `autoscale-scheduler.plugin`). So the `logger` field of each log line can now be used to disambiguate (once I fix the klog issue I mentioned above: https://github.com/neondatabase/autoscaling/pull/323#discussion_r1220840682)

pkg/plugin/plugin.go
```diff
@@ -713,25 +754,30 @@ func (e *AutoscaleEnforcer) Unreserve(
 ) {
 	e.metrics.pluginCalls.WithLabelValues("unreserve").Inc()
 
-	pName := util.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}
-	klog.Infof("[autoscale-enforcer] Unreserve: Handling request for pod %v, node %s", pName, nodeName)
+	podName := util.GetNamespacedName(pod)
+
+	logger := e.logger.With(zap.String("method", "Filter"), zap.String("node", nodeName), util.PodNameFields(pod))
```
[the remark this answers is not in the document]
sharnoff: Ah, good catch - thanks! :sweat_smile:

LOGGING.md
```diff
@@ -0,0 +1,58 @@
...
+`logger` field for lines generated by the new logger.
+
+[`logger.Named(name)`]: https://pkg.go.dev/go.uber.org/zap#Logger.Named
+
+_In general_, all tight groupings of long-lived goroutines (sometimes just one) are given a leaf
+node in the hierarchy. For example: the autoscaler-agent's `Runner`s spawn a handful of groroutines
+to handle various tasks. The goroutine responsible for updating which scheduler to communciate with
+has limited communication and a lifecycle that is notably independent from other parts of the
+`Runner`. But the part that fetches metrics is (relatively) closely tied to the part that updates
+the VM's resources, so they share a logger name (`agent.runner.main`).
```
tychoish, line 47: To be honest, hierarchical logging functionality is perhaps one of my least favorite and most bothersome logging features (and I think it's kind of retro and imagines a different sort of execution model and way of interaction with logging from the one we actually have/interact with these days.) 

In short it makes a lot of sense as a way of filtering logs, particularly when you might want to control log output at runtime (e.g. moving from info to debug transiently) or to be able to set some hierarchies with different levels. But we don't do this, and also often use much better tools (like grep! or jq! or splunk!) to search and filter logs. 

What you're left with is this "magic", potentially easy to parse, string that is supposed to tell you about the component that you end up filtering things by, and then if a component moves around or gets called from a different place, you can end up with a different hierarchy and therefore a different string. In Cloud (and this is an incredibly common  mode, we create new loggers in the hierarchy using `logger.FromLogger(logger, "workerOp")` or similar... and as things move around or get called from a different place the messages end up changing. 

We have this in cloud and it's _it's confusing_. Sometimes operations (e.g. applying a config) are triggered by background jobs, in which case the log hierarchy says "background job" at the top, but sometimes we call the same code from a web request, which means that the loggers get `publicapi_v2` as a prefix. 

We can say "don't do this" or "be very careful about creating new loggers," but adding friction to logging, means that people will be less likely to use loggers or will do something slapdash when firefighting, and the point of using better logging tools is to avoid this kind of friction and hesitation. 

-----

I've always thought that the combination of structured logging and hierarchical logging doesn't make a lot of sense. Just put the component name or system name in as a field, throw all logs into one stream and filter it out later as you need it. Structured logging gives you the ability, in practice, to not need to think about this at all. 

-----

One thing that hierarchical logging kind of gives you that's hard to recreate directly with structured logging is the _sense_ of a call stack for a logging component. This can be quite useful, but I'm inclined to acctually just have helpers that annotate specific log messages (or all log messages depending on volume or form.) with stack traces.

pkg/agent/entrypoint.go
```diff
@@ -25,67 +25,62 @@ type MainRunner struct {
...
 	if err := srv.GetOrchestrator(ctx).Add(srv.Broker(broker)); err != nil {
 		return err
 	}
 
-	watcherLogger := util.PrefixLogger{Prefix: "Scheduler Watcher: "}
-	schedulerStore, err := schedwatch.StartSchedulerWatcher(ctx, watcherLogger, r.KubeClient, watchMetrics, broker, r.Config.Scheduler.SchedulerName)
+	schedulerStore, err := schedwatch.StartSchedulerWatcher(ctx, logger, r.KubeClient, watchMetrics, broker, r.Config.Scheduler.SchedulerName)
 	if err != nil {
-		return fmt.Errorf("starting scheduler watch server: %w", err)
+		return fmt.Errorf("Starting scheduler watch server: %w", err)
```
tychoish, line 54: I wouldn't do this.. errors are hard to grep, not supposed to be sentences (e.g. linter's nag/used to nag if you put a period, etc.)

tychoish (commented):
capitalization style is inconsistent, I'd go for "lowercase basically anything where it's not flagrantly incorrect or confusing to"

sharnoff:
@tychoish and I spoke at length yesterday about this PR. I'll write more in response to the individual comments, but the tl;dr is:

I'll spend some time polishing up this PR because future incremental improvements are always harder to justify than the initial thing. Broadly it's _mostly fine_, but we need a better story around state[lessness] before merging. A lot of the things that make certain patterns bad in `neondatabase/cloud` are _less bad_ here because our architecture isn't as uniform (e.g., we have no central jobs dispatcher, but instead many bespoke solutions for repeatedly doing things)

cmd/autoscale-scheduler/main.go
sharnoff: Thought it might've been that for a minute. Did some digging, looks like klog _by default_ has `LogToStderr` set to `true`, which completely bypasses the output redirection :confused:

But calling `klog.LogToStderr(false)` fixes it, so... should be ok I think.

LOGGING.md
sharnoff: recording a proper response here for posterity (even though we talked through most of this already):

> To be honest, hierarchical logging functionality is perhaps one of my least favorite and most bothersome logging features

:disappointed:

> I think [hierarchical logging is] kind of retro and imagines a different sort of execution model and way of interaction with logging from the one we actually have/interact with these days.

I can see how the execution model piece makes sense for something like the control plane. For this repo, and _particularly_ for the design of the autoscaler-agent, we _do_ actually have many independent goroutines doing stuff — without a central jobs distributor managing them.

> In short it makes a lot of sense as a way of filtering logs, particularly when you might want to control log output at runtime (e.g. moving from info to debug transiently) or to be able to set some hierarchies with different levels. But we don't do this, and also often use much better tools (like grep! or jq! or splunk!) to search and filter logs.

A few things here:

1. I wanted it for the filtering! Filtering is good. The existing autoscaling logs uses filename-based filtering to achieve what I'd hope would be accomplished by this, and even though it's very hacky, it's _incredibly_ useful.
2. Controlling log output at runtime is... fine, I guess? I have no desire to do that here.
3. Relating to tools, IME we do not currently use much better tools (see for example the [autoscaling logs dashboard](https://neonprod.grafana.net/d/cc4f137c-f2fd-4e18-a00a-c92234e3b33a/neon-logs-autoscaling)); and even if we did, the ability to filter by the internal subsystem is always going to be difficult if we don't _explicitly_ make that information available in the logs.

> What you're left with is this "magic", potentially easy to parse, string that is supposed to tell you about the component that you end up filtering things by, and then if a component moves around or gets called from a different place, you can end up with a different hierarchy and therefore a different string. In Cloud (and this is an incredibly common mode, we create new loggers in the hierarchy using `logger.FromLogger(logger, "workerOp")` or similar... and as things move around or get called from a different place the messages end up changing.
>
> We have this in cloud and it's it's confusing. Sometimes operations (e.g. applying a config) are triggered by background jobs, in which case the log hierarchy says "background job" at the top, but sometimes we call the same code from a web request, which means that the loggers get `publicapi_v2` as a prefix.

Regarding the top-level loggers in cloud, from what you've said it does seem that the `logger` field is unhelpful under the current architecture. See also: the stuff I said above about the architecture in this repo being different.

In my view, this is not something that should be a "magic" string in any sense; there should be a fixed number of `logger` values, so that you can filter over which subsystems you'd like to view the output for (or gather metrics by "which subsystem is experiencing errors", etc.). This is why in my examples it's things like `autoscaler-agent.runner.main`.

> We can say "don't do this" or "be very careful about creating new loggers," but adding friction to logging, means that people will be less likely to use loggers or will do something slapdash when firefighting, and the point of using better logging tools is to avoid this kind of friction and hesitation.

I don't think there's any (or: at least not many) shortcuts to get to a place where the structured logging is easy to manipulate algorithmically — it requires common keys, and various other disciplined behaviors that are _easier_ with a better-designed architecture, but still aren't guaranteed! Some of this can be fixed by investing into an internal logging shim/API that provides a lot of these "common keys" as methods on the logger as a form of standardization. Ultimately I'm skeptical of the idea that there's low-effort ways to get high-quality logs.

> I've always thought that the combination of structured logging and hierarchical logging doesn't make a lot of sense. Just put the component name or system name in as a field, throw all logs into one stream and filter it out later as you need it. Structured logging gives you the ability, in practice, to not need to think about this at all.
>
> One thing that hierarchical logging kind of gives you that's hard to recreate directly with structured logging is the sense of a call stack for a logging component. This can be quite useful, but I'm inclined to acctually just have helpers that annotate specific log messages (or all log messages depending on volume or form.) with stack traces.

A final couple notes:

1. I think under the model of "logger names are a proxy for a callstack" (which you elude to here), the hierarchical logging thing _does_ fall apart. I'm not intending for the represented hierarchy to represent a callstack (nor does LOGGING.md say that). My belief is that the set of guidelines I've proposed should work reasonably for this project as it currently is, and avoid some of the pitfalls you've described.
2. The immediate question in response to "just put the component name as a field" is: "given the existing hierarchy, why aren't you just using the `.Named()` method to do this for you?". There's value in using the standardized thing, even if the API is... lacking, and doesn't grant nearly as much freedom to change the `logger` name as we'd like.

pkg/agent/entrypoint.go
sharnoff: We discussed; leaving as-is. Capitals help to draw the eyes, and given we're not matching on the content of the errors, upper vs lower case doesn't matter aside from how it's displayed.

pkg/agent/entrypoint.go
```diff
@@ -25,67 +25,62 @@ type MainRunner struct {
...
-	klog.Info("Starting VM watcher")
-	vmWatchStore, err := startVMWatcher(ctx, r.Config, r.VMClient, watchMetrics, r.EnvArgs.K8sNodeName, pushToQueue)
+	logger.Info("Starting VM watcher")
+	vmWatchStore, err := startVMWatcher(ctx, logger, r.Config, r.VMClient, watchMetrics, r.EnvArgs.K8sNodeName, pushToQueue)
 	if err != nil {
 		return fmt.Errorf("Error starting VM watcher: %w", err)
 	}
 	defer vmWatchStore.Stop()
-	klog.Info("VM watcher started")
+	logger.Info("VM watcher started")
```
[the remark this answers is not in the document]
sharnoff: Given this only happens once per lifetime of the autoscaler-agent, and it _can_ be useful if something stalls during startup (or: if starting the VM watcher is taking a long time), I'm going to leave it as-is. The log volume from this should be miniscule.

pkg/plugin/plugin.go
```diff
@@ -419,14 +428,17 @@ func (e *AutoscaleEnforcer) Filter(
...
 	} else {
-		resultString = "rejecting"
+		resultString = "Rejecting"
 	}
 
-	klog.Infof(
-		"[autoscale-enforcer] Filter: %s %s pod %v in node %s: %s; %s",
-		resultString, kind, pName, nodeName, cpuMsg, memMsg,
+	logger.Info(
+		fmt.Sprintf("%s Pod", resultString),
```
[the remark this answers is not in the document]
sharnoff: `resultString` is either "Allowing" or "Rejecting" — the result is already derivable from the verdict, but IMO it's useful to have a message that makes sense by itself, without having to look at the keys (possibly many of them) to tell what actually happened.

As written, this will either have a message of "Allowing Pod" or "Rejecting Pod". I'd like to keep the messages.

Cleaned up `resultString` so that now it's called `message` and skips the unnecessary `fmt.Sprintf` step.

pkg/util/vmlogfield.go
```diff
@@ -0,0 +1,59 @@
...
+		podName := NamespacedName{Namespace: vm.Namespace, Name: vm.Status.PodName}
+
+		return zap.Inline(nameFields{
+			virtualmachine: vmName,
+			pod:            podName,
+		})
+	}
+}
+
+func PodNameFields(pod *corev1.Pod) zap.Field {
```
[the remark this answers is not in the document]
sharnoff: Yeah... I thought about that. Realistically, there definitely _are_ places where we have duplicate keys. TBH I think it's fine as-is, as long as we're not recursively adding them (and as long as it's not being overwritten with _different_ values), we should be ok — it's easier to do this imprecisely.

sharnoff:
Ok, I think there's some remaining work that should be done on this, but given the urgent stuff has been fixed, I'd rather handle it after merging. These are:

1. Adding a new package to group logging-related things. `util.PodNameField` and `util.VMNameField` should go here.
2. Where possible, removing string names for zap fields, and instead having dedicated methods in the logging package (interface could be something like `log.VM(zap.Object, vmName)`)

.. and maybe some others, as they come up. We could also do the work on making a full "logger" (wrapper around `zapcore`), which _may_ be worth it, but I don't know. Stuff to think about.

PR #331, opened 2023-06-14
plugin: Fix readClusterState handling of completed pods

sharnoff:
This should have been handled in #260, but it was missed there.

~~**NB:** This PR builds on #323, and must not be merged before it.~~

### DIFF
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -964,6 +964,8 @@ func (p *AutoscaleEnforcer) readClusterState(ctx context.Context, logger *zap.Lo
 		} else if pod.Spec.NodeName == "" {
 			logSkip("VM pod Spec.NodeName = \"\" (maybe it hasn't been scheduled yet?)")
 			continue
+		} else if util.PodCompleted(pod) {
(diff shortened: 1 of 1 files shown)


pkg/plugin/state.go
```diff
@@ -964,6 +964,8 @@ func (p *AutoscaleEnforcer) readClusterState(ctx context.Context, logger *zap.Lo
 		} else if pod.Spec.NodeName == "" {
 			logSkip("VM pod Spec.NodeName = \"\" (maybe it hasn't been scheduled yet?)")
 			continue
+		} else if util.PodCompleted(pod) {
```
tychoish, line 967: strictly speaking you don't need any of these because there's no way past any of these if statements (there used to be one of the really annoying linters that people don't use any more that chided you for the unnecessary else, and while the linter is turned off it's still in my heart.)

sharnoff: Yeah. I often will include "unnecessary" control flow like that because the "else" is implied _anyways_ if the block above diverges (returns / continues / etc), and I tend to feel that writing the "else" explicitly makes it a bit more clear. But... :shrug:

PR #332, opened 2023-06-14
neonvm/controller: Set migration source ownerref

sharnoff:
Part of #333, adds an owner reference for the migration to the _source_ runner pod, so that it's possible to check whether a pod is involved in a migration just by checking it (without e.g. cross-referencing with the VM phase).

**NB:** This PR builds on #256, and must not be merged before it.

### DIFF
--- a/neonvm/controllers/virtualmachinemigration_controller.go
+++ b/neonvm/controllers/virtualmachinemigration_controller.go
@@ -233,9 +233,24 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 			log.Error(err, "Failed to get migration source pod")
 			return ctrl.Result{}, err
 		}
-		if err = controllerutil.SetOwnerReference(migration, sourceRunner, r.Scheme); err != nil {
-			log.Error(err, "Failed to set owner reference for source pod")
-			return ctrl.Result{}, err
+		ownedByMigration := false
+		gvk := vmv1.SchemeGroupVersion.String()
+		for _, ref := range sourceRunner.OwnerReferences {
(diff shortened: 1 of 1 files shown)


sharnoff:
I've validated that this works as expected for creation. Locally, migrations were stuck at pending, so I didn't check deletion — but AFAICT it should be ok, because of these lines:

https://github.com/neondatabase/autoscaling/blob/f947db3ea468c924a02bb11e92a454f111802521/neonvm/controllers/virtualmachinemigration_controller.go#L417-L424

(note that the owner references get completely unset, so that should remove the VMM's ref as well)

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -226,6 +226,33 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
+		sourceRunner := &corev1.Pod{}
+		err = r.Get(ctx, types.NamespacedName{Name: vm.Status.PodName, Namespace: vm.Namespace}, sourceRunner)
+		if err != nil {
+			log.Error(err, "Failed to get migration source pod")
+			return ctrl.Result{}, err
+		}
+		ownedByMigration := false
+		gvk := vmv1.SchemeGroupVersion.String()
+		for _, ref := range sourceRunner.OwnerReferences {
+			if ref.APIVersion == gvk && ref.Kind == "VirtualMachineMigration" && ref.Name == migration.Name {
```
cicdteam, line 239: may be this more reliable ?
```go
if ref.UID == migration.GetUID()
```

sharnoff: Oh, yup - thanks

cicdteam (approved):

Issue #324, opened 2023-06-07
Scheduler plugin state update on migrations is deficient


sharnoff:
Basically, `(*pluginState).startMigration` needs to be updated (as written, it only updates CPU!)

Also, it should be careful to _not_ update the state unless starting the migration actually succeeds.

... And because of that, it probably _instead_ shouldn't actually update the state directly, watching for when the VM starts migrating instead. Because then we can accurately interpret the state of VMs that are already migrating when the scheduler starts! (instead of e.g. saying "gee, haven't heard from the autoscaler-agent about this VM [that it isn't allowed to touch, because it's migrating], I guess it could be scaling at any time, so we should assume the worst!")

PR #333, opened 2023-06-14
plugin: Fix migration handling

sharnoff:
This fixes all the cases mentioned in #324. Testing is still required before enabling scheduler-triggered migration (#134).

Will not handle migrations entirely correctly without #332, but the handling without migrations will remain ok either way.

**NB:** This PR builds on #330, and must not be merged before it.


sharnoff:
I've validated that this works as expected with #332.

pkg/plugin/state.go
```diff
@@ -796,43 +860,70 @@ func (s *podState) isBetterMigrationTarget(other *podState) bool {
...
+			// TODO: it's maybe possible for this to run into name length limits? Unclear what we
+			// should do if that happens.
+			Name:      vmmName.Name,
+			Namespace: pod.name.Namespace,
 		},
 		Spec: vmapi.VirtualMachineMigrationSpec{
 			VmName: pod.vmName.Name,
+
+			// FIXME: NeonVM's VirtualMachineMigrationSpec has a bunch of boolean fields that aren't
+			// pointers, which means we need to explicitly set them when
```
tychoish, line 909: when what?

pkg/plugin/trans.go
```diff
@@ -273,6 +273,37 @@ func (r resourceTransition[T]) handleAutoscalingDisabled() (verdict string) {
 	return verdict
 }
 
+// handleStartMigration updates r.node with changes to clear any buffer and capacityPressure from
+// r.pod.
+//
+// If the pod is the migration source, this method *also* increases the node's PressureAccountedFor
+// to match the pod's resource usage.
+func (r resourceTransition[T]) handleStartMigration(source bool) (verdict string) {
+	// The first bit of this is _mostly_ the same as handleAutoscalingDisabled.
```
tychoish, line 282: I'd avoid leaving the difference that isn't covered by "mostly" as an exercise to the reader.

pkg/plugin/trans.go
```diff
@@ -273,6 +273,37 @@ func (r resourceTransition[T]) handleAutoscalingDisabled() (verdict string) {
...
+		*v -= buffer
+	}
+
+	r.node.CapacityPressure -= r.pod.CapacityPressure
+	r.pod.CapacityPressure = 0
+
+	r.node.PressureAccountedFor += r.pod.Reserved
+
+	fmtString := "pod had buffer %d, capacityPressure %d; " +
+		"node reserved %d -> %d, capacityPressure %d -> %d, pressureAccountedFor %d -> %d"
```
tychoish, line 296: I'd just build this all as a giant `fmt.Sprint` call and avoid the string concatination (so you can use the buffer pool), and then you don't have to keep the deal with having the format string in the comment below as well.

pkg/plugin/state.go
```diff
@@ -678,6 +682,66 @@ func (e *AutoscaleEnforcer) handleVMDisabledScaling(logger *zap.Logger, podName
...
+
+	e.state.lock.Lock()
+	defer e.state.lock.Unlock()
+
+	pod, ok := e.state.podMap[podName]
+	if !ok {
+		logger.Warn("Cannot find pod in podMap")
+		return
+	}
+	logger = logger.With(zap.String("node", pod.node.name), zap.Object("virtualmachine", pod.vmName))
```
tychoish, line 738: I think it's a bit odd to make new logger objects rather than just adding some fields to a log message

tychoish (approved):
my comments are all pretty small. The code seems reasonable. 

it sort of feels like a regression test or something might be nice? Is there a plan for automated migration testing?

pkg/plugin/state.go
sharnoff: Oops, fixed - thanks!

pkg/plugin/trans.go
sharnoff: Done

pkg/plugin/trans.go
sharnoff: Leaving as-is for now because that's _currently_ the pattern used in the rest of the file.

(I'd imagine the compiler is ok doing this at compile-time though?)

pkg/plugin/state.go
sharnoff: I'd say this is a case of localized weirdness in order to fit the broader pattern among similar functions — as such, going to keep it as is for now.

sharnoff:
> it sort of feels like a regression test or something might be nice? Is there a plan for automated migration testing?

Yeah. I don't know what a good way to test this would be, because the implementation here really should only be visible while migrations are ongoing.

I've done some ad-hoc "watch the logs and check it looks right" sort of testing, but... that's neither scalable nor automated, lol.

PR #334, opened 2023-06-15
agent: Add scaling metrics

sharnoff:
Broadly, this PR primarily adds metrics that focus on the change in resources -- both "requested" changes (to: scheduler, informant, NeonVM) and "approved" changes (from: scheduler, informant).

The CPU and memory changes are separate metrics. Memory is in units of Mibibytes. Upscaling vs downscaling is represented by the "direction" label, which is either "up" or "down".

There's also the addition of NeonVM request counts and the associated result.

In total, the new metrics are:

* `autoscaling_agent_scheduler_plugin_{requested,approved}_{cpu,mem}_change_total`
* `autoscaling_agent_informant_{requested,approved}_{cpu,mem}_change_total`
* `autoscaling_agent_neonvm_requested_{cpu,mem}_change_total`
* `autoscaling_agent_neonvm_outbound_requests_total`

So, 11 new metrics in total.

---

This PR is part 1 of 3, for upcoming metrics additions. The next pieces are:

* Gauges for per-node utilization in the scheduler
  - Blocked on cleaning up node info when nodes are removed, plus some extra questions on how metrics should be handled there.
* Timing histograms/etc for how long it takes for a desired scaling to become reality (i.e. duration between autoscaler-agent realizing it should scale and that scaling coming into effect in the VM)
  - Blocked on an `agent.Runner` refactor that would allow us to track this in the first place.

### DIFF
--- a/pkg/agent/prommetrics.go
+++ b/pkg/agent/prommetrics.go
@@ -198,11 +198,13 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
 	// unil the value is non-zero (because something's happened), which makes it harder to
 	// distinguish between "valid signal of nothing" vs "no signal".
 	metricsWithDirection := []resourceChangePair{
-		// sched requested/approved
-		metrics.schedulerRequestedChange, metrics.schedulerApprovedChange,
-		// informant requested/approved
-		metrics.informantRequestedChange, metrics.informantApprovedChange,
-		// neonvm requested
+		// scheduler:
+		metrics.schedulerRequestedChange,
+		metrics.schedulerApprovedChange,
+		// informant:
+		metrics.informantRequestedChange,
+		metrics.informantApprovedChange,
+		// neonvm:
 		metrics.neonvmRequestedChange,
 	}
 	for _, p := range metricsWithDirection {
--- a/pkg/agent/prommetrics.go
+++ b/pkg/agent/prommetrics.go
@@ -32,9 +32,9 @@ type resourceChangePair struct {
 }
 
 const (
-	directionLabel     = "direction"
-	directionValueUp   = "up"
-	directionValueDown = "down"
+	directionLabel    = "direction"
+	directionValueInc = "inc"
+	directionValueDec = "dec"
 )
 
 func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Registry) {
@@ -207,8 +207,8 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
 	}
 	for _, p := range metricsWithDirection {
 		for _, m := range []*prometheus.CounterVec{p.cpu, p.mem} {
-			m.WithLabelValues(directionValueUp).Add(0.0)
-			m.WithLabelValues(directionValueDown).Add(0.0)
+			m.WithLabelValues(directionValueInc).Add(0.0)
+			m.WithLabelValues(directionValueDec).Add(0.0)
 		}
 	}
 
--- a/pkg/agent/prommetrics.go
+++ b/pkg/agent/prommetrics.go
@@ -8,13 +8,27 @@ import (
 )
 
 type PromMetrics struct {
-	schedulerRequests         *prometheus.CounterVec
+	schedulerRequests        *prometheus.CounterVec
+	schedulerRequestedChange resourceChangePair
+	schedulerApprovedChange  resourceChangePair
+
 	informantRequestsOutbound *prometheus.CounterVec
 	informantRequestsInbound  *prometheus.CounterVec
-	runnerFatalErrors         prometheus.Counter
-	runnerThreadPanics        prometheus.Counter
-	runnerStarts              prometheus.Counter
-	runnerRestarts            prometheus.Counter
+	informantRequestedChange  resourceChangePair
+	informantApprovedChange   resourceChangePair
+
+	neonvmRequestsOutbound *prometheus.CounterVec
+	neonvmRequestedChange  resourceChangePair
+
+	runnerFatalErrors  prometheus.Counter
+	runnerThreadPanics prometheus.Counter
+	runnerStarts       prometheus.Counter
+	runnerRestarts     prometheus.Counter
+}
+
+type resourceChangePair struct {
+	cpu *prometheus.CounterVec
+	mem *prometheus.CounterVec
 }
 
 func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Registry) {
@@ -30,6 +44,8 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
 	metrics := PromMetrics{
 		// the util.RegisterMetric() function registers the collector and returns
 		// it so we can set it directly on the output structure.
+
+		// ---- SCHEDULER ----
 		schedulerRequests: util.RegisterMetric(reg, prometheus.NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "autoscaling_agent_scheduler_plugin_requests_total",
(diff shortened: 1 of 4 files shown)


pkg/agent/prommetrics.go
```diff
@@ -30,13 +44,49 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
...
 			},
 			[]string{"code"},
 		)),
+		schedulerRequestedChange: resourceChangePair{
+			cpu: util.RegisterMetric(reg, prometheus.NewCounterVec(
+				prometheus.CounterOpts{
+					Name: "autoscaling_agent_scheduler_plugin_requested_cpu_change_total",
+					Help: "Total change in CPU requested from the scheduler",
+				},
+				[]string{"direction"},
```
tychoish, line 62: maybe direction should be a constant

pkg/agent/prommetrics.go
```diff
@@ -45,11 +95,71 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
...
+			)),
+		},
+
+		// ---- NEONVM ----
+		neonvmRequestsOutbound: util.RegisterMetric(reg, prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "autoscaling_agent_neonvm_outbound_requests_total",
+				Help: "Number of k8s patch requests to NeonVM objects",
+			},
+			[]string{"result"},
```
tychoish, line 143: what's the cardinality of this

pkg/agent/prommetrics.go
```diff
@@ -76,6 +186,24 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
...
+		metrics.schedulerRequestedChange, metrics.schedulerApprovedChange,
+		// informant requested/approved
+		metrics.informantRequestedChange, metrics.informantApprovedChange,
+		// neonvm requested
+		metrics.neonvmRequestedChange,
+	}
+	for _, p := range metricsWithDirection {
+		for _, m := range []*prometheus.CounterVec{p.cpu, p.mem} {
+			m.WithLabelValues("up").Add(0.0)
+			m.WithLabelValues("down").Add(0.0)
```
tychoish, line 203: I'd probably go for inc/dec rather than up/down and just write a helper to do this inline when you declare the metrics

pkg/util/arith.go
```diff
@@ -34,6 +34,15 @@ func Min[T constraints.Ordered](x, y T) T {
 	}
 }
 
+// AbsDiff returns the absolute value of the difference between x and y
+func AbsDiff[T constraints.Unsigned](x, y T) T {
+	if x > y {
+		return x - y
+	} else {
```
tychoish, line 41: you don't need the else because you early return

pkg/agent/runner.go
```diff
@@ -1551,6 +1570,35 @@ func (r *Runner) doVMUpdate(
...
 
+func (r *Runner) recordResourceChange(current, target api.Resources, metrics resourceChangePair) {
+	abs := current.AbsDiff(target)
+
+	// Add CPU
+	if abs.VCPU != 0 {
+		direction := "up"
+		if target.VCPU < current.VCPU {
+			direction = "down"
+		}
```
tychoish, line 1581: I feel like a "getDirection" helper function would be nice

pkg/agent/runner.go
```diff
@@ -1551,6 +1570,35 @@ func (r *Runner) doVMUpdate(
...
+	// Add memory
+	if abs.Mem != 0 {
+		direction := "up"
+		if target.Mem < current.Mem {
+			direction = "down"
+		}
+
+		// Avoid floating-point inaccuracy.
+		byteTotal := r.vm.Mem.SlotSize.Value() * int64(abs.Mem)
+		mib := int64(1 << 20)
```
tychoish, line 1595: constant in the package?

why not wrap/convert to floats here just to make the math part on the next line easier to read?

pkg/agent/runner.go

pkg/util/arith.go
sharnoff: Keeping as-is for symmetry and consistency with other parts in the file.

pkg/agent/prommetrics.go
sharnoff: Done

pkg/agent/prommetrics.go
sharnoff: In theory it could be large, I think? It's roughly the same idea as in #287 — record "ok" or the root cause of an error.

I'm expecting that _realistically_ it should be quite small (see e.g. https://neonprod.grafana.net/d/aa1653e5-0c19-492f-9da3-b6e6336181d0/neon-autoscaling?orgId=1&viewPanel=21&from=1687110644779&to=1687118972276)

---

In any case, I've added a comment to explain that.

pkg/agent/prommetrics.go
sharnoff: Makes sense. Made "direction" and "up" and "down" all constants.

pkg/agent/runner.go
sharnoff: > constant in the package?

honestly I think it's easier to understand with the variable closer.

> why not wrap/convert to floats here just to make the math part on the next line easier to read?

So the next line very much _could_ be `float64(byteTotal)/float64(mib)` (which we could wrap with floats beforehand). But floating point math is... always iffy, and so IMO it's worthwhile here to split up `byteTotal/mib` into the integer and fractional parts .

pkg/agent/prommetrics.go
```diff
@@ -76,6 +194,24 @@ func makePrometheusParts(globalstate *agentState) (PromMetrics, *prometheus.Regi
...
+	// unil the value is non-zero (because something's happened), which makes it harder to
+	// distinguish between "valid signal of nothing" vs "no signal".
+	metricsWithDirection := []resourceChangePair{
+		// sched requested/approved
+		metrics.schedulerRequestedChange, metrics.schedulerApprovedChange,
+		// informant requested/approved
+		metrics.informantRequestedChange, metrics.informantApprovedChange,
+		// neonvm requested
+		metrics.neonvmRequestedChange,
+	}
```
tychoish, line 207: why not put hard returns after every comma? I think that would be clear too?

pkg/agent/runner.go
```diff
@@ -1507,6 +1518,8 @@ func (r *Runner) doVMUpdate(
 
 	// We couldn't update the VM
 	if err != nil {
+		r.global.metrics.neonvmRequestsOutbound.WithLabelValues(fmt.Sprintf("[error: %s]", util.RootError(err))).Inc()
```
tychoish, line 1521: cardinality concerns.

pkg/agent/runner.go
```diff
@@ -1521,6 +1534,8 @@ func (r *Runner) doVMUpdate(
 		panic(fmt.Errorf("Unexpected VM patch request failure: %w", err))
 	}
 
+	r.global.metrics.neonvmRequestsOutbound.WithLabelValues("ok").Inc()
```
tychoish, line 1537: constants for labels?

tychoish (approved):

pkg/agent/runner.go
sharnoff: This is the same metric that you've already mentioned — resolving this in favor of that thread.

pkg/agent/runner.go
sharnoff: Given this label value is only set once, in that location, I'm going to leave it

pkg/agent/prommetrics.go
sharnoff: Done

pkg/agent/prommetrics.go
sharnoff: Yeah I mean, I'd definitely be more certain about the standard library's HTTP errors rather than the K8s API errors, but at the end of the day it boils down to HTTP (ish) anyways, so I'm figuring it's probably fine.

As long as they aren't creating the base error with formatting, then there's a finite number — and I'd expect that we're never actually going to run into very many different types of errors if we do.

IME keeping this kind of thing around in metrics can be particularly useful when you're not sure how many distinct things are going wrong — like in the link above.

PR #246, opened 2023-05-09
Kubernetes 1.25

bayandin:
This PR includes the following

Version changes:
- Kubernetes 1.25.10 ~1.25.9~ (from 1.24.13)
- Kubernetes API  0.25.10 ~0.25.9~ (from 0.24.13)
- kubectl 1.26.5 ~1.26.4~ (from 1.25.9)
- Controller Tools 0.10.0 (from 0.9.2)
- ❗ Code Generator 0.25.10 ~0.25.9~ (from 0.26.0) -- Downgrade to match k8s API version ❗ 
- kind image kindest/node:v1.25.9 ~v1.25.8~ (from v1.24.12)

Autogenerated changes:
- `make generate`
- `make manifests`

Other changes:
- Uncommented changes in `neonvm/controllers/virtualmachinemigration_controller.go` (related to `controllerutil.RemoveFinalizer`)
- `make generate` passes CONTROLLER_TOOLS_VERSION and CODE_GENERATOR_VERSION via build-arg to avoid defining version in several places
- Applied changes to Cluster Autoscaler proposed in https://github.com/neondatabase/autoscaling/pull/246#issuecomment-1557715505

PR #336, opened 2023-06-16
kind and kubectl version update for 1.25 k8s

cicdteam:
- `kind` updated to v0.20.0
- kind config updated for 1.25.11 k8s
- `kubectl` updated for 1.25 k8s

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -28,8 +28,8 @@ jobs:
         env:
           # kubectl is supported within one minor version (older or newer) of k8s
           # https://kubernetes.io/releases/version-skew-policy/#kubectl
-          KUBECTL_VERSION: v1.26.5
-          KIND_VERSION: v0.19.0
+          KUBECTL_VERSION: v1.25.11
+          KIND_VERSION: v0.20.0
           KUTTL_VERSION: v0.15.0
 
       - name: Check dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -28,8 +28,8 @@ jobs:
         env:
           # kubectl is supported within one minor version (older or newer) of k8s
           # https://kubernetes.io/releases/version-skew-policy/#kubectl
-          KUBECTL_VERSION: v1.26.5
```
sharnoff, line 31: I would double-check why this was 1.26 in the first place (maybe @bayandin had a reason for it?), but otherwise looks good

sharnoff (approved):

.github/workflows/e2e-test.yaml
cicdteam: I don't see any kubectl version bindings in e2e tests or anything like that.

sharnoff: Ah, I mean because that kubectl version was changed from v1.25.9 -> v1.26.5 by #246

bayandin: > I would double-check why this was 1.26 in the first place (maybe @bayandin had a reason for it?), but otherwise looks good

There's no particular reason, it just should work, according to the comment above.
Should be ok to change it to a matching k8s version 😄

bayandin (approved):

PR #340, opened 2023-06-17
k3d support for local developmnt and tests

cicdteam:
Addons and improvements

- `make` targets changed:
  - `make kind-setup` instead of `make local-cluster`
- `make` targets added:
  - `make kind-destroy` - destroy local `kind` cluster
  - `make k3d-setup` - create and run `k3d` local cluster with Cilium CNI (built-in flannel doesn't support other CNIs)
  - `make k3d-destroy` - destroy local `k3d` cluster
  - `make load-images` (instead of `make kind-load`) - load docker images to local cluster (`kind` or `k3d` - depends on current kubectl context)
  - `Makefile` reorganized a bit, `make help` improved
  - GitHub actions runs for both `k3d` and `kind` based clusters


bayandin:
Sorry, it seems I missing some context here, why do we need k3d? Why do we need to keep both k3d and kind?

cicdteam:
> Sorry, it seems I missing some context here, why do we need k3d? Why do we need to keep both k3d and kind?

Cloud repo (console) uses k3d for local development. And it doesn't support install neonvm.

sharnoff (approved):
Broadly I think this looks good.

@bayandin:

> Why do we need to keep both k3d and kind?

I don't have an opinion on this (it's a tradeoff), but I think we should be intentional about it. As long as we're testing all options in CI (which the PR does), then IMO we're ok either way.

If either of y'all feel strongly one way or another, let's go with that.

bayandin:
To make sure I understand it right:
- We're adding k3d support to make it easier to integrate with the console repo tools, and make it more friendly for developers to work on both repos?
- Keeping kind "just in case" if we're not happy with k3d?

cicdteam:
> To make sure I understand it right:
> 


I would say that we are just adding the ability to choose which tool to use. You can use kind (`make kind-setup`) or k3d (`make k3d-setup`), all other commands are the same for both cluster types.

bayandin (approved):

PR #344, opened 2023-06-20
neonvm/controller: Better label/annotation propagation

sharnoff:
In particular, we don't want to propagate the annotation kubectl.kubernetes.io/last-applied-configuration to the runner pod, but the preexisting code also modified the VM's labels/annotations, which maybe is not what's desired.

This PR is split off from some commits in #279, because it's actually not related. ref https://github.com/neondatabase/autoscaling/pull/279#issuecomment-1597844234


tychoish (approved):
preallocate the maps, but otherwise this is fine.

cicdteam (approved):

PR #348, opened 2023-06-21
release workflow: Add whereabouts.yaml asset

sharnoff:
Noticed that whereabouts.yaml was missing while working on releasing v0.11.0 — AFAICT it was extracted from multus-eks.yaml and never added to the release workflow?

Not immediately blocking the release (I just used `make render-release` and took the yaml from there), but would be good to fix.


cicdteam:
Thanks. Also added one more line in release GA (to render that manifest)

cicdteam (approved):

PR #366, opened 2023-06-29
billing, agent/billing: Log IdempotencyKey of events

sharnoff:
In order to make this work on the agent's side, idempotency key generation needed to be *somewhat* exposed, so this PR adds the `Enrich` function, which handles the common "filling out the other fields" tasks for each event.

By abstracting this interface, we also remove the need for separate `(*Batch).AddAbsoluteEvent` vs `(*Batch).AddIncrementalEvent` methods, so now we just have a single `(*Batch).Add` method that just calls `Enrich`.

### DIFF
--- a/pkg/agent/billing/billing.go
+++ b/pkg/agent/billing/billing.go
@@ -96,7 +96,7 @@ func RunBillingMetricsCollector(
 	}
 
 	state.collect(conf, store, metrics)
-	batch := client.NewBatch()
+	batch := billing.NewBatch[billing.IncrementalEvent](client)
 
 	for {
 		select {
@@ -122,7 +122,7 @@ func RunBillingMetricsCollector(
 			//
 			// Don't reset metrics.batchSizeCurrent because it stores the *most recent* batch size.
 			// (The "current" suffix refers to the fact the metric is a gague, not a counter)
-			batch = client.NewBatch()
+			batch = billing.NewBatch[billing.IncrementalEvent](client)
 		case <-backgroundCtx.Done():
 			// If we're being shut down, push the latests events we have before returning.
 			logger.Info("Creating final billing batch")
@@ -248,7 +248,7 @@ func (s *metricsTimeSlice) tryMerge(next metricsTimeSlice) bool {
 	return merged
 }
 
-func logAddedEvent(logger *zap.Logger, event *billing.IncrementalEvent) *billing.IncrementalEvent {
+func logAddedEvent(logger *zap.Logger, event billing.IncrementalEvent) billing.IncrementalEvent {
 	logger.Info(
 		"Adding event to batch",
 		zap.String("IdempotencyKey", event.IdempotencyKey),
@@ -260,13 +260,13 @@ func logAddedEvent(logger *zap.Logger, event *billing.IncrementalEvent) *billing
 }
 
 // drainAppendToBatch clears the current history, adding it as events to the batch
-func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batch *billing.Batch) {
+func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batch *billing.Batch[billing.IncrementalEvent]) {
 	now := time.Now()
 
 	for key, history := range s.historical {
 		history.finalizeCurrentTimeSlice()
 
-		batch.Add(logAddedEvent(logger, billing.Enrich(batch, &billing.IncrementalEvent{
+		batch.Add(logAddedEvent(logger, batch.Enrich(billing.IncrementalEvent{
 			MetricName:     conf.CPUMetricName,
 			Type:           "", // set by billing.Enrich
 			IdempotencyKey: "", // set by billing.Enrich
@@ -277,7 +277,7 @@ func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batc
 			StopTime:  now,
 			Value:     int(math.Round(history.total.cpu)),
 		})))
-		batch.Add(logAddedEvent(logger, billing.Enrich(batch, &billing.IncrementalEvent{
+		batch.Add(logAddedEvent(logger, batch.Enrich(billing.IncrementalEvent{
 			MetricName:     conf.ActiveTimeMetricName,
 			Type:           "", // set by billing.Enrich
 			IdempotencyKey: "", // set by billing.Enrich
@@ -292,7 +292,7 @@ func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batc
 	s.historical = make(map[metricsKey]vmMetricsHistory)
 }
 
-func pushBillingEvents(conf *Config, batch *billing.Batch) error {
+func pushBillingEvents(conf *Config, batch *billing.Batch[billing.IncrementalEvent]) error {
 	ctx, cancel := context.WithTimeout(context.TODO(), time.Second*time.Duration(conf.PushTimeoutSeconds))
 	defer cancel()
 
--- a/pkg/agent/billing/billing.go
+++ b/pkg/agent/billing/billing.go
@@ -268,8 +268,8 @@ func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batc
 
 		batch.Add(logAddedEvent(logger, batch.Enrich(billing.IncrementalEvent{
 			MetricName:     conf.CPUMetricName,
-			Type:           "", // set by billing.Enrich
-			IdempotencyKey: "", // set by billing.Enrich
+			Type:           "", // set by batch.Enrich
+			IdempotencyKey: "", // set by batch.Enrich
 			EndpointID:     key.endpointID,
 			// TODO: maybe we should store start/stop time in the vmMetricsHistory object itself?
 			// That way we can be aligned to collection, rather than pushing.
@@ -279,8 +279,8 @@ func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batc
 		})))
 		batch.Add(logAddedEvent(logger, batch.Enrich(billing.IncrementalEvent{
 			MetricName:     conf.ActiveTimeMetricName,
-			Type:           "", // set by billing.Enrich
-			IdempotencyKey: "", // set by billing.Enrich
+			Type:           "", // set by batch.Enrich
+			IdempotencyKey: "", // set by batch.Enrich
 			EndpointID:     key.endpointID,
 			StartTime:      s.pushWindowStart,
 			StopTime:       now,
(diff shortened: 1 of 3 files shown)


pkg/billing/model.go
```diff
@@ -23,3 +26,44 @@ type IncrementalEvent struct {
...
+	setEventType()
+	idempotencyKey() *string
+}
+
+// Enrich sets the event's Type and IdempotencyKey fields, so that users of this API don't need to
+// manually set them
+//
+// Enrich is already called by (*Batch).Add, but this is used in the autoscaler-agent's billing
+// implementation, which manually calls Enrich in order to log the IdempotencyKey for each event.
+func Enrich[E enrichable](batch *Batch, event E) E {
```
tychoish, line 40: I think this doesn't need to be a generic, it's just a normal interface if you make the methods do the mutation themselves rather than needing to return  (I might, to avoid the whole "pointers everywhere" thing, have a "Init()` or `SetDefaults` method, that you can call in this or the batch interface.

pkg/billing/model.go
```diff
@@ -23,3 +26,44 @@ type IncrementalEvent struct {
...
+
+// setEventType implements enrichable
+func (e *IncrementalEvent) setEventType() {
+	e.Type = "incremental"
+}
+
+// idempotencyKey implements enrichable
+func (e *IncrementalEvent) idempotencyKey() *string {
+	return &e.IdempotencyKey
+}
```
tychoish, line 69: can you group methods in this file by type?

pkg/billing/model.go
```diff
@@ -23,3 +26,44 @@ type IncrementalEvent struct {
...
+// manually set them
+//
+// Enrich is already called by (*Batch).Add, but this is used in the autoscaler-agent's billing
+// implementation, which manually calls Enrich in order to log the IdempotencyKey for each event.
+func Enrich[E enrichable](batch *Batch, event E) E {
+	event.setEventType()
+
+	key := event.idempotencyKey()
+	if *key == "" {
+		*key = fmt.Sprintf("Host<%s>:ID<%s>:T<%s>", batch.c.hostname, uuid.NewString(), time.Now().Format(time.RFC3339))
```
tychoish, line 45: in the spirit of "encapsulation" it seems like this really just wants to be a method on batch because you have to look at it's private fields, and the only reason it can't be a method is the generic thing, which (I might be missing something subtle, but) I think is avoidable.

pkg/billing/client.go
```diff
@@ -40,24 +37,9 @@ func (b *Batch) Count() int {
...
-	e.Type = "absolute"
-	e.IdempotencyKey = b.idempotenize(e.IdempotencyKey)
-	b.events = append(b.events, &e)
-}
-
-func (b *Batch) AddIncrementalEvent(e IncrementalEvent) {
-	e.Type = "incremental"
-	e.IdempotencyKey = b.idempotenize(e.IdempotencyKey)
-	b.events = append(b.events, &e)
+func (b *Batch) Add(e enrichable) {
```
tychoish, line 40: I think exported methods that accept private interfaces (particularly interface types with unexported methods) are a code smell redflag, for me.

I might:
- Make the interface public,
- make batch generic and have it be, basically `type Batch[E IncrementalEvent | AbsoulteEvent] struct {}` and then pass the `E` variable into a method/function that will do the setup, and I think you get all of typing strictness that you want.

pkg/billing/client.go
```diff
@@ -40,24 +37,9 @@ func (b *Batch) Count() int {
...
+	Enrich(b, e) // Realistically, we're already calling Enrich before Add, but this is a good safety mechanism.
```
tychoish, line 41: a better safety mechanism might be a test ;)

I think I'd opt for calling it in only one of these places (my vote is this one, with some minor modifications)

pkg/agent/billing/billing.go
```diff
@@ -265,26 +266,26 @@ func (s *metricsState) drainAppendToBatch(logger *zap.Logger, conf *Config, batc
 	for key, history := range s.historical {
 		history.finalizeCurrentTimeSlice()
 
-		batch.AddIncrementalEvent(logAddedEvent(logger, billing.IncrementalEvent{
+		batch.Add(logAddedEvent(logger, billing.Enrich(batch, &billing.IncrementalEvent{
```
tychoish, line 269: I missed this change the first time I read this. I'm not super into this, which is easy enough to miss and ends up being "lets pass batch it itself."

tychoish (changes requested):

pkg/billing/client.go
sharnoff: It can't be just this one because we're trying to log `IdempotencyKey` in pkg/agent/billing

pkg/billing/client.go
sharnoff: Made the change. It's... maybe a bit ugly. LMK what you think.

pkg/agent/billing/billing.go
sharnoff: Not sure what you mean? Also, have since updated, so maybe that resolves this?

pkg/billing/model.go
sharnoff: Made the change.

pkg/billing/model.go
sharnoff: Not sure what you mean, can you elaborate?

pkg/billing/model.go
sharnoff: Updated.

pkg/billing/model.go
tychoish: sorry, 
```go
type foo struct{}
func (foo) thingOne() {}
func (foo) thingTwo() {}
type bar struct{}
func (bar) thingOne() {}
func (bar) thingTwo() {}
```

pkg/agent/billing/billing.go
tychoish: while you don't do this, this code sort of looks like you might be... passing (and storing?) batch inside of the Event... I think it's maybe/probably fine, it just sort of looks like a leaky abstraction or a possible circular dependency.

pkg/billing/client.go
tychoish: put the logging in the add/batch method?

pkg/billing/client.go
tychoish: I don't see the change?

sharnoff: > * Make the interface public,

https://github.com/neondatabase/autoscaling/blob/8579616217587b55150fc1ba578f9bda5b1a4e06/pkg/billing/model.go#L7-L28

> * make batch generic and have it be, basically type Batch[E IncrementalEvent | AbsoulteEvent] struct {} and then pass the E variable into a method/function that will do the setup, and I think you get all of typing strictness that you want.

https://github.com/neondatabase/autoscaling/blob/8579616217587b55150fc1ba578f9bda5b1a4e06/pkg/billing/client.go#L34-L38

pkg/billing/client.go
sharnoff: That feels weird, a little bit "spooky-action-at-a-distance"-y. Basically, I don't think the billing client should log anything, and that should be left to the API user.

(next PR will also make `Batch.events` public, in order to e.g. decrease size if we get a 413 / context expired / etc.)

pkg/billing/model.go
```diff
@@ -14,6 +42,16 @@ type AbsoluteEvent struct {
...
+// typeSetter implements eventMethods
+func (AbsoluteEvent) typeSetter() func(*AbsoluteEvent) { //nolint:unused // linter is incorrect
+	return func(e *AbsoluteEvent) { e.Type = "absolute" }
+}
+
+// idempotencyKeyGetter implements eventMethods
+func (AbsoluteEvent) idempotencyKeyGetter() func(*AbsoluteEvent) *string { //nolint:unused // linter is incorrect
+	return func(e *AbsoluteEvent) *string { return &e.IdempotencyKey }
+}
+
```
tychoish, line 54: why do these need to be pointers?

pkg/billing/model.go
```diff
@@ -4,6 +4,34 @@ import (
 	"time"
 )
 
+type Event[Self any] interface {
+	AbsoluteEvent | IncrementalEvent
+
+	eventMethods[Self]
```
tychoish, line 10: 🤮 oh god I just realized what you did.

pkg/billing/model.go
tychoish: ... oh no....

pkg/billing/model.go
tychoish: I just figured it out for myself during the meething. I'd split the comment up, and some of the comment here?

pkg/billing/model.go
```diff
@@ -4,6 +4,34 @@ import (
...
+// eventMethods exists because Go does not let you assume that a value of generic type "Foo | Bar"
+// has the fields common to both Foo and Bar. So we still have to thread through these fields.
+//
+// The methods in this interface *could* have directly been a part of Event, but then we couldn't do
+// the "implements eventMethods" assertions below (because you aren't allowed to store a value of
+// type "Foo | Bar"; it can only be used as a constraint for generics).
+//
+// eventMethods also takes Self as a parameter because we want Event[...] to be implemented for the
+// raw AbsoluteEvent and IncrementalEvent types, not pointers to them, because that would make
+// things just a bit more annoying. But we still need to set field values, which we can ONLY do by
```
tychoish, line 22: I think maybe everything should just be a pointer maybe?

pkg/billing/model.go
```diff
@@ -4,6 +4,34 @@ import (
...
+// all that stuff for us.
+type eventMethods[Self any] interface {
+	typeSetter() func(*Self)
+	idempotencyKeyGetter() func(*Self) *string
+}
+
+var (
+	_ eventMethods[AbsoluteEvent]    = AbsoluteEvent{}    //nolint:exhaustruct // just checking they impl the interface
+	_ eventMethods[IncrementalEvent] = IncrementalEvent{} //nolint:exhaustruct // just checking they impl the interface
+)
```
tychoish, line 33: while this is a fine thing to have, doesn't the compiler ensure this because you construct `Event` objects with each/either of these types?

tychoish (approved):

pkg/billing/model.go
sharnoff: We're not actually using `AbsoluteEvent`s unfortunately.

pkg/billing/model.go
sharnoff: :D

pkg/billing/model.go
sharnoff: Added a comment along the lines of "go read the eventMethods docs"

pkg/billing/model.go
sharnoff: Yeah... fixed it so we're using pointers now. Kind of sad, but it's a lot simpler, and ultimately for the best I think.

PR #367, opened 2023-06-30
informant/filecache: Close DB connections

sharnoff:
Realistically a better solution _should_ be to just connect at startup and keep the connection open, but that's a larger change, and this is at least clearly correct.

ref https://neondb.slack.com/archives/C03H1K0PGKH/p1688139786575229

IMO this should be backported + hotfix released as a single-commit change, so that we can get it out sooner.


tychoish (approved):

fprasx (approved):

fprasx:
> Realistically a better solution should be to just connect at startup and keep the connection open,  ...

This is what the monitor does - so hopefully the issue is gone in further releases

PR #368, opened 2023-07-01
agent/billing: Move push logic into separate thread

sharnoff:
This should resolve some of the ongoing issues we've had with billing push requests timing out, because *currently* the push timeout must be short in order to produce correct data.

Also, this PR removes `pkg/billing.Batch`, because this PR changes the agent to use its own batching, meaning pkg/billing.Batch is no longer required.

Billing config changes:

* Renamed `pushTimeoutSeconds` to `pushRequestTimeoutSeconds`
* Added `accumulateEverySeconds` (now distinct from pushing!)
* Added `maxBatchSize`

Billing metrics changes:

* Renamed `autoscaling_agent_billing_batch_size` to `autoscaling_agent_billing_queue_size`
* Added `autoscaling_agent_billing_last_send_duration_seconds`

**NB: We some alerts based on the batch size. They must be silenced during rollout of the first release containing this PR.**

**NB: This PR builds on #366, and must not be merged before it.**


pkg/agent/billing/queue.go
```diff
@@ -0,0 +1,79 @@
...
+		items:     nil,
+		sizeGauge: sizeGauge,
+	}
+	return eventQueuePusher[E]{internals}, eventQueuePuller[E]{internals}
+}
+
+// NB: must hold mu
+func (qi *eventQueueInternals[E]) updateGauge() {
+	qi.sizeGauge.Set(float64(len(qi.items)))
+}
```
tychoish, line 45: this probably doesn't need to be a pointer method. 

also prom gauges are threadsafe, so I think you can make the locking story more hygenic, by passing the size as an argument rather.

pkg/agent/billing/send.go
```diff
@@ -0,0 +1,147 @@
...
+	// update the metric when either (a) the loop is done, or (b) the duration so far is already
+	// longer than the previous one.
+	//
+	// This means that we remove the bias towards shorter durations, at the expense of sometimes
+	// returning higher durations for too long. IMO that's ok, and we'd rather have our metrics give
+	// a pessimistic but more accurate view.
+	lastSendDuration time.Duration
+}
+
+func (s eventSender) senderLoop(logger *zap.Logger) {
```
tychoish, line 45: I would want this to have a context, and be able to handle shutdown in some clean way.

sharnoff: So it _mostly_ does that with the `collectorFinished` signal — the reason it doesn't just take a context is because we could miss some trailing events (e.g. the "collect" thread receives the shutdown notification & packages off the last set of events, but if the "send" thread has already exited, we'll drop some)

pkg/agent/billing/queue.go
sharnoff: > this probably doesn't need to be a pointer method.

The linter gets cranky otherwise because `eventQueueInternals` contains a mutex.

> I think you can make the locking story more hygenic, by passing the size as an argument rather.

I'm not sure that's better - doing it that way, I'd need to pass in `len(qi.items)` both times `updateGauge` is called, which... idk, feels less encapsulated.

sharnoff:
Tested this locally, seems to work ok!

PR #369, opened 2023-07-01
billing: Record push error causes in agent's metrics

sharnoff:
Hopefully should be _relatively_ self-explanatory. Not sure if this is the right direction. I think this is the last billing change I had in mind.

**NB: This PR builds on #368, and must not be merged before it.**
**NB: This PR should be rebased onto main.**


pkg/billing/client.go
```diff
@@ -57,26 +57,58 @@ func Send[E Event](ctx context.Context, client Client, events []E) error {
...
 	// does the retrying, to avoid writing that logic here.
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("got code %d, posting %d events", resp.StatusCode, len(events))
+		return UnexpectedStatusCodeError{StatusCode: resp.StatusCode}
 	}
 
 	return nil
 }
+
+type JSONError struct {
```
tychoish, line 84: I would just have sentinel errors for json and requests and just use error is rather than these extra types. 

There's an `fun/erc.Merge`

sharnoff: > I would just have sentinel errors for json and requests

I'd hate to be in a position where our requests start failing but we can't tell why because the error message provides no useful information.

That said, we don't need as much information about JSON errors in the metrics, which is why this is here:

https://github.com/neondatabase/autoscaling/blob/c3a61b948885c875ed87497b61c9af603a3b9b82/pkg/agent/billing/send.go#L133-L134

tychoish: I don't see how you'd be losing information in this case: 

```go

var ErrJSON = errors.New("problem found with json") 

func thing(in any) error {
     out, err := json.Marshal(in)
     if err != nil {
          return errors.Join(err, ErrJSON)
     }
     // logic
}

// and then:
errors.Is(thing(func() {}, ErrJSON) == true
// with a string form of... 
"invalid marshal type: problem found with JSON" 
```

`errors.Join` is new in 1.20 I think; the `fun/erc` package has a function `erc.Merge` which does the same thing, more or less (different formatting, eh)

sharnoff:
Looking at finishing up this PR, left a reply above. Maybe will need to do some annoying rebase/cherry-pick shenanigans before merging if the original two commits are going to stay separate, but that can be sorted out later.

pkg/billing/client.go
sharnoff: Hm, I think I get the idea. How would you recommend replicating the functionality here?

https://github.com/neondatabase/autoscaling/blob/4eb73cde8d32d7da94878eeb32a0cac9a4999234/pkg/agent/billing/send.go#L144-L145

PR #372, opened 2023-07-03
billing: Add x-trace-id header to requests

sharnoff:
As discussed. Not sure what the header actually should be - maybe x-request-id is easier to work with? (if so, happy to change it).

**NB: This PR builds on #368, and must not be merged before it.**


sharnoff:
e2e test failure is from dockerhub rate limiting. Seems to be ongoing right now

pkg/billing/client.go
```diff
@@ -31,6 +31,12 @@ func (c Client) Hostname() string {
 	return c.hostname
 }
 
+type TraceID string
+
+func (c Client) GenerateTraceID() TraceID {
+	return TraceID(uuid.NewString())
```
Daniel-ef, line 37: Or https://github.com/lithammer/shortuuid. It's faster to select strings without a dash.
But don't insist

Daniel-ef (approved):

pkg/billing/client.go
sharnoff: Done

PR #373, opened 2023-07-03
e2e: Set strategy.fail-fast: false (keep running if other fails)

sharnoff:
We have various sources of flakiness (some avoidable, after fixing bugs, and some unavoidable), and given the success rate of e2e tests, I suspect the default fail-fast behavior is probably using our CI runners less efficiently than if we turned it off.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e-tests:
     strategy:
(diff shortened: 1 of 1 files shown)


bayandin (approved):
Agree!

sharnoff:
hah, already got an example of the desired behavior :D

sharnoff:
Started re-running to get the green checkmarks, but... there's probably no reason to. Merging.

PR #377, opened 2023-07-06
PG14_DISK_TEST_IMG docker build fix

cicdteam:
Fixes the issue when do `make docker-build-examples`

```json
{
  "errorDetail": {
    "message": "pull access denied for vm-informant, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"
  },
  "error": "pull access denied for vm-informant, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,9 +198,8 @@ docker-build-pg14-disk-test: vm-informant bin/vm-builder-generic ## Build a VM i
 	    chmod uga+rw 'vm-examples/pg14-disk-test/ssh_id_rsa' 'vm-examples/pg14-disk-test/ssh_id_rsa.pub'; \
 	fi
 
-	docker buildx build \
+	docker build \
 		--tag tmp-$(PG14_DISK_TEST_IMG) \
-		--load \
 		--file vm-examples/pg14-disk-test/Dockerfile.vmdata \
 		vm-examples/pg14-disk-test/
 	./bin/vm-builder-generic -src tmp-$(PG14_DISK_TEST_IMG) -use-inittab -dst $(PG14_DISK_TEST_IMG)


sharnoff (commented):
Out of curiosity, how does this fix the issue?

cicdteam:
> Out of curiosity, how does this fix the issue?
> 
> Edit: I saw the [message in slack](https://neondb.slack.com/archives/C039YKBRZB4/p1688642724027749?thread_ts=1688584972.832649&cid=C039YKBRZB4), but not sure how this is broken?

The buildx docker-container driver  doesn't support accessing local images, so you need to pull from a registry, whether local or remote. But `vm-informant:dev` used in `FROM` lives locally, not in registry.

sharnoff:
> The buildx docker-container driver doesn't support accessing local images

That's.. interesting. It works for me, I wonder what the difference is?

cicdteam:
@sharnoff I've mentioned wrong make target. 

can you try
```console
docker image prune -a -f
make docker-build-pg14-disk-test
```

cicdteam:
Ah, sorry, of course this PR will not fix issue @fprasx mentioned. In tests just need build `vm-informant:dev` before `vm-builder` call.

sharnoff:
Yeah, the image gets built successfully when I do as you suggested — docker-build-pg14-disk-test also builds the vm-informant.

IIRC @fprasx's issue was separate; not an underlying issue here (I think?)

cicdteam:
@sharnoff seems we have some difference in local docker setup. Because I have

```
 => [internal] load build definition from Dockerfile.vmdata                                                   0.0s
 => => transferring dockerfile: 3.05kB                                                                        0.0s
 => CANCELED [internal] load metadata for docker.io/library/alpine:3.16                                       0.6s
 => ERROR [internal] load metadata for docker.io/library/vm-informant:dev                                     0.6s
 => [auth] library/vm-informant:pull token for registry-1.docker.io                                           0.0s
------
 > [internal] load metadata for docker.io/library/vm-informant:dev:
------
Dockerfile.vmdata:1
--------------------
   1 | >>> FROM vm-informant:dev as informant
   2 |     # ^ don't do anything with this; we just want it around for later use.
   3 |     
--------------------
ERROR: failed to solve: vm-informant:dev: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
make: *** [Makefile:194: docker-build-pg14-disk-test] Error 1
```

sharnoff:
Yeah, I just retried with after `docker system prune -a` and it still works ok. I'm running docker as root (not best practice, I know), version 24.0.2. Maybe some difference there?

Omrigan:
@cicdteam is this still needed?

PR #387, opened 2023-07-10
plugin: Migration handling reliability improvements

sharnoff:
Like all good things, this PR comes in three parts:

1. If the plugin decides to trigger a migration, it no longer returns a non-nil `PluginResponse.Migrate` if the VirtualMachineMigration already existed.
    - This had the potential to cause spurious failures where the autoscaler-agent permanently shuts off communication because _it_ was told that the scheduler is going to migrate it, but actually the migration had already completed.
2. The plugin now automatically cleans up completed migrations.
    - To make this work, all migrations now have the `autoscaling.neon.tech/created-by-scheduler` label.
3. The plugin now exposes metrics about migration creation and deletion. These are:
    - `autoscaling_plugin_migrations_created_total`
    - `autoscaling_plugin_migrations_deleted_total`
    - `autoscaling_plugin_migration_create_fails_total`
    - `autoscaling_plugin_migration_delete_fails_total`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,8 +49,8 @@ linters-settings:
     exclude:
       - '^net/http\.(Client|Server)'
       - '^net\.TCPAddr$'
-      # metav1.{CreateOptions,GetOptions,ListOptions,WatchOptions,PatchOptions}
-      - '^k8s\.io/apimachinery/pkg/apis/meta/v1\.(Create|Get|List|Watch|Patch)Options$'
+      # metav1.{CreateOptions,GetOptions,ListOptions,WatchOptions,PatchOptions,DeleteOptions}
+      - '^k8s\.io/apimachinery/pkg/apis/meta/v1\.(Create|Get|List|Watch|Patch|Delete)Options$'
       - '^k8s\.io/apimachinery/pkg/apis/meta/v1\.ObjectMeta$'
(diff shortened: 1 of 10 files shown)


.golangci.yml
```diff
@@ -49,8 +49,8 @@ linters-settings:
     exclude:
       - '^net/http\.(Client|Server)'
       - '^net\.TCPAddr$'
-      # metav1.{CreateOptions,GetOptions,ListOptions,WatchOptions,PatchOptions}
-      - '^k8s\.io/apimachinery/pkg/apis/meta/v1\.(Create|Get|List|Watch|Patch)Options$'
+      # metav1.{CreateOptions,GetOptions,ListOptions,WatchOptions,PatchOptions,DeleteOptions}
+      - '^k8s\.io/apimachinery/pkg/apis/meta/v1\.(Create|Get|List|Watch|Patch|Delete)Options$'
```
sharnoff, line 53: This is worth being a separate PR; opened #388 to handle that. It should be merged before this one.

sharnoff:
Ok, going to merge this as-is because the code is probably ok _enough_, and we can make changes later if need be. The effect without migrations enabled should be quite limited, and so even though the PR itself is quite large, the potential negative impact is probably small. And it's worth getting this in sooner so we can have the better reliability :)

PR #391, opened 2023-07-10
Fix spurious "updated scaling bounds logs"

### LINKED ISSUE #386: Scheduler has spurious "updated scaling bounds for VM" log lines
Seems to occur whenever there's _any_ change to a VM, rather than just when the scaling bounds changed.

fprasx:
Fixes #386

### DIFF
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -213,7 +213,7 @@ func ExtractVmInfo(logger *zap.Logger, vm *vmapi.VirtualMachine) (*VmInfo, error
 }
 
 func (vm VmInfo) EqualScalingBounds(cmp VmInfo) bool {
-	return vm.Min() != cmp.Min() || vm.Max() != cmp.Max()
+	return vm.Min() == cmp.Min() && vm.Max() == cmp.Max()
 }
 
 func (vm *VmInfo) applyBounds(b ScalingBounds) {


fprasx:
Was there a reason the code was structured as a bunch of if->return instead of if->elif-elif->else? I've changed to that style since the bug seemed to be forgetting to return out of some if blocks, causing the control flow to "fall through" to the bottom where a the `handleUpdatedScalingBounds` callback would get called, causing the logs.

pkg/plugin/watch.go
```diff
@@ -265,35 +265,24 @@ func (e *AutoscaleEnforcer) watchVMEvents(
...
 					podName := util.NamespacedName{Namespace: newInfo.Namespace, Name: newVM.Status.PodName}
 					logger.Info("Received update changing usage for VM", zap.Object("old", oldInfo.Using()), zap.Object("new", newInfo.Using()))
 					callbacks.submitNonAutoscalingVmUsageChanged(logger, newInfo, podName.Name)
+				} else if oldVM.Status.PodName != newVM.Status.PodName {
+					// If the pod changed, then we're going to handle a deletion event for the old pod,
+					// plus creation event for the new pod. Don't worry about it - because all VM
+					// information comes from this watch.Store anyways, there's no possibility of missing
+					// an update.
+				} else if oldInfo.EqualScalingBounds(*newInfo) {
+					// If bounds didn't change, then no need to update
```
sharnoff, line 282: This seems suspect; why doesn't that work in the old version?

sharnoff (commented):
> Was there a reason the code was structured as a bunch of if->return instead of if->elif-elif->else? I've changed to that style since the bug seemed to be forgetting to return out of some if blocks

IIRC (and, looking back over it), the reason why not all of the cases return is because sometimes multiple changes that we care about will have occurred, and we need to handle all of them. Although, I think towards the end of the function things get a little bit... weird. Maybe (at some point, not necessarily this PR) the `if oldVM.Status.PodName != newVm.Status.PodName` should be moved up alongside `if newVM.Status.PodName == ""` for early returns, and get rid of the "if ... return"s elsewhere?

pkg/plugin/watch.go
fprasx: Good point.

fprasx:
Ok - the actual issue was that `EqualScalingBounds` was returning whether the bounds are *not* equal, rather than whether they are equal.

sharnoff (approved):

PR #393, opened 2023-07-10
Collect memory metrics and scale based on them

fprasx:


sharnoff (commented):
Broadly looks good. Memory _is_ more volatile than CPU (so, _maybe_ we want `memoryUsageFractionTarget` to be smaller for now? see also: [this thread](https://neondb.slack.com/archives/C03TN5G758R/p1688937736930849?thread_ts=1688931082.675569&cid=C03TN5G758R))

Some other things to note:

When you measure the _total_ memory within the VM, there's ~100 MiB that the kernel uses that isn't included — so if we create a 0.25 CPU / 1 GiB VM, it'll basically never get to 90% memory usage.

The way the vm-informant _currently_ handles this is by acknowledging that there's some amount of memory reserved for the kernel, and compensates with that for any communications with the autoscaler-agent about total resources.

fprasx:
> The way the vm-informant currently handles this is by acknowledging that there's some amount of memory reserved for the kernel, and compensates with that for any communications with the autoscaler-agent about total resources.

Got it. To clarify: we still need to account for that here right?

Also, is there any way of deducing a good value for `MemoryUsageFractionTarget`?

sharnoff:
> Got it. To clarify: we still need to account for that here right?

Yup (although it matters less once the VM is sufficiently large)

> Also, is there any way of deducing a good value for `MemoryUsageFractionTarget`?

uh, not really. Just take a guess (look at some numbers for potential decisions) and if users complain we raise/lower it, probably

pkg/api/metrics.go
```diff
@@ -61,5 +62,17 @@ func ReadMetrics(nodeExporterOutput []byte, loadPrefix string) (m Metrics, err e
...
+	if err != nil {
+		return
+	}
+	totalMem, err := getField(loadPrefix+"memory_total_bytes", "")
+	if err != nil {
+		return
+	}
+
+	// Subract off an extra 100 MiB to account for the kernel
+	m.MemoryUsageBytes = totalMem - availableMem - 100*(1<<20)
```
sharnoff, line 75: The accounting here (i.e. `totalMem - availableMem`) is going to be accurate without the extra 100 MiB. The place where the kernel comes in is that `totalMem` is about 100 MiB less than it should be.

So `m.MemoryUsageBytes` should be `totalMem - availableMem + 100*(1<<20)`, afaict?

Alternatively, just using `availableMem` might be cleaner — I'm not super sure either way.

fprasx: Ahh I see - then I agree: it should be `+100*(1<<20)`

sharnoff (approved):
Looks good!

Issue #384, opened 2023-07-08
Scheduler plugin should track usage from pods that aren't scheduled through it


sharnoff:
There's a couple things here:

1. The plugin should update the node state when a pod that's _not_ scheduled through it is bound
    1. **NB:** This check should be implemented as "do we already know about this pod", not "does it have the right `schedulerName`"
2. The `System` resource allocation can be removed, because we'll now be tracking those pods
3. The plugin should _also_ handle changes to `resources.requests` for non-VM pods. This isn't _yet_ possible to occur because we're on K8s 1.25, but it's maybe worth implementing here as well, for completeness.
4. The plugin should entirely ignore the paused pods we use for overprovisioning when:
    1. Deciding whether to migrate a VM
    2. Reporting node resource usage

This is mostly relevant for the per-node resource metrics added by #363.

See also:

* #383

PR #397, opened 2023-07-12
plugin: Track nodes even when nothing's scheduled

### LINKED ISSUE #383: Scheduler plugin should track node info even if nothing's scheduled on it
A current flaw with the per-node resource metrics added by #363 is that we only have metrics for nodes we're tracking, which means that we don't include metrics for nodes where nothing's been scheduled through the `autsocale-scheduler`.

We'd like to also include these nodes in the metrics. So: We should track these nodes in order to provide metrics, even when the _scheduler_ doesn't get asked about anything on them.

See also:
* #384

sharnoff:
Resolves #383. Should improve our metrics *somewhat*. Doesn't make sense without #384.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -127,6 +127,12 @@ func makeAutoscaleEnforcerPlugin(
 
 	hlogger := logger.Named("handlers")
 	nwc := nodeWatchCallbacks{
+		submitNodeAddition: func(logger *zap.Logger, node *corev1.Node) {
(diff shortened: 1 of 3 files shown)


sharnoff:
Thinking about this some more, it's probably net-negative. Tracking all pods is already accomplished by #399, so the only benefit we'd get is by _also_ tracking nodes that don't (yet) have any pods on them. But those upsides are probably not worth the added complication.

PR #403, opened 2023-07-13
agent/informant: make CondChannelSignal generic

fprasx:
Pre-req for #362.

### DIFF
--- a/pkg/util/signal.go
+++ b/pkg/util/signal.go
@@ -7,32 +7,32 @@ import (
 	"sync"
 )
 
-func NewSingleSignalPair() (SignalSender, SignalReceiver) {
-	sigCh := make(chan struct{})
+func NewSingleSignalPair[T any]() (SignalSender[T], SignalReceiver[T]) {
+	sigCh := make(chan T)
 	once := &sync.Once{}
 	closeSigCh := func() { once.Do(func() { close(sigCh) }) }
 
-	return SignalSender{send: closeSigCh}, SignalReceiver{sigCh: sigCh, closeSigCh: closeSigCh}
+	return SignalSender[T]{send: closeSigCh}, SignalReceiver[T]{sigCh: sigCh, closeSigCh: closeSigCh}
 }
 
-type SignalSender struct {
+type SignalSender[T any] struct {
 	send func()
 }
 
-type SignalReceiver struct {
-	sigCh      chan struct{}
+type SignalReceiver[T any] struct {
+	sigCh      chan T
 	closeSigCh func()
 }
 
-func (s SignalSender) Send() {
+func (s SignalSender[T]) Send() {
 	s.send()
 }
 
-func (s SignalReceiver) Recv() chan struct{} {
+func (s SignalReceiver[T]) Recv() chan T {
 	return s.sigCh
 }
 
-func (s SignalReceiver) Close() {
+func (s SignalReceiver[T]) Close() {
 	s.closeSigCh()
 }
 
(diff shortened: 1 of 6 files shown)


pkg/util/signal.go
```diff
@@ -7,32 +7,37 @@ import (
 	"sync"
 )
 
-func NewSingleSignalPair() (SignalSender, SignalReceiver) {
-	sigCh := make(chan struct{})
+func NewSingleSignalPair[T any]() (SignalSender[T], SignalReceiver[T]) {
+	sigCh := make(chan T)
```
sharnoff, line 11:
```suggestion
	sigCh := make(chan T, 1)
```

If this has a buffer size of 0, then the send will block until the message is received.

sharnoff:
> agent/informant: make CondChannelSignal generic

this is primarily a change in pkg/util, right?

fprasx:
Yeah. Meant that the affected parts of the system are in the agent/informant.

pkg/util/signal.go
fprasx: Ahh right. Definitely don't want to block.

pkg/util/signal.go
```diff
@@ -7,32 +7,37 @@ import (
...
 	once := &sync.Once{}
 	closeSigCh := func() { once.Do(func() { close(sigCh) }) }
 
-	return SignalSender{send: closeSigCh}, SignalReceiver{sigCh: sigCh, closeSigCh: closeSigCh}
+	return SignalSender[T]{
+		send: func(data T) {
+			sigCh <- data
+			closeSigCh()
+		},
```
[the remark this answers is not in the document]
fprasx: Ok, I think it should be fixed now.

pkg/util/signal.go
```diff
@@ -7,32 +7,38 @@ import (
...
+	sigCh := make(chan T, 1)
 	once := &sync.Once{}
 	closeSigCh := func() { once.Do(func() { close(sigCh) }) }
-
-	return SignalSender{send: closeSigCh}, SignalReceiver{sigCh: sigCh, closeSigCh: closeSigCh}
+	sendOnce := &sync.Once{}
+
+	return SignalSender[T]{
+		send: func(data T) {
+			sendOnce.Do(func() { sigCh <- data })
```
sharnoff, line 18: Having a separate `sendOnce` I don't think fully solves the panic issue, e.g.:

```go
tx, rx := NewSingleSignalPair()
rx.Close()
tx.Send(v)
```

will result in sending on the channel after it's already closed, right?

Consider having a single `sync.Once`, and using it for both the "close" and "send+close" operations?

fprasx: Ahh! The same once could potentially execute different things, I see.

pkg/util/signal.go
```diff
@@ -7,32 +7,39 @@ import (
...
+	sigCh := make(chan T, 1)
 	once := &sync.Once{}
 	closeSigCh := func() { once.Do(func() { close(sigCh) }) }
 
-	return SignalSender{send: closeSigCh}, SignalReceiver{sigCh: sigCh, closeSigCh: closeSigCh}
+	return SignalSender[T]{
+		send: func(data T) {
+			once.Do(func() {
+				sigCh <- data
+				closeSigCh()
```
sharnoff, line 19: does calling `closeSigCh` here work? I'd assume it wouldn't actually close the channel, because the `Once` is already running the function, but I don't know how reentrancy is handled

fprasx: You're probably correct. That means we need to just `close(sigCh)` instead of calling `closeSigCh`, and that's safe because both `send` and `closeSigCh` are wrapped in the same once - so we can only have one closure (pun intended), and if we are sending, we know the channel isn't closed.

sharnoff (approved):

PR #408, opened 2023-07-14
plugin: Emit k8s Event on failed ExtractVmInfo

sharnoff:
This should make observability for invalid autoscaling annotations / labels much better; previously, this information was only really visible in the logs.

Tested this locally by modifying `vm-deploy.yaml` and it seems to work as expected.

cc @cicdteam for any advice on events here.

### DIFF
--- a/pkg/plugin/watch.go
+++ b/pkg/plugin/watch.go
@@ -82,10 +82,10 @@ func (e *AutoscaleEnforcer) watchPodEvents(
 	parentLogger *zap.Logger,
 	metrics watch.Metrics,
 	callbacks podWatchCallbacks,
-) error {
+) (*watch.Store[corev1.Pod], error) {
 	logger := parentLogger.Named("pod-watch")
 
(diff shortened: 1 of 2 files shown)


cicdteam (approved):

pkg/plugin/watch.go
```diff
@@ -298,14 +298,34 @@ func (e *AutoscaleEnforcer) watchVMEvents(
...
+					logger.Error("Failed to extract VM info in update for new VM", util.VMNameFields(newVM), zap.Error(err))
+					e.handle.EventRecorder().Eventf(
+						newVM,            // regarding
+						runnerPod,        // related
+						"Warning",        // eventtype
+						"ExtractVmInfo",  // reason
+						"HandleVmUpdate", // action
+						"Failed to extract autoscaling info about VM: %s", // note
+						err,
+					)
```
sharnoff, line 323: Updated this section to include the runner pod if we can get it

PR #417, opened 2023-07-17
fix go list errors when importing cilium package to the go project

liuyuanzhi-cn:
errors desc: https://github.com/cilium/cilium/issues/24348


sharnoff (commented):
Thanks for the PR! Please run `go mod tidy` (running `go list ...` modifies `go.sum`) and then we can merge it :)

PR #421, opened 2023-07-18
agent: Fix LA-based scaling being 4x too small

sharnoff:
When we switched compute units from (1 CPU, 4 GiB) to (1/4 CPU, 1 GiB), the previous behavior of assuming that 1 CPU = 1 CU became a regression, causing us to significantly underestimate the actual amount of CPU needed to match a given load average.

Noticed this while looking at this VM: https://neonprod.grafana.net/goto/_RIiDPC4R?orgId=1
(using memory as a proxy for CU × 4, because reading "desired CPUs" is somewhat non-trivial)


fprasx (approved):
`CS != math`

PR #429, opened 2023-07-20
plugin: Standardize+improve plugin method logging

sharnoff:
Some ongoing issues on staging due to inconsistencies between the Filter and Reserve steps are proving hard to debug because there's just a bit too little information available. For example, the Filter step shows the node totals, whereas a failed Reserve step shows the incremental change.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -869,11 +869,11 @@ func (e *AutoscaleEnforcer) Reserve(
 			}
 
 			cpuVerdict := fmt.Sprintf(
-				"need %v CPU (%v -> %v raw), %v of %v used, so %v available (%s)",
+				"need %v (%v -> %v raw), %v of %v used, so %v available (%s)",
 				addCpu, &oldNodeRes.RawCPU, &newNodeRes.RawCPU, node.vCPU.Reserved, node.totalReservableCPU(), node.remainingReservableCPU(), cpuShortVerdict,
 			)
 			memVerdict := fmt.Sprintf(
-				"need %v CPU (%v -> %v raw), %v of %v used, so %v available (%s)",
+				"need %v (%v -> %v raw), %v of %v used, so %v available (%s)",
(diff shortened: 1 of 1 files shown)


pkg/plugin/plugin.go
```diff
@@ -473,16 +473,29 @@ func (e *AutoscaleEnforcer) Filter(
 	otherResources.MarginCPU = node.otherResources.MarginCPU
 	otherResources.MarginMemory = node.otherResources.MarginMemory
 
+	// As we process all pods, we should record all the pods that either aren't
```
fprasx, line 476: Note sure what this comment is trying to say. That we should record all pods that "either aren't in `e.state.podMap` or `e.state.otherPods`"

sharnoff: ah, good catch! finished that comment

pkg/plugin/plugin.go
```diff
@@ -833,12 +869,12 @@ func (e *AutoscaleEnforcer) Reserve(
...
 			cpuVerdict := fmt.Sprintf(
-				"need %v vCPU (%v -> %v raw), have %v available (%s)",
-				addCpu, &oldNodeRes.RawCPU, &newNodeRes.RawCPU, node.remainingReservableCPU(), cpuShortVerdict,
+				"need %v CPU (%v -> %v raw), %v of %v used, so %v available (%s)",
+				addCpu, &oldNodeRes.RawCPU, &newNodeRes.RawCPU, node.vCPU.Reserved, node.totalReservableCPU(), node.remainingReservableCPU(), cpuShortVerdict,
 			)
 			memVerdict := fmt.Sprintf(
-				"need %v mem slots (%v -> %v raw), have %d available (%s)",
-				addMem, &oldNodeRes.RawMemory, &newNodeRes.RawMemory, node.remainingReservableMemSlots(), memShortVerdict,
+				"need %v CPU (%v -> %v raw), %v of %v used, so %v available (%s)",
```
fprasx, line 876:
```suggestion
				"need %v Mem (%v -> %v raw), %v of %v used, so %v available (%s)",
```

sharnoff: Realized they shouldn't say the units in the first place, because the `verdictSet` already has them

PR #431, opened 2023-07-20
bump dependency.

tychoish:
I'm super interested in seeing how this runs through the tests. can revert the change to the part of the code that had the bug today.


tychoish:
same as #508

PR #435, opened 2023-07-24
plugin: Handle bypassed Reserve due to k8s bug

sharnoff:
ref https://neondb.slack.com/archives/C03F5SM1N02/p1690131944251759?thread_ts=1690130256.724109&cid=C03F5SM1N02

Basically, we noticed a bug(?) in the core k8s scheduler where a pod went through Reserve, Bind failed, got Unreserve'd, and then re-bound to the node without consulting Reserve again. While the core scheduler shouldn't do that to the plugin, fixing that is a lot more effort than just gracefully handling it ourselves.

**Tasks**

- [x] Handle for non-VM pods
- [ ] Handle for VM pods

### DIFF
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -732,11 +732,6 @@ func (e *AutoscaleEnforcer) handlePodStarted(logger *zap.Logger, pod *corev1.Pod
 		zap.String("node", nodeName),
 	)
 
-	if pod.Spec.SchedulerName == e.state.conf.SchedulerName {
-		logger.Info("Got non-VM pod start event for pod assigned to this scheduler; nothing to do")
-		return
-	}
-
 	logger.Info("Handling non-VM pod start event")
 
 	podResources, err := extractPodOtherPodResourceState(pod)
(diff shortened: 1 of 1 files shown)


sharnoff:
Happened again: https://neondb.slack.com/archives/C03TN5G758R/p1701273358276909?thread_ts=1701272545.401659&cid=C03TN5G758R

shayanh:
> Basically, we noticed a bug(?) in the core k8s scheduler where a pod went through Reserve, Bind failed, got Unreserve'd, and then re-bound to the node without consulting Reserve again. While the core scheduler shouldn't do that to the plugin, fixing that is a lot more effort than just gracefully handling it ourselves.

I tested this in a test environment by customizing the `Bind` plugin such that it fails randomly. I couldn't reproduce the mentioned bug. The `Reserve` and `Unreserve` methods were called as expected. 

Also, I looked at the [this set of logs](https://neondb.slack.com/archives/C03TN5G758R/p1701273358276909?thread_ts=1701272545.401659&cid=C03TN5G758R) and it's not clear to me that the `Reserve` and `Unreserve` methods haven't been called after the `Bind` failure.

shayanh:
An update on this: this wasn't due to a k8s bug. This is what happened:

1. Scheduler: received the pod
2. Scheduler: called Reserve
3. Scheduler: called Bind
4. API Server: bound pod to the node (etcd updated)
5. Scheduler: didn't receive the reply from API server because network 
6. Scheduler: Bind timeout
7. Scheduler: called Unreserve

Since the pod is bound to its node on etcd, it didn't go through the scheduler anymore although scheduler thinks the pod wasn't scheduled.

sharnoff:
Closing because #666 (now merged) fixed this issue.

shayanh:
I don't clearly see how #666 resolves this issue. Can you elaborate on this?

sharnoff:
Sure. IIRC before, we only handled starts for non-VM pods that weren't _supposed_ to go through our scheduler.

Now with #666, we handle starts for _all_ pods (VM or non-VM), regardless whether they were supposed to go through the scheduler, and just do nothing if we already know about the pod.

PR #441, opened 2023-07-26
neonvm/runner: Create QEMU cgroup inside current one

### LINKED ISSUE #375: Wrong cAdvisor metrics
NeonVM uses cgroups for fractional CPU scaling. And kubelet uses cgroups to manage pod resources too. As a result, something went wrong with the resource metrics for the pods where the VMs are running.

I started a VM (compute via console API) with autoscale settings `minCU=1, maxCU=7` and then put in a workload that uses a lot of CPU. Compute  was automatically scaled up to 7 CUs within 1-2 minutes. But `kubectl top` shows that the pod consumes only 1 milliCPU (I checked inside the virtual machine - the processor was almost 100% loaded).

VM right after scale up
```console
% k get neonvm compute-quiet-scene-688043
NAME                         CPUS   MEMORY   POD                                EXTRAIP   STATUS    AGE
compute-quiet-scene-688043   7      28Gi     compute-quiet-scene-688043-7qmmd             Running   2m8s
```

pod resource usage
```
% k top po compute-quiet-scene-688043-7qmmd
NAME                               CPU(cores)   MEMORY(bytes)   
compute-quiet-scene-688043-7qmmd   1m           8417Mi          
```

### LINKED ISSUE #440: neonvm-runner doesn't clean up cgroup v1 cpu controller entry on exit
e.g. on staging, `kubectl exec -it <VM pod> -- /bin/sh` and then compare `ls /sys/fs/cgroup/systemd` (current cgroups) with `ls /sys/fs/cgroup/cpu` (cgroups still tied to the CPU controller).

On a sample node, the first has 21 entries, while the second has 129. Presumably there's just some cleanup missing in the runner.

sharnoff:
This should improve a handful of things. It should fix the cAdvisor metrics issue, because now QEMU will actually be running as part of the container's cgroup (so its CPU/memory usage should be reported correctly), AND it should help with cgroup cleanup, because containerd will take care of that too! :D

Fixes #375, fixes #440.

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -583,7 +583,6 @@ func main() {
 		qemuCmd = append(qemuCmd, "-incoming", fmt.Sprintf("tcp:0:%d", vmv1.MigrationPort))
 	}
 
-	// leading slash is important
 	selfCgroupPath, err := getSelfCgroupPath()
 	if err != nil {
 		log.Fatalf("Failed to get self cgroup path: %s", err)
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -579,12 +579,17 @@ func main() {
 	}
 
 	// leading slash is important
-	cgroupPath := fmt.Sprintf("/%s-vm-runner", vmStatus.PodName)
+	selfCgroupPath, err := getSelfCgroupPath()
+	if err != nil {
+		log.Fatalf("Failed to get self cgroup path: %s", err)
+	}
+	cgroupPath := fmt.Sprintf("%s/neonvm-qemu", selfCgroupPath)
+
+	log.Printf("Using QEMU cgroup path %q", cgroupPath)
 
 	if err := setCgroupLimit(qemuCPUs.use, cgroupPath); err != nil {
 		log.Fatalf("Failed to set cgroup limit: %s", err)
 	}
-	defer cleanupCgroup(cgroupPath)
 	ctx, cancel := context.WithCancel(context.Background())
 	wg := sync.WaitGroup{}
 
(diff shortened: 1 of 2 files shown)


neonvm/runner/main.go
```diff
@@ -579,12 +584,27 @@ func main() {
 	}
 
 	// leading slash is important
```
nikitakalyanov, line 586: the comment is now outdated

neonvm/runner/main.go
```diff
@@ -579,12 +584,27 @@ func main() {
...
+	// that the cgroup '/neonvm-qemu' still works.
+	if selfCgroupPath == "/" {
+		selfCgroupPath = ""
+	}
+	// ... but also we should have some uniqueness just in case, so we're not sharing a root level
+	// cgroup if that *is* what's happening. This *should* only be relevant for local clusters.
+	//
+	// We don't want to just use the VM spec's .status.PodName because during migrations that will
+	// be equal to the source pod, not this one, which may be... somewhat confusing.
+	cgroupPath := fmt.Sprintf("%s/neonvm-qemu-%s", selfCgroupPath, selfPodName)
```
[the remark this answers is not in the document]
sharnoff: > our group could not use more resources than the parent cgroup. Is this correct?

That's correct. The limits from the parent cgroup will come from k8s container `resources.limits`, if there are any — so because VM pods aren't created with limits we should be good

neonvm/runner/main.go
sharnoff: oops, thanks. Will remove

cicdteam (approved):

sharnoff:
Merging this so it can get included in a release (and so we can validate it on staging, etc).

Issue #420, opened 2023-07-17
Epic: Find the way to connect to static port from VM-informant


vadim2404:
It's needed to create a network policy to limit the network visibility scope for VMs

fprasx:
Plan for week of 31JUL2023:
* Merge @hlinnaka  PR (#447) for an immediate fix
* After @fprasx is finished refactoring the monitor, prepare (#442) for merge. This will require restarting all VMs but is a longer term fix.

PR #442, opened 2023-07-26
agent: establish connection with monitor instead of informant

fprasx:
Addresses #420.

Since this PR is based off another PR branch, #362 (branch-ception?), that PR should get merged first, then this once should get merged after merging/rebasing any changes.

The object of this PR is to allow an agent to talk to _either_ an informant or monitor. This addresses #420 since the monitor has a fixed port. The way we achieve this is by making the monitor have some of the informant's API, namely the `/register` endpoint.

When an agent connects to the monitor's `/register` endpoint, they will receive a 404, signalling that they should connect via websocket to the `/monitor` endpoint.

From there, most of the changes for handling the `agent/informant-monitor` protocol are very similar to #362.


sharnoff (commented):
Some things that are missing:

- Update README.md / ARCHITECTURE.md to remove references to vm-informant
- Update the release workflow

Haven't finished reviewing, will try to get another pass done tomorrow.

pkg/agent/dispatcher.go
```diff
@@ -156,20 +147,16 @@ func extractField[T any](data map[string]interface{}, key string) (*T, error) {
 	return &coerced, nil
 }
 
-type messageHandlerFuncs struct {
-	handleUpscaleRequest      func(api.UpscaleRequest)
-	handleUpscaleConfirmation func(api.UpscaleConfirmation, uint64) error
-	handleDownscaleResult     func(api.DownscaleResult, uint64) error
-	handleMonitorError        func(api.InternalError, uint64) error
-	handleHealthCheck         func(api.HealthCheck, uint64) error
-}
```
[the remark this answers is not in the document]
fprasx: Yeah. Cherry pick is so cool!

Makefile
```diff
@@ -131,13 +130,12 @@ bin/vm-builder-generic: ## Build vm-builder-generic binary.
...
-vm-informant: ## Build vm-informant image
+.PHONY: vm-monitor
+vm-monitor: ## Build vm-monitor image
 	docker buildx build \
-		--tag $(VM_INFORMANT_IMG) \
+		--tag $(VM_MONITOR_IMG) \
 		--load \
-		--build-arg GIT_INFO=$(GIT_INFO) \
-		--file build/vm-informant/Dockerfile \
+		--file build/vm-monitor/Dockerfile \
```
sharnoff, line 138: Duplicated `vm-monitor` target? (there's an existing one right below this)

pkg/agent/dispatcher.go
```diff
@@ -46,18 +45,10 @@ type Dispatcher struct {
...
-	// just tries to receive off the channel and then request the upscale.
-	// A different way to do this would be to keep a backpointer to the parent
-	// `State` and then just call a method on it when an upscale is requested.
-	// However, we do not do this because it causes a reference cycle, which
-	// interferes with go's ability to run finalizers - which nhooyr/websocket
-	// uses to manage connections. Thus, if a panic were to happen, the connection
-	// might not get dropped.
-	notifier util.CondChannelSender
-
-	// nextTransactionID represents the current transaction id. When we need a new one
```
sharnoff, line 60: Unnecessary comment change on this line? ("nextTransactionID" -> "This counter")

pkg/agent/informant.go
```diff
@@ -1091,3 +1175,93 @@ func (s *InformantServer) Upscale(ctx context.Context, logger *zap.Logger, to ap
...
+	select {
+	case res := <-rx.Recv():
+		// A nil pointer means a monitor error occured
+		if res != nil {
+			return nil
+		} else {
+			return fmt.Errorf("monitor experienced an internal error")
+		}
+	case <-time.After(timeout):
+		return fmt.Errorf("timed out waiting %v for monitor response", timeout)
```
sharnoff, line 1201: `time.After` spawns a goroutine that isn't cleaned up until the time expires (IIRC). For this, it'd be better to use something like `timer := time.NewTimer(...); defer timer.Stop()`

pkg/agent/runner.go
```diff
@@ -1661,7 +1666,13 @@ func (r *Runner) doInformantDownscale(ctx context.Context, logger *zap.Logger, t
 		return nil, fmt.Errorf("%s: InformantServer is not set (this should not occur after startup)", msg)
 	}
 
-	resp, err := server.Downscale(ctx, logger, to)
+	// TODO: is it racy to read informantIsMonitor
```
sharnoff, line 1669: maybe

The read must happen while either holding r.lock or r.server.requestLock. `(*InformantServer).Downscale` acquires its request lock, so holding that while making requests here wouldn't work, unfortunately.

pkg/agent/informant.go
```diff
@@ -1091,3 +1175,93 @@ func (s *InformantServer) Upscale(ctx context.Context, logger *zap.Logger, to ap
 	logger.Info("Received successful upscale result")
 	return nil
 }
+
+func (s *InformantServer) MonitorHealthCheck(ctx context.Context, logger *zap.Logger) error {
```
sharnoff, line 1179: this function (and `MonitorUpscale` / `MonitorDownscale` should have comments explaining their usage).

Also: consider adding requestLock usage here as well, plus facilities to restart the connection (via `s.exit`) if the request fails.

pkg/agent/informant.go
```diff
@@ -1091,3 +1175,93 @@ func (s *InformantServer) Upscale(ctx context.Context, logger *zap.Logger, to ap
...
+
+	// Wait for result
+	timeout := time.Second * time.Duration(s.runner.global.config.Monitor.ResponseTimeoutSeconds)
+	select {
+	case res := <-rx.Recv():
+		// A nil pointer means a monitor error occured
+		if res != nil {
+			return nil
+		} else {
+			return fmt.Errorf("monitor experienced an internal error")
```
sharnoff, line 1198: For errors where there's no formatting required, it's preferable to use `errors.New`

pkg/agent/informant.go
```diff
@@ -413,13 +436,70 @@ func (s *InformantServer) RegisterWithInformant(ctx context.Context, logger *zap
...
+				s.runner.global.config.Informant.ServerPort,
+			)
+			logger.Info(
+				"received 404 from informant's /register endpoint; connecting to monitor",
+				zap.String("addr", addr),
+			)
+			// pre-declare disp so that err get's assigned to err from enclosing scope,
+			// overwriting original request error.
+			var disp Dispatcher
+			disp, err = NewDispatcher(logger, addr, s)
```
sharnoff, line 462: Does `NewDispatcher` make a network request?

If so, we shouldn't block `s.runner.lock` while holding it

pkg/agent/dispatcher.go
```diff
@@ -251,40 +242,9 @@ func (disp *Dispatcher) HandleMessage(
...
-				logger.Warn("monitor experienced an internal error")
-			}
-		case <-time.After(MonitorResponseTimeout):
-			logger.Warn(
-				"timed out waiting for monitor response",
-				zap.Duration("timeout", MonitorResponseTimeout),
-			)
-		}
-	}
-}
```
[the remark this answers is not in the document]
fprasx: Correct!

pkg/agent/dispatcher.go
fprasx: I think that was a rebase/merge error on my part.

pkg/agent/informant.go
fprasx: What would the advantages of using requestLock here be? All the types involved (`Dispatcher` and `CondChannelSignal`) are thread-safe.

sharnoff: The general idea behind having requestLock was to avoid making concurrent requests to the informant (because e.g. concurrent "upscale" and "downscale" could have unpredictable consequences). In practice, I think we _should_ be fine without it, but we lose some nice guarantees - in particular, if we e.g. call `server.exit(...)` while holding requestLock, we know that no further requests will be made.

This was more relevant for the informant, where the /register request was handled more like other requests, so we _did_ actually have to prevent concurrent register attempts. But it may be useful to keep here as well.

pkg/agent/dispatcher.go
```diff
@@ -46,18 +45,10 @@ type Dispatcher struct {
...
-	// interferes with go's ability to run finalizers - which nhooyr/websocket
-	// uses to manage connections. Thus, if a panic were to happen, the connection
-	// might not get dropped.
-	notifier util.CondChannelSender
-
-	// nextTransactionID represents the current transaction id. When we need a new one
+	// The InformantServer that this dispatcher is part of
+	server *InformantServer
+
+	// nextTransactionID  the current transaction id. When we need a new one
```
sharnoff, line 51: > nextTransactionID the current transaction id.

missing word?

pkg/agent/informant.go
```diff
@@ -236,6 +246,15 @@ func NewInformantServer(
 			case <-ticker.C:
 			}
 
+			// Are we talking to a monitor?
+			if server.informantIsMonitor {
```
sharnoff, line 250: is it possible for `server.informantIsMonitor` to change after this point? If so, reading the field should be done with a lock acquired.

pkg/agent/informant.go
```diff
@@ -1059,3 +1143,169 @@ func (s *InformantServer) Upscale(ctx context.Context, logger *zap.Logger, to ap
...
+		} else {
+			err := errors.New("monitor experienced an internal error")
+			exit(err)
+			return err
+		}
+	case <-timer.C:
+		err := fmt.Errorf("timed out waiting %v for monitor response", timeout)
+		exit(err)
+		return err
+	}
```
sharnoff, line 1196: A thought: it seems like this control flow here is pretty common. Maybe either (a) bundle that into `Dispatcher` (so it handles the timeouts for you), or make your own wrapper around this?

(the only parts that change seem to be the argument to `s.dispatcher.Call` and the logic handling the result)

pkg/agent/runner.go
```diff
@@ -1660,7 +1665,12 @@ func (r *Runner) doInformantDownscale(ctx context.Context, logger *zap.Logger, t
 		return nil, fmt.Errorf("%s: InformantServer is not set (this should not occur after startup)", msg)
 	}
 
-	resp, err := server.Downscale(ctx, logger, to)
+	downscale := server.Downscale
+	if server.informantIsMonitor {
```
sharnoff, line 1669: Same question here about "can this change because we're not holding the lock?" (same for upscale as well)

sharnoff (approved):
Looks good! +validated that this continues to work with old computes

PR #454, opened 2023-08-02
Bump VM kernel to 6.4.7

sharnoff:
The goal here is that *maybe* the kernel panic issue has already been fixed, and so we won't need to do the difficult work of fixing it ourselves. I couldn't reproduce it locally, so this is going to live on staging for testing with known reproduction cases.

Supporting both v5 and v6 kernel versions required separating the major version out so we can construct the download URL correctly (maybe we don't need to? oh well).

For reviewing, may be useful to diff `linux-config-5.15.80` and `linux-config-6.4.7`.


cicdteam:
I think it's Ok. Just some points...

1. How i calculate major kernel version (to avoid specify it in separate variable)
```bash
MAJOR=`echo ${KERNEL_VERSION} | sed -E 's/^([0-9]+)\.[0-9]+\.[0-9]+$/\1/'`
```
2. Some info about kernel sizes with your config and with  generated config by "all-no" option.
```bash
generated config (use 5.15.80 as base):

$ KCONFIG_ALLCONFIG=../linux-config-5.15.80 make allnoconfig
$ make -j `nproc`
$ ls -lah $(make image_name)
-rw-rw-r-- 1 andrey andrey 6.3M Aug  2 16:05 arch/x86/boot/bzImage


config from this PR:

$ make mrproper
$ cp ../linux-config-6.4.7 .config
$ make -j `nproc`
$ ls -lah $(make image_name)
-rw-rw-r-- 1 andrey andrey 6.9M Aug  2 16:13 arch/x86/boot/bzImage
```

cicdteam (approved):

sharnoff:
Ah, I'll retry with `make allnoconfig` instead. Maybe also why the compile time was longer.

cicdteam:
@sharnoff are we still need this PR ? fyi: I have working config for 6.6.2 version (latest stable)

sharnoff:
Yeah, no longer needed - thanks

PR #455, opened 2023-08-02
informant: trigger informant shutdown on server error

tychoish:


sharnoff:
The informant has since been removed

PR #459, opened 2023-08-07
agent: Track runner state metrics as it goes

### LINKED ISSUE #198: agent "number of X" metrics should be tracked globally
The autoscaler-agent exposes a number of metrics that are pulled from each `Runner` (or `podStatus`) individually. For some of these, we can track the value globally, which would be more efficient.

See also: https://github.com/neondatabase/autoscaling/pull/180#discussion_r1167626715

sharnoff:
... instead of the previous method of calculating them as needed at scrape time.

There's some ongoing issues with metrics collection on staging, maybe this will help with it.

Resolves #198.

~~Haven't tested it or anything, still need to self-review.~~

## This PR makes breaking changes to metrics

All the following metrics are removed:

- `autoscaling_errored_vm_runners_current`
- `autoscaling_panicked_vm_runners_current`
- `autoscaling_agent_tracked_vms_current`
- `autoscaling_vms_unsuccessful_communication_with_informant_current`
- `autoscaling_billed_vms_unsuccessful_communication_with_informant_current`

... replaced by:

- `autoscaling_agent_runners_current`

with the previous metrics replaced with combinations of labels on the new one.

### DIFF
--- a/pkg/agent/globalstate.go
+++ b/pkg/agent/globalstate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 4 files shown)


sharnoff:
Self-reviewed and tested locally, should be ok.

Issue #218, opened 2023-04-24
Switch to klog in neonvm/runner


sharnoff:
Low priority. We're using klog in other components, so _may_ be worthwhile.

ref https://github.com/neondatabase/autoscaling/pull/172#discussion_r1175628635

sharnoff:
No longer desired (we switched from klog to zap) and neonvm/runner has since been switched to zap with #460.

PR #460, opened 2023-08-08
neonvm/runner: Switch to zap for logging

sharnoff:
I've seen some people get confused about where the runner's logs are coming from (e.g. "setting cgroup to ..."), and there was previously a plan to standardize the logging repo-wide (see e.g. #218) before we switched various components to zap. So: I figure it's worthwhile to switch the runner to zap as well.

While going through this, there were some places with "log error with context; return error" that I think _should_ be fixed, but I'll leave that to a follow-up PR, because I wanted this to _mostly_ match the original, just with JSON logging and more than two levels :)


neonvm/runner/main.go
```diff
@@ -492,12 +494,12 @@ func main() {
...
 		if vmSpec.Guest.RootDisk.Size.Cmp(*imageSizeQuantity) == 1 {
-			log.Printf("resizing rootDisk from %s to %s", imageSizeQuantity.String(), vmSpec.Guest.RootDisk.Size.String())
+			logger.Info(fmt.Sprintf("resizing rootDisk from %s to %s", imageSizeQuantity.String(), vmSpec.Guest.RootDisk.Size.String()))
 			if err := execFg(QEMU_IMG_BIN, "resize", rootDiskPath, fmt.Sprintf("%d", vmSpec.Guest.RootDisk.Size.Value())); err != nil {
-				log.Fatal(err)
+				logger.Fatal("Failed to resize rootDisk", zap.Error(err))
 			}
 		} else {
-			log.Printf("rootDisk.size (%s) should be more than size in image (%s)", vmSpec.Guest.RootDisk.Size.String(), imageSizeQuantity.String())
+			logger.Info(fmt.Sprintf("rootDisk.size (%s) is greater than image size (%s)", vmSpec.Guest.RootDisk.Size.String(), imageSizeQuantity.String()))
```
cicdteam, line 502: seems like a false statement. There we have to report that the size in the VM specification is smaller than the size in qcow2. We can only resize up, not down.

neonvm/runner/main.go
```diff
@@ -703,17 +712,17 @@ func listenForCPUChanges(ctx context.Context, port int32, cgroupPath string, wg
...
-			log.Println("cpu_change server closed")
+			logger.Info("cpu_change server closed")
 		} else if err != nil {
-			log.Fatalf("error starting server: %s", err)
+			logger.Fatal("cpu_change exited with error", zap.Error(err))
 		}
 	case <-ctx.Done():
 		err := server.Shutdown(context.Background())
-		log.Printf("shut down cpu_change server: %v", err)
+		logger.Info("shut down down cpu_change server", zap.Error(err))
```
cicdteam, line 721: down down

neonvm/runner/main.go
```diff
@@ -986,28 +996,28 @@ func execFg(name string, arg ...string) error {
...
 	}
 	if err := netlink.LinkAdd(bridge); err != nil {
-		log.Fatalf("could not create bridge interface: %s", err)
+		logger.Fatal("could not create bridge interface", zap.Error(err))
 		return nil, err
 	}
 	ipPod, ipVm, mask, err := calcIPs(cidr)
 	if err != nil {
-		log.Fatalf("could not parse IP: %s", err)
+		logger.Fatal("could not parse IP: %s", zap.Error(err))
```
cicdteam, line 1020: should be just `could not parse IP`

sharnoff: Good catch - fixed.

neonvm/runner/main.go
sharnoff: Oops lol, fixed.

neonvm/runner/main.go
sharnoff: Ah right, yeah - fixed.

cicdteam (approved):

PR #468, opened 2023-08-09
vm-builder: remove monitor

fprasx:
This PR introduces changes associated with the monitor being started by `compute_ctl`. It contains two main changes:
- _only_ creating a cgroup, and not starting `compute_ctl` in it. This is necessary so that the monitor does not freeze itself when it freezes the cgroup.
- not starting the monitor, as this is done by `compute_ctl`.

This should _not_ pass CI as there is no longer anything doing autoscaling! In the future we'll have to change our testing to actually run Neon so that the monitor is brought in in some capacity.

*Note*: merge this after #442 - there will be some deletion conflicts. Maybe I shouldn't have also deleted references to the informant here; we'll see.


sharnoff (commented):
Hm, I see that removing the monitor outright is failing CI. Maybe add a config option to disable it? That way _for now_, we can still mostly convert normal docker images into VMs, but we have the option to handle the monitor's functionality elsewhere (i.e. inside compute_ctl).

I hadn't considered the long-term implications on our CI of moving the monitor into `neondatabase/neon`... I'm not sure what the right path is here, tbh.

Maybe: we make a stub implementation of the monitor that just approves everything, and then use that here for our "basic" CI, with the full Neon stack's e2e tests added to this repo's CI to test the monitor?

fprasx:
> Maybe add a config option to disable it?

I think this a good idea. As for maintaining a stub an the version in `compute_ctl`, I think that'll probably be a lot of work to keep in sync. By "approves everything" do you mean "accepts all requests (i.e, always downscales/upscales when asked)"? If so, we can just use the monitor as it is now in `neondatabase/vm-monitor`. Removing the logic for actually deciding can we or can we not downscale does not but a big dent in code size.

fprasx:
Closing in favour of #477

fprasx:
Closing in favour of #477

PR #476, opened 2023-08-15
plugin: Penalize underfull nodes during Score

### LINKED ISSUE #411: Scheduler plugin should bias 'Score' against under-full nodes
As currently written, the scheduler plugin's 'Score' method will rank nodes higher if they're less full. This is good when the cluster as a whole is _mostly_ full, because it keeps things balanced. But it also makes scaling down much harder: the scheduler plugin will naturally evenly fill all available nodes, which means that even if we _should_ scale down the cluster (which may require some manual work), all nodes are equally bad "scale-down" targets.

We still don't want to over-fill nodes, so doing this correctly requires a bit of a careful balancing act — perhaps some simulations might be useful here. The key information is that the "core" scheduler (the thing interpreting our plugin's responses) will just take the scores for each node as given; if a node is scored higher, it will try to place the pod onto that node.

See also:

- #412
- [`TargetLoadPacking`](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/trimaran/targetloadpacking/README.md) plugin

sharnoff:
us-west-2 is currently at ~30-35% cluster-wide utilization, so I'm figuring that #411 should be handled sooner rather than later.

Implemented this after some brainstorming with @fprasx in a call earlier today.

Should resolve #411.

### DIFF
--- a/pkg/plugin/config.go
+++ b/pkg/plugin/config.go
@@ -99,10 +99,10 @@ type nodeConfig struct {
 	//
 	// This corresponds to y₁ in the desmos link above.
 	MaxUsageScore float64 `json:"maxUsageScore"`
-	// ScorePeak gives the fraction at which the "midpoint" should be -- or, in other words, it
-	// gives a notion of a "target" usage.
+	// ScorePeak gives the fraction at which the "target" or highest score should be, with the score
+	// sloping down on either side towards MinUsageScore at 0 and MaxUsageScore at 1.
 	//
-	// The midpoint has the maximum score attached to it, and corresponds to xₚ in the desmos link.
+	// This corresponds to xₚ in the desmos link.
 	ScorePeak float64 `json:"scoreMidpoint"`
 }
 
--- a/pkg/plugin/config.go
+++ b/pkg/plugin/config.go
@@ -185,6 +185,14 @@ func (c *nodeConfig) validate() (string, error) {
 		return "computeUnit", err
 	}
 
+	if c.MinUsageScore < 0 || c.MinUsageScore > 1 {
+		return "minUsageScore", errors.New("value must be between 0 and 1, inclusive")
+	} else if c.MaxUsageScore < 0 || c.MaxUsageScore > 1 {
+		return "maxUsageScore", errors.New("value must be between 0 and 1, inclusive")
+	} else if c.ScorePeak < 0 || c.ScorePeak > 1 {
+		return "scorePeak", errors.New("value must be between 0 and 1, inclusive")
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/plugin/config.go
```diff
@@ -82,6 +82,28 @@ type nodeConfig struct {
...
+	// MaxUsageScore gives the ratio of the score at the maximum usage (i.e. full) relative to the
+	// score at the midpoint, which will have the maximum.
+	//
+	// This corresponds to y₁ in the desmos link above.
+	MaxUsageScore float64 `json:"maxUsageScore"`
+	// ScoreMidpoint gives the fraction full at which the "midpoint" should be -- or, in other
+	// words, it gives a notion of a "target" usage.
+	//
+	// The midpoint has the maximum score attached to it, and corresponds to xₚ in the desmos link.
+	ScoreMidpoint float64 `json:"scoreMidpoint"`
```
fprasx, line 106: I would rename this field to `ScorePeak` (maybe `ScorePeakFraction`?) since it's unintuitive for the midpoint not to be in the middle. 

Also, is the word "full" in the comment supposed to be there?

pkg/plugin/plugin.go
```diff
@@ -702,23 +702,52 @@ func (e *AutoscaleEnforcer) Score(
...
-	totalMem := e.state.maxTotalReservableMemSlots
+	cpuRemaining := node.remainingReservableCPU()
+	cpuTotal := node.totalReservableCPU()
+	memRemaining := node.remainingReservableMemSlots()
+	memTotal := node.totalReservableMemSlots()
+
+	cpuFraction := 1 - cpuRemaining.AsFloat64()/cpuTotal.AsFloat64()
+	memFraction := 1 - float64(memRemaining)/float64(memTotal)
+	cpuScale := node.totalReservableCPU().AsFloat64() / e.state.maxTotalReservableCPU.AsFloat64()
+	memScale := float64(node.totalReservableMemSlots()) / float64(e.state.maxTotalReservableMemSlots)
```
fprasx, line 713: The formatting hurts my eyes. Curse you gofmt! :rofl:

pkg/plugin/config.go
fprasx: Don't forget to add validation as well.

fprasx (changes requested):

pkg/plugin/plugin.go
sharnoff: yeah.... :sweat_smile:

fprasx (approved):

PR #477, opened 2023-08-16
vm-builder: add flag to enable informant/monitor at VM startup and remove cgexec

fprasx:
This change enables more flexibility rolling out the monitor. We can keep using the monitor during `e2e` tests, while not enabling it for release builds.

This PR also removes cgexec. This functionality will be delegated to `compute_ctl`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ docker-build-scheduler: ## Build docker image for (autoscaling) scheduler
 
 .PHONY: docker-build-examples
 docker-build-examples: vm-informant bin/vm-builder ## Build docker images for testing VMs
-	./bin/vm-builder -src postgres:15-bullseye -dst $(E2E_TESTS_VM_IMG)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -196,7 +196,7 @@ docker-build-scheduler: ## Build docker image for (autoscaling) scheduler
 
 .PHONY: docker-build-examples
 docker-build-examples: vm-informant bin/vm-builder ## Build docker images for testing VMs
-	./bin/vm-builder -src postgres:15-bullseye -dst $(E2E_TESTS_VM_IMG)
+	./bin/vm-builder -src postgres:15-bullseye -dst $(E2E_TESTS_VM_IMG) -enable-informant
```
sharnoff, line 199: Need to also `-enable-monitor`?

sharnoff (commented):
Seems ok, should also mention cgexec in the PR title / body / final commit message.

Makefile
fprasx: Yup, just realized this. Am testing locally as it should be a little faster.

sharnoff (approved):

PR #480, opened 2023-08-17
vm-builder: add flag to configure uid of neon-postgres cgroup

fprasx:
This may fix a bug where the `postgres` user running compute_ctl cannot start postgres in the neon-postgres cgroup, due to inadequate permissions


sharnoff (commented):
s/cgroupUid/cgroupUID/ (becuase UID is an acronym/initialism, and Go style is to have the same case for all letters), but otherwise looks good.

sharnoff (approved):

PR #481, opened 2023-08-18
plugin: randomize plugin scores

### LINKED ISSUE #412: Scheduler plugin's 'Score' results should be (weighted) random
Deterministic scoring makes the system more fragile. We can improve this by making the score random in the range `[min score + 1, true score]`. The reason to make the floor just above the minimum score is because we use the minimum score for nodes where there isn't room, which shouldn't really ever be selected.

See also:

- #411

fprasx:
Randomizes scores in the range `[minScore + 1, trueScore]` to "soften" scaling decisions and make the system less fragile. Resolves #412.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -763,12 +763,28 @@ func (e *AutoscaleEnforcer) NormalizeScore(
 	state *framework.CycleState,
 	pod *corev1.Pod,
 	scores framework.NodeScoreList,
-) *framework.Status {
-	for _, nodeScore := range scores {
-		score := nodeScore.Score
+) (status *framework.Status) {
+	ignored := e.state.conf.ignoredNamespace(pod.Namespace)
+
+	e.metrics.IncMethodCall("NormalizeScore", ignored)
+	defer func() {
+		e.metrics.IncFailIfNotSuccess("NormalizeScore", ignored, status)
+	}()
+
+	logger := e.logger.With(zap.String("method", "NormalizeScore"), util.PodNameFields(pod))
+	logger.Info("Handling NormalizeScore request")
+
+	for _, node := range scores {
+		// logger := e.logger.With(zap.String("method", "Score"), zap.String("node", nodeName), util.PodNameFields(pod))
+		nodeScore := node.Score
+		nodeName := node.Name
 
 		// rand.Intn will panic if we pass in 0
-		if score == 0 {
+		if nodeScore == 0 {
+			logger.Info(
+				fmt.Sprintf("Ignoring node %s as it was assigned a score of 0", nodeName),
+				zap.String("node", nodeName),
+			)
 			continue
 		}
 
@@ -780,14 +796,24 @@ func (e *AutoscaleEnforcer) NormalizeScore(
 		// We want to pick a score in the range [minScore, score], so use
 		// score _+ 1_ - minscore, as rand.Intn picks a number in the _half open_
 		// range [0, n)
-		nodeScore.Score = int64(rand.Intn(int(score+1-minScore))) + minScore
+		newScore := int64(rand.Intn(int(nodeScore+1-minScore))) + minScore
+		logger.Info(
+			fmt.Sprintf(
+				"Randomly chose score %d from range [minScore=%d, trueScore=%d]",
+				newScore, minScore, nodeScore,
+			),
+			zap.String("node", nodeName),
+			zap.Int64("score", newScore),
+		)
+		node.Score = newScore
 	}
 	return nil
 }
 
-// ScoreExtensions is required for framework.ScorePlugin, and can return nil if it's not used
+// ScoreExtensions is required for framework.ScorePlugin, and can return nil if it's not used.
+// However, we do use it, to randomize scores.
 func (e *AutoscaleEnforcer) ScoreExtensions() framework.ScoreExtensions {
-	return nil
+	return e
 }
 
 // Reserve signals to our plugin that a particular pod will (probably) be bound to a node, giving us
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -773,11 +773,6 @@ func (e *AutoscaleEnforcer) NormalizeScore(
 	logger := e.logger.With(zap.String("method", "NormalizeScore"), util.PodNameFields(pod))
 	logger.Info("Handling NormalizeScore request")
 
-	if !e.state.conf.RandomizeScores {
-		logger.Info("randomizeScores not set, skipping randomizations")
-		return
-	}
-
 	for _, node := range scores {
 		nodeScore := node.Score
 		nodeName := node.Name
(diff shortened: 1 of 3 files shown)


pkg/plugin/plugin.go
```diff
@@ -754,6 +756,35 @@ func (e *AutoscaleEnforcer) Score(
...
+		// We want to pick a score in the range [minScore, score], so use
+		// score _+ 1_ - minscore, as rand.Intn picks a number in the _half open_
+		// range [0, n)
+		nodeScore.Score = int64(rand.Intn(int(score+1-minScore))) + minScore
+	}
+	return nil
+}
+
 // ScoreExtensions is required for framework.ScorePlugin, and can return nil if it's not used
 func (e *AutoscaleEnforcer) ScoreExtensions() framework.ScoreExtensions {
```
sharnoff, line 789: IIRC you also need to return `*AutoscaleEnforcer` from this function in order for the NormalizeScore method to get used

pkg/plugin/plugin.go
```diff
@@ -754,6 +756,35 @@ func (e *AutoscaleEnforcer) Score(
...
 }
 
+// NormalizeScore weights scores uniformly in the range [minScore, trueScore], where
+// minScore is framework.MinNodeScore + 1.
+func (e *AutoscaleEnforcer) NormalizeScore(
+	ctx context.Context,
+	state *framework.CycleState,
+	pod *corev1.Pod,
+	scores framework.NodeScoreList,
+) *framework.Status {
```
sharnoff, line 766: It'd also be good to have metrics & logs here about what's going on (i.e. metrics for "NormalizeScore was called!" and logs for "changed score ; node=..., pod=..., old score=..., new score=...").

sharnoff (commented):
Awesome! It'd be great to make enabling this configurable so we can try it out on half the regions at a time — I realized we _probably_ should have done that for #476, but I didn't think of it at the time.

And now it's [possible](https://github.com/neondatabase/autoscaling/blob/f2cfdc210b6750f89db10039e434469e91cde914/build.sh) to have [per-region configuration](https://github.com/neondatabase/autoscaling/blob/f2cfdc210b6750f89db10039e434469e91cde914/patches.json), so that would actually be feasible on the ops side :sweat_smile:

pkg/plugin/plugin.go
fprasx: Are you sure? I don't see anything here: https://pkg.go.dev/k8s.io/kubernetes/pkg/scheduler/framework#ScoreExtensions

sharnoff: I might be wrong, but my intuition is that given the ScoreExtensions method exists (and especially because there's no corresponding `NormalizeScorePlugin` interface), the scheduler won't call the NormalizeScore method without you telling it to.

You should be able to tell either way with logs though ;)

fprasx: Ohhh I see. I interpreted "this function" as the new one, instead of the `ScoreExtensions` one. Should've paid closer attention to the section of code you selected. Indeed, the docs say:
```
// ScoreExtensions returns a ScoreExtensions interface if it implements one, or nil if does not.
ScoreExtensions() ScoreExtensions
```

pkg/plugin/plugin.go
fprasx: Fixed in 72b53c3a0193e812ea9d19aee2d39ce750ae54ed

pkg/plugin/plugin.go
fprasx: Fixed in 72b53c3a0193e812ea9d19aee2d39ce750ae54ed

pkg/plugin/plugin.go
```diff
@@ -60,6 +61,7 @@ var _ framework.PreFilterPlugin = (*AutoscaleEnforcer)(nil)
 var _ framework.PostFilterPlugin = (*AutoscaleEnforcer)(nil)
 var _ framework.FilterPlugin = (*AutoscaleEnforcer)(nil)
 var _ framework.ScorePlugin = (*AutoscaleEnforcer)(nil)
+var _ framework.ScoreExtensions = (*AutoscaleEnforcer)(nil)
```
sharnoff, line 64: I think this is no longer necessary now? because we're returning `*AutoscaleEnforcer` from `ScoreExtensions()`

sharnoff (commented):
Some logging nits. + should have a config flag to enable the randomness — other than that, looks good! 

To elaborate a bit on "why a config flag": One of the things I'm concerned about is that enabling the randomness _may_ nullify our other scoring work, because random steps towards the target utilization increase the chance of further steps in that direction. I'm really not sure though, and without a simulation to go off of (and because staging's scheduling load isn't representative of prod), the next best thing is A/B testing :)

pkg/plugin/plugin.go
```diff
@@ -754,11 +756,65 @@ func (e *AutoscaleEnforcer) Score(
...
+		// actual score we assign a node is thus framework.MinNodeScore + 1
+		minScore := framework.MinNodeScore + 1
+
+		// We want to pick a score in the range [minScore, score], so use
+		// score _+ 1_ - minscore, as rand.Intn picks a number in the _half open_
+		// range [0, n)
+		newScore := int64(rand.Intn(int(nodeScore+1-minScore))) + minScore
+		logger.Info(
+			fmt.Sprintf(
+				"Randomly chose score %d from range [minScore=%d, trueScore=%d]",
```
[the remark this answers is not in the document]
fprasx: I was going for a "verdict" style message here. Would it be ok to remove the actual score from the message but keep the range? That information should be available from the `Score` step logs anyways.

pkg/plugin/plugin.go
```diff
@@ -754,11 +755,67 @@ func (e *AutoscaleEnforcer) Score(
...
+		e.metrics.IncFailIfNotSuccess("NormalizeScore", ignored, status)
+	}()
+
+	logger := e.logger.With(zap.String("method", "NormalizeScore"), util.PodNameFields(pod))
+	logger.Info("Handling NormalizeScore request")
+
+	if !e.state.conf.RandomizeScores {
+		logger.Info("randomizeScores not set, skipping randomizations")
+		return
+	}
```
sharnoff, line 779: I think it might be easier / clearer / etc to implement this in `ScoreExtensions` — that way the metrics also show whether NormalizeScore is enabled. (and so we don't get a bunch of repeated log lines just to say "doing nothing because this feature isn't enabled")

pkg/plugin/plugin.go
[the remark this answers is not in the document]
sharnoff: Hm, yeah honestly I'd probably just put `minScore` and `trueScore` as additional fields, and have the message read `"Randomly choosing from range [minScore, trueScore]"` — _in general_ the benefit from making the messages constant is that it's possible to search for things with logql queries that look like

```
{pod="..."} | json | msg = "Doing XYZ with foobar"
```

(and then "foobar" can be extracted separately)

Of course this is not necessarily the case outside of our stack, and striking a balance between having good information in the message while still making it searchable is hard! The specific problem with searchability here is that regex, at least for Loki, tends to be slower than exact string matching.

---

Back to the specifics here though, I might even say something like `"Randomly choosing newScore from range [minScore, trueScore]"` (with zap fields `newScore`, `minScore`, and `trueScore`), just to make the meaning of the action clearer — it's up to you though.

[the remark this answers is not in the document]
fprasx: Ahh ok, fair point. Given that the logs are so processed I can understand that the actual message text is not as important.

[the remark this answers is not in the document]
sharnoff: Yup -- saw cfc0a18, looks good

pkg/plugin/plugin.go
fprasx: I see; that's better.

sharnoff (approved):

PR #485, opened 2023-08-21
vm-builder: use passed in cgroup-uid in more places

fprasx:
This should hopefully resolve permission errors causing neondatabase/neon#4920 (start postgres in cgroup via compute_ctl) to fail CI.


fprasx:
Closed as it doesn't work :(

PR #489, opened 2023-08-22
agent/informant: Update lastSuccessfulInformantComm for monitor comms

sharnoff:
This should fix false positives for the "stuck VMs" alert on staging. Validated the old & new (fixed) behavior locally.

### DIFF
--- a/pkg/agent/informant.go
+++ b/pkg/agent/informant.go
@@ -1203,7 +1203,23 @@ func (s *InformantServer) MonitorHealthCheck(ctx context.Context, logger *zap.Lo
 	}
 
 	_, err = s.awaitMonitorResponse(ctx, rx)
-	return err
+	if err != nil {
+		return err
+	}
+
+	s.runner.lock.Lock()
+	defer s.runner.lock.Unlock()
+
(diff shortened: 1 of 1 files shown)


pkg/agent/informant.go
```diff
@@ -1203,7 +1203,23 @@ func (s *InformantServer) MonitorHealthCheck(ctx context.Context, logger *zap.Lo
...
 	_, err = s.awaitMonitorResponse(ctx, rx)
-	return err
+	if err != nil {
+		return err
+	}
+
+	s.runner.lock.Lock()
+	defer s.runner.lock.Unlock()
+
+	// Update our record of the last successful time we heard from the monitor, if the "server"
```
fprasx, line 1213: I would drop the "if the server is confirmed part" as connected to monitor implies server is confirmed, and we don't even check for this anyways.

fprasx (approved):

pkg/agent/informant.go
sharnoff: Makes sense - done! 4f04f81

sharnoff:
Going to merge without waiting fore CI because it previously passed and the newest commit is just a comment.

PR #499, opened 2023-08-22
agent: Move awaitMonitorResponse into (*Dispatcher).Call

sharnoff:
This is an initial refactoring that should make some of the metrics parts easier.

Part of #492.


fprasx (approved):
Looks good.

sharnoff:
I'll wait to merge this, pending review of #500, in case there's changes needed there that would be better left to this PR.

PR #500, opened 2023-08-22
agent/dispatcher: Add request(ish) message metrics

sharnoff:
Builds on #499. Two parts to this PR:

- [x] "request" metrics (transactions initiated by the agent)
- [x] "response" metrics (transactions initiated by the monitor)

### DIFF
--- a/pkg/agent/dispatcher.go
+++ b/pkg/agent/dispatcher.go
@@ -164,7 +164,12 @@ func (disp *Dispatcher) unregisterWaiter(id uint64) {
 
 // Make a request to the monitor and wait for a response. The value passed as message must be a
 // valid value to send to the monitor. See the docs for SerializeInformantMessage for more.
-func (disp *Dispatcher) Call(ctx context.Context, timeout time.Duration, messageType string, message any) (*MonitorResult, error) {
+func (disp *Dispatcher) Call(
+	ctx context.Context,
+	timeout time.Duration,
+	messageType string,
+	message any,
+) (*MonitorResult, error) {
 	id := disp.lastTransactionID.Add(2)
 	sender, receiver := util.NewSingleSignalPair[waiterResult]()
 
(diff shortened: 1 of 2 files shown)


pkg/agent/dispatcher.go
```diff
@@ -233,20 +248,47 @@ func (disp *Dispatcher) HandleMessage(
...
+		// waiting for it.
+		var err error
+
+		panicPayload := recover()
+		if panicPayload != nil {
+			err = errors.New("panicked")
+		} else if rootErr != nil {
+			err = rootErr
+		} else {
+			err = errors.New("unknown")
```
fprasx, line 267: Wouldn't we want to just return here as we're neither panicking nor is there an error?

fprasx: Hmm ok, I would just add a comment then. Or maybe even set the error message to `unknown error, did you forget to set rootError?`. Might be a bit too cheeky lol

sharnoff: added a comment in 928d423.

pkg/agent/dispatcher.go
```diff
@@ -142,9 +147,14 @@ func (disp *Dispatcher) unregisterWaiter(id uint64) {
 
 // Make a request to the monitor and wait for a response. The value passed as message must be a
 // valid value to send to the monitor. See the docs for SerializeInformantMessage for more.
-func (disp *Dispatcher) Call(ctx context.Context, timeout time.Duration, message any) (*MonitorResult, error) {
+func (disp *Dispatcher) Call(ctx context.Context, timeout time.Duration, messageType string, message any) (*MonitorResult, error) {
```
fprasx, line 150:
```suggestion
func (disp *Dispatcher) Call(
    ctx context.Context,
    timeout time.Duration,
    messageType string,
    message any
) (*MonitorResult, error) {
```
Line is getting a bit long.

sharnoff: Good call. done 93bc7cf

PR #501, opened 2023-08-23
agent/dispatcher: Only generate even IDs

sharnoff:
This should help avoid collisions with IDs generated by the monitor, which is one way of resolving potential bugs that #495 introduced.

See also: neondatabase/vm-monitor#4, neondatabase/neon#5073


fprasx (approved):
Clever, I like it :)

PR #505, opened 2023-08-25
pull vm-monitor from neon repository

fprasx:
This keeps autoscaling and neon 'N Sync.

### DIFF
--- a/build/vm-monitor/Dockerfile
+++ b/build/vm-monitor/Dockerfile
@@ -8,14 +8,14 @@ RUN apk add musl-dev git
 #
 # Note: unlike in the stackoverflow post we access the branches endpoint so that
 # `latest_commit` will change upon a commit in any branch
-ADD "https://api.github.com/repos/neondatabase/vm-monitor/branches" latest_commit
+ADD "https://api.github.com/repos/neondatabase/neon/branches" latest_commit
 
-RUN git clone https://github.com/neondatabase/vm-monitor.git 
-RUN cargo build --release --manifest-path vm-monitor/Cargo.toml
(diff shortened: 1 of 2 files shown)


fprasx:
Well, guess the testing issues were just local :)

build/vm-monitor/Dockerfile
```diff
@@ -8,14 +8,14 @@ RUN apk add musl-dev git
 #
 # Note: unlike in the stackoverflow post we access the branches endpoint so that
 # `latest_commit` will change upon a commit in any branch
-ADD "https://api.github.com/repos/neondatabase/vm-monitor/branches" latest_commit
+ADD "https://api.github.com/repos/neondatabase/neon/branches" latest_commit
```
sharnoff, line 11: H, probably we should only care about the branch we're building from - otherwise the rate of new commits on _any_ branch in `neondatabase/neon` will constantly invalidate the cache :sweat_smile:

build/vm-monitor/Dockerfile
```diff
@@ -8,14 +8,14 @@ RUN apk add musl-dev git
...
 
-RUN git clone https://github.com/neondatabase/vm-monitor.git 
-RUN cargo build --release --manifest-path vm-monitor/Cargo.toml
+RUN git clone https://github.com/neondatabase/neon.git
```
sharnoff, line 13: A couple things:

1. Should be `--depth 1` so we're not pulling in all of the history as well
2. Branch should be configurable (e.g. via build arg), to make it easier to test vm-monitor changes with autoscaling changes.

sharnoff (approved):
Looks great :)

PR #507, opened 2023-08-29
neonvm: Add EmptyDisk.Discard option

sharnoff:
It enables the "discard" mount option, which can be used instead of fstrim to propagate hole punches down to the backing file (I think).

See also:

- https://chrisirwin.ca/posts/discard-with-kvm-2020/
- https://serverfault.com/questions/896448/qemu-trim-and-discard-on-a-physical-ssd-device

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -310,6 +310,8 @@ type DiskSource struct {
 
 type EmptyDiskSource struct {
 	Size resource.Quantity `json:"size"`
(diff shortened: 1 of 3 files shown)


sharnoff:
Perf analysis: This seems to be more demanding on the host filesystem (which... makes sense). On my machine, it seems about 30-50% slower, on both top speed and when the disk queue is saturated (although medium-speed SSD + full-disk encryption means it's easy to saturate disk queue, so it's hard to tell).

So, opened #509 which should help with that, and other places where we've run into throttling, more generally.

So we should be good here.

PR #508, opened 2023-08-30
error: upgrade fun, simplify error handling

tychoish:
Thought I'd do one last bump of this, and make a few changes to take advantage of 

I wrote up the following about the various error handling options. 

(The other differnce between using `ers.{Wrap,Join,Collect}`. and
using `%w` is that the concatination of the errors with `%w` happens
in the format of the error itself, with the ers tools, it's a property
of `ers.Stack` (which underpins all of the ers/erc tooling?), so you
can disassemble a stack-based error and have each individual error be
tractable and usable?)

------

The basic overview is:

- `ers.Join` (like `errors.Join`) combines a collection of errors,
  however, rather than using a slice, it uses the `ers.Stack`, which
  as a size-tracking linked list, can be used more incrementally and
  has more intuitive Wrapping/Unwrapping sementics. Like `errors.Join`
  this function drops nil errors, and returns nil when there are no
  errors. Where `errors.Join` has a string output form that joins
  constituent errors with a `\n`, `ers.Join` uses `: ` which is
  consistent with the way we think about wrapped errors (and typically
  format), it also reduces the amount of escaping that has to happen
  in log output.

- this `ers.Stack` is the same type that's been at the heart of the
  `erc.Collector` forever, so things don't really change. While
  `ers.Stack` isn't thread-safe, it's just a size-tracking linked
  list, for errors (thereby avoiding the preallocation temptation,)
  with helpers (e.g. Push/Add) much like the Collector. 
  
  Pros:
  - minimal overhead
  - good ergonomics (push/add ignore nil errors).
  - reasonable semantics, and tools (e.g. `ers.AsStack` and
    `ers.Unwind`) to introspect error values.
  - respects `Is` and `As` interfaces, while also avoiding the
    awkwardness of having an `Unwrap` method that returns a slice.
      
  Cons:
  - Stack objects implement `error`, so if you return an error stack
    as an error, it's non-nil even when it's empty. ~~I'm going to add a
    `Resolve()` method before merging this just to reduce confusion.~~
  - not in the standard library, (though fully compatible.)
  
- `erc.Collector` just wraps the ers.Stack object with a mutex
  (essentially,) and provides some higher level interfaces. The mutex
  is useful for collecting errors from operations in different worker
  functions. However, because the the collector checks to see if the
  error is non-nil before taking the mutex, any mutex cost/contention
  happens outside of the happy path.

- Nothing that happens here uses it, but [following from this
  post](https://dave.cheney.net/2016/04/07/constant-errors) the `ers`
  package has an `Error` type which wraps `string` and makes it
  possible to declare sentinel errors as constants. There are some
  other goodies in the `ers` package: helpers to attach and access
  timestamps from errors, the Wrap (eaiser than `%w` for common cases)
  and When (for making errors conditional to avoid really verbose
  aggregation forms) an their `Whenf`/`Wrapf` helpers that build their
  strings lazily. A `Strings()` helper for converting a slice of
  errors into a slice of strings (dropping nil errors), and some
  panics-to-errors tooling.

(As always, fun doesn't have any non-stdlib dependencies, and the ers
package is upstream of (nearly) all packages in the fun module, and
the test coverage on this stuff is pretty darn good. While my impulse
is to fix problems or confusion with the API rather than let broken
semantics remain in the name of stability, the dependency and testing
will help avoid bit rot.)

### DIFF
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -261,10 +261,10 @@ func (b ResourceBounds) validate(path string, memSlotSize *resource.Quantity) er
 	// ers.Whenf is nil if the condition is false; also does not
 	// resolve/format the string unless the condition is true.
 	es.Push(ers.Whenf(b.CPU.IsZero(), "%s.cpu: must have a non-zero value", path))
-	es.Push(ers.Whenf(b.Mem.IsZero() || b.Mem.Value() < 0),
-		"%s.mem: must have value greater than zero", path)
-	es.Push(ers.Whenf(Mem.Value()%memSlotSize.Value() != 0,
-		"%s.mem: must be divisiable by VM memory slot size %s", memSlotSize))
+	es.Push(ers.Whenf(b.Mem.IsZero() || b.Mem.Value() < 0,
+		"%s.mem: must have value greater than zero", path))
+	es.Push(ers.Whenf(b.Mem.Value()%memSlotSize.Value() != 0,
+		"%s.mem: must be divisiable by VM memory slot size %s", path, memSlotSize))
 
 	// ers.Stack is always error, and is non-nil here: calling
 	// resolve returns nil if none of the above are errors:
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -4,10 +4,9 @@ package api
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 
-	"github.com/tychoish/fun/erc"
+	"github.com/tychoish/fun/ers"
 	"go.uber.org/zap"
 
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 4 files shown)


pkg/api/vminfo.go
```diff
@@ -285,20 +286,19 @@ type ScalingConfig struct {
...
+	es.Push(ers.Whenf(c.LoadAverageFractionTarget < 0.0, "%s must be set to value >= 0", ".loadAverageFractionTarget"))
+	es.Push(ers.Whenf(c.LoadAverageFractionTarget >= 2.0, "%s must be set to value < 2 ", ".loadAverageFractionTarget"))
 
 	// Make sure c.MemoryUsageFractionTarget is between 0 and 1
-	erc.Whenf(ec, c.MemoryUsageFractionTarget < 0.0, "%s must be set to value >= 0", ".memoryUsageFractionTarget")
-	erc.Whenf(ec, c.MemoryUsageFractionTarget >= 1.0, "%s must be set to value < 1 ", ".memoryUsageFractionTarget")
+	es.Push(ers.Whenf(c.MemoryUsageFractionTarget < 0.0, "%s must be set to value >= 0", ".memoryUsageFractionTarget"))
+	es.Push(ers.Whenf(c.MemoryUsageFractionTarget >= 1.0, "%s must be set to value < 1 ", ".memoryUsageFractionTarget"))
 
 	// heads-up! some functions elsewhere depend on the concrete return type of this function.
```
tychoish, line 300: I'm pretty sure this isn't true, based on some poking around, and while the stack type moved from `erc` to `ers` it should all work the same. 

There's also now an `ers.AsStack` helper that has some strong "dwim" vibes for converting/casting an error to a stack, which would be useful if this were true!

tychoish:
I no longer have write access to this branch, and this is pretty stale, so I think the best course of action is to close it and let you all deal with this on your own timing/priority.

PR #509, opened 2023-08-31
neonvm/runner: Add overcommit factor for CPU limits

sharnoff:
Copy-pasting the added comment:

> cpuLimitOvercommitFactor sets the amount above the VM's
> spec.guest.cpus.use that we set the QEMU cgroup's CPU limit to.
> e.g. if cpuLimitOvercommitFactor = 3 and the VM is using 0.5 CPUs, we
> set the cgroup to limit QEMU+VM to 1.5 CPUs.
>
> This exists because setting the cgroup exactly equal to the VM's CPU
> value is overly pessimistic, results in a lot of unused capacity on
> the host, and particularly impacts operations that parallelize between
> the VM and QEMU, like heavy disk access.
>
> See also: https://neondb.slack.com/archives/C03TN5G758R/p1693462680623239

With the factor at 4x, this _should_ also resolve neondatabase/cloud#6243 as well (given that the slow starts were only happening when the cgroup was moved from 1 CPU to 0.25 CPUs, and this would bump it up to 4 -> 1 CPUs). But we should wait & see first.

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -68,6 +68,17 @@ const (
 	// in microseconds. Min 1000 microseconds, max 1 second
 	cgroupPeriod     = uint64(100000)
 	cgroupMountPoint = "/sys/fs/cgroup"
+
+	// cpuLimitOvercommitFactor sets the amount above the VM's spec.guest.cpus.use that we set the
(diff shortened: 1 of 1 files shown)


sharnoff:
Oh, I think this also supersedes #404 - @hlinnaka to confirm.

cicdteam (approved):

nikitakalyanov (approved):

PR #520, opened 2023-09-06
add serviceAccountName field

cicdteam:
1. first there was this - https://github.com/neondatabase/neonvm/pull/12
2. then was this - https://github.com/neondatabase/autoscaling/pull/140

and now we need serviceAccount again to have ability use IRSA and let applications inside VM have access to AWS resources (s3)

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -69,11 +69,12 @@ type VirtualMachineSpec struct {
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds"`
 
-	NodeSelector  map[string]string           `json:"nodeSelector,omitempty"`
-	Affinity      *corev1.Affinity            `json:"affinity,omitempty"`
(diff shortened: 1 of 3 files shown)


sharnoff:
This won't actually propagate the service account token to become visible within the VM though, right? See e.g. what I wrote on the original PR to remove it here:

> As I understand it, the current issue with `ServiceAccount` is that it won't forward the added mounts from the service account to the VM, so the feature doesn't actually do anything right now (this is what neondatabase/neonvm#103 is about). IIRC there's some subtle tricky things there as well, so I think it's not as simple as "just" adding the config.

Also, because service account tokens are rotated _somewhat_ frequently by updating the configmap, we also need to make sure that configmap updates get forwarded through the iso9660 mount to become visible in the VM.

cicdteam:
@sharnoff 

yes, any serviceAccount related stuff (token, CA, k8s api hostname) not mounted into VM. And for IRSA it's not necessary. Application in VM should be able to retrieve temporary AWS creds from instance metadata service - IMDS (it lives at special link-local IP address 169.254.169.254). 

I hope that attaching runner pod to SA will be enough to get access to AWS S3 resources (corresponding IAM role is linked to SA)
.

sharnoff (approved):
Fair enough, let's try it :)

PR #521, opened 2023-09-10
agent: Add support for metrics provided by vm-monitor

sharnoff:
The prometheus output provided by node_exporter / vector is a lot more than we need, and parsing it in the autoscaler-agent (because that's currently how we get metrics) is a little hacky. We already maintain communication with the vm-monitor, so let's just fetch the metrics there too — this also means we can more easily customize the information in the future, like providing stats relevant to the `neon-postgres` cgroup.

This should also help somewhat with providing per-endpoint metrics.

**NB:** This PR currently overrides vm-monitor builds to point to the `sharnoff/monitor-metrics` branch of `neondatabase/neon`. That should be reverted before merging.


sharnoff:
Copying from my comment closing neondatabase/neon#5272:

> Closing because I no longer think this is necessary / worthwhile. All the metrics are already available from vector running inside the VM, and it's better to pipe new metrics through there anyways.

PR #523, opened 2023-09-18
speculative workaround for high unclosed connection count between compute smgr & pageserver

problame:
One possible explanation for our high ESTABLISHED connection count
between computes and pageserver's page_service.rs is that we kill the
VMs before the FIN/RST packets from the `pg_ctl -m stop` leave the VM's
kernel.

The sleep added in this PR would give them a good chance.

Roll this out and observe.

Context: https://neondb.slack.com/archives/C036U0GRMRB/p1694427077510779?thread_ts=1694425364.531029&cid=C036U0GRMRB

### DIFF
--- a/neonvm/tools/vm-builder/main.go
+++ b/neonvm/tools/vm-builder/main.go
@@ -260,6 +260,12 @@ action=/neonvm/bin/powerdown
 	scriptPowerDown = `#!/neonvm/bin/sh
 
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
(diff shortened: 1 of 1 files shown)


neonvm/tools/vm-builder/main.go
```diff
@@ -260,6 +260,7 @@ action=/neonvm/bin/powerdown
 	scriptPowerDown = `#!/neonvm/bin/sh
 
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
+sleep 1
```
sharnoff, line 263: Let's add a brief comment here explaining what's going on

sharnoff (approved):
One nit, otherwise all good

JFYI this won't be in effect until we (a) cut a release here [happy to do that today], and (b) update `neondatabase/neon`'s usage of vm-builder to that release.

vadim2404 (approved):

PR #525, opened 2023-09-22
acpi shutdown: correct and robust graceful shutdown

problame:
We're seeing unclosed connections from VM-based computes to pageservers.
They are reproducibly due to `suspend_compute` actions in console.

An earlier attempt to mitigate this was made in
https://github.com/neondatabase/autoscaling/pull/523
but it didn't show any effect.

See the `README.md` added in this PR for details.

### DIFF
--- a/neonvm/tools/vm-builder/main.go
+++ b/neonvm/tools/vm-builder/main.go
@@ -265,8 +265,8 @@ if [ -e /neonvm/vmstart.allowed ]; then
 	echo "Error: could not remove vmstart.allowed marker, might hang indefinitely during shutdown" >2&
 fi
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
-# wait for ongoing command to exit
-flock /neonvm/vmstart.lock
+# wait for ongoing command to exit by grabbing the lock
+flock /neonvm/vmstart.lock true
 `
 
 	scriptVmInit = `#!/neonvm/bin/sh
--- a/neonvm/tools/vm-builder/main.go
+++ b/neonvm/tools/vm-builder/main.go
@@ -262,11 +262,15 @@ action=/neonvm/bin/poweroff
 	scriptVmShutdown = `#!/neonvm/bin/sh
 rm /neonvm/vmstart.allowed
 if [ -e /neonvm/vmstart.allowed ]; then
-	echo "Error: could not remove vmstart.allowed marker, might hang indefinitely during shutdown" >2&
+	echo "Error: could not remove vmstart.allowed marker, might hang indefinitely during shutdown" 1>&2
 fi
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
 # wait for ongoing command to exit by grabbing the lock
-flock /neonvm/vmstart.lock true
+if flock /neonvm/vmstart.lock true; then
+	echo "vmstart workload shut down cleanly" 1>&2
+else
+	echo "error: vmshutdown flock failed" 1>&2
+fi
 `
 
 	scriptVmInit = `#!/neonvm/bin/sh
--- a/neonvm/tools/vm-builder/main.go
+++ b/neonvm/tools/vm-builder/main.go
@@ -231,16 +231,14 @@ else
     /neonvm/bin/echo -n {{range .Cmd}}' '{{.}}{{end}} >> /neonvm/bin/vmstarter.sh
 fi
 
-/neonvm/bin/cat <<'EOF' >>/neonvm/bin/vmstarter.sh
-
-if [ -e "/neonvm/vmstart_command_finished.fifo" ]; then
-	echo ' ' > /neonvm/vmstart_command_finished.fifo
-fi
-EOF
-
 /neonvm/bin/chmod +x /neonvm/bin/vmstarter.sh
 
 /neonvm/bin/su-exec {{.User}} /neonvm/bin/sh /neonvm/bin/vmstarter.sh
+
+# signal vmshutdown that vmstarter.sh has exited
+if [ -e "/neonvm/vmstart_command_finished.fifo" ]; then
+	echo ' ' > /neonvm/vmstart_command_finished.fifo
+fi
 `
 
 	scriptInitTab = `
@@ -268,6 +266,7 @@ action=/neonvm/bin/poweroff
 	scriptVmShutdown = `#!/neonvm/bin/sh
 mkfifo /neonvm/vmstart_command_finished.fifo || exit 2
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
+# wait for vmstart to signal exit of vmstarter.sh
 cat /neonvm/vmstart_command_finished.fifo
 `
 
--- a/neonvm/tools/vm-builder/main.go
+++ b/neonvm/tools/vm-builder/main.go
@@ -169,10 +169,11 @@ RUN set -e \
 ADD inittab   /neonvm/bin/inittab
 ADD vminit    /neonvm/bin/vminit
 ADD vmstart   /neonvm/bin/vmstart
+ADD vmshutdown /neonvm/bin/vmshutdown
 ADD vmacpi    /neonvm/acpi/vmacpi
 ADD vector.yaml /neonvm/config/vector.yaml
 ADD powerdown /neonvm/bin/powerdown
-RUN chmod +rx /neonvm/bin/vminit /neonvm/bin/vmstart /neonvm/bin/powerdown
+RUN chmod +rx /neonvm/bin/vminit /neonvm/bin/vmstart /neonvm/bin/vmshutdown /neonvm/bin/powerdown
 
 FROM vm-runtime AS builder
 ARG DISK_SIZE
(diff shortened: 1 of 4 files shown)


neonvm/tools/vm-builder/main.go
```diff
@@ -250,24 +258,18 @@ fi
 ::respawn:su -p nobody -c '/usr/local/bin/pgbouncer /etc/pgbouncer.ini'
 ::respawn:su -p nobody -c 'DATA_SOURCE_NAME="user=cloud_admin sslmode=disable dbname=postgres" /bin/postgres_exporter --auto-discover-databases --exclude-databases=template0,template1'
 ttyS0::respawn:/neonvm/bin/agetty --8bits --local-line --noissue --noclear --noreset --host console --login-program /neonvm/bin/login --login-pause --autologin root 115200 ttyS0 linux
+::shutdown:/neonvm/bin/vmshutdown
```
sharnoff, line 261: does `::shutdown` actually exist? I couldn't find it, but maybe I was looking at the wrong man page :sweat_smile:

neonvm/tools/vm-builder/main.go
```diff
@@ -250,24 +258,18 @@ fi
...
 `
 
 	scriptVmAcpi = `
 event=button/power
-action=/neonvm/bin/powerdown
+action=/neonvm/bin/poweroff
```
sharnoff, line 266: s/poweroff/shutdown/ ? (or, maybe leave powerdown as it is?)

neonvm/tools/vm-builder/main.go
```diff
@@ -169,10 +169,11 @@ RUN set -e \
 ADD inittab   /neonvm/bin/inittab
 ADD vminit    /neonvm/bin/vminit
 ADD vmstart   /neonvm/bin/vmstart
+ADD vmshutdown /neonvm/bin/vmshutdown
 ADD vmacpi    /neonvm/acpi/vmacpi
 ADD vector.yaml /neonvm/config/vector.yaml
 ADD powerdown /neonvm/bin/powerdown
-RUN chmod +rx /neonvm/bin/vminit /neonvm/bin/vmstart /neonvm/bin/powerdown
+RUN chmod +rx /neonvm/bin/vminit /neonvm/bin/vmstart /neonvm/bin/vmshutdown /neonvm/bin/powerdown
```
sharnoff, line 176: nb: `/neonvm/bin/powerdown` no longer exists because powerdown was renamed to vmshutdown below

sharnoff (commented):
Need to think about this some more. Figured I'd point out the stuff that came up on the first few passes.

(also, I'm mostly going to comment on `vm-builder` instead of `vm-builder-generic`, but I _think_ the changes apply to both, unfortunately)

problame:
(Added a mermaid diagram to the PR description)

neonvm/tools/vm-builder/main.go
problame: No, what I have here is correct. See the sequence diagram and code links I added to the PR description.

neonvm/tools/vm-builder/main.go
problame: Yes it does, the online documentation is terrible. I think it's been around in several init implementation, and I know it's present in the busybox init.
See the source code links I added to the PR description.

neonvm/tools/vm-builder/main.go
```diff
@@ -231,6 +231,13 @@ else
     /neonvm/bin/echo -n {{range .Cmd}}' '{{.}}{{end}} >> /neonvm/bin/vmstarter.sh
 fi
 
+/neonvm/bin/cat <<'EOF' >>/neonvm/bin/vmstarter.sh
+
+if [ -e "/neonvm/vmstart_command_finished.fifo" ]; then
+	echo ' ' > /neonvm/vmstart_command_finished.fifo
+fi
+EOF
```
sharnoff, line 239: nb: `vmstarter.sh` is fully generated by vm-builder, and run as the image's USER (for us, postgres). May be worthwhile to move this into `vmstart` instead? (both because it's running as root, and it's static)

sharnoff (commented):
The diagram was super helpful - thanks!

From what you've said, it seems like it's possible _in theory_ for a race condition to result in `vmshutdown` hanging, but I'm not entirely sure. Here's what I think can happen:

1. `vmstarter.sh`/`vmstart` checks `-e "/neonvm/vmstart_command_finished.fifo"` and the file doesn't exist OR `vmstarter.sh`/`vmstart` gets killed.
2. ACPI shutdown starts, init gets SIGUSR2 and stops respawning `vmstart`
3. `vmstarter.sh`/`vmstart` exits, and doesn't get restarted
4. `vmshutdown` creates `/neonvm/vmstart_command_finished.fifo`, and later blocks on reading from it, without anything doing that

I guess it could also happen if `vmstarter.sh`/`vmstart` gets killed during shutdown, while blocking on writing to the fifo, but I think that's less likely.

Not sure how bad the risk of this is, but we _probably_ should have a shutdown timeout anyways, and that could serve a secondary purpose here to make sure that we handle potential deadlocks during shutdown a bit better. Up to you.

problame:
Yeah I considered that race condition and decided it doesn't matter because I hope there is a timeout implemented outside the VM anyway.
Basically a timeout on VM responding to ACPI shutdown signal.
We need such a timeout anyway because we cannot trust the VM to cooperate.

neonvm/tools/vm-builder/main.go
problame: Good idea

problame:
I'm learning TLA+ at the moment.
So, I took your complaint about the possible deadlock as a challenge, modeled shutdown in PlusCal, and switched to a different implementation that passes model check
* no deadlocks
* the temporal properties at the bottom of the model file

I updated the Mermaid diagram and moved it into the README.

Even without TLA+, I think this "permit file + flock" approach is much simpler to reason about.

problame:
What I'd need confirmed before we merge this is that the k8s-level timeout exists.
From the added docs/README.md

```
Neon Control Plane's `suspend_compute` relies on ACPI shutdown
signalling for graceful shutdown of the NeonVM.
If the NeonVM doesn't shut down timely, the pod that contains
the qemu process gets SIGKILLed.
```

neonvm/tools/vm-builder/main.go
```diff
@@ -250,24 +251,22 @@ fi
...
-action=/neonvm/bin/powerdown
+action=/neonvm/bin/poweroff
 `
 
-	scriptPowerDown = `#!/neonvm/bin/sh
-
+	scriptVmShutdown = `#!/neonvm/bin/sh
+rm /neonvm/vmstart.allowed
+if [ -e /neonvm/vmstart.allowed ]; then
+	echo "Error: could not remove vmstart.allowed marker, might hang indefinitely during shutdown" >2&
```
sharnoff, line 265:
```suggestion
	echo "Error: could not remove vmstart.allowed marker, might hang indefinitely during shutdown" >&2
```
assuming this was meant to go to stderr, right? (rather than running "write to the file '2' in the background")

neonvm/tools/vm-builder/main.go
```diff
@@ -250,24 +251,22 @@ fi
...
 su -p postgres --session-command '/usr/local/bin/pg_ctl stop -D /var/db/postgres/compute/pgdata -m fast --wait -t 10'
-# The poweroff below is the busybox poweroff which goes straight to the kernel, i.e., LINUX_REBOOT_CMD_POWER_OFF / RB_POWER_OFF.
-# Our experiments have shown that, generally, this type of hard shutdown will not FIN/RST existing TCP connections (i.e., state ESTABLISHED).
-# Now, for the particular case of NeonVM, we just did 'pg_ctl stop'.
-# But, the libpagestore.c is currently not explicitly shutting down the TCP connection to pageservers, relying on the kernel
-# to implicitly close the socket on exit.
-# This sleep here is to give the kernel time to send the FIN/RST.
-sleep 1
-/neonvm/bin/poweroff
+# wait for ongoing command to exit by grabbing the lock
```
sharnoff, line 268: I _think_ there's still a race condition here, if I'm not mistaken.

Because `vmstarter.sh` → `postgres` isn't atomic, we can remove `vmstart.allowed` and run `pg_ctl stop` _after_ compute_ctl starts (again), but before postgres does, so pg_ctl does nothing, and we wait on flock until compute_ctl exits on its own, with the signal ignored.

sharnoff (commented):
> What I'd need confirmed before we merge this is that the k8s-level timeout exists

Looks like it does exist. VMs have a `TerminationGracePeriodSeconds` field (matching normal pods) that, for VMs, is 5s by default. The field gets propagated to the runner pod, so it'll SIGKILL after the duration. The control plane doesn't override that.

using flock does seem easier to reason about to me as well :)

neonvm/tools/vm-builder/main.go
problame: Yeah, noticed independently, but even with the fix 03ae60de02e49137bc35bd1eea23e669f97110c9 , the log messages don't show up in staging logs

problame: Gotta correct myself, it works: https://neonprod.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22xHHYY0dVz%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22xHHYY0dVz%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bneon_region%3D%5C%22us-east-2%5C%22%7D%20%7C%3D%20%60compute-royal-firefly-37505771%60%20%7C%3D%20%60workload%20shut%20down%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221695580354571%22,%22to%22:%221695580426866%22%7D%7D

problame: Going to retry a few times to make sure

neonvm/tools/vm-builder/main.go
problame: You’re correct, will try to model this and come up with a fix.

I think ultimately the only right way to do it is to teach compute_ctl to do proper shutdown on SIGTERM.

Then we’ll send SIGTERM to vmstarter.sh and have it propagate it to the workload.

problame: fixed in https://github.com/neondatabase/autoscaling/pull/525/commits/4e58bd38443217f4f093fd1cf19651f3b8846cbc , please resolve this conversation if you're convinced and open a new conversation if the fix introduces new problems / you find other race conditions.

problame:
Tested manually that with this PR (and #528 deployed in prod), the compute_ctl doesn't get restarted anymore.

So, I think if there aren't any race conditions left, this is good to go.

sharnoff (approved):
Looks good to me; don't see any other potential races/deadlocks :)

PR #527, opened 2023-09-24
neonvm/controller: Close QMP connection if Connect() fails

sharnoff:
Looking for ways that we could be leaving these connections open, due to certain errors that are consistently showing up. All the usages of QmpConnect have the corresponding `defer mon.Disconnect()` lines when `err != nil` (so they're probably not the cause), and this is theoretically a possible cause - although not totally sure.

cc @problame


problame (approved):
LGTM, let's merge & deploy AFTER #528 has been deployed & verified in staging.

cicdteam (approved):

PR #534, opened 2023-09-27
fixup #525: handle SIGKILLED postmaster process correctly

problame:
After rolling out #525 to prod, we noticed that when the postmaster process gets OOM killed, it fails to restart.

The `init` would try to respawn `vmstart` but it would get stuck trying to `flock` the file, before starting `vmstarter.sh` / the new postgres processes.

Investigation showed that we were inherting the file descriptor and flock for `vmstart.lock`  to the child process tree.
Including the `postgres: Local free space monitor`  process, which fails to stop when postmaster gets killed.
Once we'd manually kill it, the respawned `vmstart`'s blocked `flock` would get unblocked.

We never intended to inherit the vmstart.lock file descriptor or flock, so, this PR disables that behavior.
That makes the restart-after-OOM-kill work again.

But, reviewing the file cache code, there's actually a data corruption possiblity here, because after respawn, the old free space monitor process will still be running, able to interfere with the restarted postgres processes.

Created issue https://github.com/neondatabase/neon/issues/5405 for that.

### DIFF
--- a/neonvm/tools/vm-builder-generic/main.go
+++ b/neonvm/tools/vm-builder-generic/main.go
@@ -44,12 +44,14 @@ RUN set -e \
 		su-exec \
 		e2fsprogs-extra \
 		blkid \
+		flock \
 	&& mv /sbin/acpid         /neonvm/bin/ \
 	&& mv /sbin/udevd         /neonvm/bin/ \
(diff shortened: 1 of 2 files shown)


sharnoff:
Squashing+merging myself to get the changes out there sooner.

PR #535, opened 2023-09-29
util: Move patch.go into its own package

sharnoff:
Also, rename some bits so that it more closely aligns with github.com/evanphx/json-patch, so it's simpler for us to use in the control plane.


sharnoff:
The change in the control plane ended up going in a different direction, figured this PR is worthwhile anyways.

PR #540, opened 2023-10-04
Makefile: Check kubectl context before applying

### LINKED ISSUE #319: `make deploy` (and other kubectl targets) should check context is `kind-kind`, not staging/prod
ref https://neondb.slack.com/archives/C03TN5G758R/p1686067295384499?thread_ts=1686063160.407649&cid=C03TN5G758R

There should be a way to disable this if need be, but it should be abundantly clear about what's going on.

sharnoff:
The idea is that we *really* don't want people to accidentally `make deploy` to prod (or even staging), and the best way to do that is to check the context in `make deploy` et al itself.

Implemented as its own target so we don't have to copy-paste. Not sure if I missed any (reviewers: please check!)

Resolves #319.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,12 +248,16 @@ kernel: ## Build linux kernel.
 		--file neonvm/hack/Dockerfile.kernel-builder \
 		neonvm/hack
 
+.PHONY: check-local-context
+check-local-context: ## Asserts that the current kubectl context is pointing at k3d or kind, to avoid accidentally applying to prod
+	if [ "$$($(KUBECTL) config current-context)" != 'k3d-neonvm' ] && [ "$$($(KUBECTL) config current-context)" != 'kind-kind' ]; then echo "kubectl context is not pointing to local k3d or kind cluster (must be k3d-neonvm or kind-kind)"; fi
+
 .PHONY: install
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -248,12 +248,16 @@ kernel: ## Build linux kernel.
 		--file neonvm/hack/Dockerfile.kernel-builder \
 		neonvm/hack
 
+.PHONY: check-local-context
+check-local-context: ## Asserts that the current kubectl context is pointing at k3d or kind, to avoid accidentally applying to prod
+	if [ "$$($(KUBECTL) config current-context)" != 'k3d-neonvm' ] && [ "$$($(KUBECTL) config current-context)" != 'kind-kind' ]; then echo "kubectl context is not pointing to local k3d or kind cluster (must be k3d-neonvm or kind-kind)"; fi
```
bayandin, line 253: Should we add exit with an error in case of unexpected context? Or does `make` do it automagically?

bayandin (approved):
Nice!

Makefile
sharnoff: Oh, I completely missed that :sweat_smile:

I don't think it does, no.

PR #545, opened 2023-10-05
billing: Change idempotency key format

sharnoff:
Quoting a conversation about the current format:

> It’s quite big but mostly used only for deduplication purposes so that
> it can be any unique value.

And

> Pageservers use
> ```
> <timestamp with milliseconds>-<pageserver_id>-<number of event in the batch>
> ```

So the new format is

```
<RFC3339 timestamp with microseconds>-<pod name>-<number in batch>/<total batch size>
```


Daniel-ef (approved):

sharnoff:
Tested locally by adding agent's billing config and replacing `Send` implementation with just printing ­— idempotency keys seem to be generated as expected.

PR #547, opened 2023-10-06
neonvm: Add support for spec-defined sysctl settings

sharnoff:
Adds a new field, Guest.Settings, in case there's other tunables aside from sysctl that we may want to change later.

The motivation is so that we can test things like disabling memory overcommit as a feature flag in the control plane, rather than e.g. changing it the compute image globally.
Ref https://neondb.slack.com/archives/C03TN5G758R/p1696359675029629

Tested locally by adding some sample settings to vm-deploy.yaml:

```yaml
settings:
  sysctl:
    - 'vm.overcommit_ratio = 30'
    - 'vm.panic_on_oom = 1'
```


cicdteam:
LGTM

cicdteam (approved):

PR #561, opened 2023-10-12
release workflow: Add 'free disk space' step

sharnoff:
Latest release attempt has been hitting issues running out of disk partway through, so this should unblock that.

Long-term we probably want to switch this to our self-hosted runners but there's some known issues preventing that for now.

Ref https://neondb.slack.com/archives/C059ZC138NR/p1697120311648149?thread_ts=1697120074.180899

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,6 +25,8 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
+      - name: free disk space
+        uses: jlumbroso/free-disk-space@v1.3.0
       - name: git checkout
         uses: actions/checkout@v3
       - name: get version and git info


.github/workflows/release.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
+      - name: free disk space
+        uses: jlumbroso/free-disk-space@f68fdb76e2ea636224182cfb7377ff9a1708f9b8 # v1.3.0
```
bayandin, line 29: To not break self-hosted runners occasionally, I would add a condition like this (a bit ugly, but I didn't find a way to access `runs-on:` from `if:`):

```suggestion
    strategy:
      matrix:
        os: [ ubuntu-latest ]
    runs-on: ${{ matrix.os }}
    steps:
      - name: free disk space
        if: matrix.os == 'ubuntu-latest'
        uses: jlumbroso/free-disk-space@f68fdb76e2ea636224182cfb7377ff9a1708f9b8 # v1.3.0
```

bayandin (approved):
LGTM, just one optional comment

PR #564, opened 2023-10-16
agent/executor: Use a single update listener thread, spawn new ones per each request

sharnoff:
Follow-up change to #371 that I'd wanted to make at the time, but focused instead on finishing it, because this is really just a refactoring.

Marked WIP until full cleanup items are done.


sharnoff:
In theory this is a good change, but in a lot of ways it's just churn for no reason, because it's fine to have a lot of goroutines.

PR #571, opened 2023-10-19
vm-builder: Turn off --auto-discover-databases

ololobus:
It causes exporter connection to every database in the Postgres and it was disabled for pods a while ago, but I forgot that VMs have a different build path.

See this for details:
  https://github.com/neondatabase/cloud/commit/0a60057f3380e570b540cc823975e07400cd640c

Also remove `--exclude-databases` it seems to be bogus as well, see:
  https://github.com/prometheus-community/postgres_exporter/issues/588


vadim2404 (approved):

sharnoff:
Looks like downscale is failing because the vm-monitor is hitting internal errors. Unfortunately we don't have logs for the VMs themselves, but it's unlikely that it's failing because of this PR — each vm-monitor is pulled from neondatabase/neon's current main, so maybe some recent change or something.

I'll dig into it further.

ololobus:
OK, I'd avoid stacking PR on top of failing tests and it's not urgent, so let me know when it makes sense to re-run / safe to merge

sharnoff (approved):
Problem was fixed with neondatabase/neon#5606, all good to merge.

sharnoff:
Filed #577 to track the long term solution for moving postgres_exporter configuration into the control plane (among other things).

There's more immediate changes that could be made to make that simpler, if we anticipate needing to do this frequently or with urgency.

PR #598, opened 2023-10-31
ip acquirization fix

cicdteam:
This PR:

- avoid situation when `ipam.AcquireIP` return `<nil>` as result and `<nil>` as error to prevent controller crash.

### DIFF
--- a/neonvm/pkg/ipam/ipam.go
+++ b/neonvm/pkg/ipam/ipam.go
@@ -233,8 +233,17 @@ func (i *IPAM) acquireORrelease(ctx context.Context, vmName string, vmNamespace 
 	case <-leCtx.Done():
 		err = fmt.Errorf("context got timeout while waiting to become leader")
 	}
+	if err != nil {
+		return ip, err
+	}
 
 	wg.Wait()
+
+	// if ip.String() still <nil> then probably something wrong with leader election
+	if len(ip.String()) == 0 {
+		return ip, fmt.Errorf("something wrong, probably with leader election")
+	}
+
 	return ip, ipamerr
 }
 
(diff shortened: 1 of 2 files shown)


neonvm/pkg/ipam/ipam.go
```diff
@@ -233,8 +233,17 @@ func (i *IPAM) acquireORrelease(ctx context.Context, vmName string, vmNamespace
...
 		err = fmt.Errorf("context got timeout while waiting to become leader")
 	}
+	if err != nil {
+		return ip, err
+	}
 
 	wg.Wait()
+
+	// if ip.String() still <nil> then probably something wrong with leader election
+	if len(ip.String()) == 0 {
```
sharnoff, line 243: > len(...) == 0

typo? should be checking if `ip.String() == "<nil>"` or if `ip == nil` ?

see also: https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/net/ip.go;l=474-487

sharnoff (commented):
One comment, otherwise looks good

neonvm/pkg/ipam/ipam.go
cicdteam: `ip` variable is struct `net.IPNet` and it not comparable directly with `nil`. Also `ip.String()` could be empty string (`""`), I always prefer compare `len() == 0` to check string is nil or empty.

sharnoff: ah ok. the reason I say `"<nil>"` specifically is that it's what `ip.String()` returns if there's issues:

```
nn, m := networkNumberAndMask(n)
if nn == nil || m == nil {
    return "<nil>"
}
```

(see the linked source above)

Presumably there's other ways to check if it's valid though?

cicdteam: `return "<nil>"` what the... then yes, `len() == 0` is incorrect

sharnoff (approved):

PR #610, opened 2023-11-05
update in resources for neonvm controllers

cicdteam:
This PR:
- add more resources (requests and limits) to neonvm-controller
- add more limits to neonvm-vxlan-contrroller

this improve stability and speed when thousands VMs running.


cicdteam:
I'll merge without code review as no code here.

PR #612, opened 2023-11-05
agent/schedwatch: Fix deadlock

sharnoff:
This was noticed on staging dev-eu-west-1-zeta ([ref](https://neondb.slack.com/archives/C03TN5G758R/p1699208662336279)), and with some digging, realized that the following sequence of events could have caused the observed behavior:

1. Scheduler deploy causes >1 scheduler pod to be present in the cluster
2. New VM with autoscaling enabled appears (either existing or fresh)
3. Initial `agent.Runner` calls `WatchSchedulerUpdates`
4. `WatchSchedulerUpdates` calls `eventBroker.Subscribe()` but exits with error without calling `eventBroker.Unsubscribe()`.
5. When the next scheduler event comes along, the event broker blocks forever on trying to send the event to a channel that no one's listening to.

For now, this issue can be circumvented by restarting all autoscaler-agents after making a change to the scheduler deployment.

### DIFF
--- a/pkg/agent/schedwatch/trackcurrent.go
+++ b/pkg/agent/schedwatch/trackcurrent.go
@@ -65,6 +65,13 @@ func WatchSchedulerUpdates(
 	store *watch.Store[corev1.Pod],
 ) (SchedulerWatch, *SchedulerInfo, error) {
 	events := eventBroker.Subscribe(ctx)
+	finishOk := false
+	defer func() {
+		if !finishOk {
+			eventBroker.Unsubscribe(ctx, events)
+		}
+	}()
+
 	readyQueue := make(chan SchedulerInfo)
 	deleted := make(chan SchedulerInfo)
 	cmd := make(chan watchCmd)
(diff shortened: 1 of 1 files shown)


cicdteam:
@sharnoff 

may be change scheduler deployment from `RollingUpdate` to `Recreate` ? It will prevent multiply pods during rollout.

sharnoff:
I hadn't considered that - but, I'd be worried about the possible side-effects: If we switch to `Recreate`, then if anything goes wrong during deploy (e.g. [image pull takes a long time](https://neondb.slack.com/archives/C033QLM5P7D/p1699042860716079?thread_ts=1699041653.691329&cid=C033QLM5P7D)), we'd have no scheduler at all. Currently, the autoscaler-agents only care about scheduler pods that are `Running`, so the period of time with >1 scheduler is typically quite limited.

---

Going to merge because I think that's an orthogonal change to this one, but still curious for your thoughts.

PR #618, opened 2023-11-14
enable /proc/kallsyms, this should enable symbolified backtraces

problame:
After creating this PR, I triggered a kernel build: https://github.com/neondatabase/autoscaling/actions/runs/6862156705

It passed, so, the kernel built with the new config is already on DockerHub and will be used in new releases.

IIUC, to test this in staging, we need to release a new neonvm-controller.
So, let's get this merged and a new neonvm-controller deployed to staging.


sharnoff (approved):
tested this in dev-eu-west-1-zeta, seems to work ok — made some progress on neondatabase/neon#5678 from it.
ref https://neondb.slack.com/archives/C04KGFVUWUQ/p1699979549705169?thread_ts=1697642632.378239&cid=C04KGFVUWUQ

PR #625, opened 2023-11-16
neonvm: Add ability to append to kernel command line

problame:
Stacked atop #621

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -1165,6 +1165,10 @@ func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
 		})
 	}
 
+	if virtualmachine.Spec.Guest.AdditionalKernelCommandLine != nil {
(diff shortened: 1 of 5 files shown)


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1165,6 +1165,10 @@ func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
 		})
 	}
 
+	if virtualmachine.Spec.Guest.AdditionalKernelCommandLine != nil {
+		pod.Spec.Containers[0].Args = append(pod.Spec.Containers[0].Args, "-additionalKernelCommandLine="+*virtualmachine.Spec.Guest.AdditionalKernelCommandLine)
```
sharnoff, line 1169:
```suggestion
		pod.Spec.Containers[0].Args = append(pod.Spec.Containers[0].Args, fmt.Sprintf("-additionalKernelCommandLine=%s", *virtualmachine.Spec.Guest.AdditionalKernelCommandLine))
```

In general, `fmt.Sprint[f]` is faster than string addition because it uses a buffer pool, but string addition does not.

sharnoff (approved):

PR #628, opened 2023-11-17
neonvm/runner: treat guest-side disk flush requests as no-ops

problame:
The NeonVM k8s nodes are currently write-IOPS-bottlenecked.
We can mitigate that by using cheaper & faster storage than EBS.

But, also, we can avoid most synchronous writes by ignoring
fua/flush/sync requests. That's what `cache=unsafe` does.

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -565,8 +565,8 @@ func main() {
 	}
 
 	// disk details
-	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=none", rootDiskPath))
-	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=none", runtimeDiskPath))
+	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=unsafe", rootDiskPath))
+	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=unsafe", runtimeDiskPath))
 	for _, disk := range vmSpec.Disks {
 		switch {
 		case disk.EmptyDisk != nil:
@@ -579,7 +579,7 @@ func main() {
 			if disk.EmptyDisk.Discard {
 				discard = ",discard=unmap"
 			}
-			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=disk,cache=none%s", disk.Name, dPath, discard))
+			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=disk,cache=unsafe%s", disk.Name, dPath, discard))
 		case disk.ConfigMap != nil || disk.Secret != nil:
 			dPath := fmt.Sprintf("%s/%s.qcow2", mountedDiskPath, disk.Name)
 			mnt := fmt.Sprintf("/vm/mounts%s", disk.MountPath)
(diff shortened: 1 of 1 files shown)


hlinnaka (approved):
Makes sense. One tenant might now use up the host's memory for the cache though. But seems like a good idea overall.

Seems that libvirt discourages live migration of VMs with cache=unsafe: https://doc.opensuse.org/documentation/leap/virtualization/html/book-virtualization/cha-cachemodes.html#sec-cache-mode-live-migration. Any insight on why that is?

problame:
We're using live migration in staging already, let's not merge this until I've read up on the interaction of this PR with live migration; wasn't aware of it, good catch @hlinnaka

neonvm/runner/main.go
```diff
@@ -565,8 +565,8 @@ func main() {
 	}
 
 	// disk details
-	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=none", rootDiskPath))
-	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=none", runtimeDiskPath))
+	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=unsafe", rootDiskPath))
+	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=unsafe", runtimeDiskPath))
```
cicdteam, line 569: it's `media=cdrom`, so I think we can leave `cache=none`there but `unsafe` also Ok (as no writes there)

neonvm/runner/main.go
```diff
@@ -579,15 +579,15 @@ func main() {
...
+			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=disk,cache=unsafe%s", disk.Name, dPath, discard))
 		case disk.ConfigMap != nil || disk.Secret != nil:
 			dPath := fmt.Sprintf("%s/%s.qcow2", mountedDiskPath, disk.Name)
 			mnt := fmt.Sprintf("/vm/mounts%s", disk.MountPath)
 			logger.Info("creating iso9660 image", zap.String("diskPath", dPath), zap.String("diskName", disk.Name), zap.String("mountPath", mnt))
 			if err := createISO9660FromPath(logger, disk.Name, dPath, mnt); err != nil {
 				logger.Fatal("Failed to create ISO9660 image", zap.Error(err))
 			}
-			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=cdrom,cache=none", disk.Name, dPath))
+			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=cdrom,cache=unsafe", disk.Name, dPath))
```
cicdteam, line 590: the same `media=cdrom`

cicdteam:
> We're using live migration in staging already, let's not merge this until I've read up on the interaction of this PR with live migration; wasn't aware of it, good catch @hlinnaka

we are not using shared storage to store VM images, so `cache=unsafe` should not prevent live migration. But if root cause is high IOPS then may be need move from EBS gp3 volume back to instance store -  `m6id.metal` instances have 4x1900 NVMe disks.

problame:
> we are not using shared storage to store VM images, so cache=unsafe should not prevent live migration.

That is how I parsed the wording in post on the page that Heikki linked as well, but @sharnoff  parsed it differently. I'm planning to spend some time in the qemu source code to make sure, but, if you have direct practical experience with this, please weigh in once more.

> But if root cause is high IOPS then may be need move from EBS gp3 volume back to instance store - m6id.metal instances have 4x1900 NVMe disks.

That is the wrong mental approach, IMO.

NeonVM doesn't provide durable storage. Hence we shouldn't spend money on it.

problame:
🤔 The cause for the write operations by neonvm can only be local file cache writes or Postgres itself using external sorting, etc in some scratch files.
OTOH I can't really imagine why either of these would need to do fsync / issue flush requests to the disk.

sharnoff (approved):
Did some basic tests in dev-eu-west. Live migration is _generally_ not really working, with or without this PR (I _think_ due to lack of end-to-end networking support?) but this PR was able to survive a few migrations, so good enough IMO :)
If we hit issues later on with live migration, we can switch to a different solution to handle IO.

PR #623, opened 2023-11-14
neonvm: enable using bpftrace

problame:
Hopefully this is enough to make bpftrace work.

PR #630, opened 2023-11-17
switch 5.15.80 kernel to near-ubuntu-mainline config

problame:
This PR changes the 5.15.80 kernel config, following the recipe from the recent commits in https://github.com/neondatabase/autoscaling/pull/623 where I did the same for 6.5.11 with good results for the 0-tps-issue (neondatabase/neon#5678).

Let's see whether we can solve 0-tps-issue on the 5.15.80 kernel by using this config.

Recipe:

(In the docker container set up via the instructions I added in https://github.com/neondatabase/autoscaling/pull/622 )

```

extract kernel config from the linux-modules deb in ubuntu: https://kernel.ubuntu.com/mainline
open `make menuconfig` and save, without any changes
default hostname + =y instead of =m for isofs
sed -i 's/=m/=n/g' ./neonvm/hack/linux-config-6.5.11; then make menuconfig
uncheck "Automatically sign all modules"
uncheck 'Enable different security model' (SECURITY=n)
  this avoids module signing stuff and other complexity we don't
  care about inside neonvm, like selinux
set empty string for 'Additional X.509 keys for default system keyring'
set empty string for 'X.509 certificates to be preloaded into the system blacklist keyring'
```


problame:
kernel build & push to docker under same 5.15.80 tag: https://github.com/neondatabase/autoscaling/actions/runs/6903092257

cicdteam:
just small notice: now `vmlinuz` about 10Mb, before it was about 6Mb

sharnoff:
Yeah, I think this was just used for testing and not intended to be merged, but cc @problame to cofirm

problame:
The idea was to merge & ship this if it fixes the 0-tps issue on 5.15.80 (less risk of introducing other regressions).

A few MiBs of kernel don't matter right now (neither as storage nor DRAM footprint).

cicdteam:
@problame
I think we can close this, as we moved to 6.1.63 kernel (and it builds in ubuntu:23.10  environment)

PR #632, opened 2023-11-19
neonvm: Allow per-VM custom runner image

### LINKED ISSUE #629: ability to override neonvm-runner image

We just added the ability to override the kernel image for a single VM (see #621) in order to make it easier to quickly test changes on staging without redeploying the entire world.

It'd be useful to have a similar override neonvm-runner in a similar way, in order to test #628.

## Feature idea(s) / DoD

We should be able to override neonvm-runner image for a single VM.

## Implementation ideas

New field on `VirtualMachine`, `.spec.overrideRunnerImage`

sharnoff:
Similar to #621; resolves #629 (see there for more info).

One notable difference from #621 is that here, the new field is `.spec.runnerImage`, whereas there it was `.spec.guest.kernelImage` - i.e. this is on the spec, whereas the other was on the Guest.


sharnoff:
Merging this so we can test #628 on staging.

PR #633, opened 2023-11-19
Update to Rust 1.74

sharnoff:
Right now, this is only used to build vm-builder from `neondatabase/neon`, for the testing images.

I don't _think_ there's ordering requirements here with neondatabase/build#73 or neondatabase/neon#5873 (seems like no backwards-incompatible changes), but cc @arpad-m just in case.


arpad-m (approved):

arpad-m:
https://github.com/neondatabase/neon/pull/5873 is merged now

PR #636, opened 2023-11-20
update whereabouts CNI plugin to v0.6.2

cicdteam:
This PR fixes issue with leader election in whereabouts CNI plugin (see https://github.com/k8snetworkplumbingwg/whereabouts/pull/221)


nikitakalyanov:
How does this work? I see the fix is more than a year old, and it was already present in v0.6.1

sharnoff (approved):
+1 to **@﻿nikitakalyanov**'s question, but generally in favor of upgrading

cicdteam:
Yes, I checked the code and it is present in version tagged 0.6.1. Then, perhaps, the main reason of issue was different, but it is better to update it anyway.

PR #637, opened 2023-11-20
CI: Use the latest kernel in e2e tests

bayandin:
- Build VM Kernel for e2e tests (the naive approach without caches takes ~3m)
- Make `vm-kernel` job to trigger e2e with this freshly built kernel
- Allow to trigger e2e tests with specified kernel

Things to improve (in the future PRs):
- `e2e-tests:` Use cached kernel version if it has no changes
- `vm-kernel`: Do not push kernel image if e2e tests with it fails

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -4,6 +4,12 @@ on:
   push:
     branches:
       - main
+  workflow_call:
+    inputs:
+      kernel-image:
+        type: string
+        description: 'The kernel image to use for the VMs'
+        required: false
 
 jobs:
   e2e-tests:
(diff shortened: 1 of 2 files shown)


cicdteam (approved):

sharnoff (approved):
lgtm :)

One concern I still have is that we might accidentally do a release with a different kernel than intended

https://github.com/neondatabase/autoscaling/blob/25b9d3b46c78216fe4c8a1f1cc44bdba04fe1839/.github/workflows/release.yaml#L65-L70

but... that can be a separate PR; doesn't need to block merging this one.

sharnoff:
For historical purposes, afaict the motivation for this PR was https://neondb.slack.com/archives/C059ZC138NR/p1700487216332599?thread_ts=1700435685.425779&cid=C059ZC138NR

PR #644, opened 2023-11-27
WIP [neonvm]: use multus v4.0.2

cicdteam:


cicdteam:
Tested on staging and found it more buggy then version we currently use.

PR #647, opened 2023-11-29
agent/execbridge: Fix plugin approved resource metrics

sharnoff:
In short: Regardless what the response was, we recorded metrics as if the plugin had approved everything up to the target, which was not necessarily true. It's more correct to use the actual response.


shayanh (approved):
lgtm

Issue #555, opened 2023-10-10
agent→plugin requests should include agent's perceived last permit


sharnoff:
ref https://github.com/neondatabase/autoscaling/pull/371#issuecomment-1752110131

> Requests to the scheduler plugin must include what the agent believes the last permit was. It's always possible for the scheduler to get killed and immediately restart, without the agent believing there was any disconnect, which could lead to unintentional overcommitting of resources from the `Buffer` values if too many agents request upscaling on the first request to the scheduler. (This would also remove the need for separate "informative" requests.)

PR #649, opened 2023-11-30
agent,plugin: requests include agent's perceived last permit

shayanh:
- Added LastPermit to AgentRequest
- Scheduler plugin updates its pod and node reserved values based on agent's last permit. This allows a newly running scheduler to recover the previous scheduler's state.
- There are some subtleties in the agent <-> plugin protocol that I discussed on them in the comments.

resolves https://github.com/neondatabase/autoscaling/issues/555


shayanh:
addresses #555

pkg/plugin/trans.go
```diff
@@ -21,8 +21,12 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
+// resourceTransition maintains the current state of its resource and handles the transition
+// into a new state.
 type resourceTransition[T constraints.Unsigned] struct {
-	node    *nodeResourceState[T]
+	// node represents the current state of the node
+	node *nodeResourceState[T]
+	// oldNode is the old state of the node, which we just keep for convenience
```
sharnoff, line 29: let's split this into a separate PR? afaict, the changes aren't related to the rest of this PR, and don't need to be held up on the review on the rest of it

pkg/plugin/trans.go
```diff
@@ -80,6 +84,40 @@ func (s verdictSet) MarshalLogObject(enc zapcore.ObjectEncoder) error {
...
+// from the Buffer values if too many agents request upscaling on the first
+// request to the scheduler.
+func (r resourceTransition[T]) handleLastPermit(lastPermit T) (verdict string) {
+	if lastPermit <= r.pod.Reserved {
+		r.node.Reserved -= r.pod.Reserved - lastPermit
+		r.pod.Reserved = lastPermit
+		verdict = fmt.Sprintf(
+			"pod reserved (%v -> %v), node reserved (%v -> %v)",
+			r.oldPod.reserved, r.pod.Reserved, r.oldNode.reserved, r.node.Reserved,
+		)
```
sharnoff, line 103: IIUC, we also need to update `r.{node,pod}.Buffer`?

pkg/plugin/trans.go
```diff
@@ -80,6 +84,40 @@ func (s verdictSet) MarshalLogObject(enc zapcore.ObjectEncoder) error {
...
+		//   2. scheduler dies
+		//   3. vm bounds decrease (the max value is `b` and we have `b < a`).
+		//   4. new scheduler reads the cluster state and sets up the buffer values
+		//   => agent’s last permit is greater (`a`) than plugin’s reserved value (`b`)
+		// This might also happen in case of processing a stale request from an agent
+
+		verdict = fmt.Sprintf(
+			"unexpected last permit, no changes: last permit (%v) is greater than pod reserved (%v)",
+			lastPermit, r.pod.Reserved,
+		)
```
sharnoff, line 116: I'd be worried that doing nothing here might leave us in a permanently degraded state. Thoughts?

Or, is this mostly "revert to previous behavior" because we call `handleRequested` afterwards anyways?

pkg/plugin/run.go
```diff
@@ -261,11 +262,23 @@ func (e *AutoscaleEnforcer) handleResources(
...
+			"Handled last permit info from pod",
+			zap.Object("verdict", verdictSet{
+				cpu: vCPUVerdict,
+				mem: memVerdict,
+			}),
+		)
+	}
+
 	vCPUVerdict := vCPUTransition.handleRequested(req.VCPU, startingMigration, !supportsFractionalCPU)
 	memVerdict := memTransition.handleRequested(req.Mem, startingMigration, false)
```
sharnoff, line 278: reusing the `resourceTransition` across multiple function calls smells unsound (iirc, the methods are implemented with the (undocumented) assumption that at the start, the `oldNode` fields are equal to those in `node`, and also for `oldPod` and `pod`).

Was there a particular reason to reuse the `resourceTransition`s here?

---

aside: I think it probably makes sense to split out the `oldNode` and `oldPod` fields into values that are only created & used inside the methods, rather than stored alongside the mutable fields. That would make this sound, and probably easier to understand too.

sharnoff (commented):
broadly looks good! some nits on the specifics

pkg/plugin/run.go
shayanh: Makes sense. I'm going to rename `resourceTransition` to `resourceTransitioner` because it not an object that represents a transition, but actually it's performs transition on a resource state. I'll remove `oldPod` and `oldNode` from here. I'm just wondering why did you made those inline structs and just didn't make a copy of `nodeResourceState` and `podResourceState`?

pkg/plugin/trans.go
shayanh: I had the latter in mind, though we might later call this without having `handleRequested` be called afterwards. I don't think updating the reserved value in this case would be right, cause it's not the most recent value anymore.

Anyway, now I think it's better to enforce that `handleRequested` will be called later. I'm going to remove this function and make it part of `handleRequested`.

shayanh:
@sharnoff this is ready for review

PR #651, opened 2023-11-30
release workflow: Build kernel instead of downloading

sharnoff:
ref https://github.com/neondatabase/autoscaling/pull/637#pullrequestreview-1750637301. This would also have prevented the need for #643.

Broadly: It's ok to build the kernel now because we're using the big runner for releases, and the guarantees around actually releasing what's in main is more than worthwhile.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,7 +15,6 @@ env:
   IMG_VXLAN:                "neondatabase/neonvm-vxlan-controller"
   IMG_RUNNER:               "neondatabase/neonvm-runner"
   VM_KERNEL_IMAGE:          "neondatabase/vm-kernel"
-  VM_KERNEL_VERSION:        "6.1.63"
 
   CLUSTER_AUTOSCALER_IMAGE: "neondatabase/cluster-autoscaler-neonvm"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -62,12 +61,9 @@ jobs:
...
 
-      - name: load vm kernel
-        run: |
-          docker pull --quiet ${{ env.VM_KERNEL_IMAGE }}:${{ env.VM_KERNEL_VERSION }}
-          ID=$(docker create ${{ env.VM_KERNEL_IMAGE }}:${{ env.VM_KERNEL_VERSION }} true)
-          docker cp ${ID}:/vmlinuz neonvm/hack/vmlinuz
-          docker rm -f ${ID}
+      - name: build vm kernel
+        run:
+          make kernel
```
bayandin, line 66: I would prefer to call the `vm-image` workflow from the release workflow (this requires adding `workflow_call` to `vm-image`). Then we have vm-kernel image as an artifact (published to docker hub, we can even add a custom release tag to the image). 
What do you think? (I'm happy to make all workflows modifications for this)

sharnoff: > What do you think?

Yeah, that sounds like a good idea :)

+1 to adding a custom release tag, so that we don't accidentally end up with another image sneaking in.

> I'm happy to make all workflows modifications for this

that'd be wonderful, thanks :)

bayandin: Experimenting in this PR https://github.com/neondatabase/autoscaling/pull/679.

bayandin:
Superseded by https://github.com/neondatabase/autoscaling/pull/679

PR #652, opened 2023-12-01
Only keep one kernel version

sharnoff:
In practice, we only use a single kernel version at a time. Previous versions can be found with git history.


neonvm/hack/Dockerfile.kernel-builder
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:23.10 AS build-deps
 
-ARG KERNEL_VERSION
+ARG KERNEL_VERSION=6.1.63
```
bayandin, line 3: Is it required to re-generate linux-config for each kernel update, major update? If so, then it'll be convenient to keep linux-config file version-suffixed, just to avoid possible missynchronisation.

sharnoff: Ah yeah, makes sense. I'll keep it version-suffixed.

bayandin (approved):

PR #654, opened 2023-12-01
plugin: Remove config nodeOverrides

sharnoff:
Also renames `nodeDefaults` to `nodeConfig`.

This was a very old feature that was never used, and would have needed to be significantly changed to be useful anyways (checking by node name is not very useful; instead we'd want to use labels, if anything).

Split out from #653.


shayanh (approved):
lgtm

pkg/plugin/config.go
```diff
@@ -22,12 +22,7 @@ import (
 
 type Config struct {
 	// NodeDefaults is the default node configuration used when not overridden
```
shayanh, line 24: Update the comment

PR #657, opened 2023-12-01
plugin: Move `(*AutoscaleEnforcer).checkSchedulerName()` outside of locks

sharnoff:
I _think_ this was a holdover from when we still had live config reloading, which was removed in #87.

Split off from #653.


shayanh (commented):
I'm not sure how I can make sure we're not causing any problem by moving this out of lock.

sharnoff:
> I'm not sure how I can make sure we're not causing any problem by moving this out of lock.

Opened #668 to document the existing locking behavior.

shayanh (approved):
lgtm. feel free to merge

PR #658, opened 2023-12-01
plugin: Remove `(*nodeState).totalReservable{CPU,MemSlots}`

sharnoff:
Since #399, "total reservable CPU/memory" has just been equal to the value of nodeResourceState.Total, so these methods haven't been doing anything.

Split out from #653.


shayanh (approved):
lgtm

Issue #660, opened 2023-12-02
Make CI faster


shayanh:
## Problem description / Motivation

As of now, the entire CI process [takes about 20 minutes](https://github.com/neondatabase/autoscaling/actions/runs/7066178893/job/19237582034?pr=649), which is not the worst thing in the world but still it's long and annoying. 

## Feature idea(s) / DoD

As a starting point, a DoD can be decreasing CI time to less than 10m.

## Implementation ideas

Most expensive parts of the pipeline are `make kernel` and `make example-vms`, which I think we can use pre-built image and avoid building them every time.

bayandin:
https://github.com/neondatabase/autoscaling/pull/661 brings e2e-test workflow to around 10m.
https://github.com/neondatabase/autoscaling/pull/642 shaves off another 1.5m

sharnoff:
Since merging #748, _at first glance_ it appears e2e-test times are around 8-9 minutes (at least, when there's no contention for runners).

@shayanh wdyt, is this done? should we wait a little longer to validate e2e-test times are still low?

shayanh:
<10 is nice. Great progress. I close the issue. We can re-open if the times go up again.

PR #662, opened 2023-12-02
CI: use self-hosted runner for vm-kernel workflow

bayandin:
Switch from GitHub's `gha-ubuntu-22.04-64cores` runner to self-hosted `large` runner (32 cores / RAM 128 GiB).
- Before: 2m 21s total (1m 38s 'build-linux kernel' step) [a link](https://github.com/neondatabase/autoscaling/actions/runs/7014316396/job/19081849534)
- After: 1m 85s total (1m 43s 'build-linux kernel' step) [a link](https://github.com/neondatabase/autoscaling/actions/runs/7072729343/job/19251884151)

Also, fix `env. VM_KERNEL_VERSION` usage, because we don't have this variable anymore, get it from parsing `neonvm/hack/linux-config-*` (maybe we should get it from another place?)

Part of https://github.com/neondatabase/autoscaling/issues/660

### DIFF
--- a/.github/workflows/vm-kernel.yaml
+++ b/.github/workflows/vm-kernel.yaml
@@ -26,29 +26,22 @@ jobs:
       - name: git checkout
         uses: actions/checkout@v3
 
+      # Use custom DOCKER_CONFIG directory to avoid conflicts with default settings
+      # The default value is ~/.docker
       - name: set custom docker config directory
         run: |
           mkdir -p .docker-custom
           echo DOCKER_CONFIG=$(pwd)/.docker-custom >> $GITHUB_ENV
-      - name: login to docker hub
-        run: |
-          DOCKERHUB_AUTH=$(echo -n "${{ secrets.NEON_DOCKERHUB_USERNAME }}:${{ secrets.NEON_DOCKERHUB_PASSWORD }}" | base64)
-          echo "::add-mask::${DOCKERHUB_AUTH}"
-
-          cat <<-EOF > .docker-custom/config.json
-            {
-              "auths": {
-                "https://index.docker.io/v1/": {
-                  "auth": "${DOCKERHUB_AUTH}"
-                }
-              }
-            }
-          EOF
 
       - name: docker - install qemu
         uses: docker/setup-qemu-action@v2
       - name: docker - setup buildx
         uses: docker/setup-buildx-action@v2
+      - name: login to docker hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.NEON_DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/vm-kernel.yaml
```diff
@@ -21,22 +21,42 @@ jobs:
...
-        uses: docker/login-action@v2
-        with:
-          username: ${{ secrets.NEON_DOCKERHUB_USERNAME }}
-          password: ${{ secrets.NEON_DOCKERHUB_PASSWORD }}
+
+      - name: get kernel version
+        id: get-kernel-version
+        run: |
+          linux_config=$(ls neonvm/hack/linux-config-*)  # returns something like "neonvm/hack/linux-config-6.1.63"
+          kernel_version=${linux_config##*-}             # returns something like "6.1.63"
```
sharnoff, line 57: neat! Opened #663 to use this same technique for `make kernel` so that the filename can be an authoritative source for the kernel version we're using.

re: "maybe we should get it from another place" - yeah, I'm not totally sure either, but this can work for now :)

.github/workflows/vm-kernel.yaml
```diff
@@ -21,22 +21,42 @@ jobs:
...
+
+          cat <<-EOF > .docker-custom/config.json
+            {
+              "auths": {
+                "https://index.docker.io/v1/": {
+                  "auth": "${DOCKERHUB_AUTH}"
+                }
+              }
+            }
+          EOF
```
sharnoff, line 46: I'm not too familiar with github actions semantics, so at first glance, I'd be worried about leaking credentials here - but as long as it doesn't, I have no objections

sharnoff (approved):
nice! would this docker config method could also work for the e2e tests?

.github/workflows/vm-kernel.yaml
areyou1or0: @bayandin I am not worried about the logs file however why do we save the credentials in a config file? Can't we just set it as an environment variable to avoid saving it in a file?

bayandin: I didn't find a way to use an environment variable for docker authorisation. However, I found that we don't need to configure this file manually, we can use `docker/login-action@v2` (as we used before)

A workflow run: https://github.com/neondatabase/autoscaling/actions/runs/7129968930

bayandin:
> nice! would this docker config method could also work for the e2e tests?

Yep, with custom `DOCKER_CONFIG` we can do it for e2e tests as well

areyou1or0:
@bayandin ? https://github.com/neondatabase/autoscaling/pull/662#discussion_r1418765740

PR #663, opened 2023-12-03
Makefile: Set KERNEL_VERSION from linux-config filename

sharnoff:
Copying the method used from #662.

Also, remove the default value of the ARG from the top of the Dockerfile, so that the `linux-config-*` filename is the authoritative source for the kernel version we're using.

See also #652, which this PR _partially_ reverts in order to move the source of the kernel version back out of the Dockerfile.

---

Note for review: Not entirely sure that this is the right way to do it, but after seeing the technique in #662, I figured it might make sense to go all-in on that way. At some point in the future, we _might_ want to switch from `${linux_config##*-}` to `${linux_config#*linux-config-}`, but it probably doesn't matter _too_ too much.

### DIFF
--- a/.github/workflows/vm-kernel.yaml
+++ b/.github/workflows/vm-kernel.yaml
@@ -55,7 +55,7 @@ jobs:
         id: build-linux-kernel
         uses: docker/build-push-action@v3
         with:
-          build-args: KERNEL_VERSION=6.1.63 # Temporary: Will be overridden by #662.
+          build-args: KERNEL_VERSION=${{ steps.get-kernel-version.outputs.VM_KERNEL_VERSION }}
           context: neonvm/hack
           platforms: linux/amd64
           # Push only if this is a scheduled run or if the workflow_dispatch input push is set to true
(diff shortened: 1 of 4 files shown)


.github/workflows/vm-kernel.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
         id: build-linux-kernel
         uses: docker/build-push-action@v3
         with:
+          build-args: KERNEL_VERSION=6.1.63 # Temporary: Will be overridden by #662.
```
sharnoff, line 45: @bayandin IIUC this would kind of conflict with #662 - just want to make sure you're aware. Happy to handle it in this PR or that one, whichever one merges second.

bayandin: After rebasing on main you can use 

```suggestion
          build-args: KERNEL_VERSION=${{ steps.get-kernel-version.outputs.VM_KERNEL_VERSION }}
```

sharnoff: Done, thanks :) c544b951c2d707a78a9fecb6a116479d3aad84d1

PR #635, opened 2023-11-20
CI: do not delete e2e pods on CI to catch logs

bayandin:
If `CI` env variable is set (it's set on [GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)), then add [`--skip-delete`](https://kuttl.dev/docs/cli.html#flags) flag to out `kuttl` test command. This should keep resources and allow us to get logs.

Afterwards, everything will be deleted in the `Cleanup` step:
https://github.com/neondatabase/autoscaling/blob/4c9defeb921bfec14293a1498e7debff9dd287e3/.github/workflows/e2e-test.yaml#L106-L108

Closes https://github.com/neondatabase/autoscaling/issues/634

Issue #634, opened 2023-11-20
e2e tests should dump logs for deleted containers


sharnoff:
## Problem description / Motivation

If there's problems starting VMs ([ref](https://neondb.slack.com/archives/C059ZC138NR/p1700435685425779)), the pods all get deleted by the time the e2e tests are done, which means our current "dump logs for existing pods" doesn't catch them.

## DoD

The current interface is good enough, we just need to _also_ have logs from pods that have since been deleted.

## Implementation ideas

Logs exist on disk for a little while after the pods are deleted. _Maybe_ we can periodically collect those logs in the background, and then display them all at once when tests are done.

Not sure if we need different solutions for kind vs k3d, but should be manageable either way.

```[tasklist]
# Tasks
- [ ] https://github.com/neondatabase/autoscaling/pull/635
- [ ] https://github.com/neondatabase/autoscaling/pull/664
```

sharnoff:
Closing, all tasks have been completed

PR #664, opened 2023-12-04
neonvm/controller: Don't delete runner pods in CI

sharnoff:
This is implemented by adding a build tag 'nodelete', which disables deletion of VM runner pods - at least, while the VM or VMM object exists.

We *could* have actually kept the runner pods around even if the VM/VMM is deleted, but there's already ways to keep the objects around after e2e tests end (see #635), and this way the results would be a little less catastrophic if we accidentally released it.

The 'nodelete' tag is enabled by the Makefile based on the `E2E` env var.

Part of #634.

---

Note for review: I was going to use the `CI` env var like #635, but it's set on all workflows and the point is to build images with different config for e2e tests only. I'm not sure that the new `E2E` env var is the right name; open to suggestions there or alternate solutions (maybe something to do with "debug" / "test"?)

Tested both ways locally, but not super thoroughly — at first glance, it appears to be working as intended.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -17,6 +17,9 @@ on:
         description: 'The kernel image to use for the VMs'
         required: false
 
+env:
+  E2E: "true"
+
 jobs:
   e2e-tests:
     strategy:
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -18,7 +18,7 @@ on:
         required: false
 
 env:
-  E2E: "true"
+  PRESERVE_RUNNER_PODS: "true"
 
 jobs:
   e2e-tests:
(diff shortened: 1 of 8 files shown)


cicdteam:
@sharnoff 

>The 'nodelete' tag is enabled by the Makefile based on the E2E env var.

Does this mean that if I manually run `make e2e` the pods will not be removed? but `kuttl` will remove namespaces anyway (with pods inside), so I suggest relying on the `CI` variable rather than introducing a new `E2E` variable.

bayandin:
> Note for review: I was going to use the CI env var like https://github.com/neondatabase/autoscaling/pull/635, but it's set on all workflows and the point is to build images with different config for e2e tests only.

Good catch! I didn't think about it.

> I'm not sure that the new E2E env var is the right name; open to suggestions there or alternate solutions (maybe something to do with "debug" / "test"?)

I would go with something more specific, `PRESERVE_RUNNER_PODS` or something like this

cicdteam (approved):

PR #666, opened 2023-12-04
plugin: Unify reserve and unreserve logic

sharnoff:
We had the same thing implemented in a few places, with some TODOs to unify them eventually - so, here it is.

Summary of changes:

- Merge all handling of pod/VM "start" / "reserve" logic to go through `(*AutoscaleEnforcer).reserveResources()`
  - Combined `handlePodStarted`/`handleVMStarted` -> `handleStarted`
- Merge all handling of pod/VM "deletion" / "unreserve" logic to go through `(*AutoscaleEnforcer).unreserveResources()`
- Pass around `*corev1.Pod` objects 

TODO:

- [x] Add comments explaining new functions and their parameters
- [x] Add comments on substantially modified functions that didn't already have them (like `handlePodStarted`)

### DIFF
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -569,91 +569,185 @@ func (e *AutoscaleEnforcer) handleNodeDeletion(logger *zap.Logger, nodeName stri
 	logger.Info("Deleted node")
 }
 
-func (e *AutoscaleEnforcer) handlePodStarted(logger *zap.Logger, pod *corev1.Pod) {
-	podName := util.GetNamespacedName(pod)
+func (e *AutoscaleEnforcer) handleStarted(logger *zap.Logger, pod *corev1.Pod) {
 	nodeName := pod.Spec.NodeName
 
 	logger = logger.With(
 		zap.String("action", "Pod started"),
-		zap.Object("pod", podName),
 		zap.String("node", nodeName),
+		util.PodNameFields(pod),
 	)
-
-	if pod.Spec.SchedulerName == e.state.conf.SchedulerName {
-		logger.Info("Got non-VM pod start event for pod assigned to this scheduler; nothing to do")
-		return
+	if migrationName := util.TryPodOwnerVirtualMachineMigration(pod); migrationName != nil {
+		logger = logger.With(zap.Object("virtualmachinemigration", *migrationName))
 	}
 
-	logger.Info("Handling non-VM pod start event")
+	logger.Info("Handling Pod start event")
 
-	podResources := extractPodResources(pod)
+	_, _, _ = e.reserveResources(context.TODO(), logger, pod, "Pod started", false)
+}
+
+func (e *AutoscaleEnforcer) reserveResources(
+	ctx context.Context,
+	logger *zap.Logger,
+	pod *corev1.Pod,
+	action string,
+	allowDeny bool,
+) (ok bool, _ *verdictSet, _ error) {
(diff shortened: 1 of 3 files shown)


sharnoff:
Goal: include this in the next release, after #653. That change was relatively mechanical, this is much higher risk. Want to evenly distribute risk across releases.

pkg/plugin/state.go
```diff
@@ -569,98 +569,205 @@ func (e *AutoscaleEnforcer) handleNodeDeletion(logger *zap.Logger, nodeName stri
...
+	logFields, kind, migrating, verdict := e.unreserveResources(logger, podName)
+
+	logger.With(logFields...).Info(
+		fmt.Sprintf("Deleted %s Pod", kind),
+		zap.Bool("migrating", migrating),
+		zap.Object("verdict", verdict),
+	)
+}
+
+func (e *AutoscaleEnforcer) unreserveResources(
```
shayanh, line 767: As an outsider to me reserve and unreserve methods are symmetric operations. I don't get why here their inputs and outputs are so different.

pkg/plugin/state.go
```diff
@@ -569,98 +569,205 @@ func (e *AutoscaleEnforcer) handleNodeDeletion(logger *zap.Logger, nodeName stri
...
+	logger *zap.Logger,
+	podName util.NamespacedName,
+) (_ []zap.Field, kind string, migrating bool, _ verdictSet) {
```
shayanh, line 770: To me having partially named return arguments looks slightly ugly. I'd rather suggest naming all or none.

pkg/plugin/state.go
```diff
@@ -569,98 +569,205 @@ func (e *AutoscaleEnforcer) handleNodeDeletion(logger *zap.Logger, nodeName stri
...
+	_, _, _ = e.reserveResources(context.TODO(), logger, pod, "Pod started", false)
+}
+
+func (e *AutoscaleEnforcer) reserveResources(
+	ctx context.Context,
+	logger *zap.Logger,
+	pod *corev1.Pod,
+	action string,
+	allowDeny bool,
+) (ok bool, _ *verdictSet, _ error) {
```
shayanh, line 595: Why we return both an error and a boolean here?

pkg/plugin/state.go
shayanh: Returning log fields here looks sorta hacky.

pkg/plugin/state.go
sharnoff: Added some comments in f4b9d26cd4a19be1baec6fdf0899f04c79d2b9ac — wdyt?

pkg/plugin/state.go
sharnoff: Elaborated in a comment in f4b9d26cd4a19be1baec6fdf0899f04c79d2b9ac — wdyt?

pkg/plugin/state.go
sharnoff: I'd left a comment on the function. I agree parts of this are a little ugly/hacky. I don't see a cleaner solution tbh, although maybe I'm missing something.

wdyt? 

(In terms of named/unnamed returns, IMO having a mix allows documenting some fields while making it clear that you shouldn't use a naked return. Here, I think it's necessary to name `kind` and `migrating`.)

shayanh: Sounds good.

shayanh (approved):

PR #670, opened 2023-12-06
neonvm: added pprof endpoint

### LINKED ISSUE #608: neonvm-controller should have pprof endpoint

While trying to debug neondatabase/neon#5678, we noticed some potential stalling in neonvm-controller, but had to `kill -6` from the kubernetes node in order to get stack traces.

We should expose pprof information like other components so that this is easier (and doesn't require restart)

## Implementation ideas

`pkg/util/pprof.go` has a utility function to handle some of this. May be useful here.

shayanh:
added pprof endpoint for neonvm

fixes #608

### DIFF
--- a/neonvm/main.go
+++ b/neonvm/main.go
@@ -17,8 +17,12 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -35,6 +39,8 @@ import (
 
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 	"github.com/neondatabase/autoscaling/neonvm/controllers"
+	"github.com/neondatabase/autoscaling/pkg/util"
+	"github.com/tychoish/fun/srv"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -48,6 +54,11 @@ func init() {
 
 	utilruntime.Must(vmv1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
+}
+
+func logPanic(err error, msg string, keysAndValues ...any) {
+	setupLog.Error(err, msg, keysAndValues...)
+	panic(msg)
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


neonvm/main.go
```diff
@@ -50,6 +56,11 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func logPanic(err error, msg string, keysAndValues ...any) {
+	setupLog.Error(err, msg, keysAndValues...)
+	panic(msg)
+}
+
```
sharnoff, line 63: Can this be split into a separate PR?

sharnoff (commented):
broadly looks good - one note about not bundling in unrelated changes.

Also: there are a few different "neonvm" binaries deployed (neonvm-controller, neonvm-vxlan-controller, neonvm-runner). Could you clarify in the PR title that this PR only affects the controller?

neonvm/main.go
shayanh: I see your point and it's good to separate different changes in different PRs, but at the same time having a different PR for every small change is too much. I'll revise the PR message to elaborate on logging here.

neonvm/main.go
```diff
@@ -95,48 +106,58 @@ func main() {
...
+	if err := orca.Add(srv.HTTP("scheduler-pprof", time.Second, util.MakePPROF("0.0.0.0:7777"))); err != nil {
+		logPanic(err, "Failed to add pprof service")
 	}
 
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
+	if err := mgr.Start(ctx); err != nil {
+		logPanic(err, "problem running manager")
```
sharnoff, line 161: These lines will result in artificially increased severity for problems encountered at runtime.

For example, [errors from failed leadership election](https://neondb.slack.com/archives/C03F5SM1N02/p1701819152996709?thread_ts=1701817648.282449&cid=C03F5SM1N02) would result in panicking here, which - when someone reads the logs - are likely to be interpreted as a more significant problem than actually exists (especially because the stack trace of the panic does not add useful information about the origin of the error).

A more complete solution would probably involve something like how the scheduler sets up a "real" main that holds the deferred functions and is allowed return error:

https://github.com/neondatabase/autoscaling/blob/9b1e55c7cb55f503223d9817930a387e8622f297/cmd/autoscale-scheduler/main.go#L25-L38

sharnoff (changes requested):

neonvm/main.go
shayanh: Resolved in c89494ff5ffc7f996e415aaa283d36350f245531. I decided to keep the general structure of main.go intact since it's generated by kubebuilder and has some markers for future code insertions. So, I didn't feel good changing it.

sharnoff (approved):
looks like a clean solution - nice!

PR #672, opened 2023-12-06
util/watch: Remove resourceVersion from List calls

sharnoff:
The added comment has some detail, but in short: The synchronization guarantees we *intended* to provide with Relist() were not actually guaranteed, because we needed to not set the ResourceVersion on list requests in order to get the most recent data (rather than just "not older than the resource version").

Noticed this while trying to provide justification for some documentation here: https://github.com/neondatabase/autoscaling/pull/669#discussion_r1418002027

---

Notes for review: It seems like sending `AllowWatchBookmarks = true` on List requests was already working fine, so _in theory_ we could have just temporarily unset the `ResourceVersion`, but I figured it'd make things clearer to just have two separate sets of options.

In the future, we may want to continue to set `ResourceVersion` for List calls if there wasn't an explicit relist request, for the sake of efficiency. I don't know if it actually matters.

### DIFF
--- a/pkg/util/watch/watch.go
+++ b/pkg/util/watch/watch.go
@@ -113,7 +113,7 @@ func Watch[C Client[L], L metav1.ListMetaAccessor, T any, P Object[T]](
 	config Config,
 	accessors Accessors[L, T],
 	mode InitMode,
-	opts metav1.ListOptions,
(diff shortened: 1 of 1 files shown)


shayanh (approved):
good catch

PR #597, opened 2023-10-31
CI: add k8s logs from e2e-tests to GitHub Summary

bayandin:
This PR attaches logs from `neonvm-system` and `kuttl-test-*` namespaces to GitHub Step Summary. 
Here's an example: https://github.com/neondatabase/autoscaling/actions/runs/6960120849?pr=597#summary-18938907450

GitHub Step Summary is limited by 1Mb, so if we have more logs then we won't be able to display it. I guess we can truncate it by 1MB, but it can be done later, for `neonvm-system` and `kuttl-test-*` namespaces 1MB is enough.

All the information is still available in the regular job output (in addition to logs and events from other namespaces)

Note that currently `kuttl-test-*` namespace got deleted automatically by `kuttl`, this will be changed in https://github.com/neondatabase/autoscaling/pull/635

PR #674, opened 2023-12-07
CI/e2e: Add some kube-system logs to GH summary

sharnoff:
Realized after the fact that these were missing from #597 - the autoscaler-agents and autoscale-scheduler are both in kube-system, so weren't included before.

The logs from those components are a bit more verbose. _Hopefully_ this doesn't cause us to go over the 1MB limit.

---

Notes for review: In order to support "only include some pods in kube-system" I split `tee_if_needed` into a value for the namespace (`tee_ns_if_needed`) and one for the pod (`tee_pod_if_needed`).

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -126,24 +126,31 @@ jobs:
 
           namespaces=$(kubectl get namespaces -o jsonpath='{.items[*].metadata.name}')
           for namespace in $namespaces; do
-            if [[ "$namespace" == "neonvm-system" ]] || [[ "$namespace" == kuttl-test-* ]]; then
-              tee_if_needed=$GITHUB_STEP_SUMMARY
+            if [[ "$namespace" == "neonvm-system" ]] || [[ "$namespace" == kuttl-test-* ]] || [[ "$namespace" == "kube-system" ]]; then
+              tee_ns_if_needed=$GITHUB_STEP_SUMMARY
             else
-              tee_if_needed=/dev/null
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -120,24 +120,31 @@ jobs:
...
             fi
 
             {
               echo "<details>"
               echo "<summary>Namespace=$namespace</summary>"
-            } | tee -a $tee_if_needed
+            } | tee -a $tee_ns_if_needed
 
             pods=$(kubectl get pods -n $namespace -o jsonpath='{.items[*].metadata.name}')
             for pod in $pods; do
```
bayandin, line 135: It looks like we get all logs & events inside this for-loop. Let's move the `if` under it, and then we can have all conditions(and it won't be required to split `tee_if_needed`).

It'll require dropping the outer scoping by namespaces, but it's probably fine, we still mention namespaces for each entry.

bayandin (commented):
it seems something went wrong, and no logs available in the Summary https://github.com/neondatabase/autoscaling/actions/runs/7123993810
(try to expand any namespace)

bayandin:
And, a bit off-topic, but still related. 
I'm thinking about moving logs from GitHub Summary to a GitHub comment (something as we have for `neondatabase/neon` PRs). 
The limit for GitHub Summary is 1 MB, and for GitHub Comment is 16 MB.

I haven't planned this yet, but I would like to do it at some point.

sharnoff:
Closing because no progress planned on this for now. We can reopen or recreate if/when it's useful.

Issue #676, opened 2023-12-07
local k8s clusters (kind and k3d) should have $USER in name


cicdteam:
## Problem description / Motivation

Currently, `make kind-setup` creates a cluster named `kind-kind`, and `make k3d-setup` creates `k3d-neonvm`. On shared servers will be a mess if multiple users run local clusters.

## Implementation ideas

Add `$USER` to the local cluster name to avoid confusion.

cc: @sharnoff @shayanh

sharnoff:
@cicdteam this was fixed in #677, IIUC?

PR #677, opened 2023-12-07
add username to local cluster name

cicdteam:
solves #676


cicdteam:
> Does `make e2e` require any adjustments?
> I haven't checked it, and I am curious how `kuttl` picks up a correct cluster

@bayandin 

`kuttl` uses current context. Added `check-local-context` dependency for `e2e` as we using in `make deploy`

bayandin (approved):
Haven't tried it, but the idea makes sense to me.

PR #679, opened 2023-12-07
CI: call vm-kernel from release and e2e-test workflows

bayandin:
Make `vm-kernel` a reusable workflow and call it from `e2e-test` and `release` workflows.

Other changes:
- add `kernel-image-tag-override` to `vm-kernel` to make it possible reuse previously built kernel.
- do not trigger `e2e-test` from `vm-kernel`: since now it happens the other way around

Caching kernel image is going to be the next PR.

The PR supersedes #651.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -17,18 +17,21 @@ on:
           - k3d
           - kind
         default: k3d
-  workflow_call:
-    inputs:
-      kernel-image:
-        type: string
-        description: 'The kernel image to use for the VMs'
-        required: false
 
 env:
   PRESERVE_RUNNER_PODS: "true"
 
 jobs:
+  vm-kernel:
+    uses: ./.github/workflows/vm-kernel.yaml
+    with:
+      push: true
+      tag: ${{ github.run_id }}
+      kernel-image-tag-override: ${{ inputs.kernel-image }}
+    secrets: inherit
+
   e2e-tests:
+    needs: vm-kernel
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -17,18 +17,21 @@ on:
...
-        type: string
-        description: 'The kernel image to use for the VMs'
-        required: false
 
 env:
   PRESERVE_RUNNER_PODS: "true"
 
 jobs:
+  vm-kernel:
+    uses: ./.github/workflows/vm-kernel.yaml
```
sharnoff, line 26: Out of curiosity: How does `uses` work here? does it inline the entire contents of `vm-kernel.yaml`, or does it only use the `vm-kernel` job from `vm-kernel.yaml` ?
(and then, later on, when we say `vm-kernel.outputs`, does that refer to the job, or the entire workflow?)

.github/workflows/vm-kernel.yaml
```diff
@@ -58,21 +99,14 @@ jobs:
           build-args: KERNEL_VERSION=${{ steps.get-kernel-version.outputs.VM_KERNEL_VERSION }}
           context: neonvm/hack
           platforms: linux/amd64
-          # Push only if this is a scheduled run or if the workflow_dispatch input push is set to true
-          push: ${{ github.event_name == 'schedule' && 'true' || ( github.event_name == 'workflow_dispatch' && inputs.push || 'false' ) }}
+          # Push kernel image only for schedules builds or if workflow_dispatch/workflow_call input is true
```
sharnoff, line 102:
```suggestion
          # Push kernel image only for scheduled builds or if workflow_dispatch/workflow_call input is true
```
typo?

sharnoff (approved):
Looks good! One non-blocker question, one nit. I don't have enough expertise with github actions to independently validate correctness, but checking the CI logs, it looks like it works well enough :)

.github/workflows/e2e-test.yaml
bayandin: It calls all jobs from `vm-kernel.yaml`, and `vm-kernel.outputs` refers to workflow outputs. So, there's no external access to jobs in `vm-kernel.yaml` workflow.

sharnoff: Makes sense - thanks!

PR #683, opened 2023-12-08
billing: Don't hard-code usage_events suffix

sharnoff:
**NOTE: This requires changing billing config when deploying!**

This commit removes the existing `/usage_events` suffix from the billing URL used, which means that values of the autoscaler-agent config's `.billing.url` must now add `/usage_events` to the end.

ref https://neondb.slack.com/archives/C03H1K0PGKH/p1702050181471839?thread_ts=1701881238.622479

**NB: This PR builds on #681 and must not be merged before it.**

### DIFF
--- a/pkg/agent/billing/send.go
+++ b/pkg/agent/billing/send.go
@@ -110,8 +110,9 @@ func (s eventSender) sendAllCurrentEvents(logger *zap.Logger) {
 
 		logger.Info(
 			"Pushing billing events",
-			zap.String("traceID", string(traceID)),
 			zap.Int("count", count),
+			zap.String("traceID", string(traceID)),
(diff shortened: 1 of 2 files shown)


antonyc (commented):
not sure this is a change that simplifies anything. We got an API entrypoint now, and can easily add new endpoints if we want. This change removes the flexibility and also requires to reconfigure everything. Why?

shayanh (commented):
The changes are small, but I haven't been involved with the billing and control plane integration. So I can't really help much here unless you really need my opinion, and then you gotta give me the context. Generally, I'm fine moving forward with what you guys agree on.

sharnoff:
Closing because no need

PR #688, opened 2023-12-11
agent/core: Deduplicate `State` and `StateDump` fields by sharing them

sharnoff:
This PR comes in two parts, intended to be merged via rebase so we can add the first to `.git-blame-ignore-revs`.

Basically, we currently have a bunch of duplicated fields between `State` and `StateDump` in order to implement JSON-serializable deep copying without exposing those from `State` itself.

This is _kind of silly_; we can just add a layer of indirection in `State` so that the JSON-serializable contents aren't visible, and then share them. So that's what this PR does.

---

Notes for review: I'd recommend looking at the two commits separately. The first is just a bunch of renames (done via LSP), and the second is the actual meat of the simplification. 

Because we're now using the same types in both places, the copying code can actually be much simpler — various pointer-to-structs that _were_ initialized with conditionals can now just use shallow copying.

### DIFF
--- a/pkg/agent/core/dumpstate.go
+++ b/pkg/agent/core/dumpstate.go
@@ -3,6 +3,7 @@ package core
 // Implementation of (*State).Dump()
 
 import (
+	"encoding/json"
 	"time"
 
 	"github.com/neondatabase/autoscaling/pkg/api"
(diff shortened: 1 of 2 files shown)


pkg/agent/core/dumpstate.go
```diff
@@ -3,142 +3,83 @@ package core
...
-	return monitorStateDump{
-		OngoingRequest:     ongoingRequest,
-		RequestedUpscale:   requestedUpscale,
-		DeniedDownscale:    deniedDownscale,
-		Approved:           shallowCopy(s.approved),
-		DownscaleFailureAt: shallowCopy(s.downscaleFailureAt),
-		UpscaleFailureAt:   shallowCopy(s.upscaleFailureAt),
+func (s *monitorState) deepCopy() monitorState {
+	return monitorState{
+		OngoingRequest: shallowCopy(s.OngoingRequest, func(r ongoingMonitorRequest) ongoingMonitorRequest {
```
shayanh, line 64: I don't understand how this function `func(r ongoingMonitorRequest) ongoingMonitorRequest` helps here. If this is going to be a shallow copy then copying the object value and making a pointer to the new value should be enough.

pkg/agent/core/state.go
```diff
@@ -88,83 +88,86 @@ type LogConfig struct {
 // State holds all of the necessary internal state for a VM in order to make scaling
 // decisions
 type State struct {
-	// ANY CHANGED FIELDS MUST BE UPDATED IN dumpstate.go AS WELL
+	internal internalState
+}
 
-	config Config
+// one level of indirection below State so that the fields can be public, and JSON-serializable
+type internalState struct {
```
shayanh, line 95: This is more sort of a personal opinion, but you may want to name this `state` instead. `internalState` is kinda redundant, since `state` implies that it's internal because it's unexported.

shayanh (commented):
Looks good in general. Just two small comments.

pkg/agent/core/dumpstate.go
sharnoff: Yeah, that's fair. There's a number of those. I was _kind of_ thinking along the lines of "how do we make sure this is still correct if we add a field that needs deeper copying?".
But maybe that's more hassle than it's worth, and it's better to just be simple here.

pkg/agent/core/state.go
sharnoff: Hm, yeah, that makes sense. I would _kind of_ find having two "state"s confusing, but perhaps it's worth it. I'll make the change.

sharnoff: aab8174a4cc76a7b17b0c236d90866e47998fb55 - overwritten by later rebase.

pkg/agent/core/dumpstate.go
sharnoff: 55dc0b786b558aac16ebf8c200a877f948c31ece - overwritten by later rebase.

shayanh: Can we remove the extra argument (`f ...func(T) T`) from `shallowCopy`? We haven't used it anywhere and it's confusing.

sharnoff: c4b5dc5e1fe28e912ce6aa2312e66e317aa678aa - same deal w/ rebase.

shayanh (approved):
looks good. let's merge!

PR #690, opened 2023-12-11
plugin: Fix nil deref on failed ExtractVmInfo on VM update

### LINKED ISSUE #689: Bug: Scheduler derefs nil pointer when failed ExtractVmInfo on VM without pod
## Environment

Observed on staging after v0.20.0 (e.g. [here](https://neonprod.grafana.net/explore?schemaVersion=1&panes=%7B%22HNs%22:%7B%22datasource%22:%22xHHYY0dVz%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22xHHYY0dVz%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bcontainer%3D%5C%22autoscale-scheduler%5C%22,cluster%3D%5C%22dev-us-east-2-beta%5C%22%7D%5Cn%20%20%7C%20json%5Cn%20%20%7C%20__error__%20%21%3D%20%5C%22%5C%22%22,%22queryType%22:%22range%22,%22refId%22:%22A%22%7D%5D,%22range%22:%7B%22from%22:%221702101600000%22,%22to%22:%221702274400000%22%7D%7D%7D&orgId=1))

This doesn't appear release-blocking because the specific errors triggering this are because of the one very old VM on staging that doesn't work quite right, and we're hitting this more often because of the `(*watch.Store).Relist()` fixes from #667 resulting in more update update events (which causes us to re-fail handling the wonky VM).

## Steps to reproduce

With normal cluster setup, apply the following diff to cause `autoscaling.neon.tech/bounds` parsing to fail

```diff
diff --git a/vm-deploy.yaml b/vm-deploy.yaml
index 93fa211..dff543e 100644
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -5,7 +5,7 @@ metadata:
   name: postgres14-disk-test
   annotations:
     # In this example, these bounds aren't necessary. So... here's what they look like :)
-    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.25, "mem": "5Gi" } }'
+    autoscaling.neon.tech/bounds: '{ "min": { "cpu": 0.25, "mem": "1Gi" }, "max": { "cpu": 1.25, "mem": { "foo": "bar" } } }'
   labels:
     autoscaling.neon.tech/enabled: "true"
     # Set to "true" to continuously migrate the VM (TESTING ONLY)
```

... and then `kubectl apply -f vm-deploy.yaml` causes the scheduler to panic with

```
{"level":"error","ts":1702272964.3934977,"logger":"autoscale-scheduler.vm-watch","caller":"plugin/watch.go:280","msg":"Failed to extract VM info in update for new VM","virtualmachine":{"namespace":"default","name":"postgres14-disk-test"},"error":"Error unmarshaling annotation \"autoscaling.neon.tech/bounds\": quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'","stacktrace":"github.com/neondatabase/autoscaling/pkg/plugin.(*AutoscaleEnforcer).watchVMEvents.func2\n\t/workspace/pkg/plugin/watch.go:280\ngithub.com/neondatabase/autoscaling/pkg/util/watch.handleEvent[...]\n\t/workspace/pkg/util/watch/watch.go:549\ngithub.com/neondatabase/autoscaling/pkg/util/watch.Watch[...].func4\n\t/workspace/pkg/util/watch/watch.go:306"}
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd3b5a9]

goroutine 216 [running]:
k8s.io/api/core/v1.(*Pod).GetObjectKind(0x1f310e0?)
	<autogenerated>:1 +0x9
k8s.io/client-go/tools/reference.GetReference(0x1fb6675?, {0x2256908, 0x0})
	/go/pkg/mod/k8s.io/client-go@v0.25.16/tools/reference/ref.go:59 +0x110
k8s.io/client-go/tools/events.(*recorderImpl).Eventf(0xc0005b2040, {0x22568e0?, 0xc000a6f000}, {0x2256908?, 0x0}, {0x1f68233, 0x7}, {0x1f70a9d, 0xd}, {0x1f7232c, ...}, ...)
	/go/pkg/mod/k8s.io/client-go@v0.25.16/tools/events/event_recorder.go:54 +0x12f
github.com/neondatabase/autoscaling/pkg/plugin.(*AutoscaleEnforcer).watchVMEvents.func2(0xc000a6ec00, 0xc000a6f000)
	/workspace/pkg/plugin/watch.go:281 +0x799
github.com/neondatabase/autoscaling/pkg/util/watch.handleEvent[...](0xc0004d4d40, {0xc00013be50, 0xc000606140?, 0xc00013be60?}, {0xc000b90fb0?, 0x8a2665422199c594?}, {0xc0007b8c90?, 0x24}, 0xc000a6f000)
	/workspace/pkg/util/watch/watch.go:549 +0x5f1
github.com/neondatabase/autoscaling/pkg/util/watch.Watch[...].func4()
	/workspace/pkg/util/watch/watch.go:306 +0x14cd
created by github.com/neondatabase/autoscaling/pkg/util/watch.Watch[...] in goroutine 1
	/workspace/pkg/util/watch/watch.go:206 +0xc0e
```

In our code on the scheduler side, we try to emit a k8s event about the `VmInfo` extraction failed (due to the intentionally malformed annotation in this example). To link the event to the runner pod as well, we try to fetch it if available:

https://github.com/neondatabase/autoscaling/blob/436edea2649b11bec85c078fa5fd65ddd718956c/pkg/plugin/watch.go#L269-L289

Unfortunately, this means that when using `runnerPod` as a `runtime.Object` — even if there's no pod associated with the VM — `Eventf` finds that the "related" object [isn't nil, because it's a typed nil](https://go.dev/play/p/bXfB3iT0AOv):

```go
func (recorder *recorderImpl) eventf(logger klog.Logger, regarding runtime.Object, related runtime.Object, eventtype, reason, action, note string, args ...interface{}) {

    // -- omitted for brevity --

    var refRelated *v1.ObjectReference
    if related != nil {
    	refRelated, err = reference.GetReference(recorder.scheme, related)
    	if err != nil {
    		logger.V(9).Info("Could not construct reference", "object", related, "err", err)
    	}
    }

    // -- omitted for brevity --
}
```

source: https://github.com/kubernetes/kubernetes/blob/0c645922edcc06adff43c70c02fb56751364bbb5/staging/src/k8s.io/client-go/tools/events/event_recorder.go#L73-L79

sharnoff:
Fixes #689. tl;dr: we need to avoid typed nils to not get nil derefs when trying to create a k8s event.

Tested locally with the reproduction steps from #689, and this appears to fix the issue.

### DIFF
--- a/pkg/plugin/watch.go
+++ b/pkg/plugin/watch.go
@@ -11,6 +11,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	vmapi "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 
@@ -268,13 +269,18 @@ func (e *AutoscaleEnforcer) watchVMEvents(
 				if err != nil {
 					// Try to get the runner pod associated with the VM, if we can, but don't worry
 					// about it if we can't.
-					var runnerPod *corev1.Pod
+					var runnerPod runtime.Object
 					if podName := newVM.Status.PodName; podName != "" {
 						// NB: index.Get returns nil if not found, so we only have a non-nil
 						// runnerPod if it's currently known.
-						runnerPod, _ = podIndex.GetIndexed(func(index *watch.NameIndex[corev1.Pod]) (*corev1.Pod, bool) {
+						rp, _ := podIndex.GetIndexed(func(index *watch.NameIndex[corev1.Pod]) (*corev1.Pod, bool) {
 							return index.Get(newVM.Namespace, podName)
 						})
+						// avoid typed nils by only assigning if non-nil
+						// See <https://github.com/neondatabase/autoscaling/issues/689> for more.
+						if rp != nil {
+							runnerPod = rp
+						}
 					}
 
 					logger.Error("Failed to extract VM info in update for new VM", util.VMNameFields(newVM), zap.Error(err))


shayanh (approved):
lgtm

PR #693, opened 2023-12-14
IOPS and throughput limits

cicdteam:
This PR allows you to specify the IOPS and throughput limit for the VM `.guest.rootDisk` and for disks of type `disks[].emptyDisk`.

Limits successfully tested

### tool used

Flexible I/O tester - [fio](https://fio.readthedocs.io/en/latest/index.html)

### fio args

- random read-write
```
fio --filename=/TESTFILE --rw=randrw --bs=4k --runtime=60 --numjobs=4 --time_based --name=iops --size=512m --group_reporting --direct=1
```
- sequential read
```
fio --filename=/TESTFILE --rw=read --bs=4k --runtime=60 --numjobs=4 --time_based --name=iops --size=512m --group_reporting --direct=1
```
### VM spec

```yaml
apiVersion: vm.neon.tech/v1
kind: VirtualMachine
metadata:
  name: andrey-disk-speed
spec:
  restartPolicy: Never
  guest:
    cpus:
      min: 1
    memorySlotSize: 1Gi
    memorySlots:
      min: 4
    rootDisk:
      image: neondatabase/vm-debian:11
      size: 20Gi
```

#### random read-write
```
read: IOPS=54.0k, BW=215MiB/s (225MB/s)(12.6GiB/60001msec)
write: IOPS=54.9k, BW=215MiB/s (225MB/s)(12.6GiB/60001msec); 0 zone resets
```
####  sequential read
```
read: IOPS=119k, BW=466MiB/s (489MB/s)(27.3GiB/60001msec)
```

### IOPS limited to 3000

```yaml
    rootDisk:
      image: neondatabase/vm-debian:11
      size: 20Gi
      iops: 3000
```

#### random read-write

expected:
- iops summary = 3000, ~1500 for read and ~1500 for write
- throughput summary = 11.7 MiB/s (4 KiB block size * 3000 iops), ~6 MiB/s for read and ~6 MiB/s for write

result:
```
read: IOPS=1499, BW=5999KiB/s (6143kB/s)(352MiB/60002msec)
write: IOPS=1504, BW=6019KiB/s (6163kB/s)(353MiB/60002msec); 0 zone resets
```

#### sequential read

expected:
- iops summary = 3000
- throughput summary = 11.7 MiB/s (4 KiB block size * 3000)

result:
```
read: IOPS=3004, BW=11.7MiB/s (12.3MB/s)(704MiB/60002msec)
```

### throughput limited to 100 MiB/s

```yaml
    rootDisk:
      image: neondatabase/vm-debian:11
      size: 20Gi
      throughput: 100
```
#### random read-write

expected:
- iops summary = 25.6k (100 MiB/s / 4 KiB block size), ~12.8 for read and ~12.8k for write
- throughput summary = 100 MiB/s, ~50 MiB/s for read and ~50 MiB/s for write

result:
```
read: IOPS=12.8k, BW=50.0MiB/s (52.5MB/s)(3001MiB/60002msec)
write: IOPS=12.8k, BW=50.1MiB/s (52.6MB/s)(3008MiB/60002msec); 0 zone resets
```

#### sequential read

expected:
- iops summary 25.6k (100 MiB/s / 4 KiB block size)
- throughput summary = 100 MiB/s

result:
```
read: IOPS=25.6k, BW=100MiB/s (105MB/s)(6010MiB/60002msec)
```

### IOPS limited to 3000 and throughput limited to 100 MiB/s

```yaml
    rootDisk:
      image: neondatabase/vm-debian:11
      size: 20Gi
      iops: 3000
      throughput: 100
```

expected:
- the same as the IOPS-limited test, since in this case the throughput is limited by the IOPS

result:
- same results as the test with limited IOPS


sharnoff (commented):
nice!

My main concern is about how we expect this to be used in the broader system — AFAICT this requires integration with the control plane, and if we want to have IOPS proportional to CU, then we'd want to adjust throttling at runtime (can we do that?) and control it via autoscaling. I also think - whatever route we go with - we should try to avoid making the control plane responsible for managing cluster IO, although maybe that's not possible.

Some other APIs that come to mind:
1. Controlling IOPS globally for a VM, with the ability to adjust it at runtime (and setting that via autoscaling)
    - We could maybe implement that with this PR as it is, but that's a decent amount of work.
2. Setting IOPS on the VM spec as a multiple of CPU, setting that from the control plane and applying it in neonvm-runner)
3. Handling this entirely internally, hard-coded in neonvm-runner 

Didn't do a detailed review for now, want to hear your thoughts.

IIUC this won't solve our immediate concerns[^wont-solve], and we're planning other ways to resolve the IO concerns.[^other-ways] But it'd still be really good to have these as a safety measure, especially when looking at enabling overcommit.

[^wont-solve]: https://neondb.slack.com/archives/C03F5SM1N02/p1702581440790379?thread_ts=1702558673.445669
[^other-ways]: https://neondb.slack.com/archives/C03F5SM1N02/p1702570735499099?thread_ts=1702558673.445669

cicdteam:
@sharnoff 

>then we'd want to adjust throttling at runtime (can we do that?) 

yes, possible change IO throttling settings via QMP

>Controlling IOPS globally for a VM, with the ability to adjust it at runtime

It is also possible to create a throttling group and assign all disks to this group. Then the total number of I/O operations per second for all disks cannot be greater than that specified in the group.

>Setting IOPS on the VM spec as a multiple of CPU

Yes,. Had the same idea, but I didn't know how many IOPS to set on the CPU core. We can also rely on disk size, like AWS does for EBS gp2 (volumes 33.33 GiB and smaller are provisioned with the minimum of 100 IOPS, volumes larger than 33.33 GiB are provisioned with 3 IOPS per GiB).

In general, I initially added this throttling option to be able to set limits just to prevent situations where a noisy tenant was overloading the disk on the Kubernetes node. If this is not necessary now, we can close this PR and design the I/O throttling idea before implementing it.

sharnoff:
> In general, I initially added this throttling option to be able to set limits just to prevent situations where a noisy tenant was overloading the disk on the Kubernetes node.

Ah, yeah that makes sense.

> If this is not necessary now, we can close this PR and design the I/O throttling idea before implementing it.

Fair enough! Maybe let's leave as a draft so it's still around.

In general, adding IOPS throttling for VMs is one of the prerequisites to enabling overcommit (although, maybe with real SSDs it doesn't need to be) - but yeah, let's talk design & we can get this [or something like it] merged in January. I'd like to be able to move away from the container-io-limiter-hack & allow IOPS proportional to CPUs.

Omrigan:
Not needed at the moment.

Issue #686, opened 2023-12-09
Bug: agent's per-VM metrics don't include autoscaling bounds annotation


sharnoff:
## Environment

Any

## Steps to reproduce

Port-forward to get VM metrics:

```sh
kubectl port-forward -n kube-system pod/<name> 9101:9101
```

and

```sh
$ curl -sSf http://localhost:9101/metrics | grep 'vm_name="compute-weathered-wood-82166743"'
autoscaling_vm_resources{endpoint_id="",resource="cpu",value="max",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 8
autoscaling_vm_resources{endpoint_id="",resource="cpu",value="min",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 0.25
autoscaling_vm_resources{endpoint_id="",resource="cpu",value="spec_use",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 1
autoscaling_vm_resources{endpoint_id="",resource="cpu",value="status_use",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 1
autoscaling_vm_resources{endpoint_id="",resource="mem",value="max",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 3.4359738368e+10
autoscaling_vm_resources{endpoint_id="",resource="mem",value="min",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 1.073741824e+09
autoscaling_vm_resources{endpoint_id="",resource="mem",value="spec_use",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 4.294967296e+09
autoscaling_vm_resources{endpoint_id="",resource="mem",value="status_use",vm_name="compute-weathered-wood-82166743",vm_namespace="default"} 4.294967296e+09
```

Currently, the `value="min"` and `value="max"` time series are based on the VM spec, and there's no way to get a version of the min/max values that includes the `autoscaling.neon.tech/bounds` annotation.

Because this is missing, these metrics have limited utility for us in their current form.

## Other logs, links

- https://neondb.slack.com/archives/C061BGTBZMG/p1702016182162559?thread_ts=1701900191.812849

PR #695, opened 2023-12-14
agent: Include autoscaling bounds annotation in per vm metrics

shayanh:
Fixes https://github.com/neondatabase/autoscaling/issues/686


pkg/agent/watch.go
```diff
@@ -202,6 +204,18 @@ func (e vmEvent) Format(state fmt.State, verb rune) {
 	}
 }
 
+func extractAutoScalingBounds(vm *vmapi.VirtualMachine) *api.ScalingBounds {
```
sharnoff, line 207:
```suggestion
func extractAutoscalingBounds(vm *vmapi.VirtualMachine) *api.ScalingBounds {
```

sharnoff, line 207: It'd be good to have a comment here explaining what it does and why we have this even though `api.ExtractVmInfo` exists and does some of it

pkg/agent/watch.go
```diff
@@ -230,11 +245,24 @@ func makeVMCPUMetrics(vm *vmapi.VirtualMachine) []vmMetric {
...
 	}
 
+	// metrics from autoscaling bounds annotation
+	if bounds := extractAutoScalingBounds(vm); bounds != nil {
+		boundPairs := []pair[vmResourceValueType, resource.Quantity]{
+			{vmResourceValueAutoscalingMin, bounds.Min.CPU},
+			{vmResourceValueAutoscalingMax, bounds.Max.CPU},
+		}
+		for _, p := range boundPairs {
+			m := makeVMMetric(vm, p.first, vmapi.MilliCPUFromResourceQuantity(p.second).AsFloat64())
```
sharnoff, line 261: worth explaining why we're converting to `vmapi.MilliCPU` first instead of directly converting `resource.Quantity` to float

sharnoff (commented):
generally looks good

pkg/agent/watch.go
shayanh: I wrote it this way to keep it consistent with the other CPU metrics. Do you have another reason in mind?

sharnoff: oh, I think that's how it should be - I'm just saying that there should be a comment explaining why to do it this way (because `resource.Quantity.AsApproximateFloat64()` is quite inaccurate)

sharnoff: https://github.com/neondatabase/autoscaling/blob/90e79ff655aac6774fa806e7928c4a3f68bbd273/neonvm/apis/neonvm/v1/virtualmachine_types.go#L200-L206

shayanh: fixed in 4ff769668e14024cee2b8679e79a7cd12c149664

pkg/agent/watch.go
```diff
@@ -202,6 +204,22 @@ func (e vmEvent) Format(state fmt.State, verb rune) {
 	}
 }
 
+// extractAutoscalingBounds extracts the ScalingBounds from a VM's autoscaling
+// annotation. We have defined this function for the purpose of metrics, so we
+// could track scaling bounds changes on a VM. api.ExtractVmInfo doesn't provide
+// this data, since its CPU and memory values might come from vm.Spec.
```
sharnoff, line 210:
```suggestion
// extractAutoscalingBounds extracts the ScalingBounds from a VM's autoscaling
// annotation, for the purpose of exposing it in per-VM metrics.
//
// We're not reusing api.ExtractVmInfo even though it also looks at the bounds
// annotation, because its data is less precise - CPU and memory values might
// come from the VM spec without us knowing.
```

I'd reword + simplify + add a line break.

I found it hard to parse the original because logically distinct thoughts started mid-line, and felt there wasn't a clear connection of why we might want to use `api.ExtractVmInfo` in the first place.
IIUC, the bit about "so we could track scaling bound changes on a VM" isn't _quite_ accurate, because in theory that data is fully available in the control plane database - it's just easier to provide it all in one place.

Feel free to reword to fit your style — figured it'd be easier to show what I meant, rather than tell :)

sharnoff (approved):
one comment nit. feel free to merge once changed

PR #699, opened 2023-12-18
agent: added project ID label to per vm metrics

shayanh:
Requested here: https://neondb.slack.com/archives/C061BGTBZMG/p1702897129242809

### DIFF
--- a/pkg/agent/watch.go
+++ b/pkg/agent/watch.go
@@ -34,8 +34,10 @@ type vmEvent struct {
 	endpointID string
 }
 
-const endpointLabel = "neon/endpoint-id"
-const projectLabel = "neon/project-id"
+const (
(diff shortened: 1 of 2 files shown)


pkg/agent/watch.go
```diff
@@ -35,6 +35,7 @@ type vmEvent struct {
 }
 
 const endpointLabel = "neon/endpoint-id"
+const projectLabel = "neon/project-id"
```
sharnoff, line 38: nit: should be grouped, like
```go
const (
    ...
)
```

sharnoff (commented):
one nit, no other objections. Wondering if this is actually needed (replied in the slack thread).

sharnoff (approved):

PR #701, opened 2023-12-19
fix iptables rules for traffic from localhost

### LINKED ISSUE #692: Bug: Cannot access NeonVM pods metrics endpoint
## Environment

local development, staging

(idk if it's the case in prod)

## Steps to reproduce

```
kubectl --context dev-us-east-2-beta port-forward compute-wispy-surf-w25ffuy9-hwvdf 9100:9100
```

and then `curl http://localhost:9100/metrics` gets stuck.

## Expected result

Seeing metrics via curl.

## Actual result

No response.

cicdteam:
### This PR
- adds additional `iptables` rules that allow traffic from localhost (in the runner pod) to ports on the VM.

Now possible do `kubectl port-forward vm-runner-pod-name 9100` and then run `curl localhost:9100/metrics`
 or just `kubectl exec -it vm-runner-pod-name -- curl localhost:9100/metrics`

### Internal details:

These rules will be added for port 9100 as an example

```
iptables -t nat -A OUTPUT -m addrtype --src-type LOCAL --dst-type LOCAL -p tcp --dport 9100 -j DNAT --to-destination 169.254.254.254:9100
iptables -A OUTPUT -s 127.0.0.1 -d 169.254.254.254 -p tcp --dport 9100 -j ACCEPT
```

and this rule for all ports
```
iptables -t nat -A POSTROUTING -m addrtype --src-type LOCAL --dst-type UNICAST -j MASQUERADE
```

fixes #692

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -1137,17 +1137,50 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 	}
 
 	// pass incoming traffic to .Guest.Spec.Ports into VM
+	iptablesArgs := []string{}
 	for _, port := range ports {
-		logger.Info(fmt.Sprintf("setup DNAT for incoming traffic, port %d", port.Port))
-		iptablesArgs := []string{
+		logger.Info(fmt.Sprintf("setup DNAT rule for incoming traffic to port %d", port.Port))
+		iptablesArgs = []string{
 			"-t", "nat", "-A", "PREROUTING",
 			"-i", "eth0", "-p", fmt.Sprint(port.Protocol), "--dport", fmt.Sprint(port.Port),
 			"-j", "DNAT", "--to", fmt.Sprintf("%s:%d", ipVm.String(), port.Port),
 		}
 		if err := execFg("iptables", iptablesArgs...); err != nil {
-			logger.Error("could not set up DNAT for incoming traffic", zap.Error(err))
+			logger.Error("could not set up DNAT rule  for incoming traffic", zap.Error(err))
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


neonvm/runner/main.go
```diff
@@ -1137,17 +1137,50 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
...
+			"-t", "nat", "-A", "OUTPUT",
+			"-m", "addrtype", "--src-type", "LOCAL", "--dst-type", "LOCAL",
+			"-p", fmt.Sprint(port.Protocol), "--dport", fmt.Sprint(port.Port),
+			"-j", "DNAT", "--to-destination", fmt.Sprintf("%s:%d", ipVm.String(), port.Port),
+		}
+		if err := execFg("iptables", iptablesArgs...); err != nil {
+			logger.Error("could not set up DNAT rule for traffic from localhost", zap.Error(err))
+			return nil, err
+		}
+		logger.Info(fmt.Sprintf("setup ACCEPT rule for traffic originating from localhost to port %d", port.Port))
```
shayanh, line 1163: why do we need this accept rule here?

neonvm/runner/main.go
```diff
@@ -1137,17 +1137,50 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
...
+		iptablesArgs = []string{
```
shayanh, line 1164: Why do we need the accept rule here?

shayanh (approved):
lgtm

neonvm/runner/main.go
cicdteam: it's  OUTPUT chain needed because in neon-compute we have additional iptables rules that prevent any traffic to 169.254.252.252/30

PR #703, opened 2023-12-21
qmp: add manual QMP access

Omrigan:
Introduce another QMP port, and a convenience script to forward it for manual debugging.


cicdteam:
@Omrigan 

It's nice to be able to debug QMP commands using scripts, but why need another port? why not use the current one?

Omrigan:
> @Omrigan
> 
> It's nice to be able to debug QMP commands using scripts, but why need another port? why not use the current one?

Unfortunately, QEMU only allows one connection per QMP port [1]. That's also consistent with the behaviour I was observing: when connected by hand to the only port, scaling stops working. 

1: https://lists.gnu.org/archive/html/qemu-devel/2016-10/msg02210.html

cicdteam:
heh.. I thought `server,nowait` was meant for multiple connections :)

cicdteam (approved):

PR #704, opened 2023-12-21
qmp: unify up/down memory scaling 

### LINKED ISSUE #451: Bug: Memory scaling breaks with "attempt to add duplicate property 'memslot<N>' ..."
## Environment

Local testing while trying to reproduce the kernel panic (#450). A quick search shows it's widely present in prod as well.

## Steps to reproduce

There's a handful of ways to do this. But broadly having some kind of memory usage in the VM that prevents downscaling memory + continually scaling up & down memory slots[^memory-slots] (manually; not via autoscaling) will result in the error. I used some of the stuff from https://github.com/neondatabase/autoscaling/issues/450#issuecomment-1678074941. Here's an example of the error:

```
attempt to add duplicate property 'memslot10' to object (type 'container')
```

... which typically occurs after many rounds of these:

```
All memory devices busy, unable to unplug any, will modify .spec.guest.memorySlots.use instead
```

NB: in this example, 'memslot10' shouldn't ever be used — I was using a VM with `spec.guest.memorySlots.max = 10` so there should only be memslots 0 through 9, but because `qom-list` is returning a value for all memory slots (even if not all of them are actually in use), then we make it through this loop without finding anything (and use `slots + 1` as the id):

https://github.com/neondatabase/autoscaling/blob/f622fd96a3106d8a432f3b2f345c76d084020262/neonvm/controllers/virtualmachine_qmp_queries.go#L354-L370

I suspect that either (a) our usage of qom-list is incorrect (maybe presence via qom-list doesn't necessarily mean it's online?), or (b) QEMU's behavior is incorrect.

## Expected result

We should continue scaling up and down as normal, subject to occasionally not being able to unplug a memory slot.

## Actual result

We gradually exhaust all memory slots, and because of an off-by-one error end up repeatedly failing with this "attempt to add duplicate property" error, which prevents adding any further memory slots. I think _maybe_ the situation self-resolves when the memory usage drops lower AND it coincides with requesting to unplug memory, but I'm not sure.

[^memory-slots]: For specifics: I was using 0.1 CPU (larger CPU means the problem is less likely to occur) and alternating between 2 and 3 memory slots, 5s apart. I was using the `allocate-loop` binary that's included in `postgres14-disk-test`, run as `allocate-loop 1024 1536`, which _doesn't_ get OOM-killed during normal operation at 2 memory slots, but _does_ when scaling.

Omrigan:
The new implementation has several advantages over the previous one:
1. It handles the case, when there are leftover memory backends from
previous incomplete unplugs (see #451). Those backends are reused for
new DIMM devices.
2. It allows for simultaneous addition/removal of multiple devices.
3. It adds symmetrical k8s event reporting for both memory backends
and DIMM devices.

Fixes #451

### DIFF
--- a/neonvm/controllers/virtualmachine_qmp_queries.go
+++ b/neonvm/controllers/virtualmachine_qmp_queries.go
@@ -343,12 +343,16 @@ func QmpQueryMemoryBackendIds(mon *qmp.SocketMonitor) (map[int]struct{}, error) 
 	return backends, nil
 }
 
+type QMPRunner interface {
+	Run([]byte) ([]byte, error)
+}
+
 // QmpAddMemoryBackend adds a single memory slot to the VM with the given size.
 //
 // The memory slot does nothing until a corresponding "device" is added to the VM for the same memory slot.
 // See QmpAddMemoryDevice for more.
 // When unplugging, QmpDelMemoryDevice must be called before QmpDelMemoryBackend.
-func QmpAddMemoryBackend(mon *qmp.SocketMonitor, idx int, sizeBytes int64) error {
+func QmpAddMemoryBackend(mon QMPRunner, idx int, sizeBytes int64) error {
 	cmd := []byte(fmt.Sprintf(
 		`{"execute": "object-add",
 		  "arguments": {"id": "memslot%d",
--- a/neonvm/controllers/virtualmachine_qmp_queries.go
+++ b/neonvm/controllers/virtualmachine_qmp_queries.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -56,6 +58,15 @@ type QmpMemoryDevice struct {
 		Node         int64  `json:"node"`
 		Id           string `json:"id"`
 	} `json:"data"`
+}
+
+type QmpObjects struct {
+	Return []QmpObject `json:"return"`
+}
+
+type QmpObject struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
 }
 
 type QmpMigrationInfo struct {
@@ -269,6 +280,105 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
 	return result.Return, nil
 }
 
+func QmpMatchName(reg *regexp.Regexp, name string) (int, bool) {
+	// extracts the id from the name
+	matches := reg.FindStringSubmatch(name)
+	if len(matches) < 2 {
+		return 0, false
+	}
+	id, err := strconv.Atoi(matches[1])
+	if err != nil {
+		return 0, false
+	}
+	return id, true
+}
+
+var memBackendRegexp = regexp.MustCompile(`memslot(\d+)`)
+
+func QmpMatchMemBackend(name string) (int, error) {
+	// "/objects/memslot3" -> 3
+	name = strings.ReplaceAll(name, "/objects/", "")
+	idx, ok := QmpMatchName(memBackendRegexp, name)
+	if !ok {
+		return 0, fmt.Errorf("failed to parse memory device id: %s", name)
+	}
+	return idx, nil
+}
+
+func QmpQueryMemoryBackendIds(mon *qmp.SocketMonitor) (map[int]struct{}, error) {
+	cmd := []byte(`{"execute": "qom-list", "arguments": {"path": "/objects"}}`)
+	raw, err := mon.Run(cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	var result QmpObjects
+	if err := json.Unmarshal(raw, &result); err != nil {
+		return nil, fmt.Errorf("error unmarshaling json: %w", err)
+	}
+	backends := map[int]struct{}{}
+	for _, o := range result.Return {
+		if o.Name == "pc.ram" { // Non-hotplugged memory
+			continue
+		}
+		if o.Type != "child<memory-backend-ram>" {
+			continue
+		}
+
+		idx, err := QmpMatchMemBackend(o.Name)
+		if err != nil {
+			return nil, err
+		}
+		backends[idx] = struct{}{}
+	}
+	return backends, nil
+}
+
+// QmpAddMemoryBackend adds a single memory slot to the VM with the given size.
+//
+// The memory slot does nothing until a corresponding "device" is added to the VM for the same memory slot.
+// See QmpAddMemoryDevice for more.
+// When unplugging, QmpDelMemoryDevice must be called before QmpDelMemoryBackend.
+func QmpAddMemoryBackend(mon *qmp.SocketMonitor, idx int, sizeBytes int64) error {
+	cmd := []byte(fmt.Sprintf(
+		`{"execute": "object-add",
+		  "arguments": {"id": "memslot%d",
+						"size": %d,
+						"qom-type": "memory-backend-ram"}}`, idx, sizeBytes,
+	))
+	_, err := mon.Run(cmd)
+	return err
+}
+
+func QmpDelMemoryBackend(mon *qmp.SocketMonitor, idx int) error {
+	cmd := []byte(fmt.Sprintf(
+		`{"execute": "object-del",
+		  "arguments": {"id": "memslot%d"}}`, idx,
+	))
+	_, err := mon.Run(cmd)
+	return err
+}
+
+func QmpAddMemoryDevice(mon *qmp.SocketMonitor, idx int) error {
+	cmd := []byte(fmt.Sprintf(
+		`{"execute": "device_add",
+		  "arguments": {"id": "dimm%d",
+						"driver": "pc-dimm",
+						"memdev": "memslot%d"}}`, idx, idx,
+	))
+	_, err := mon.Run(cmd)
+	return err
+}
+
+func QmpDelMemoryDevice(mon *qmp.SocketMonitor, idx int) error {
+	cmd := []byte(fmt.Sprintf(
+		`{"execute": "device_del",
+		  "arguments": {"id": "dimm%d"}}`, idx,
+	))
+	_, err := mon.Run(cmd)
+	return err
+}
+
 func QmpPlugMemory(virtualmachine *vmv1.VirtualMachine) error {
 	// slots - number of pluggable memory slots (Max - Min)
 	slots := *virtualmachine.Spec.Guest.MemorySlots.Max - *virtualmachine.Spec.Guest.MemorySlots.Min
@@ -291,29 +401,28 @@ func QmpPlugMemory(virtualmachine *vmv1.VirtualMachine) error {
 	defer mon.Disconnect() //nolint:errcheck // nothing to do with error when deferred. TODO: log it?
 
 	// try to find empty slot
-	var slot int32
-	for slot = 1; slot <= slots; slot++ {
-		cmd := []byte(fmt.Sprintf(`{"execute": "qom-list", "arguments": {"path": "/objects/memslot%d"}}`, slot))
-		_, err = mon.Run(cmd)
-		if err != nil {
+	backends, err := QmpQueryMemoryBackendIds(mon)
+	if err != nil {
+		return err
+	}
+	var slot int
+	for slot = 1; slot <= int(slots); slot++ {
+		_, ok := backends[slot]
+		if !ok {
 			// that mean such object wasn't found, or by other words slot is empty
 			break
 		}
 	}
 
-	// add memdev object
-	cmd := []byte(fmt.Sprintf(`{"execute": "object-add", "arguments": {"id": "memslot%d", "size": %d, "qom-type": "memory-backend-ram"}}`, slot, virtualmachine.Spec.Guest.MemorySlotSize.Value()))
-	_, err = mon.Run(cmd)
+	err = QmpAddMemoryBackend(mon, slot, virtualmachine.Spec.Guest.MemorySlotSize.Value())
 	if err != nil {
 		return err
 	}
 	// now add pc-dimm device
-	cmd = []byte(fmt.Sprintf(`{"execute": "device_add", "arguments": {"id": "dimm%d", "driver": "pc-dimm", "memdev": "memslot%d"}}`, slot, slot))
-	_, err = mon.Run(cmd)
+	err = QmpAddMemoryDevice(mon, slot)
 	if err != nil {
 		// device_add command failed... so try remove object that we just created
-		cmd = []byte(fmt.Sprintf(`{"execute": "object-del", "arguments": {"id": "memslot%d"}}`, slot))
-		mon.Run(cmd) //nolint:errcheck // already have one error, ignoring the second.
+		_ = QmpDelMemoryBackend(mon, slot)
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


sharnoff:
Reviewing commit-by-commit (thanks for separating the steps!)

- First commit looks good! Let's split that into a separate PR, it'll be quicker to merge & simplify this one — and that way, the squashed history is better.
    - Alternatively, there's a way to rebase & merge so each commit keeps the PR number on the end. Happy to explain that one a bit more. I mention this is because it's helpful to have separate changes in separate commits after merging, but we usually squash+merge because then the commits all reference the PR. 
- Second commit - need to look at this more thoroughly. First impressions: it could some comments, in particular: a brief (few sentences) explanation of how the new functions fit together (+ what's the difference between devices vs objects/backends), as well as brief comments on the functions saying what they do
- Third commit - roughly same situation as the second commit. Need to look at it more thoroughly, but at first glance, it could use some comments.

Omrigan:
@sharnoff 

Let's go forward with "rebase & merge". I'll update the subjects before merging with the PR number.

I've added some comments and expanded commit message for the 2nd commit. If you think more explanation is needed, let me know where should I put it.

sharnoff:
Sounds good. I'll try to get a more thorough review done today.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,17 +257,93 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	for _, o := range result.Return {
+		idx, err := QmpMatchMemBackend(o.Name)
+		if err == nil {
+			backends[idx] = struct{}{}
+		}
+	}
+	return backends, nil
+}
+
+func QmpAddMemoryBackend(mon *qmp.SocketMonitor, idx int, size int64) error {
```
sharnoff, line 305: Needs documentation:

- What does this function do?
- what's the units of `size`?
- what's the relationship between this function and `QmpAddMemoryDevice`? (which one goes first?)

I happen to know the answers here. I ask because the comments should give a reasonable overview for someone who's not intimately familiar with QEMU — otherwise we risk limiting the people who can work on neonvm-controller to those familiar with both QEMU and Kubernetes.

I would imagine something like:

> QmpAddMemoryBackend adds a single memory slot to the VM with the given size.
>
> The memory slot does nothing until a corresponding "device" is added to the VM for the same memory slot. See QmpAddMemoryDevice for more.
> When unplugging, QmpDelMemoryDevice must be called before QmpDelMemoryBackend.
>
> The size is in units of bytes.

The other functions below (`Qmp{Add,Del}Memory{Backend,Device}`) should have similar comments

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,17 +257,93 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	json.Unmarshal(raw, &result)
+	backends := map[int]struct{}{}
+	for _, o := range result.Return {
+		idx, err := QmpMatchMemBackend(o.Name)
+		if err == nil {
+			backends[idx] = struct{}{}
+		}
+	}
+	return backends, nil
+}
```
sharnoff: I had a couple thoughts:

1. We have a not-insignificant amount of code for parsing the `memslot<N>` IDs — and regex for such a small thing always feels a little unfortunate.
2. We're repeatedly writing `"memslot%d", idx` and passing around this `idx` value to various functions - what does `idx` _really_ represent though?
3. The calls to `QmpMatchMemBackend` silently discard the error because some objects may not be memory slots, which leaves us at risk of hard-to-debug problems.

So: We already have the type of the objects included in `QmpObject` — maybe we should filter for `memory-backend-ram` objects, and pass around the (qmp) `id` as a string.
What do you think?

(it may be slightly less efficient, but I think it'll be easier to explain what's going on)

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,17 +257,93 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	cmd := []byte(fmt.Sprintf(`{"execute": "device_del",
+								"arguments": {"id": "dimm%d"}}`, idx))
+	_, err := mon.Run(cmd)
+	return err
+}
+
 func QmpPlugMemory(virtualmachine *vmv1.VirtualMachine) error {
 	// slots - number of pluggable memory slots (Max - Min)
-	slots := *virtualmachine.Spec.Guest.MemorySlots.Max - *virtualmachine.Spec.Guest.MemorySlots.Min
+	slots := int(*virtualmachine.Spec.Guest.MemorySlots.Max - *virtualmachine.Spec.Guest.MemorySlots.Min)
```
sharnoff: this commit (ead54af9ec0010a6c8cc213b78020b46551a191b) already has a lot of changes — let's either (a) keep these variables as `int32` or (b) make the switch to `int` a separate commit.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -327,19 +410,19 @@ func QmpSyncMemoryToTarget(vm *vmv1.VirtualMachine, migration *vmv1.VirtualMachi
...
 		}
 		// now add pc-dimm device
-		cmd = []byte(fmt.Sprintf(`{"execute": "device_add", "arguments": {"id": "%s", "driver": "pc-dimm", "memdev": "%s"}}`, m.Data.Id, memdevId))
-		_, err = target.Run(cmd)
+		err = QmpAddMemoryDevice(target, memdevIdx)
 		if err != nil {
 			// device_add command failed... so try remove object that we just created
-			cmd = []byte(fmt.Sprintf(`{"execute": "object-del", "arguments": {"id": "%s"}}`, m.Data.Memdev))
-			target.Run(cmd)
+			_ = QmpDelMemoryBackend(target, memdevIdx)
```
sharnoff: not a blocker: Do we need to also `QmpDelMemoryDevice` here?

I know the previous version doesn't, but is it possible for e.g. adding the memory device to "fail" due to timeout, but the operation to go through? I guess: what happens if we try to remove the backend while the device is still present?

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -368,8 +451,11 @@ func QmpUnplugMemory(ip string, port int32) error {
 	var merr error
 	for i = plugged - 1; i >= 0; i-- {
 		// remove pc-dimm device
-		cmd := []byte(fmt.Sprintf(`{"execute": "device_del", "arguments": {"id": "%s"}}`, memoryDevices[i].Data.Id))
-		_, err = mon.Run(cmd)
+		idx, err := QmpMatchMemBackend(memoryDevices[i].Data.Memdev)
+		if err != nil {
+			merr = errors.Join(merr, err)
+		}
+		err = QmpDelMemoryDevice(mon, idx)
```
sharnoff: not a blocker:
The `continue` just below this seems... scary to me.

IIRC I've seen unplug "timeout" (from QMP's perspective) but eventually occur inside the VM if the guest's ACPI handler just takes a while to migrate pages.
If that happens, is it possible to accidentally unplug many memory slots? i.e. start unplugging many memory slots because none of them have succeeded yet, and then end up unplugging more than you intended?

sharnoff (commented):
Some thoughts on the second commit

neonvm/controllers/virtualmachine_qmp_queries.go
Omrigan: Good point. 

I'd rather keep it as-is in this commit. I've addressed it in the next one, where this logic gets overhauled.


It is modified  such that we attempt to unplug no more than the intended number of slots.

neonvm/controllers/virtualmachine_qmp_queries.go
shayanh: I suggest going with the common practice in Go. For exported types, functions, etc, you should have a comment providing necessary the documentation (details [here](https://tip.golang.org/doc/comment)). The comment should start with the name of the exported entity.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,57 +262,345 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	return attempts.DidSomething()
+}
+
+func QmpSetMemorySlots(ctx context.Context, vm *vmv1.VirtualMachine, targetCnt int, recorder record.EventRecorder) (int, error) {
+	// Attempts to plug/unplug memory slots to match targetCnt.
+	// Returns the number of slots, which the function managed to plug.
+	// Ideally, it matches targetCnt, but can be less or more if there are
+	// errors.
+	//
+	// Returns -1 if failed to connect to QMP, otherwise, the return value is
```
shayanh, line 560: This function is returning an error. The common Go practice to figure out if an error happened is to use the returned error value (if present). So I'd say in this case we can eliminate this guarantee (returning -1 if we failed) and stick to the returned error.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,57 +262,345 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+
+	backends, err := QmpQueryMemoryBackendIds(r.mon)
 	if err != nil {
-		// device_add command failed... so try remove object that we just created
-		cmd = []byte(fmt.Sprintf(`{"execute": "object-del", "arguments": {"id": "memslot%d"}}`, slot))
-		mon.Run(cmd)
 		return err
 	}
 
+	for b, _ := range backends {
```
shayanh, line 396: Simplify it as:

```suggestion
	for b := range backends {
```

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,57 +262,345 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	for idx, _ := range r.memBackends {
+		r.maxBackend = max(r.maxBackend, idx)
+	}
+	r.log.Info("QMP memory state", "backends", r.memBackends, "maxBackend", r.maxBackend)
 	return nil
 }
 
+func (r *QmpMemorySetter) Disconnect() {
+	if r.mon != nil {
+		r.mon.Disconnect()
```
shayanh, line 410: The error returned from `r.mon.Disconnect` is ignored. Probably you want to return it here.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,57 +262,345 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
-		mon.Run(cmd)
 		return err
 	}
 
+	for b, _ := range backends {
+		if _, ok := r.memBackends[b]; !ok {
+			r.memBackends[b] = false
+		}
+	}
+	for idx, _ := range r.memBackends {
```
shayanh, line 401: Same as above

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -246,57 +262,345 @@ func QmpQueryMemoryDevices(ip string, port int32) ([]QmpMemoryDevice, error) {
...
+	log := log.FromContext(ctx)
+
+	setter := &QmpMemorySetter{
+		vm:        vm,
+		targetCnt: targetCnt,
+		recorder:  recorder,
+		log:       log,
+	}
+	defer setter.Disconnect()
+	setter.BuildState()
```
shayanh, line 582: The returned error is ignored here

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -327,112 +631,26 @@ func QmpSyncMemoryToTarget(vm *vmv1.VirtualMachine, migration *vmv1.VirtualMachi
...
-	_, err = mon.Run(cmd)
-	if err != nil {
-		// that mean such object wasn't found, let's add it
-		cmd = []byte(fmt.Sprintf(`{"execute": "object-add", "arguments": {"id": "memslot%d", "size": %d, "qom-type": "memory-backend-ram"}}`, plugged+1, size))
-		_, err = mon.Run(cmd)
+		// now add pc-dimm device
+		err = QmpAddMemoryDevice(target, memdevIdx)
 		if err != nil {
+			// device_add command failed... so try remove object that we just created
+			_ = QmpDelMemoryBackend(target, memdevIdx)
```
shayanh, line 646: What if the delete operation fails here? No error handling

shayanh (changes requested):
Some comments about comments and error handlnig.

shayanh:
This comment is about the qmp logic in general, not this PR specifically. I think we should make a `qmp` package and put all the qmp related stuff in there (separation of concerns). Prefixing related functions with a name (like `Qmp`) is a practice coming from languages such as C that didn't have package like Go. With the `qmp` package we would have a nicer API, for example, `qmp.DelMemoryBackend`.

neonvm/controllers/virtualmachine_qmp_queries.go
Omrigan: Right, thanks. We should have a linter for this.

sharnoff:
> I think we should make a `qmp` package and put all the qmp related stuff in there

Opened #716.

neonvm/controllers/virtualmachine_qmp_queries.go
shayanh: Which commit you mean? I think I'm looking at the latest version.

neonvm/controllers/virtualmachine_qmp_queries.go
```diff
@@ -229,72 +245,380 @@ searchForEmpty:
...
+		}
+		r.recorder.Event(r.vm, "Normal", "ScaleDown",
+			fmt.Sprintf("Removed memslot%d from VM %s", idx, r.vm.Name))
+		delete(r.memBackends, idx)
+	}
+	return attempts.DidSomething()
+}
+
+func (r *QmpMemorySetter) run() (int, error) {
+	fmt.Println("start run")
```
shayanh, line 558: probably you want to remove this or keep it with proper logging

neonvm/controllers/virtualmachine_qmp_queries.go
shayanh: I see, makes sense.

neonvm/controllers/virtualmachine_qmp_queries.go
Omrigan: You are looking at the latest version, but at the isolated 2nd commit. It's gone in the third commit https://github.com/neondatabase/autoscaling/pull/704/commits/98856c0f0287c899a8d9d831bd73020778a95086#diff-acc89b7771770ec55e48e065d33e10243a15901645298c0b241e06f7f67ed369L392

GitHub interface is quite confusing to do per-commit review :disappointed:

neonvm/controllers/virtualmachine_qmp_queries.go
Omrigan: Removed, thx

neonvm/controllers/virtualmachine_qmp_queries.go
shayanh: We can add docs here.

shayanh (approved):

Issue #706, opened 2023-12-21
Epic: Move ComputeUnit source of truth from scheduler to autoscaler-agent


sharnoff:
## Problem description / Motivation

The autoscaler-agent makes scaling decisions. It's weird that it needs to successfully communicate with the scheduler to be able to decide what it wants ­— it only knows the compute unit (and therefore, how much to scale) once it successfully receives a `PluginResponse`.

Also, moving this from the scheduler to the autoscaler-agent unblocks #615.

## Feature idea(s) / DoD

ComputeUnit is defined in the autoscaler-agent's config, and nowhere else.

## Tasks

```[tasklist]
# Tasks
- [ ] #707
- [ ] #743
```

```[tasklist]
# Follow-ups
- [ ] https://github.com/neondatabase/autoscaling/pull/744
```

shayanh:
Could you a little bit elaborate on this part?

> It's weird that it needs to successfully communicate with the scheduler to be able to decide what it wants

sharnoff:
> Could you a little bit elaborate on this part?

Sure — currently, because the compute unit is defined by the scheduler and the autoscaler-agent only knows the compute unit when it receives a response from the plugin, it's unable to decide how to scale the VM until after it's communicated with the scheduler plugin.

This also means that _in theory_ the compute unit could change between requests, which is actually the reason why this is required for #680 (otherwise we'd lose the guarantees around vm-monitor requests that #615 will provide)

sharnoff:
status: Waiting on review, low priority.

sharnoff:
status: Unchanged.

sharnoff:
status: Final PR has 1 approval, [just commented](https://github.com/neondatabase/autoscaling/pull/743#issuecomment-1951617959) asking if I should hold off merging to give a chance for another review.

PR #707, opened 2023-12-21
agent,plugin: Don't send ComputeUnit from plugin

sharnoff:
Instead, define it in the autoscaler-agent config.

This change starts the removal process of ComputeUnit from PluginResponse, so that the new protocol version expects that it won't be set - but to preserve backwards-compatibility during the upgrade, we need to keep the old behavior for now (hence the field is `omitempty`).

A follow-up change will update the plugin so that it *never* sends ComputeUnit, removing the field and the plugin's config along with it.

Part of #706.

---

Notes for review: This overlaps _a little_ with #653 — I think it's _probably_ easier to merge this first, release, and then merge that one.

~~Marking as a draft for now, until I've tested that e2e tests pass with old agent + new scheduler.~~ Tested locally and it appears to work.

### DIFF
--- a/pkg/agent/core/state_test.go
+++ b/pkg/agent/core/state_test.go
@@ -105,6 +105,7 @@ func Test_DesiredResourcesFromMetricsOrRequestedUpscaling(t *testing.T) {
 				ScalingEnabled: true,
 			},
 			core.Config{
+				ComputeUnit: api.Resources{VCPU: 250, Mem: 1},
 				DefaultScalingConfig: api.ScalingConfig{
 					LoadAverageFractionTarget: 0.5,
 					MemoryUsageFractionTarget: 0.5,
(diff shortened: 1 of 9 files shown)


pkg/api/types.go
```diff
@@ -58,9 +58,18 @@ const (
 	//
 	// * added AgentRequest.LastPermit
 	//
-	// Currently the latest version.
+	// Last used in release version <todo>.
```
sharnoff, line 61: > `<todo>`

Leaving this for now, will set it before merging

pkg/agent/core/state_test.go
```diff
@@ -1418,7 +1415,23 @@ func TestMetricsConcurrentUpdatedDuringDownscale(t *testing.T) {
 	}
 
 	// Send initial scheduler request - without the monitor active, so we're stuck at 4 CU for now
-	doInitialPluginRequest(a, state, clock, duration("0.1s"), DefaultComputeUnit, nil, resForCU(4))
+	a.
```
shayanh, line 1418: I guess this test is not directly related to this PR but it's fine.

shayanh (approved):
lgtm

pkg/api/types.go
Omrigan: And update `pkg/api/VERSIONING.md` probably

pkg/agent/core/state_test.go
sharnoff: Ah, right - I did this because the expected warnings changed here. But looking at this again, just realized that I can actually just pass `a.WithWarnings(...)` into `doInitialPluginRequest`, which should make this simpler.

pkg/api/types.go
sharnoff: Ah, yep! good catch - thanks!

Issue #453, opened 2023-08-02
Bug: VM memory scale-up leaves "Scaling" phase before updating status resources


sharnoff:
## Environment

Local development

## Steps to reproduce

Start VM with e.g. 1 memory slot, 0.1 CPUs. In one window, run `kubectl get -w neonvm`, and in another (while the first is ongoing), increase the memory slots:

```sh
kubectl patch $VM --type=json --patch '[{"op": "replace","path": "/spec/guest/memorySlots/use","value": 2}]'
```

The output of the `kubectl get -w` ("watch") shows something like:

```
postgres14-disk-test   100m   1Gi      postgres14-disk-test-295fb             Running   26s
postgres14-disk-test   100m   1Gi      postgres14-disk-test-295fb             Running   54s
postgres14-disk-test   100m   1Gi      postgres14-disk-test-295fb             Scaling   54s
postgres14-disk-test   100m   1Gi      postgres14-disk-test-295fb             Running   54s
postgres14-disk-test   100m   2Gi      postgres14-disk-test-295fb             Running   54s
```

## Expected result

The VM should leave the "Scaling" phase only once the resource usage in the status has been adjusted. (either at the same time, or after).

This isn't a _super_ important (especially because k8s is eventually consistent), but IMO the resource usage should not change unless we're in the "Scaling" phase.

## Actual result

The VM's phase is set to "Running" before the resources are updated.

## Links
- #350

shayanh:
This is intended. I mean it's not a bug and NeonVM controller intentionally works this way. Here is how NeonVM controller reconcile loop works given a VM object.

If VM is in `Running` phase, then update the resource values in status and update the status to  `Scaling`, if needed.
If VM is in `Scaling` phase, then update QEMU and cgroup stuff. If successful then update the status to `Running`. Note that *we don't update status resource values here*.

So here is the trace:

1. `vm.Status.Phase = Running` -> update `vm.Status.Phase` to `Scaling`.
2. `vm.Status.Phase = Scaling` ->  update qemu and cgroups, and update `vm.Status.Phase` to `Running`.
3. `vm.Status.Phase = Running` -> update `vm.Status.CPUs` and `vm.Status.MemorySize`.

I'd say this is not high priority issue. We can update the `vm.Status.CPUs` and `vm.Status.MemorySize` value while in the `Scaling` phase. I gotta think how to do it without breaking anything or making things more complicated.

sharnoff:
Status: waiting on further review from @cicdteam

PR #708, opened 2023-12-21
neonvm: Update VM resources status after scaling phase is done

shayanh:
Fixes https://github.com/neondatabase/autoscaling/issues/453.

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -435,9 +435,10 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 			var currentCPUUsage vmv1.MilliCPU
 			if supportsCgroup {
 				if cgroupUsage.VCPUs.RoundedUp() != pluggedCPU {
-					// Only report the mismatch here. We will resolve it in the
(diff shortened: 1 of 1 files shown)


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -694,6 +713,18 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 
 		// set VM phase to running if everything scaled
 		if cpuScaled && ramScaled {
+			// update status by CPUs used in the VM
```
cicdteam, line 716: when `cpuScaled=true` and  `memScaled=true`, it means that we did not perform any action in `VmScaling` since it was already scaled in the previous reconcile cycle. In other words - there is nothing to do and no need to update any statues other than moving `.status.phase` to `VmRunning`.

cicdteam (changes requested):

neonvm/controllers/virtualmachine_controller.go
shayanh: You're right. However, the fact that we didn't update status use in the scaling phase causes #453. Also see my comment on the issue. What do you think?

cicdteam (approved):

Issue #680, opened 2023-12-08
Bug: autoscaler-agent incorrectly assumes unsuccessful requests didn't succeed


sharnoff:
## Problem description / Motivation

I know this seems strange, but hear me out:

With the way that `pkg/agent/core/` is currently implemented, when it makes a request to an external entity (scheduler plugin / NeonVM API / vm-monitor), if there's an error in the request, it assumes that the request was not successfully processed.

Unfortunately this is not necessarily the case: It's entirely possible for example, that the autoscaler-agent could send a request to the scheduler plugin, and hit a client timeout before receiving the result of a successfully handled request. The scheduler plugin may not detect that the connection was closed in time.

This can lead to a possibly inconsistent state if the autoscaler-agent doesn't retry that *exact* request to completion: the scheduler (or whoever) could believe that there's a different amount reserved for a VM than the autoscaler-agent does.

## Implementation ideas

We already track the "current" or "approved" values for the three entities we communicate with. We could also track the lower/upper bound for amounts that they _may_ think we've requested and use e.g. the lower bound for what they've seen as the upper bound for what they've approved (or vice versa).

Here, it's important to be aware that the current NeonVM state is used both as an upper _and_ lower bound, for communications with the scheduler and vm-monitor, respectively. We can't "just" change the meaning of a single set of resources — we'll have to separately track a second set of resources as well.

shayanh:
What could go wrong with having inconsistent state between agent <-> scheduler or agent <-> vm-monitor?

My main concern is that agent's codebase already is the most complicated part of the system and this change makes it even more complicated.

sharnoff:
> What could go wrong with having inconsistent state between agent <-> scheduler or agent <-> vm-monitor?

for the scheduler, see above:

> This can lead to a possibly inconsistent state if the autoscaler-agent doesn't retry that exact request to completion: the scheduler (or whoever) could believe that there's a different amount reserved for a VM than the autoscaler-agent does.

After that point, we (a) may unintentionally overcommit, and (b) may end up with unexpected results from requests.

For the vm-monitor, this would mean that we could indefinitely run in a degraded state (file cache too big / too small), because currently there's no guarantee that we retry failed requests (it doesn't make sense to keep asking for scaling that you no longer think is correct).

sharnoff:
Status update: AFAICT:

1. implementing this is a blocker for #350.
2. this is blocked on #615 (and transitively, #706)

Issue #350, opened 2023-06-22
autoscaling e2e test is flaky


sharnoff:
It occasionally fails; haven't done any investigation as to why.

sharnoff:
From conversation: I think this is due to the same underlying problem as #23 — the agent assumes that (a) a successful NeonVM patch request means the VM has already been updated, and (b) that the VM's spec is only being changed by the agent.

(a) is incorrect because the request finishing is just updating the _k8s object_, but has no relation to the reconcile loop. (b) is incorrect because of the example highlighted above :)

sharnoff:
self-assigned; will be able to fix this after #371 is merged.

sharnoff:
Current status of this issue:
cc @shayanh

The immediate cause of this bug is that the autoscaler-agent currently assumes that (a) completion of requests to update VirtualMachine objects means that the underlying VM has been updated, and (b) no one else ever changes the VirtualMachine object.

https://github.com/neondatabase/autoscaling/blob/81083e20d647857b8ead3573581667cad5ac5424/pkg/agent/core/state.go#L913-L924

https://github.com/neondatabase/autoscaling/blob/81083e20d647857b8ead3573581667cad5ac5424/pkg/agent/core/state.go#L1090-L1095

Both of these assumptions are clearly false, but exist because resolving them is non-trivial, even after #371 (before, it was infeasible).

At the core of the difficulty is that we _want_ to provide simple, strong guarantees to the other components that the autoscaler-agent talks to — that the VM never uses more than the scheduler permits, and that it never goes below what the vm-monitor approves.

But because _in theory_ any change can happen to the VM at any time (and in practice, the control plane already modifies them directly), there's some design work needed to decide how the autoscaler-agent should respond to these edge cases.

One possible option is something like
- https://github.com/neondatabase/autoscaling/issues/554

---

In any case: I'm _generally_ working on this when I can find time, and will close this issue once the autoscaler-agent uses the actual values of the VM spec as reported by k8s events, rather than assuming its request was immediately enacted.

There's a bunch of work to be done after that, that I'll leave for other issues. See:

- https://github.com/neondatabase/autoscaling/issues/592
- https://github.com/neondatabase/autoscaling/issues/593
  - There's actually a similar issue here with downscaling and requests to the scheduler. The same approach will probably fix both.
- https://github.com/neondatabase/autoscaling/issues/594

PR #713, opened 2023-12-22
agent/core: Send monitor requests in 1 CU increments

### LINKED ISSUE #615: autoscaler-agent should request downscaling from high → low

Currently, the policy inside `pkg/agent/core.State` is that if the desired state is more than one compute unit less than what was last approved by the vm-monitor, then requests should be made from low to high.

This means that if there's a significant discrepancy between what the autoscaler-agent believes should be the desired resources vs what the vm-monitor is willing to accept, then there will periodically be as many requests as the size of the difference, rather than just a single request that gets denied.

## Implementation ideas

The most annoying part of fixing this will probably be updating the tests, which rely on specifics that'll change here.

sharnoff:
Before, we would request the "full" downscaling or upscaling that we desired - for downscaling, we'd retry with a higher amount if the initial request was denied.

To reduce load, it's better to only request decreasing by 1 CU at a time. Also, bounding the size of changes from vm-monitor requests will also help with managing uncertainty (see #680).

Fixes #615, building towards #350.

~~**NB: This PR builds on #707 and must not be merged before it.**~~

### DIFF
--- a/deploy/agent/config_map.yaml
+++ b/deploy/agent/config_map.yaml
@@ -24,7 +24,6 @@ data:
           "unhealthyStartupGracePeriodSeconds": 20,
           "maxHealthCheckSequentialFailuresSeconds": 30,
           "retryDeniedDownscaleSeconds": 5,
-          "deniedDownscaleFollowUpWaitSeconds": 1,
           "requestedUpscaleValidSeconds": 10,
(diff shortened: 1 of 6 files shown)


shayanh (approved):
LGTM! Only needs to be updated with the main branch.

PR #715, opened 2023-12-27
vm-kernel: cache image

bayandin:
This PR adds caching logic for `neondatabase:vm-kernel` image (it uses the latest git commit sha related to the kernel directory as a cache key)

Changes:
- All the stuff related to the kernel (Dockerfile + linux-config file) moved to a separate directory `neonvm/hack/kernel`

`vm-kernel` workflow:
- Gets the last commit sha of `neonvm/hack/kernel`
  - if an image with such a tag exists — return it to be used in e2e tests.
  - if the image doesn't exist — build an image and add cache tag for it, return new image for e2e tests
- accepts `force-rebuild` parameter, which forces CI to rebuild the image

Part of https://github.com/neondatabase/autoscaling/issues/660


.github/workflows/vm-kernel.yaml
```diff
@@ -28,39 +33,97 @@ on:
...
           else
             IMAGE=""
           fi
+
           echo "image=${IMAGE}" >> $GITHUB_OUTPUT
 
+      - name: login to docker hub
+        if: |
+          steps.get-kernel-image.outputs.image != '' &&
+          inputs.tag && !inputs.force-rebuild && !inputs.kernel-image-tag-override
```
sharnoff, line 105: Took a while for me to validate that this is correct - would be good to have a comment here explaining that we only need to login if

1. there's a preexisting image to use (`steps.get-kernel-image.outputs.image != ''`), and we don't need to rebuild it (`!inputs.force-rebuild`); and
2. we explicitly want to tag the image (`inputs.tag != ''`)

(didn't include `kernel-image-tag-override` because I'm not sure I understand the semantics)

.github/workflows/vm-kernel.yaml
```diff
@@ -87,25 +150,42 @@ jobs:
...
       - name: build linux kernel
         id: build-linux-kernel
         uses: docker/build-push-action@v3
         with:
           build-args: KERNEL_VERSION=${{ steps.get-kernel-version.outputs.VM_KERNEL_VERSION }}
-          context: neonvm/hack
+          context: neonvm/hack/kernel
           platforms: linux/amd64
           # Push kernel image only for scheduled builds or if workflow_dispatch/workflow_call input is true
           push: ${{ github.event_name == 'schedule' && 'true' || ( inputs.push || 'false' ) }}
```
sharnoff, line 183: If we're using this for caching, we now want to push always, right? (and I guess we can get rid of the schedule)

.gitignore
```diff
@@ -25,5 +25,6 @@ testbin/*
 
 *.qcow2
 neonvm/hack/vmlinuz
+neonvm/hack/kernel/vmlinuz
```
sharnoff, line 28: I'm torn whether to leave the old gitignore rule or not, but would lean more towards removing it and notifying people in slack.
Thoughts?

sharnoff (approved):
Just a few small things - otherwise looks good! Thanks!

.gitignore
bayandin: I'll create a follow-up PR which will remove old paths (we also have one in `make kernel` job), we can merge it in a month, for example

.github/workflows/vm-kernel.yaml
bayandin: Moved this to a separate step with comments. I also renamed a bunch of stuff to make it clear. Let me know what do you think

bayandin:
Examples:
- build custom kernel (with tag `test-pr-715`): [link](https://github.com/neondatabase/autoscaling/actions/runs/7463438768)
- trigger e2e tests with custom kernel (with tag `test-pr-715`): [link](https://github.com/neondatabase/autoscaling/actions/runs/7463537825)

.github/workflows/vm-kernel.yaml
sharnoff: Looks good :)

sharnoff (approved):

.gitignore
Omrigan: Why wait? What happens if we remove it now?

PR #718, opened 2023-12-29
lint: delete --fix from golangci-lint

Omrigan:
--fix is not supposed to be used in CI.

It just silently fixes the issue and doesn't commit it.


Omrigan:
I think we are not supposed to run --fix in the CI.

It just silently fixes the issue and doesn't commit it.

sharnoff (approved):
looks good (and good catch!). I came across the same import grouping issue you did while working on #721 and opened #720 to fix. We can go with either.

PR #721, opened 2023-12-29
Enable golangci-lint on neonvm/

sharnoff:
Supersedes #400. With the recent PRs making changes to neonvm, it seemed like enabling golangci-lint would be more valuable. And with the amount of work required, I figured it'd be easier for someone more involved to tackle.

~~This PR builds on #720, but #718 could be used instead.~~

---

Notes for review:

Happy to clarify any weird bits. There's a lot here.

There were a few function signatures changed - either because they were returning a `(struct, error)` tuple (which trips exhaustruct - ends up beings simpler to return `(*struct, error)`), or because they had an unused parameter (or, newly unused, after switching some of the functions in `virtualmachine_controller.go` to use `*api.VCPUCgroup`).

Because of the size of #704, it may be worthwhile to get this merged before it. However, it's probably easier to resolve merge conflicts if that merges first. cc @omrigan for your thoughts.

PR #722, opened 2023-12-29
lint: Remove unused lints and exemptions

sharnoff:
The additional printf functions have not been necessary since we switched away from klog (#323), and the exemption for containerd/cgroups types hasn't been necessary since we removed the vm-informant (#506).

Split out from #721.


Omrigan (approved):

Omrigan:
@sharnoff merge?

sharnoff:
Oops, thanks for the reminder :)

PR #723, opened 2023-12-29
neonvm-controller: Fix pprof server name (remove "scheduler")

sharnoff:
I must've missed this one during review, oops :sweat_smile:


shayanh (approved):
Thanks!

PR #717, opened 2023-12-29
runner: pass logs through virtio-serial

Omrigan:
virtio-serial is much more performant than /dev/console emulation, therefore, is much more suitable for the verbose postgres log.

A more comprehensive log solution is pending until we have systemd+syslog support to separate logs of different components.

Fixes neondatabase/cloud#8602

PR #724, opened 2024-01-03
runner: pass logs through virtio-serial

Omrigan:
virtio-serial is much more performant than /dev/console emulation, therefore, is much more suitable for the verbose postgres log.

A more comprehensive log solution is pending until we have systemd+syslog support to separate logs of different components.

Fixes neondatabase/cloud#8602


Omrigan:
rnamed branch
Old PR: https://github.com/neondatabase/autoscaling/pull/717

vm-examples/pg14-disk-test/10-neon-log-port.rules
```diff
@@ -0,0 +1 @@
+ATTR{name}=="tech.neon.log.0" MODE="0666"
```
sharnoff, line 1: nb: missing newline

neonvm/runner/main.go
```diff
@@ -779,6 +786,47 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
 	}
 }
 
+// writes from socket line by line
+func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) error {
```
sharnoff, line 790: Some of the control flow in this function seems undesirable.

As far as I can tell:

- If there's no logs from the VM, we wait for up to 3s after termination to decide we're done (corollary: This will increase runner pod termination time [and therefore suspend_compute time], by up to 3 seconds — may be fine in practice, but worth being confident that it's fine)
- If we get an error reading from the socket, we retry 3s later (IMO we should have exponential backoff here, with a _limit_ of a few seconds)
- If the context expires while we're reading log lines, then we will drop the rest (probably not what we want to do?), leaving the logs from QEMU shutdown unread.
    - imo: This is ok if we get an error. But if we're still reading logs, we should print those!

Not 100% sure on these, please correct me if I made a mistake.

It'd also be good to know what QEMU's behavior around the socket is when the VM exits — e.g. do we get a particular kind of error?

neonvm/runner/main.go
```diff
@@ -779,6 +786,47 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
...
+			reader = bufio.NewReader(conn)
+		}
+
+		select {
+		case <-ctx.Done():
+			conn.Close()
+			return ctx.Err()
+		default:
+			conn.SetReadDeadline(time.Now().Add(delay))
+			str, err := reader.ReadString('\n')
```
sharnoff, line 816: a couple things:

1. It's probably cheaper to keep the output as `[]byte` rather than converting to `string` and printing (in general: if the data is already `[]byte`, then strings are less efficient because they're guaranteed read-only)
2. AFAICT there's no length limit here — in theory, someone could send an arbitrarily long failed transaction, and then print several GB which we'd be forced to keep in memory. IMO we must have a limit to the amount we're buffering here.

sharnoff (commented):
Some comments, some questions.

Also: What's the purpose of the `tech.neon.log.0`? (do we need to redirect various program's output through that? or will we get the speed improvements even when continuing to write to `/dev/console`?)

Omrigan:
> do we need to redirect various program's output through that?

Yes. I've redirected postgres's output. 
 
> or will we get the speed improvements even when continuing to write to /dev/console?)

No, we won't.

neonvm/runner/main.go
```diff
@@ -807,6 +815,86 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
...
+		if errors.Is(err, os.ErrDeadlineExceeded) {
+			// We've hit the deadline, meaning the reading session was successful.
+			b.Reset()
+			return
+		}
+
+		if err != nil {
+			logger.Error("failed to read from socket", zap.Error(err))
+			conn = nil
+		}
```
sharnoff, line 877: Should we handle (EOF if ctx.Err() != nil) here and below? When trying this locally, the final logs are:

```
{"level":"info","ts":1706032966.0483866,"logger":"neonvm-runner.terminate-qemu-on-sigterm","caller":"runner/main.go:1111","msg":"got signal, sending powerdown command to QEMU"}
{"level":"info","ts":1706032966.0497675,"logger":"neonvm-runner.terminate-qemu-on-sigterm","caller":"runner/main.go:1132","msg":"system_powerdown command sent to QEMU"}
{"timestamp":"2024-01-23T18:02:46.051992Z","level":"ERROR","fields":{"message":"cgroup watcher terminated"},"target":"vm_monitor::runner","filename":"libs/vm_monitor/src/runner.rs","line_number":206}
The system is going down NOW!
Sent SIGTERM to all processes
udhcpc: received SIGTERM
2024-01-23T18:02:46.966903Z  INFO vector: Vector has stopped.
exiting
Sent SIGKILL to all processes
Requesting system poweroff
[   11.666647] ACPI: PM: Preparing to enter system sleep state S5
[   11.667749] reboot: Power down
{"level":"error","ts":1706032969.183614,"logger":"neonvm-runner","caller":"runner/main.go:875","msg":"failed to read from socket","error":"EOF","stacktrace":"main.forwardLogs.func1\n\t/workspace/neonvm/runner/main.go:875\nmain.forwardLogs\n\t/workspace/neonvm/runner/main.go:881"}
{"level":"info","ts":1706032969.2671506,"logger":"neonvm-runner","caller":"runner/main.go:718","msg":"QEMU exited without error"}
{"level":"error","ts":1706032969.2673178,"logger":"neonvm-runner","caller":"runner/main.go:890","msg":"failed to read from socket","error":"EOF","stacktrace":"main.forwardLogs\n\t/workspace/neonvm/runner/main.go:890"}
{"level":"info","ts":1706032969.2674258,"logger":"neonvm-runner","caller":"runner/main.go:814","msg":"shut down cpu_change server"}
```

note the "failed to read from socket" error logs, with "EOF" as the error.

neonvm/runner/main.go
```diff
@@ -807,6 +815,86 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
...
+	var conn net.Conn
+	var reader *bufio.Reader
+	b := &backoff.Backoff{
+		Min:    100 * time.Millisecond,
+		Max:    delay,
+		Factor: 2,
+		Jitter: true,
+	}
+
+	tryForward := func() {
```
sharnoff, line 848: I found the control flow here confusing. WDYT about moving `tryForward` so it's defined and immediately called inside the loop? something like
```go
func () {
    // ...
}()
```

sharnoff (approved):
Looks good! A handful of small comments

shayanh (approved):

neonvm/runner/main.go
```diff
@@ -583,6 +586,9 @@ func main() {
 		"-qmp", fmt.Sprintf("tcp:0.0.0.0:%d,server,wait=off", vmSpec.QMP),
 		"-qmp", fmt.Sprintf("tcp:0.0.0.0:%d,server,wait=off", vmSpec.QMPManual),
 		"-qmp", fmt.Sprintf("unix:%s,server,wait=off", qmpUnixSocketForSigtermHandler),
+		"-device", "virtio-serial",
+		"-chardev", fmt.Sprintf("socket,path=%s,server,nowait,id=log", logSerialSocket),
```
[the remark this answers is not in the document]
Omrigan: Thx, fixed in 326d605

neonvm/runner/main.go
```diff
@@ -807,6 +815,86 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
 	}
 }
 
+func drainLogsReader(reader *bufio.Reader) error {
+	for {
+		slice, err := reader.ReadSlice('\n')
```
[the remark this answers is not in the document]
Omrigan: Actually, `ReadSlice` returns no more than a size of `bufio.Reader`'s buffer. Apologies, I should've explained it in a comment, as I was researching specifically for this one.

Made it more explicit  in cd30ef980aeaaf81f630b52e4daa479a71814ac1

neonvm/runner/main.go
```diff
@@ -807,6 +815,86 @@ func listenForCPUChanges(ctx context.Context, logger *zap.Logger, port int32, cg
...
+		if err != nil {
+			if len(slice) != 0 {
+				// Print and add a newline to the slice we got
+				fmt.Printf("%s\n", slice)
+			}
+
+			return err
+		}
+
+		fmt.Printf("%s", slice)
```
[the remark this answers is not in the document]
Omrigan: Fixed in 09f3cfc665cbe88d30888480254b44cd04497953

neonvm/runner/main.go
Omrigan: sure, d65efa744be2d0a9e288f2b972b7224d5d5acb81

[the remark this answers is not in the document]
Omrigan: I agree, but I already changed it so definition and callsite are the same. Should be better now.

Omrigan:
Everything fixed according to the comments. 

During the last round of testing I realized `udevadm` doesn't work reliably. So instead I got rid of it in  5197cabdabecc24f7829a4e341fe74ea2ea1485d. What do think about these changes?

neonvm/tools/vm-builder/files/Dockerfile.img
```diff
@@ -28,6 +28,7 @@ RUN set -e \
 		flock \
 	&& mv /sbin/acpid         /neonvm/bin/ \
 	&& mv /sbin/udevd         /neonvm/bin/ \
+	&& mv /bin/udevadm        /neonvm/bin/ \
```
sharnoff, line 31: I guess this is no longer needed?

sharnoff (approved):
lgtm!

neonvm/tools/vm-builder/files/vminit
```diff
@@ -11,6 +11,10 @@ done
 
 # udev rule for auto-online hotplugged CPUs
 mkdir -p /lib/udev/rules.d
+# taken from 50-udev-default.rules
+echo 'SUBSYSTEM=="virtio-ports", KERNEL=="vport*", ATTR{name}=="?*", SYMLINK+="virtio-ports/$attr{name}"' > /lib/udev/rules.d/99-virtio-ports.rules
+
+echo 'ATTR{name}=="tech.neon.log.0" MODE="0666"' > /lib/udev/rules.d/99-neon-log-port.rules
```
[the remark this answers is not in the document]
Omrigan: I'd rather keep it as I plan to revert back the temporary change once `systemd` support is there

PR #726, opened 2024-01-05
neonvm: enable ssh access into the VMs

shayanh:
Simple SSH access into a NeonVM by running the `ssh guest-vm` from its runner pod.

## How it works

1. NeonVM Controller makes a secret for each VM. The secret has a pair of public and private keys.
2. NeonVM Controller mounts the secret on the runner pod. 
3. Runner pod shares the public key with the VM by mounting a cdrom disk. Runner pod have access to the private key and have ssh client installed.
4. VM Builder sets up an SSH server inside the VM and uses the shared public as an authorized key. The SSH server is accessible through the private network bridge between the runner pod and the VM (`169.254.254.252/30`). For convenience, we add VM's IP address to runner pod's `/etc/hosts` as `guest-vm`. VM's SSH server is not supposed to be exposed to the rest of the cluster. Currently we don't enforce this and one might expose the SSH server port through `.Spec.Guest.Ports`.

## API Changes

- Added `.Spec.EnableSSH` to VM resource object
- Added `.Status.SSHSecretName` to VM resource object

## Migration

Virtual machine migration controller requires the SSH secret to exist because we cannot unmount and mount the new secret into the VM after the live migration. If a VM's SSH secret is deleted accidentally then live migration won't be possible. We can prevent accidental deletion of SSH secrets by implementing a custom finalizer, but I leave that for another task.

**Other changes:**
- Removing legacy SSH stuff from Makefile, vm-examples	and scripts

---

I tested migration with presence of old running VMs and then deploying the new version of controller with SSH support. Haven't automated it yet but this is the scenario:

```
git checkout main
make deploy
kubectl apply -f vm-deploy.yaml

git checkout shayanh/vm-ssh
make deploy
kubectl apply -f vm-deploy-migration.yaml
```

and it works as expected.

Resolves https://github.com/neondatabase/cloud/issues/8029.


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -955,6 +990,44 @@ func (r *VirtualMachineReconciler) podForVirtualMachine(
...
+	// Set the ownerRef for the Secret
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
+	if err := ctrl.SetControllerReference(virtualmachine, secret, r.Scheme); err != nil {
+		return nil, err
+	}
+	return secret, nil
+}
+
+func sshSecretSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Secret, error) {
+	// with bitSize = 2048 it takes ~100ms
```
sharnoff, line 1008: I know this is still WIP - wanted to comment on this.

100ms is actually quite a lot of time for the global controller to spend CPU-bound during VM startup / creation. Sometimes (particularly, at midnight UTC) the console / control plane creates many VMs all at once.
(See e.g. [this logs query](https://neonprod.grafana.net/explore?schemaVersion=1&panes=%7B%22EL3%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22sum%20by%28cluster%29%20%28count_over_time%28%5Cn%7Bjob%3D%5C%22integrations%2Fkubernetes%2Feventhandler%5C%22%7D%5Cn%20%20%20%20%7C%20logfmt%5Cn%20%20%20%20%7C%20sourcecomponent%20%3D%20%5C%22virtualmachine-controller%5C%22%5Cn%20%20%20%20%7C%20msg%20%3D~%20%5C%22VirtualMachine%20.%2A%20created,%20pod%20.%2A%5C%22%5Cn%20%20%20%20%5B1s%5D%5Cn%29%29%22,%22queryType%22:%22range%22,%22refId%22:%22A%22,%22step%22:%221s%22,%22legendFormat%22:%22%7B%7Bcluster%7D%7D%22%7D%5D,%22range%22:%7B%22from%22:%221704326340000%22,%22to%22:%221704326460000%22%7D%7D%7D&orgId=1) that visualizes the number of VMs created each second [kind of - it's close enough])

From what I can remember (+ confirmed with a quick search, and then benchmarking locally), elliptic curve keys appear to be substantially quicker to generate for the same strength. For example, on my machine, 4096-bit RSA keys ~1s to generate (2048-bit is ~85ms, 3072-bit is ~390ms), whereas Ed25519 keys take ~13 µs (~16 µs if you include the marshaling).

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -319,6 +324,11 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		meta.SetStatusCondition(&virtualmachine.Status.Conditions, metav1.Condition{Type: typeAvailableVirtualMachine, Status: metav1.ConditionUnknown, Reason: "Reconciling", Message: "Starting reconciliation"})
 	}
 
+	// Generate ssh secret name
+	if len(virtualmachine.Status.SSHSecretName) == 0 {
+		virtualmachine.Status.SSHSecretName = virtualmachine.Name
```
Omrigan, line 329: Maybe you need to generate smth like `fmt.Sprintf("ssh-%s", virtualmachine.Name)`?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -364,8 +374,30 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		vmRunner := &corev1.Pod{}
 		err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
 		if err != nil && apierrors.IsNotFound(err) {
+			// Check if the ssh secret already exists, if not create a new one
+			sshSecret := &corev1.Secret{}
+			err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.SSHSecretName, Namespace: virtualmachine.Namespace}, sshSecret)
```
Omrigan, line 379: I'd define the struct here on separate line, and one line per field. Otherwise, line is too long.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -364,8 +374,30 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		vmRunner := &corev1.Pod{}
 		err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
 		if err != nil && apierrors.IsNotFound(err) {
+			// Check if the ssh secret already exists, if not create a new one
```
Omrigan, line 377: This block gets too nested, maybe extract it to a function?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -379,8 +411,9 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 			log.Info("Runner Pod was created", "Pod.Namespace", pod.Namespace, "Pod.Name", pod.Name)
 
 			r.Recorder.Event(virtualmachine, "Normal", "Created",
-				fmt.Sprintf("VirtualMachine %s created, pod %s",
-					virtualmachine.Name, pod.Name))
+				fmt.Sprintf("VirtualMachine %s created, Pod %s, SSH Secret %s",
```
Omrigan, line 414: This is not your change, but `virtualmachine.Name` is redundant, it's already included in an event.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -955,6 +989,43 @@ func (r *VirtualMachineReconciler) podForVirtualMachine(
 	return pod, nil
 }
 
+func (r *VirtualMachineReconciler) sshSecretForVirtualMachine(virtualmachine *vmv1.VirtualMachine) (*corev1.Secret, error) {
```
Omrigan, line 992: You should probably use same formatting as for the `podForVirtualMachine` function

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -955,6 +989,43 @@ func (r *VirtualMachineReconciler) podForVirtualMachine(
...
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      virtualmachine.Name,
+			Namespace: virtualmachine.Namespace,
+		},
+		Immutable: ptr(true),
+		Type:      corev1.SecretTypeSSHAuth,
+		Data: map[string][]byte{
+			"ssh-publickey":  publicKey,
```
Omrigan, line 1021: you sometimes use `ssh-publickey` and sometimes `ssh-public-key`. Maybe it should be universalized and defined as a constant, so that we don't mix it up?

neonvm/runner/main.go
```diff
@@ -1217,6 +1229,16 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 		return nil, err
 	}
 
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	record := fmt.Sprintf("%v neonvm-br\n", ipVm)
```
Omrigan, line 1237: why `-br`?

neonvm/runner/main.go
```diff
@@ -574,6 +578,14 @@ func main() {
 	// disk details
 	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=none", rootDiskPath))
 	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=none", runtimeDiskPath))
+	{
+		diskName := "ssh-authorized-keys"
+		diskPath := fmt.Sprintf("%s/%s.qcow2", mountedDiskPath, "ssh-authorized-keys")
```
Omrigan, line 583: 1. ISO9660 is not `.qcow2`, probably `.iso`
2. "ssh-authorized-keys" ->  diskName. And it probably can be a global constant.

neonvm/README.md
```diff
@@ -68,6 +70,12 @@ root@neonvm:~# curl -k https://kubernetes/version
 <press CTRL-a k to exit screen session>
 ```
 
+#### SSH
+
+```sh
+kubectl exec -it $(kubectl get neonvm vm-debian -ojsonpath='{.status.podName}') -- screen /dev/pts/0
```
Omrigan, line 76: I think you want to do ssh here

scripts/ssh-into-vm.sh
```diff
@@ -1,134 +0,0 @@
...
-DEFAULT_SSHCLIENT_IP="10.77.77.213"
-SSHCLIENT_IP="$( ( set -u; echo "$SSHCLIENT_IP" ) 2>/dev/null || echo "$DEFAULT_SSHCLIENT_IP" )"
-
-NODE_NAME="$NODE_NAME" # because -u isn't set yet, this sets NODE_NAME to "" if it's undefined
-
-# can't enable `set -u` for get_vm_name because we might not have $VM_NAME
-vm_name="$(get_vm_name)"
-set -u
-
-SECRET_NAME='vm-ssh'
```
Omrigan, line 31: Maybe it worth keeping the script up until this line, and replacing the rest with `kubectl exec $vm_name ssh ...`

That's basically what I did for qemu-monitor script.

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -218,8 +218,24 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 		targetRunner := &corev1.Pod{}
 		err := r.Get(ctx, types.NamespacedName{Name: migration.Status.TargetPodName, Namespace: vm.Namespace}, targetRunner)
 		if err != nil && apierrors.IsNotFound(err) {
+			// We require the SSH secret to exist because we cannot unmount and
+			// mount the new secret into the VM after the live migration. If a
+			// VM's SSH secret is deleted accidentally then live migration is
+			// not possible.
+			if len(vm.Status.SSHSecretName) == 0 {
```
Omrigan, line 225: Maybe you can abstract away this code, because you have very similar code in `doReconcile`. Not sure, maybe there are subtle differences.

neonvm/runner/main.go
```diff
@@ -574,6 +578,14 @@ func main() {
 	// disk details
 	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=rootdisk,file=%s,if=virtio,media=disk,index=0,cache=none", rootDiskPath))
 	qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=runtime,file=%s,if=virtio,media=cdrom,readonly=on,cache=none", runtimeDiskPath))
+	{
```
Omrigan, line 581: Cool. Didn't know Go had these scopes outside of `if`, `for`, etc

neonvm/runner/main.go
```diff
@@ -574,6 +578,14 @@ func main() {
...
+		diskName := "ssh-authorized-keys"
+		diskPath := fmt.Sprintf("%s/%s.qcow2", mountedDiskPath, "ssh-authorized-keys")
+		if err := createISO9660FromPath(logger, diskName, diskPath, "/vm/ssh"); err != nil {
+			logger.Fatal("Failed to create ISO9660 image", zap.Error(err))
```
Omrigan, line 585: This crashes the program, no?

neonvm/README.md
Omrigan: Why do we even have two very similar sections in this doc.

neonvm/controllers/virtualmachine_controller.go
shayanh: I felt that might be redundant since secrets type are `kubernetes.io/ssh-auth`. For example if you run `kubectl get secrets`, the output looks like this:
```
NAME                   TYPE                     DATA   AGE
example                kubernetes.io/ssh-auth   2      17h
postgres14-disk-test   kubernetes.io/ssh-auth   2      17h
```

Omrigan: In that case, maybe you don't need a separate `SSHSecretName` variable, and you can just use `virtualmachine.Name` directly?

neonvm/controllers/virtualmachine_controller.go
shayanh: I wanted to write it the same way the rest of code is written in the controller. I agree we should refactor controller's codebase with the right set of abstractions but I'd leave that for another task. I'll make an issue for it.

neonvm/controllers/virtualmachine_controller.go
shayanh: Good point. We should fix this in the entire controller codebase. I'll make an issue for it.

neonvm/controllers/virtualmachine_controller.go
shayanh: I tried to do the same, but I wanted to be a little bit more explicit so I named it `sshSecretForVirtualMachine` rather than `secretForVirtualMachine`.

neonvm/controllers/virtualmachine_controller.go
shayanh: resolved in 612191b8cfeff762ae8a499b3340a287521c6251

neonvm/controllers/virtualmachine_controller.go
shayanh: good catch. resolved in 584358c0fa8835e68bc81858d78efd9bfcc2cd48

neonvm/runner/main.go
shayanh: To show it's on the bridge network between the pod and the VM (it's named `br-def`). Better names are welcome.

neonvm/controllers/virtualmachine_controller.go
Omrigan: I am just saying, if there is no way `SSHSecretName` can ever be different, I'd not create another variable. Instead it can be made explicit by using setter and getter for the keys. 

But that's just ideas, I am fine with any way forward.

neonvm/runner/main.go
shayanh: resolved in 8bcd4de3773cd3089993246647a915bc1d7af8dd and 63d86fa16fed9b79ccd43de7ae242a2a504e290f.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: In the other controller we generate a ssh secret name but here we return an error.

neonvm/runner/main.go
shayanh: Why?

neonvm/controllers/virtualmachine_controller.go
shayanh: It's more than the codebase. I want to make the dependency explicit on the k8s object such that we can see it during operation as well.

cicdteam:
So, IIUC we will create another k8s object every time we run the computes - secrets with ssh keys. It seems that another step has been added to slow down the start of computes.

neonvm/controllers/virtualmachinemigration_controller.go
cicdteam: secrets mounted to VM as disks (qcow2) ? then they will be transferred to target QEMU on live migration as other block devices

neonvm/README.md
shayanh: fixed in cad4d78e271c58aa5fad51715780e0b707d6c83a. I'll make a separate issue for your second comment.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: Yes, we mount it as a disk (iso not qcow2). Code is here: https://github.com/neondatabase/autoscaling/blob/cad4d78e271c58aa5fad51715780e0b707d6c83a/neonvm/runner/main.go#L589

neonvm/runner/main.go
Omrigan: That's what https://pkg.go.dev/go.uber.org/zap#Logger.Fatal does.

neonvm/controllers/virtualmachine_controller.go
Omrigan: I meant the line too long and arguments should be on new line each. But now I see it's only one argument :fearful: 

Also, I think argument should be named `vm` and not `virtualmachine`.  "Variable names in Go should be short rather than long" after all [1]. 

I know that's not consistent throughout this file, but we should make it shorter everywhere.  

1: [Variable names in Go should be short rather than long](https://go.dev/wiki/CodeReviewComments#variable-names)

neonvm/runner/main.go
Omrigan: Maybe `neonvm-local`, dunno

neonvm/controllers/virtualmachine_controller.go
Omrigan: Can you explain this a bit in details?

neonvm/controllers/virtualmachine_controller.go
Omrigan: Why didn't you define a constant for those? Doing this prevents such inconsistencies.

neonvm/controllers/virtualmachinemigration_controller.go
Omrigan: > secrets mounted to VM as disks (qcow2) ? then they will be transferred to target QEMU on live migration as other block devices

Yes, but k8s secret is still needed to put private key on destination.

neonvm/controllers/virtualmachine_controller.go
shayanh: Let's say you're investigating a VM object and run `kubectl get neonvm vmname -o yaml`. Now in the status part you'll see the `sshSecretName` and the `podName`. The dependencies here are clear.

neonvm/controllers/virtualmachine_controller.go
shayanh: I agree but I want to keep it consistent with the rest of code in the controller.

neonvm/runner/main.go
shayanh: Yes that was intended. So you suggest just printing an error instead of exiting the program here?

neonvm/controllers/virtualmachinemigration_controller.go
sharnoff: I think actually we don't need to check the secret here.

IIUC: We only get to `vmv1.VmmPending` iff the VM is set to `vmv1.VmPreMigrating`, which can only happen if it first goes through `vmv1.VmRunning`, which requires the secret to be created.

There's an open question about what we should do for old VMs, that don't have an ssh secret created because they're already in `VmRunning`. IMO the correct way to handle that is to have a default-false flag for whether to enable SSH, with `SSHSecretName` getting generated iff it's empty and ssh is enabled. This would unfortunately require _also_ modifying the control plane as part of the rollout here, but I think the operational overhead is worth it for the clarity around how old VMs are handled.

neonvm/runner/main.go
```diff
@@ -588,7 +602,7 @@ func main() {
 			}
 			qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=virtio,media=disk,cache=none%s", disk.Name, dPath, discard))
 		case disk.ConfigMap != nil || disk.Secret != nil:
-			dPath := fmt.Sprintf("%s/%s.qcow2", mountedDiskPath, disk.Name)
+			dPath := fmt.Sprintf("%s/%s.iso", mountedDiskPath, disk.Name)
```
sharnoff, line 605: why this change? if unrelated, let's make it a separate PR (this one is already quite complex in terms of number of files / points in code changed)

neonvm/runner/main.go
```diff
@@ -1217,6 +1231,16 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 		return nil, err
 	}
 
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
sharnoff, line 1234: It's not immediately clear what this section does. Let's add a brief comment?

neonvm/runner/main.go
sharnoff: IMO: `guest` or `neonvm-guest` or something. That way, the ssh looks like
```sh
kubectl exec -it <pod name> -- ssh guest
# ... or ...
kubectl exec -it <pod name> -- ssh neonvm-guest
```

basically: the thing we're connecting to isn't the bridge network itself, it's the VM.

maybe even just `ssh vm` ?

neonvm/tools/vm-builder/files/vminit
```diff
@@ -50,3 +50,11 @@ resize2fs /dev/vda
 # networking
 ip link set up dev lo
 ip link set up dev eth0
+
+# ssh
+/neonvm/bin/ssh-keygen -A
```
sharnoff, line 55: How long does this take? From the docs of `-A`:

> Generate host keys of all default key types (rsa, ecdsa, and ed25519) if they do not already exist

Maybe we should only use ed25519 for efficiency?

neonvm/tools/vm-builder/files/vminit
```diff
@@ -50,3 +50,11 @@ resize2fs /dev/vda
...
+
+# setup sshd user to support sshd UsePrivilegeSeparation
+mkdir -p /var/empty
+/neonvm/bin/addgroup sshd
+/neonvm/bin/adduser -D -H -G sshd -g 'sshd privsep' -s /neonvm/bin/nologin sshd
```
sharnoff, line 60: Does this need to run at startup? or can it be built into the image

neonvm/runner/main.go
Omrigan: Yeah, programs should not exit from the middle of the stack. There might be defers, which should be executed, proper outer logging. We might at some point decide to recover from the error, e.g. skip ssh entirely.

neonvm/controllers/virtualmachine_controller.go
shayanh: Changed in to `ssh-neonvm-<name>`: 6383f4e7a2e9a73569799ee85a8dc667358afcd4

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: > IIUC: We only get to vmv1.VmmPending iff the VM is set to vmv1.VmPreMigrating, which can only happen if it first goes through vmv1.VmRunning, which requires the secret to be created.

Right, but as I wrote in the comment above the secret might get deleted accidentally (for example, by human). So that's why I have the check here.

> There's an open question about what we should do for old VMs, that don't have an ssh secret created because they're already in VmRunning. IMO the correct way to handle that is to have a default-false flag for whether to enable SSH, with SSHSecretName getting generated iff it's empty and ssh is enabled. This would unfortunately require also modifying the control plane as part of the rollout here, but I think the operational overhead is worth it for the clarity around how old VMs are handled.

Good point about old VMs that don't have SSH secret. We can make an assumption here that if the `sshSecretName` is empty then the VM doesn't have an SSH secret and we should generate one here. I didn't get your point about control plane involvement.

neonvm/controllers/virtualmachine_controller.go
shayanh: Opened https://github.com/neondatabase/autoscaling/issues/733

neonvm/controllers/virtualmachine_controller.go
shayanh: Opened https://github.com/neondatabase/autoscaling/issues/733

neonvm/runner/main.go
shayanh: Currently we don't have defers in runner's main.go and we have used `log.Fatal` everywhere in there. The question is if we want to consider this error fatal or not. @sharnoff what do you think?

.github/workflows/lint.yml
```diff
@@ -53,6 +53,9 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: 'go.mod'
```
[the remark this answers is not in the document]
shayanh: We hit the issue here, didn't want to create a PR for one line change but I can do that.

neonvm/README.md
```diff
@@ -68,6 +70,12 @@ root@neonvm:~# curl -k https://kubernetes/version
 <press CTRL-a k to exit screen session>
 ```
 
+#### SSH
+
+```sh
+kubectl exec -it $(kubectl get neonvm vm-debian -ojsonpath='{.status.podName}') -- ssh neonvm-br
+```
+
```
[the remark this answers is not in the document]
shayanh: Fixed in 721ad6b7a9896a0f68e9f91fa03affd52869b895

.github/workflows/lint.yml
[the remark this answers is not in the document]
sharnoff: eh, no reason not to - one line PRs are quick to review :wink:

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -390,6 +390,8 @@ type VirtualMachineStatus struct {
 	CPUs *MilliCPU `json:"cpus,omitempty"`
 	// +optional
 	MemorySize *resource.Quantity `json:"memorySize,omitempty"`
+	// +optional
+	SSHSecretName string `json:"secretName,omitempty"`
```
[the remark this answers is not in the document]
shayanh: Fixed in 7a5b1640895dcd468a9964db5190a155c13f408f

neonvm/runner/main.go
sharnoff: Yeah, we're already using `log.Fatal` in a bunch of places. Practically speaking, it doesn't _really_ matter whether we do cleanup because the entire container gets cleaned up when `neonvm-runner` exits.

But this is still kind of "spooky action at a distance" and we should mark it as tech debt and plan to resolve sometime.

Omrigan: Ah, that's `main()`, I see. Not used to code being in the `main()`. More used to smth like:
```
func run() error {
 ....
}

func main() {
  err := run()
  if err !=  nil {
     log.Error("error happened")
  }
}
```

The reason the code should be written like this is that when someone decides to move code somewhere else, like in a library, and there is no need to replaces `Fatal`s with returning error.

But since that's how code was before you, you can leave it as-is. 

> The question is if we want to consider this error fatal or not

I don't think it is very realistic this error happens, or even if it does, there is not much reason to run the VM without ssh. So it can be fatal.

neonvm/controllers/virtualmachinemigration_controller.go
sharnoff: > secret might get deleted accidentally

IIRC we can prevent that from happening though — I'm not sure if the owner ref alone does it, but a finalizer certainly would.
In any case, I don't think we should accommodate that; it's highly unlikely that someone will run `kubectl delete secret` in prod unless they really know what they're doing (most of that stuff is automated by helm!)

sharnoff: > We can make an assumption here that if the sshSecretName is empty then the VM doesn't have an SSH secret and we should generate one here. I didn't get your point about control plane involvement.

basically: VMs _can't_ support ssh access unless they have the vm-builder update, right? So there's no benefit from adding complexity to regenerate ssh keys for VMs that don't already have them, because the image won't support it either way.
So, I'm suggesting that we should have another boolean field on the VM spec to enable ssh access, that defaults to false (i.e. all existing VMs don't have them). The downside is that we'd have to _also_ update the control plane to enable this boolean, but IMO that's worth added clarity we get about handling old VMs.

shayanh: > basically: VMs can't support ssh access unless they have the vm-builder update, right? So there's no benefit from adding complexity to regenerate ssh keys for VMs that don't already have them, because the image won't support it either way.

Ah right, forgot about vm-builder image.

> So, I'm suggesting that we should have another boolean field on the VM spec to enable ssh access, that defaults to false (i.e. all existing VMs don't have them). The downside is that we'd have to also update the control plane to enable this boolean, but IMO that's worth added clarity we get about handling old VMs.

Makes sense. Nothing better comes to my mind.

neonvm/runner/main.go
shayanh: Fixed in 9c2169d044dc64d2921ae400688a79bc0bf1df25

neonvm/runner/main.go
shayanh: I'd love to go to with something concise (like `vm`). The caveat is that we cover this DNS name. For example, in the future we can't have a k8s service named `vm` and access that through DNS.

sharnoff: That's a good point. Maybe let's ask in #dev for feedback? this will impact anyone who connects into a VM, so probably worthwhile. Can just be a simple poll

neonvm/controllers/virtualmachinemigration_controller.go
sharnoff: pasting something else from DMs: I changed my mind - IMO actually it's worth the weirdness to have ssh secret recreation in the migration controller — we do runner pod recreation in both places, and so we should be willing to recreate their dependencies (i.e. the ssh secret) in both places as well.
In a future change, we can unify both into a single function.

.github/workflows/lint.yml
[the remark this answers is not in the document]
shayanh: Opened #734.

neonvm/runner/main.go
shayanh: Opened #735

neonvm/tools/vm-builder/files/vminit
shayanh: Openssh is quite fast, this takes ~15ms. I'll try to see how it works with only one pair of keys.

neonvm/tools/vm-builder/files/vminit
shayanh: I don't think can be built into the image since the contents of `/etc/group` and `/etc/passwd` depends on the base image that vm-builder uses.

sharnoff: Right, but we can run it as part of vm-builder, right? e.g. in `vm-builder/files/Dockerfile.img`, similarly to how we install `vector`

neonvm/tools/vm-builder/files/vminit
sharnoff: Hm... when trying this out in a debian container locally, I see that on install, it generates keys, but manually specifying a different path to generate them in causes it to take ~0.5s:

```
$ docker run -it debian:latest bash
root@6eb1acea5942:/# apt update && apt install -y openssh-server
... output omitted ...
root@6eb1acea5942:/# mkdir -p foo/etc/ssh
root@6eb1acea5942:/# time ssh-keygen -A -f foo
ssh-keygen: generating new host keys: RSA ECDSA ED25519 

real	0m0.659s
user	0m0.655s
sys	0m0.004s
```

(got varied results on re-runs)

neonvm/controllers/virtualmachinemigration_controller.go
Omrigan: > I'm suggesting that we should have another boolean field on the VM spec to enable ssh access

Why not treat presence or absence of `SSHSecretName` as an indicator for the ssh access?

neonvm/tools/vm-builder/files/vminit
shayanh: We install vector by copying its files into the rootdisk:
https://github.com/neondatabase/autoscaling/blob/a39cf89f7f799143e60f622ef527ced5bb13efc6/neonvm/tools/vm-builder/files/Dockerfile.img#L59-L62

https://github.com/neondatabase/autoscaling/blob/a39cf89f7f799143e60f622ef527ced5bb13efc6/neonvm/tools/vm-builder/files/Dockerfile.img#L84

I don't want to mess up with base image's `/etc/passwd`.

neonvm/runner/main.go
shayanh: I think it's overkill to ask in #dev. Let's go with something safe like `guest-vm`.

neonvm/controllers/virtualmachinemigration_controller.go
sharnoff: If we exclusively use presence/absence of `SSHSecretName`, semantics are less clear: If a VM is created with `SSHSecretName`, should it auto-generate a new secret? or should it rely on an existing one. IMO the behavior should not depend on the current cluster state (e.g. whether the secret exists). There's also concerns if many VMs use the same secret — if the secret doesn't already exist, we'd (depending on behavior) end up auto-generating a secret that's only owned by a single one of those VMs.

neonvm/tools/vm-builder/files/vminit
sharnoff: I don't quite understand - what's the difference between adding a user in the image vs doing it on startup? is it not the same deterministic change, just made at different times?`

neonvm/runner/main.go
sharnoff: Sure, `guest-vm` works.
Part of the reason I suggest #dev is that it also adds visibility onto upcoming improvements. Many people care about this, would have useful feedback, and would be excited to hear about the work you're doing.
But ultimately I'll leave it up to you.

neonvm/tools/vm-builder/files/vminit
shayanh: Anyway it's better to only use `ed25519` keys since the rest would be useless. Resolved here 4c4e56a6b9cd41a4c50a5ca8662dcee8e265d4e1

neonvm/tools/vm-builder/files/vminit
shayanh: Moved user creation to docker build: 1a13ff5e5a60f33ccd694a3aa7a53e0e6deb66bc

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: @sharnoff a VM cannot be created with `SSHSecretName` since it's part of the status and readonly for the user.

When we create a runner pod for a VM, we should make it in the same way in both controller (vm and migration). Currently, migration controller requires the VM to have the SSH secret, otherwise, it returns an error. It turned out that this approach doesn't work with the existing VMs. Now, It's better to treat the existing and new VMs the same way when we create pods for them. This helps with being consistent avoid trapping into edge cases.

So I'll go with creating secret in the migration controller.

neonvm/runner/main.go
shayanh: I'll announce once we deployed on staging.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: implemented in 896a8b298f257208132f79aee8f23a12b099f68e

neonvm/runner/main.go
shayanh: renamed in dbd1f507ad4f095903c1fa5fa9cffd719232649b

neonvm/tools/vm-builder/files/vminit
```diff
@@ -50,3 +50,6 @@ resize2fs /dev/vda
 # networking
 ip link set up dev lo
 ip link set up dev eth0
+
+# ssh
+/neonvm/bin/ssh-keygen -t ed25519 -f /etc/ssh/ssh_host_ed25519_key -N ""
```
sharnoff, line 55: Do we need `-N ""` here? if so, why?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1401,3 +1513,44 @@ func getEnvVarValue(envVarName string) (string, error) {
...
+	sshPublicKey, err := ssh.NewPublicKey(publicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	pubKeyBytes := ssh.MarshalAuthorizedKey(sshPublicKey)
+	return pubKeyBytes, nil
+}
+
+func ptr[T any](t T) *T { return &t }
```
sharnoff, line 1556: couple things about this function

1. elsewhere in the controllers package, it seems like instead of using e.g. `ptr[T](...)` we use `&[]T{...}[0]` — maybe just go with that instead of this function? (although in the future, it'd probably be better to use a function for this in all the places)
2. this is defined quite far from any other usages - if we keep, let's relocate?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1401,3 +1513,44 @@ func getEnvVarValue(envVarName string) (string, error) {
 	}
 	return value, nil
 }
+
+func sshKeygen() ([]byte, []byte, error) {
```
sharnoff, line 1517: let's either add docs or name the return values or both? right now, I have to read the function body to know what the return values mean.

neonvm/runner/main.go
```diff
@@ -234,8 +237,11 @@ func createISO9660runtime(diskPath string, command, args, sysctl []string, env [
 		}
 	}
 
+	var mounts []string
+	mounts = append(mounts, "/neonvm/bin/mkdir -p /mnt/ssh")
+	mounts = append(mounts, "/neonvm/bin/mount -o ro,mode=0644 $(/neonvm/bin/blkid -L ssh-authorized-keys) /mnt/ssh")
```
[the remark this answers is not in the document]
shayanh: 1. Before mounting a disk to a directory, we have to make sure that the directory exists.
2. We can modify the disks although I think it's better to not mutate `VM.Spec` and handle special cases like this explicitly.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1073,7 +1147,7 @@ func imageForVmRunner() (string, error) {
 	return image, nil
 }
 
-func podSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Pod, error) {
+func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*corev1.Pod, error) {
```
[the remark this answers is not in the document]
shayanh: This kind of ensures that the secret exists before Pod creation. So creating the pod won't result in the "secret not exists" error.

neonvm/tools/vm-builder/files/Dockerfile.img
```diff
@@ -61,22 +61,38 @@ RUN set -e \
...
+COPY --from=vm-runtime /neonvm /neonvm
+# setup sshd user to support sshd UsePrivilegeSeparation
+RUN /neonvm/bin/addgroup sshd || true \
+	&& /neonvm/bin/adduser -D -H -G sshd -g 'sshd privsep' -s /neonvm/bin/nologin sshd || true
+
 FROM vm-runtime AS builder
 ARG DISK_SIZE
-COPY --from=rootdisk / /rootdisk
+COPY --from=rootdisk-mod / /rootdisk
 COPY --from=vm-runtime /neonvm /rootdisk/neonvm
```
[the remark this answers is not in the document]
shayanh: You're right, we can remove this line actually.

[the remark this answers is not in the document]
shayanh: Resolved in 448d658c292c6d0cfa5176d9c33be4283d329897

neonvm/tools/vm-builder/files/vminit
shayanh: We don't want to set a passphrase for our keys here. I'll add a comment.

shayanh: fixed in fe99970e9c88fe1c4fa450c5503324e726da41ac

neonvm/controllers/virtualmachine_controller.go
shayanh: fixed in fe99970e9c88fe1c4fa450c5503324e726da41ac

neonvm/runner/ssh_config
```diff
@@ -0,0 +1,3 @@
+Host *
+    StrictHostKeyChecking accept-new
```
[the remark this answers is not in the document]
shayanh: fixed in fe99970e9c88fe1c4fa450c5503324e726da41ac

neonvm/controllers/virtualmachine_controller.go
shayanh: Removed the `ptr` function and used the array hack in 837a8fc1fc93a32fd12ef2a780dafa9d40fd208c. Will do the improvements in #733.

neonvm/tools/vm-builder/files/vminit
sharnoff: My understanding is that was already the case- what happens without it?

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Fair enough, makes sense

neonvm/runner/ssh_config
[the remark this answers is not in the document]
sharnoff: awesome, comment is super clear

neonvm/runner/main.go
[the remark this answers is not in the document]
shayanh: we do it here: https://github.com/neondatabase/autoscaling/blob/837a8fc1fc93a32fd12ef2a780dafa9d40fd208c/neonvm/runner/main.go#L246

neonvm/tools/vm-builder/files/vminit
shayanh: VM gets blocked on this command

neonvm/runner/main.go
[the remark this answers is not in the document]
sharnoff: Oh oops, nevermind then :sweat_smile:

thanks for the clarity

neonvm/tools/vm-builder/files/vminit
sharnoff: Huh, funky. Fair enough

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -111,6 +111,11 @@ type VirtualMachineSpec struct {
 	// Override for normal neonvm-runner image
 	// +optional
 	RunnerImage *string `json:"runnerImage,omitempty"`
+
+	// Enable SSH on the VM. It works only if the VM image is built using VM Builder that
+	// has SSH support (TODO: mention VM Builder version).
+	// +optional
```
sharnoff, line 117: let's also make this default to `false`?

Then, when we go to release this, there can be a one-line change to make it default to `true`

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -319,6 +324,16 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		meta.SetStatusCondition(&virtualmachine.Status.Conditions, metav1.Condition{Type: typeAvailableVirtualMachine, Status: metav1.ConditionUnknown, Reason: "Reconciling", Message: "Starting reconciliation"})
 	}
 
+	enableSSH := false
+	if virtualmachine.Spec.EnableSSH != nil && *virtualmachine.Spec.EnableSSH {
+		enableSSH = true
+	}
```
sharnoff, line 330: IIUC, if we set the default via kubebuilder, we should be ok to assume that it's always set here.

Might also be reasonable to have a `sshEnabled(*vmv1.VirtualMachine)` sort of function to handle this - then when we deprecate `EnableSSH` to make it always-enabled, we're just changing one function.

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -218,8 +218,31 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
+			if enableSSH {
+				// We require the SSH secret to exist because we cannot unmount and
+				// mount the new secret into the VM after the live migration. If a
+				// VM's SSH secret is deleted accidentally then live migration is
+				// not possible.
+				if len(vm.Status.SSHSecretName) == 0 {
+					err := errors.New("VM .Status.SSHSecretName is empty")
+					log.Error(err, "Failed to get VM's SSH Secret")
+					return ctrl.Result{}, err
+				}
```
sharnoff, line 235: Not sure if it's possible to get here, but I don't really see any issue with checking this

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -218,8 +218,31 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
+				enableSSH = true
+			}
+			var sshSecret *corev1.Secret
+			if enableSSH {
+				// We require the SSH secret to exist because we cannot unmount and
+				// mount the new secret into the VM after the live migration. If a
+				// VM's SSH secret is deleted accidentally then live migration is
+				// not possible.
+				if len(vm.Status.SSHSecretName) == 0 {
+					err := errors.New("VM .Status.SSHSecretName is empty")
```
sharnoff, line 232: It'd be good to mention that we're expecting it not to be empty because SSH is enabled

sharnoff (approved):
a couple comments, but otherwise lgtm! great stuff!

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: As for now yes, but things might change in the future so it's better to check it.

neonvm/apis/neonvm/v1/virtualmachine_types.go
shayanh: Makes sense. I'll make the change.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: fixed here c2a8064f6bc63524d9e645fe33e9c3a92b6b0fff

neonvm/apis/neonvm/v1/virtualmachine_types.go
shayanh: added the default value: b8de84026b60a658e05673f703946682e5580c57

neonvm/controllers/virtualmachine_controller.go
shayanh: changed it here: b8de84026b60a658e05673f703946682e5580c57

PR #730, opened 2024-01-08
lint: add lint to Makefile

Omrigan:
It is convenient to run when developing locally.


sharnoff (approved):
it'd be nice to also include codespell in this (of the things included in the lints workflow, it's the only one that's _relatively_ cheap to run).

But for now, probably good enough :)

Issue #712, opened 2023-12-22
Bug: VM clocks drift (need to have ntpd or equivalent)


sharnoff:
## Environment

Production

## Steps to reproduce

_Presumably_, leave a VM running for a long time — haven't yet validated.

## Expected result

The clock in the VM should not significantly drift — at minimum, it shouldn't be off by more than 0.1s, but _ideally_ it'd remain within 1ms. There may be difficulties in practice from cgroup CPU throttling.

## Actual result

On startup, the clock is roughly synchronized, but we've seen noticeable differences in the past (IIRC, 0.2s), and a user recently reported a more significant difference - ref https://neondb.slack.com/archives/C04DGM6SMTM/p1703259018496809.

## Other logs, links

- ...

sharnoff:
Notes from discussion:
1. There's a way to get QEMU to do the synchronization for us
2. AWS has a clock syncing service
3. @sharnoff to work on this

m0r0zzz:
In the case of KVM virtual machines, running ntpd in the guest OS to synchronize and syntonize its clock to the host one may be a suboptimal solution due to the usage of simple software skb timestamps in the virtio net driver - when combined with the non-deterministic nature of hypervisor time stealing, this can lead to increased guest clock offset and frequency jitter.

Using solutions provided by QEMU (i.e. setting time periodically via qemu-guest-agent or using its PV RTC clock to periodically set the CLOCK_REALTIME clock in a custom agent) may provide less jitter but may lead to problems related to the incorrect behavior of various software in an environment with frequent realtime clock steps (in my experience, some programs can't handle time "jumps" at all; this may lead to timed events being missed because the required time value has been "cut out" from the CLOCK_REALTIME time axis by the aforementioned jump).

There are other time synchronization solutions that can be recommended in this use case (in increasing precision order):
[...]

PR #732, opened 2024-01-10
neonvm: clock synchronization using kvm_ptp

shayanh:
Synchronizing VM clocks to host using kvm_ptp. We enable PTP clock (and the KVM related directive) on the kernel and use chrony on the VM as a server. More info [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/chap-kvm_guest_timing_management#sect-KVM_guest_timing_management-Host-guest-time-sync).

`chronyd` gradually synchronizes the clock by speeding it up or down. If the clock drift is large then this might take a long time. However, here we avoid jumps in the time, which is not ideal from user's perspective. Chrony's progress can be tracked using `chronyc tracking` command.

Fixes https://github.com/neondatabase/autoscaling/issues/712


m0r0zzz:
I'd like to provide a small clarification here, concerning the usage of "PTP protocol" and "hardware timestamping":
In this use case neither PTP protocol, nor hardware TS is used. Instead, chronyd uses PTP hardware clock (timer) directly as reference clock, periodically reading its value and synchronising system clock to it.

In real network hardware, IEEE 1588-compatible Ethernet PHY devices always contain a built-in timer clocked directly from Ethernet link frequency (25/125 MHz), and can be instructed to place a "packet start timestamp" at arbitrary offset of any received network packet. Also, in most of them the aforementioned timer can be read directly, either via MDIO interface or using special link-control Ethernet frames - in this case it is referred as "PHC" (PTP hardware clock).

When PTP protocol is used, Ethernet MAC instructs PHY to place the RX timestamp in the incoming PTP packet, at the relevant field offset (e.g. receiveTimestamp in DelayResp message), which allows to attain the ultimate packet timestamping accuracy and as such minimize the clock delay acquisition jitter.

In this case, Chrony does not use PTP protocol at all. Instead, it uses PHC time value directly as reference and synchronizes system time to it. When "real" PHC is used as reference clock there is a problem of unknown offset between PHC time and system time, so "pps" refclock option must be used - in this case chrony will only synchronize clock frequency, an won't try to correct the offset. kvm_ptp PHC is special, as it essentially returns UNIX time of the host machine when read, so full clock synchronization is the optimal solution here.

PHC is used here by KVM as convenient device model for the standalone reference clock, supported as-is by several time sync daemons. AFAIK, you can't use it for PTP protocol message timestamping.

shayanh:
@m0r0zzz thanks for the detailed explanation. I updated my PR description above. IIUC you mentioned kvm_ptp is special, so in this case only syncing to PHC would be enough on the chrony server (`refclock PHC /dev/ptp0 poll 2`).

neonvm/README.md
```diff
@@ -74,6 +74,14 @@ root@neonvm:~# curl -k https://kubernetes/version
 kubectl delete neonvm vm-debian
 ```
 
+### Clock synchronization
+
+We Synchronize VM clocks to host using PTP with hardware time stamping. We enabled PTP clock (and the KVM related directive) on the kernel and use chrony on the VM as a server. You can checkout chronyd log at `/var/log/chrony/chrony.log` and use the commands below:
```
Omrigan, line 79: You might want to remove the mention of hardware time stamping from here.

shayanh: fixed in d58e7963034d93e614bd8b510d051b73a8444606

Omrigan (approved):

neonvm/tools/vm-builder/files/vminit
```diff
@@ -50,3 +50,4 @@ resize2fs /dev/vda
 # networking
 ip link set up dev lo
 ip link set up dev eth0
+
```
sharnoff, line 53: unrelated change?

neonvm/tools/vm-builder/files/Dockerfile.img
```diff
@@ -61,19 +61,47 @@ RUN set -e \
...
 COPY vminit      /neonvm/bin/vminit
 COPY vmstart     /neonvm/bin/vmstart
 COPY vmshutdown  /neonvm/bin/vmshutdown
 COPY vmacpi      /neonvm/acpi/vmacpi
 COPY vector.yaml /neonvm/config/vector.yaml
+COPY chrony.conf /neonvm/config/chrony.conf
 RUN chmod +rx /neonvm/bin/vminit /neonvm/bin/vmstart /neonvm/bin/vmshutdown
 
+# rootdisk modification
+FROM rootdisk AS rootdisk-mod
```
sharnoff, line 93: this `rootdisk-mod` is the same change as in #726, right?
out of curiosity, what's the plan w.r.t. these two PRs?

sharnoff (approved):
lgtm. Were you able to reproduce the clock drift locally? (if so, what's the before/after like?)

neonvm/tools/vm-builder/files/Dockerfile.img
shayanh: I'll resolve the conflict manually.

PR #734, opened 2024-01-11
Use the right Go version in lint workflow

shayanh:


sharnoff (approved):
cc @bayandin in case there's some subtlety I'm not aware of

bayandin (approved):
Makes sense!

`make manifests` uses `controller-gen`, which installs the binary via `go install ...`.

I wonder, did it break something with default go (from `ubuntu-latest`?)

PR #735, opened 2024-01-11
neonvm: Use the right file extension for iso images

shayanh:


sharnoff (approved):
lgtm, afaik. Only concern would be how this interacts with live migration — when we migrate from an old neonvm-runner to a new one, do the new disk paths have to match? otherwise, should be good.

cc @cicdteam / @Omrigan

shayanh:
@sharnoff afaik this doesn't interfere with live migration. Once we attach a disk and mount on the VM, it'll be there over live migrations.

Omrigan:
> this doesn't interfere with live migration. Once we attach a disk and mount on the VM, it'll be there over live migrations.

@shayanh BTW, how does migration with disks actually work? 

I know there is `virsh migrate  --copy-storage-all`, but I am not sure how it works in our case/on QEMU level. Does it have to run block jobs for every disk attached?

@cicdteam Do you know?

shayanh:
@Omrigan I don't know the details.

Omrigan:
@cicdteam thanks. I guess those are deprecated, but can be replaced 1:1 with [the migration capability](
https://github.com/qemu/qemu/blob/master/qapi/migration.json#L472), so no need to tinker with blockjobs.

Omrigan (approved):

PR #667, opened 2023-12-05
util/watch: fix triggered relist on watch

shayanh:
- fixing triggered relist case and handle watchers

fixes #631

PR #736, opened 2024-01-12
util/watch: Add relist rate limiting

sharnoff:
This will be required for the autoscaler-agent as we try to use the VM spec as a source of truth (ref #350); we need relist rate limiting because we need relists in order to make a consistent read of the source of truth.

(For the scheduler, we're mostly fine without this - in practice, rate of relists is quite low.)

Like #727, marked as draft until there's a PR that actually uses it.

Also haven't tested this; need to do so before merging — some kind of ad-hoc testing like what was used for #667 should suffice.


Omrigan:
@sharnoff to check if this is still needed

Issue #729, opened 2024-01-08
Bug: autoscaler-agent scaling algorithm is too volatile for larger computes


sharnoff:
## Problem description / Motivation

One of the blockers for allowing larger computes (ref neondatabase/cloud#9103) is improving the scaling algorithm.

Currently, because the scaling algorithm (a) recalculates the "goal" CU every 5s, via updated metrics, and (b) does not factor into account past metrics when calculating the "goal" CU:

1. It's easy to cause the goal CU to oscillate, resulting in a lot of effort spent scaling, with little net benefit
2. As computes get larger, the same percentage change in metrics is more likely to produce a change in (integer) goal CU — meaning that each 5s the metrics update is more likely to prompt scaling, and by a larger amount

In a perfect world, maybe this'd be fine. But in practice, the process of scaling actually consumes resources, and so is _generally_ something we want to avoid doing frivolously.

See also: https://neondb.slack.com/archives/C03ETHV2KD1/p1704319422570509?thread_ts=1704316837.680979

## Feature idea(s) / DoD

Scaling algorithm should be more stable over _some_ time period, under _some_ conditions.

This isn't a super well-defined goal — so this issue mostly just exists to track _some_ improvement.

## Implementation ideas

There's a couple directions we could take this.

One is to _still_ not include any scaling history, and instead limit the size of a change (e.g., by no more than 1 CU at a time) and introduce rate-limiting on scaling. This wouldn't necessarily stop oscillation, but may reduce the impact.

The other is to include some history around recent metrics so that we have a longer time period to use for decision-making. This solution would probably be harder, but likely easier to understand and easier to produce better outcomes.

One _possibly_ annoying piece of this is that we may need to change a substantial portion of the tests for `pkg/agent/core`. We probably want a way to override the "goal CU" and directly provide that.

## Tasks

```[tasklist]
# Pre-requisites
- [ ] #1129
```

```[tasklist]
# Implementation
- [ ] ... add tasks here as they come up
```

```[tasklist]
# Follow-ups
- [ ] ... add tasks here as they come up
```

sharnoff:
There's an open RFC that will _partially_ address this issue here: https://www.notion.so/neondatabase/131f189e004780b2915ef2fdb95bae6a

In short: the approach should reduce volatility by ~60% from what we have today, but it's only a fractional decrease — probably insufficient for very volatile workloads on _much_ larger computes.

Omrigan:
Testing #1148 


[...]

PR #737, opened 2024-01-13
agent/core: Dependency-inject `ScalingAlgorithm`

sharnoff:
In short, this extracts the "goal CU" calculation from within `(*state).desiredResourcesFromMetricsOrRequestedUpscaling(...)` to a generic interface that receives metrics and outputs the desired CU.

There's two reasons for this:

1. Extracting goal CU calculation allows improvements to be more isolated, helping to reduce future code complexity.
2. Dependency-injecting to enable a much simpler version means that we can test goal CU calculation and the rest of the logic separately.

For (2), this PR switches `pkg/agent/core/state_test.go` to mostly use a new `ScalingAlgorithm` implementation that uses the desired CU as the associated "metrics". This should allow substantial changes to goal CU calculation without needing to change complex, unrelated tests.

---

This is a pre-implementation step for #729, and more generally, improvements in the scaling algorithm.

~~This will conflict with #653 (which should probably be merged first), but figured I'd open this anyways while it was on my mind.~~

To _somewhat_ reduce the size of the diff, I separated some of the bulk changes into an intermediate commit that could be added to `.git-blame-ignore-revs`, although... I'm undecided whether it's worth it.


pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+
+	"github.com/neondatabase/autoscaling/pkg/api"
+	"github.com/neondatabase/autoscaling/pkg/util"
+)
+
+type Metrics struct {
```
Omrigan, line 18: You can embed `api.Metrics` here.

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+
+	goalCU := util.Max(cpuGoalCU, memGoalCU)
+	return goalCU, true
+}
+
+// ReadMetrics generates Metrics from node_exporter output, or returns error on failure
+//
+// This function could be more efficient, but realistically it doesn't matter. The size of the
+// output from node_exporter/vector is so small anyways.
+func ReadMetrics(nodeExporterOutput []byte, loadPrefix string) (m Metrics, err error) {
```
Omrigan, line 127: Can you add an example input to know what does it parse?

Feels like this parsing should belong somewhere else. Is this standard prometheus format? Will
https://pkg.go.dev/github.com/prometheus/common/expfmt work?

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/neondatabase/autoscaling/pkg/util"
+)
+
+type Metrics struct {
+	LoadAverage1Min  float32
+	LoadAverage5Min  float32 // unused. To be removed when api.Metrics.LoadAverage5Min is removed.
+	MemoryUsageBytes float32
+}
+
+// ToAPI converts the Metrics to the subset used for communications with the scheduler plugin and
```
Omrigan, line 24: Why can't we have only `api.Metrics`? Feels like introducing separate structures will introduce confusion. Is it because multiple components depend on API so it is hard to update?

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// GoalCU returns the desired number of compute units that the VM should be using, given data
+	// from prior calls to Update and the configured resources equal to one compute unit.
+	GoalCU(scalingConfig api.ScalingConfig, computeUnit api.Resources) (_ uint32, ok bool)
+
+	// DeepCopy produces a copy of the value, intended for use when serializing State
+	DeepCopy() ScalingAlgorithm[M]
+}
+
+// ToApiMetrics is the interface that ScalingAlgorithm's metrics must satisfy - namely, that they
```
Omrigan, line 55: Why do we need to have multiple implementations of `Metrics`? Can't we have a large struct and put there everything that is needed by any of `ScalingAlgorithm`s?

Omrigan (commented):
Looked through about a half, will complete later.

pkg/agent/core/metrics.go
sharnoff: From the PR description:

> this PR switches `pkg/agent/core/state_test.go` to mostly use a new `ScalingAlgorithm` implementation that uses the desired CU as the associated "metrics". This should allow substantial changes to goal CU calculation without needing to change complex, unrelated tests.

The basic idea is that we have a lot of complex tests in `state_test.go` that don't _really_ care about the actual metrics, just the implications on goal CU. So it's better to just pass the goal CU itself as the "metric" so we can make changes to our scaling algorithm without needing to change all the tests.

pkg/agent/core/metrics.go
sharnoff: In the future, we will probably have only 1 or 2 fields in `api.Metrics`, but many fields in `core.Metrics`. And yeah, it's annoying to have something that _should_ be an implementation detail of the autoscaler-agent (`core.Metrics`) also affect backwards-compatibility of communications between components.

pkg/agent/core/metrics.go
sharnoff: I'll add some tests.

For context, this was directly copied from `main`'s current `pkg/api/metrics.go`.

Yeah, this is standard prometheus format. I _think_ I took a look at that package while prepping this PR, but it seemed complicated and better to leave as a follow-up :sweat_smile:

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/neondatabase/autoscaling/pkg/api"
+	"github.com/neondatabase/autoscaling/pkg/util"
+)
+
+type Metrics struct {
+	LoadAverage1Min  float32
+	LoadAverage5Min  float32 // unused. To be removed when api.Metrics.LoadAverage5Min is removed.
```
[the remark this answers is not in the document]
sharnoff: To avoid unexpected behavior for now, I'd rather keep the field and remove it later. Opened #750 to do that.

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// ScalingAlgorithm is the abstract interface provided by the default scaling algorithm, so that a
+// simpler version can be dependency-injected for easier testing.
+//
+// And also, separation of concerns is nice. Where state.go (and State within it) are more concerned
+// with the overall process of scaling approval and notification by communicating with other
+// components, the implementation(s) of ScalingAlgorithm can be more focused, making the calculation
+// of desired state abundantly clear.
+type ScalingAlgorithm[M ToAPIMetrics] interface {
```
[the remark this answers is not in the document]
sharnoff: Yeah- basically my thinking here was that more advanced scaling algorithms will definitely need more information, but adding more info now would just be unnecessary tech debt.

Originally I was going to include the time associated with the metrics in `Update`, but decided to leave that out for now :)

pkg/agent/core/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+// Update implements ScalingAlgorithm[Metrics]
+func (s *SimpleFactorScaling) Update(metrics Metrics) {
+	s.Metrics = &metrics
+}
+
+// GoalCU implements ScalingAlgorithm[Metrics], refer there for more.
+func (s *SimpleFactorScaling) GoalCU(scalingConfig api.ScalingConfig, computeUnit api.Resources) (_ uint32, ok bool) {
```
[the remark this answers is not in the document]
sharnoff: Happy to explain — the idea behind these return values is that naming one of them `_` prevents naked returns (which, yeah are generally not great), while allowing us to provide names associated with the return values.

Otherwise, IMO it's hard to tell what the return values mean from `func GoalCU(...) (uint32, bool)`.

pkg/agent/core/metrics.go
shayanh, line 43: I don't understand why `ScalingAlgorithm` is coupled to `ToAPIMetrics`.

pkg/agent/core/metrics.go
shayanh: To me metrics is a data object rather an entity, so I agree with @Omrigan that I can't easily see a case where having metrics as an interface can be useful. Also, I didn't quite get that why can't we use dummy values for metrics in tests.

pkg/agent/core/metrics.go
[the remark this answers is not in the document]
shayanh: Having named return values without using naked return is not uncommon (source: [Go stdlib](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/sync/map.go;l=330)).

In this particular case, I think having `(_ uint32, ok bool)` doesn't give us any more info. I think a comment for clarification would be good.

sharnoff:
Discussion with @Omrigan : tl;dr is that dependency injecting the scaling algorithm is probably a good idea, but generic metrics types are overly complicated. One way to make testing easier is to have a scaling algorithm implementation that just takes the desired CU from some external value, bypassing any metrics provided to it.

Closing this PR for now, maybe will reopen/recreate it if it's useful for #872.

Issue #711, opened 2023-12-22
Epic: investigate VM compute cold start time and ideally decrease it


kelvich:
## Motivation

We have pools of pre-created computes to amortize k8s api latency and vm boot latency. This is how the current boot time distribution looks like:
<img width="914" alt="image" src="https://github.com/neondatabase/autoscaling/assets/284219/d5a51f1f-681c-4ab3-8010-5acbe2ff720b">
in case of pool miss end user will experience bigger start latency. Since we are moving everything to VMs soon it would be good to improve VM start speed.

## DoD

Understand where time is actually spend during the boot. Fireckracker start times are below 1s, so we should understand what are the main reasons for our start times to be slower and whether there are any easy improvements that we can do right away.

So for this task DoD is a text explaining the difference in boot times and list of suggestion on how to improve them without losing ability to scale and live migrate VM's.

## Implementation ideas

* start firecracker with their default kernel/image
* start firecracker with our image -- will it be slower? how much?
* look into qemu-microvm 

## Tasks

```[tasklist]
- [x] Preliminary logs assessment
- [ ] https://github.com/neondatabase/autoscaling/issues/747
- [ ] https://github.com/neondatabase/autoscaling/issues/759
- [ ] https://github.com/neondatabase/autoscaling/issues/860
```

## Other related tasks, Epics, and links
Part of the https://github.com/neondatabase/company_projects/issues/130

cicdteam:
@kelvich 
>Fireckracker start times are below 1s

I'd like to clarify what exactly we mean by `start time`. Time to boot the kernel and start the init process (pid 1)? Our VMs do this in about 0.4s ([proof link](https://neonprod.grafana.net/goto/lBGzHFdSg?orgId=1)), which I don't think is the major overhead that VMs have. Need to understand why startup processes inside a VM are slower than in a Pod.

shayanh:
Earlier, I drew a workflow diagram to show what happens when we create a VM resource object. There other steps before getting into Kubernetes API server, which Stas has mentioned in [his Slack message](https://neondb.slack.com/archives/C03TN5G758R/p1703235575205199).

<img width="2242" alt="image" src="https://github.com/neondatabase/autoscaling/assets/1623278/6aaf658a-6aed-4950-9f1c-2c72f8e55ed5">

[...]

PR #738, opened 2024-01-15
neonvm: Use crictl to change container CPU, ditch cgroup

### LINKED ISSUE #591: Bug: VMs have low-priority CPU use, distributed equally under load (not based on VM size!)
This issue is largely adapted from a [message in slack](https://neondb.slack.com/archives/C03TN5G758R/p1693462680623239) from a couple months ago.

As of 2023-10-29, this is not urgent. It is a theoretical issue that is highly unlikely to occur, but should be considered in light of certain planned changes.

## Background

Kubernetes has "Quality of Service" (QoS) classifications for pods. The QoS given to a pod is determined exclusively by the resources.requests and resources.limits used (or not) for its containers.

For the most part, QoS is descriptive, rather than prescriptive. The QoS classes are roughly:
1. "Guaranteed" — all containers have cpu/memory requests equal to limits
2. "Burstable" — at least one container has cpu _or_ memory requests _or_ limits
3. "BestEffort" — no containers have CPU/memory requests _or_ limits

As a refresher, Kubernetes guarantees that any "requested" resources are always available to the container, should it need them.

What's actually going on is more clear if we look at how it's implemented in EKS (at least, as of 2023-10-29, k8s 1.25).

Cgroups are fundamental to the implementation of containers, so we can gather a lot about how certain kubernetes features are implemented by looking at how they affect the relevant cgroups. If we ssh into a kubernetes node and run `ls /sys/fs/cgroup/cpu/kubepods.slice`, we might see something like:

```sh-session
$ ls /sys/fs/cgroup/cpu/kubepods.slice
cgroup.clone_children  cpuacct.usage_percpu_sys   cpu.rt_period_us           kubepods-pod1035a83e_09cb_46e0_b56b_0ba59af9465c.slice  notify_on_release
cgroup.procs           cpuacct.usage_percpu_user  cpu.rt_runtime_us          kubepods-pod17f65edb_b99c_4530_8d93_3ff62c64f50f.slice  tasks
cpuacct.stat           cpuacct.usage_sys          cpu.shares                 kubepods-pod22bf9367_60ee_4481_a762_addb5f83a056.slice
cpuacct.usage          cpuacct.usage_user         cpu.stat                   kubepods-pod7bbd817f_b4d6_4e6f_8ac1_50bc6e270a2d.slice
cpuacct.usage_all      cpu.cfs_period_us          kubepods-besteffort.slice  kubepods-pod7daf1144_d3b5_4531_9a59_221f96a36a2b.slice
cpuacct.usage_percpu   cpu.cfs_quota_us           kubepods-burstable.slice   kubepods-podcb6dc9c2_eda7_4e8a_8e4a_948169e57cdc.slice
```

There's a few things of note here:

1. The `kubepods-besteffort.slice` cgroup
4. The `kubepods-burstable.slice` cgroup
5. The individual `kubepods-pod<UUID>` cgroups

The pods are relatively easy to track down: the UUID is just the UID of the pod; we can find the last one, for example, with:
```sh-session
$ kubectl get pod -A -o json | jq -r '.items[] | select(.metadata.uid == "cb6dc9c2-eda7-4e8a-8e4a-948169e57cdc") | "\(.metadata.namespace)/\(.metadata.name)"'
overprovisioning/overprovisioning-6f7d78c958-2228t
```
These are pods with a "Guaranteed" QoS.

The other cgroups mentioned are relatively self-explanatory. Inside them, we can see a similar situation: the "burstable" cgroup is full of `kubepods-burstable-pod<UUID>` cgroups, and the "besteffort" cgroup is full of `kubepods-besteffort-pod<UUID>`.

**VMs get the "BestEffort" QoS because we can't set resources.requests** (because changing a pod's resources.requests is only supported for k8s 1.27+), so: what cgroup settings are used for the BestEffort pods?

The relevant setting for resources.requests is the `cpu.shares` setting, which determines how the kernel's scheduler assigns CPU time if there isn't enough to go around. Basically, if the host is at 100% CPU usage, then all cgroups with runnable processes will get CPU time proportional to their number of `cpu.shares`, relative to whatever the total count is.

If we look at an arbitrary "Guaranteed" pod (say, e.g. the same one as before), we can see that if it has 2 CPUs:

```sh-session
$ kubectl get pod -n overprovisioning overprovisioning-6f7d78c958-2228t -o jsonpath='{.spec.containers[*].resources.requests.cpu}'
2
```

... then it gets 2048 CPU shares:

```sh-session
$ cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-podcb6dc9c2_eda7_4e8a_8e4a_948169e57cdc.slice/cpu.shares
2048
```

So we'd expect about 1 CPU share for 1m of CPU in in resources.requests.

If we look at some pods the "Burstable" QoS, we see that this is roughly the same:

```sh-session
$ cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod050b4a85_f401_4ddb_a585_3d2120ee0132.slice/cpu.shares
276
$ kubectl get pod -A -o json | jq -r '.items[] | select(.metadata.uid == "050b4a85-f401-4ddb-a585-3d2120ee0132") | "\(.metadata.namespace)/\(.metadata.name)"'
default/compute-orange-recipe-53139292-5d85955d54-5hk7x
$ kubectl get pod compute-orange-recipe-53139292-5d85955d54-5hk7x -o jsonpath='{.spec.containers[*].resources.requests.cpu}'
250m 10m 10m
```

We can also see that the total number of shares given to the cgroup containing "Burstable" pods is approximately equal to sum of the shares _inside_ the cgroup (i.e. there's no dilution)

```sh-session
$ cd /sys/fs/cgroup/cpu/kubepods.slice/kubepods-burstable.slice/
$ cat cpu.shares
32076
$ # Sum of all cpu.shares for child cgroups:
$ sh -c 'for c in kubepods-burstable-*; do cat $c/cpu.shares; done' | awk '{s+=$1} END {print s}'
32052
```

**Unfortunately, there's no such guarantees for BestEffort pods.** This, however, makes sense - after all, they haven't asked for any resources to be guaranteed!

If we look at the cgroup containing the BestEffort pods, we can see that it gets just 2 CPU shares! Essentially the equivalent of 0.002 CPUs guaranteed, split between all BestEffort pods.

```sh-session
$ cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/cpu.shares
2
```

And within that cgroup, all of the BestEffort pods are given an equal number of CPU shares:

```sh-session
$ cat /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/kubepods-*/cpu.shares | sort | uniq
2
```

## Current state of affairs

What all of that background means is this:

If a Kubernetes node is under high CPU load, there's basically no guarantees that _any_ VM will get CPU time (aside from that miniscule amount), and any CPU time allotted to VMs as a whole will be split equally between them, regardless of their total size.

So:

1. High CPU usage by _any non-VM pod_ can starve VMs
2. CPU starvation will result in bigger users being disproportionately throttled

**Note:** This is not _currently_ an issue, because our node CPU usage is so low. Once we implement overcommit (see #517), this may change and we may want to resolve this before we have an incident because of it.

## Possible solutions

There's a variety of ideas here, ranging in scope and impact. Here's an assortment:

1. Upgrade to k8s 1.27 and accurately set resources.requests on VM runner pods according to their `spec.guest.{cpus,memorySlots}.use` values
2. Set resources.requests according to the minimum values in the VM's spec
3. Revert #441 (move QEMU cgroup out of kubepods.slice) and set cpu.shares appropriately
6. Ditch kubernetes entirely :stuck_out_tongue:

### LINKED ISSUE #755: Bug: neonvm-runner creates qemu cgroup outside pod if k8s using cgroups v2
## Environment

Local (_but not staging or production_).

## Steps to reproduce

Set up a local VM on a machine using cgroups v2 (that's really all there is; the rest is informative).

You can check via `mount | grep /sys/fs/cgroup', which should show
```
cgroup2 on /sys/fs/cgroup type cgroup2
```
cgroups v1 will show a series of
```
cgroup on /sys/fs/cgroup/<controller> type cgroup
```
"hybrid" systems will tend to have the cgroups v1 output, in addition to
```
cgroup2 on /sys/fs/cgroup/unified type cgroup2
```

## Expected result

From the k8s node, the process running qemu should be deep in some `kubepods.slice/...` cgroup. You can check by - from the node - running `ps aux | grep qemu`, and then `cat /proc/<pid>/cgroup`.

## Actual result

The qemu process runs at the top level of `/sys/fs/cgroup` on the host, meaning that it's running outside the pod/container.

## Other logs, links

- Some discussion here: https://neondb.slack.com/archives/C03TN5G758R/p1705092611188719

## Tasks

```[tasklist]
## Tasks
- [ ] https://github.com/neondatabase/autoscaling/pull/749
- [ ] https://github.com/neondatabase/autoscaling/pull/738
```

sharnoff:
We recently realized[^1] that under cgroups v2, kubernetes uses cgroup namespaces which has a few effects:

1. The output of /proc/self/cgroup shows as if the container were at the root of the hierarchy
2. It's very difficult for us to determine the actual cgroup that the container corresponds to on the host
3. We still can't directly create a cgroup in the container's namespace because /sys/fs/cgroup is mounted read-only

So, neonvm-runner currently *does not* work as expected with cgroups v2; it creates a new cgroup for the VM, at the top of the hierarchy, and doesn't clean it up on exit.

How do we fix this? The aim of this PR is to remove the special cgroup handling entirely, and "just" go through the Container Runtime Interface (CRI) exposed by containerd to modify the existing container we're running in.

This requires access to `/run/containerd/containerd.sock`, which a malicious user could use to perform priviledged operations on the host (or in any other container on the host). Obviously we'd like to prevent that as much as possible, so the CPU handling is now runs alongside neonvm-runner as a separate container. neonvm-runner does not have access to the containerd socket.

On the upside, one key benefit we get from this is being able to set cpu shares, the abstraction underlying container resources.requests. The other options weren't looking so great[^2], so if this works, this would be a nice compromise.

---

~~Honestly have no clue whether this'll work. Haven't tested it _at all_ (except for some ad-hoc interactions with `crictl` from ssh-ing into a `kind` node).~~ Wanted to open this for visibility, in case there's something silly I missed.

---

Fixes #755.
Fixes #591.

[^1]: https://neondb.slack.com/archives/C03TN5G758R/p1705092611188719
[^2]: https://github.com/neondatabase/autoscaling/issues/591

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ test: fmt vet envtest ## Run tests.
 build: fmt vet bin/vm-builder ## Build all neonvm binaries.
 	go build -o bin/controller       neonvm/main.go
 	go build -o bin/vxlan-controller neonvm/tools/vxlan/controller/main.go
-	go build -o bin/runner           neonvm/runner/main.go
+	go build -o bin/runner           neonvm/runner/*.go
 
 .PHONY: bin/vm-builder
 bin/vm-builder: ## Build vm-builder binary.
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -1322,60 +1322,62 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*co
 					},
 				},
 			},
-			Containers: []corev1.Container{{
-				Image:           image,
-				Name:            "neonvm-runner",
-				ImagePullPolicy: corev1.PullIfNotPresent,
-				// Ensure restrictive context for the container
-				// More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
-				SecurityContext: &corev1.SecurityContext{
-					Privileged: &[]bool{false}[0],
-					Capabilities: &corev1.Capabilities{
-						Add: []corev1.Capability{
-							"NET_ADMIN",
-							"SYS_ADMIN",
-							"SYS_RESOURCE",
+			Containers: []corev1.Container{
+				{
+					Image:           image,
+					Name:            "neonvm-runner",
+					ImagePullPolicy: corev1.PullIfNotPresent,
+					// Ensure restrictive context for the container
+					// More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &[]bool{false}[0],
+						Capabilities: &corev1.Capabilities{
+							Add: []corev1.Capability{
+								"NET_ADMIN",
+								"SYS_ADMIN",
+								"SYS_RESOURCE",
+							},
 						},
 					},
-				},
-				Ports: []corev1.ContainerPort{{
-					ContainerPort: virtualmachine.Spec.QMP,
-					Name:          "qmp",
-				}, {
-					ContainerPort: virtualmachine.Spec.QMPManual,
-					Name:          "qmp-manual",
-				}},
-				Command: []string{
-					"runner",
-					"-vmspec", base64.StdEncoding.EncodeToString(vmSpecJson),
-					"-vmstatus", base64.StdEncoding.EncodeToString(vmStatusJson),
-				},
-				Env: []corev1.EnvVar{{
-					Name: "K8S_POD_NAME",
-					ValueFrom: &corev1.EnvVarSource{
-						FieldRef: &corev1.ObjectFieldSelector{
-							FieldPath: "metadata.name",
+					Ports: []corev1.ContainerPort{{
+						ContainerPort: virtualmachine.Spec.QMP,
+						Name:          "qmp",
+					}, {
+						ContainerPort: virtualmachine.Spec.QMPManual,
+						Name:          "qmp-manual",
+					}},
+					Command: []string{
+						"runner",
+						"-vmspec", base64.StdEncoding.EncodeToString(vmSpecJson),
+						"-vmstatus", base64.StdEncoding.EncodeToString(vmStatusJson),
+					},
+					Env: []corev1.EnvVar{{
+						Name: "K8S_POD_NAME",
+						ValueFrom: &corev1.EnvVarSource{
+							FieldRef: &corev1.ObjectFieldSelector{
+								FieldPath: "metadata.name",
+							},
+						},
+					}},
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      "virtualmachineimages",
+							MountPath: "/vm/images",
+						},
+						{
+							Name:      "sysfscgroup",
+							MountPath: "/sys/fs/cgroup",
+							// MountPropagationNone means that the volume in a container will
+							// not receive new mounts from the host or other containers, and filesystems
+							// mounted inside the container won't be propagated to the host or other
+							// containers.
+							// Note that this mode corresponds to "private" in Linux terminology.
+							MountPropagation: &[]corev1.MountPropagationMode{corev1.MountPropagationNone}[0],
 						},
 					},
-				}},
-				VolumeMounts: []corev1.VolumeMount{
-					{
-						Name:      "virtualmachineimages",
-						MountPath: "/vm/images",
-					},
-					{
-						Name:      "sysfscgroup",
-						MountPath: "/sys/fs/cgroup",
-						// MountPropagationNone means that the volume in a container will
-						// not receive new mounts from the host or other containers, and filesystems
-						// mounted inside the container won't be propagated to the host or other
-						// containers.
-						// Note that this mode corresponds to "private" in Linux terminology.
-						MountPropagation: &[]corev1.MountPropagationMode{corev1.MountPropagationNone}[0],
-					},
+					Resources: virtualmachine.Spec.PodResources,
 				},
-				Resources: virtualmachine.Spec.PodResources,
-			}},
+			},
 			Volumes: []corev1.Volume{
 				{
 					Name: "virtualmachineimages",
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -47,6 +47,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/tools/record"
 
@@ -74,10 +75,9 @@ type VirtualMachineReconciler struct {
 	client.Client
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
+	Config   *ReconcilerConfig
 
 	Metrics ReconcilerMetrics `exhaustruct:"optional"`
-
-	MaxConcurrentReconciles int
 }
 
 // The following markers are used to generate the rules permissions (RBAC) on config/rbac using controller-gen
@@ -88,6 +88,7 @@ type VirtualMachineReconciler struct {
 //+kubebuilder:rbac:groups=vm.neon.tech,resources=virtualmachines/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=vm.neon.tech,resources=virtualmachines/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=core,resources=nodes,verbs=list
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=pods/status,verbs=get;list;watch
@@ -686,7 +687,7 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 				fmt.Sprintf("One CPU was unplugged from VM %s",
 					virtualmachine.Name))
 		} else if supportsCgroup && *specCPU != cgroupUsage.VCPUs {
-			log.Info("Update runner pod cgroups")
+			log.Info("Update runner pod cgroups", "runner", cgroupUsage.VCPUs, "spec", *specCPU)
 			if err := setRunnerCgroup(ctx, virtualmachine, *specCPU); err != nil {
 				return err
 			}
@@ -1061,7 +1062,7 @@ func (r *VirtualMachineReconciler) podForVirtualMachine(
 	virtualmachine *vmv1.VirtualMachine,
 	sshSecret *corev1.Secret,
 ) (*corev1.Pod, error) {
-	pod, err := podSpec(virtualmachine, sshSecret)
+	pod, err := podSpec(virtualmachine, sshSecret, r.Config)
 	if err != nil {
 		return nil, err
 	}
@@ -1257,7 +1258,7 @@ func imageForVmRunner() (string, error) {
 	return image, nil
 }
 
-func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*corev1.Pod, error) {
+func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret, config *ReconcilerConfig) (*corev1.Pod, error) {
 	runnerVersion := api.RunnerProtoV1
 	labels := labelsForVirtualMachine(virtualmachine, &runnerVersion)
 	annotations := annotationsForVirtualMachine(virtualmachine)
@@ -1322,8 +1323,9 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*co
 					},
 				},
 			},
-			Containers: []corev1.Container{
-				{
+			// generate containers as an inline function so the context isn't isolated
+			Containers: func() []corev1.Container {
+				runner := corev1.Container{
 					Image:           image,
 					Name:            "neonvm-runner",
 					ImagePullPolicy: corev1.PullIfNotPresent,
@@ -1346,11 +1348,20 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*co
 						ContainerPort: virtualmachine.Spec.QMPManual,
 						Name:          "qmp-manual",
 					}},
-					Command: []string{
-						"runner",
-						"-vmspec", base64.StdEncoding.EncodeToString(vmSpecJson),
-						"-vmstatus", base64.StdEncoding.EncodeToString(vmStatusJson),
-					},
+					Command: func() []string {
+						cmd := []string{"runner"}
+						// intentionally add this first, so it's easier to see among the very long
+						// args that follow.
+						if config.UseContainerMgr {
+							cmd = append(cmd, "-skip-cgroup-management")
+						}
+						cmd = append(
+							cmd,
+							"-vmspec", base64.StdEncoding.EncodeToString(vmSpecJson),
+							"-vmstatus", base64.StdEncoding.EncodeToString(vmStatusJson),
+						)
+						return cmd
+					}(),
 					Env: []corev1.EnvVar{{
 						Name: "K8S_POD_NAME",
 						ValueFrom: &corev1.EnvVarSource{
@@ -1359,12 +1370,12 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*co
 							},
 						},
 					}},
-					VolumeMounts: []corev1.VolumeMount{
-						{
+					VolumeMounts: func() []corev1.VolumeMount {
+						images := corev1.VolumeMount{
 							Name:      "virtualmachineimages",
 							MountPath: "/vm/images",
-						},
-						{
+						}
+						cgroups := corev1.VolumeMount{
 							Name:      "sysfscgroup",
 							MountPath: "/sys/fs/cgroup",
 							// MountPropagationNone means that the volume in a container will
@@ -1373,19 +1384,83 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret) (*co
 							// containers.
 							// Note that this mode corresponds to "private" in Linux terminology.
 							MountPropagation: &[]corev1.MountPropagationMode{corev1.MountPropagationNone}[0],
+						}
+
+						if config.UseContainerMgr {
+							return []corev1.VolumeMount{images}
+						} else {
+							// the /sys/fs/cgroup mount is only necessary if neonvm-runner has to
+							// do is own cpu limiting
+							return []corev1.VolumeMount{images, cgroups}
+						}
+					}(),
+					Resources: virtualmachine.Spec.PodResources,
+				}
+				containerMgr := corev1.Container{
+					Image: image,
+					Name:  "neonvm-container-mgr",
+					Command: []string{
+						"container-mgr",
+						"-port", strconv.Itoa(int(virtualmachine.Spec.RunnerPort)),
+						"-init-milli-cpu", strconv.Itoa(int(*virtualmachine.Spec.Guest.CPUs.Use)),
+					},
+					Env: []corev1.EnvVar{
+						{
+							Name: "K8S_POD_UID",
+							ValueFrom: &corev1.EnvVarSource{
+								FieldRef: &corev1.ObjectFieldSelector{
+									FieldPath: "metadata.uid",
+								},
+							},
+						},
+						{
+							Name:  "CRI_ENDPOINT",
+							Value: fmt.Sprintf("unix://%s", config.criEndpointSocketPath()),
 						},
 					},
-					Resources: virtualmachine.Spec.PodResources,
-				},
-			},
-			Volumes: []corev1.Volume{
-				{
+					LivenessProbe: &corev1.Probe{
+						InitialDelaySeconds: 10,
+						ProbeHandler: corev1.ProbeHandler{
+							HTTPGet: &corev1.HTTPGetAction{
+								Path: "/healthz",
+								Port: intstr.FromInt(int(virtualmachine.Spec.RunnerPort)),
+							},
+						},
+					},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("50m"),
+							corev1.ResourceMemory: resource.MustParse("50Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("1"), // cpu limit > request, because usage is spiky
+							corev1.ResourceMemory: resource.MustParse("50Mi"),
+						},
+					},
+					// socket for crictl to connect to
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							Name:      "containerdsock",
+							MountPath: config.criEndpointSocketPath(),
+						},
+					},
+				}
+
+				if config.UseContainerMgr {
+					return []corev1.Container{runner, containerMgr}
+				} else {
+					// Return only the runner if we aren't supposed to use container-mgr
+					return []corev1.Container{runner}
+				}
+			}(),
+			Volumes: func() []corev1.Volume {
+				images := corev1.Volume{
 					Name: "virtualmachineimages",
 					VolumeSource: corev1.VolumeSource{
 						EmptyDir: &corev1.EmptyDirVolumeSource{},
 					},
-				},
-				{
+				}
+				cgroup := corev1.Volume{
 					Name: "sysfscgroup",
 					VolumeSource: corev1.VolumeSource{
 						HostPath: &corev1.HostPathVolumeSource{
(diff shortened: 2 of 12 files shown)


Omrigan:
Looks good. I guess the difference will be is that `neonvm-runner` will always be subject to cgroup restrictions, not just QEMU process. But that shouldn't be a problem. Right?

Also, why do we have to set container limits through `crictl`, isn't there a proper k8s way of setting pod's resource limits on a cluster scale?

sharnoff:
> `neonvm-runner` will always be subject to cgroup restrictions, not just QEMU process

Yeah- this is different, but I figured probably better anyways — it's more thorough resource limiting, or something like that.

> isn't there a proper k8s way of setting pod's resource limits on a cluster scale?

Unfortunately, k8s <1.27 doesn't allow changing a pod/container's resources after it's created (partly why we're in this whole mess in the first place), and the upgrade process is complex. See also #591, neondatabase/cloud#7922

areyou1or0:
Replying to your questions from Slack:

> Big risk comes from providing direct access to containerd API from within the pod, but in a separate, dedicated container. 

In case of exploiting a vulnerability on the CPU handling container, an attacker could gain access to the CRI and potentially escalate their privileges. (container breakout, bypassing network controls etc.)

One way to prevent this scenario besides isolating the CPU handling container is to scanning the CPU handling container for vulnerabilities (which we already do) & monitoring pod activity regularly. (I'm skeptical on the current capability with vuln.scanner, perhaps guardduty may cover this as well)

From the pod-to-pod exploitation, in case of a cross-container attack where an attacker exploit a vulnerability on another container in the pod, could access to CRI and to CPU handling container. 

Most obvious way to prevent this scenario is implementing least privilege, restricting access to the CRI., and implementing strong network security controls. 

> Wondering whether it adequately prevents privilege escalation (are there exploits that allow moving laterally within a pod but not further?)

Moving CPU handling to a separate container naturally helps preventing PE scenarios within the pod itself at some level, but in case of gaining unauthorized access to the CRI via another possible vulnerability can still lead to escalate privileges in theory.

sharnoff:
>  I still don't understand why we need this whole cgroup dance if we have `cpuLimitOvercommitFactor = 4` (IIUC, which means that CPU cgroup limits are not relevant, since qemu limits the guest itself by CPU cores)

There's two reasons:

1. QEMU can use more CPU time than the number of CPUs would indicate, often when doing IO (this is actually the original reason for the change - see #509)
2. We _do_ want to have some upper bound on the CPU usage, just in case.

This also would resolve #591 (see there for more info).

Omrigan:
#711 suggests a hypothesis the number of auxiliary containers might impact compute cold start time. So this might make the situation worse.

sharnoff:
@omrigan IIUC the impact is from adidtional Spec.InitContainers, not Spec.Containers, right? (iiuc, impact is specifically because k8s doesn't create the next InitContainer until the previous one finished, so sandbox creation isn't pipelined)

shayanh:
Update reserved volume names here:
https://github.com/neondatabase/autoscaling/blob/65e43930749eb12ed0a8f6da6440e5c20fe9f16f/neonvm/apis/neonvm/v1/virtualmachine_webhook.go#L118

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+}
+
+type containerInfo struct {
+	podID        string
+	containerID  string
+	lastMilliCPU atomic.Uint32
+}
+
+func (c *containerInfo) listenForCPUChanges(ctx context.Context, logger *zap.Logger, crictl *crictlConfig, port int32) {
```
shayanh, line 177: This name is also a bit confusing. We provide an interface to get and set CPU limits rather listening on cpu changes. Moreover, here we handle health check requets as well that is not related.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+
+	body, err := json.Marshal(resp)
+	if err != nil {
+		logger.Error("could not marshal body", zap.Error(err))
+		w.WriteHeader(500)
+		return
+	}
+
+	w.Header().Add("Content-Type", "application/json")
+	w.Write(body) //nolint:errcheck // Not much to do with the error here. TODO: log it?
```
shayanh, line 312: Let's log the error here. Better than ignoring it.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+			logger.Fatal("cpu_change server exited with error", zap.Error(err))
+		}
+	case <-ctx.Done():
+		err := server.Shutdown(context.Background())
+		logger.Info("shut down cpu_change server", zap.Error(err))
+	}
+}
+
+func (c *containerInfo) handleCPUChange(logger *zap.Logger, crictl *crictlConfig, w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
```
shayanh, line 217: We can use `http.MethodPost` and it's nicer

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+		logger.Error("unexpected method", zap.String("method", r.Method))
+		w.WriteHeader(400)
+		return
+	}
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		logger.Error("could not read body", zap.Error(err))
+		w.WriteHeader(400)
```
shayanh, line 225: We can use `http.StatusBadRequest` and it's nicer

shayanh: and everywhere else where used a plain status number

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+
+func (c *containerInfo) listenForCPUChanges(ctx context.Context, logger *zap.Logger, crictl *crictlConfig, port int32) {
+	mux := http.NewServeMux()
+	loggerHandlers := logger.Named("http-handlers")
+	cpuChangeLogger := loggerHandlers.Named("cpu_change")
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+		_ = r.Body.Close()
+		w.WriteHeader(200)
+	})
+	mux.HandleFunc("/cpu_change", func(w http.ResponseWriter, r *http.Request) {
```
shayanh, line 185: We can make this API look nicer (a rest-like equivalent is `POST /cpu` and `GET /cpu`), but doesn't matter much now and we can leave for another PR.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+	if criPodID == "" {
+		logger.Fatal("could not find CRI pod with matching UID", zap.String("uid", selfPodUID))
+	}
+
+	logger.Info("Got CRI ID for pod", zap.String("podID", criPodID))
+
+	var criRunnerContainerID string
+	for criRunnerContainerID == "" {
+		var containers crictlContainers
+		if err := crictl.run(logger, &containers, "ps", "-p", criPodID, "-o", "json"); err != nil {
```
shayanh, line 93: Let's add a comment describing what does this `crictl` command do. Another idea is providing a higher level abstraction on top of `run`, smth like `crictl.ListContainers(...)`. Also this helps us with type safety to some extent.

shayanh (commented):
Looks good in general. Added several small things. Also I have a question about testing. Have you checked out cAdvisor metrics to work as expected here?

Omrigan:
> @Omrigan IIUC the impact is from adidtional Spec.InitContainers, not Spec.Containers, right? (iiuc, impact is specifically because k8s doesn't create the next InitContainer until the previous one finished, so sandbox creation isn't pipelined)

Not sure yet. At least, additional "main" container will put more pressure on the `containerd`, potentially delaying other parallel requests.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+	// number of CPU shares per vCPU to use when updating a container
+	sharesPerCPU = 1024
+	cpuPeriod    = 100000
+)
+
+// mainContainerMgr is the 'main' function if neonvm-runner is called with the container-mgr bin path
+//
+// This typically runs as a secondary container alongside neonvm-runner, with privileged access to
+// conatinerd via crictl.
+func mainContainerMgr() {
```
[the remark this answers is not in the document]
sharnoff: Mostly to save on needing to distribute another image, but it's also nice to have the thing running alongside the runner be located alongside it in the code

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+	// cpuLimitOvercommitFactor sets the amount above the VM's spec.guest.cpus.use that we set the
+	// QEMU cgroup's CPU limit to. e.g. if cpuLimitOvercommitFactor = 3 and the VM is using 0.5
+	// CPUs, we set the cgroup to limit QEMU+VM to 1.5 CPUs.
+	//
+	// This exists because setting the cgroup exactly equal to the VM's CPU value is overly
+	// pessimistic, results in a lot of unused capacity on the host, and particularly impacts
+	// operations that parallelize between the VM and QEMU, like heavy disk access.
+	//
+	// See also: https://neondb.slack.com/archives/C03TN5G758R/p1693462680623239
+	cpuLimitOvercommitFactor = 4
```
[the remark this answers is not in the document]
sharnoff: Yeah, that's a good question.
See also https://neondb.slack.com/archives/C061BGTBZMG/p1705680836711229?thread_ts=1705672228.411229 — after the switch to m6id nodes, maybe we can lower this (maybe down to 2?). Although we should be careful about the interaction between CPU throttling and time drift.

sharnoff:
@shayanh I'm going to resolve the comments you left about the http handling for now. All the code there was directly copied from the existing implementation in `runner/main.go`, and I'd rather either move or modify, not both (this also means that it keeps the same API).

Happy to address things in a follow-up.

neonvm/runner/containermgr.go
sharnoff: Copied from neonvm-runner, happy to address in a follow-up.

https://github.com/neondatabase/autoscaling/blob/32e407a6ba1ae49c341e9de898d62489914fff74/neonvm/runner/main.go#L774

neonvm/runner/containermgr.go
sharnoff: Copied from neonvm-runner. We _may_ want to address in a follow-up, but this is also the current API. Changing it would require support for both versions in neonvm-controller.

https://github.com/neondatabase/autoscaling/blob/32e407a6ba1ae49c341e9de898d62489914fff74/neonvm/runner/main.go#L779

neonvm/runner/containermgr.go
sharnoff: Copied from neonvm-runner, happy to address in a follow-up.

https://github.com/neondatabase/autoscaling/blob/32e407a6ba1ae49c341e9de898d62489914fff74/neonvm/runner/main.go#L718

neonvm/runner/containermgr.go
sharnoff: Copied from neonvm-runner, happy to address in a follow-up

https://github.com/neondatabase/autoscaling/blob/32e407a6ba1ae49c341e9de898d62489914fff74/neonvm/runner/main.go#L726

see also https://github.com/neondatabase/autoscaling/pull/744#discussion_r1459384439

neonvm/runner/containermgr.go
sharnoff: Copied from neonvm-runner, happy to address in a follow-up

https://github.com/neondatabase/autoscaling/blob/32e407a6ba1ae49c341e9de898d62489914fff74/neonvm/runner/main.go#L771

shayanh:
@sharnoff makes sense. I'll make a note to open an issue regarding them later.

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
shayanh: You can still ship both binaries on the same image and then override their entrypoint on the pod definition. I understand we want to keep the code next to each other but it feels we have coupled two separate things into the a single binary. An alternative is the following:
```
.
├── cmd/
│   ├── runner/
│   │   └── main.go
│   └── containermgr/
│       └── main.go
├── runner.go
└── containermgr.go
```

wdyt?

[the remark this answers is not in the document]
sharnoff: Hm, yeah fair enough - I'll go with that.
Originally I was thinking it might be useful to share code between the two, but thinking about it further, I don't think we'll end up needing it.

Makefile
```diff
@@ -123,7 +123,7 @@ test: fmt vet envtest ## Run tests.
 build: fmt vet bin/vm-builder ## Build all neonvm binaries.
 	go build -o bin/controller       neonvm/main.go
 	go build -o bin/vxlan-controller neonvm/tools/vxlan/controller/main.go
-	go build -o bin/runner           neonvm/runner/main.go
+	go build -o bin/runner           neonvm/runner/*.go
```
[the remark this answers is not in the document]
sharnoff: comment now outdated (moved container-mgr to a separate directory), but I tried this and it doesn't work for some reason (maybe `...` just isn't allowed when compiling a binary?)

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
sharnoff: Done - ended up putting container-mgr stuff in a directory inside `runner`.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	cinfo.listenForCPUChanges(context.TODO(), logger, crictl, int32(httpPort))
+}
+
+type crictlConfig struct {
+	endpoint string
+}
+
+func (c *crictlConfig) run(logger *zap.Logger, output any, args ...string) error {
```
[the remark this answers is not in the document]
sharnoff: Makes sense - fixed in d8f1b19b553fed039ecdaeb9f176e053c00d83af, lmk what you think

neonvm/runner/containermgr.go
sharnoff: I ended up going with this approach in d8f1b19b553fed039ecdaeb9f176e053c00d83af - resolving this comment in favor of the one above.

neonvm/runner/containermgr.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+
+		if err := json.Unmarshal(out, output); err != nil {
+			return fmt.Errorf("error parsing JSON: %w", err)
+		}
+		return nil
+	}
+}
+
+type containerInfo struct {
```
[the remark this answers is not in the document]
sharnoff: Makes sense - renamed it to `cpuServerState` in e92c10082dcee400f6d58bd497391b31fd8cb46f (+ additional modifications in 5acf53663499c8d16eeae2a388d3a36658a2f0ca). wdyt?

neonvm/runner/main.go
```diff
@@ -673,38 +645,14 @@ func main() {
...
-	//
-	// We don't want to just use the VM spec's .status.PodName because during migrations that will
-	// be equal to the source pod, not this one, which may be... somewhat confusing.
-	cgroupPath := fmt.Sprintf("%s/neonvm-qemu-%s", selfCgroupPath, selfPodName)
-
-	logger.Info("Determined QEMU cgroup path", zap.String("path", cgroupPath))
-
-	if err := setCgroupLimit(logger, qemuCPUs.use, cgroupPath); err != nil {
-		logger.Fatal("Failed to set cgroup limit", zap.Error(err))
-	}
```
sharnoff, line 696: PR in its current state hasn't replicated this — so, right now there's a regression where a 0.25 CPU VM starts at 1 CPU and needs to scale down.

Probably need to add milli CPU as an arg to container-mgr or something, and then do a similar "set on startup". (and actually, that'd also remove the need for nil-able cpu returned from the runner pod as well, so can revert 5acf53663499c8d16eeae2a388d3a36658a2f0ca)

sharnoff: fixed in 2af7af5adcb738d736eac1095f098511b8e691f4

Makefile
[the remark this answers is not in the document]
shayanh: I see. Here we don't build `container-mgr` though.

neonvm/runner/container-mgr/crictl.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type Crictl struct {
+	endpoint string
+}
+
+// Pods calls 'crictl pods' and, if successful, returns the parsed output
+//
+// This command lists all running pods.
+func (c *Crictl) Pods(logger *zap.Logger) (*crictlPods, error) {
```
shayanh, line 22: Is there a reason that `crictPods` is not exported? It's a bit weird that an exported function returnes an unexported type.

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
shayanh: Looks good, put a new comment there.

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
shayanh: Rename looks good. However, I'm not quite sure why we want to have nil-able runner cgroup vCPUs.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +747,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+	case "", corev1.PodPending:
+		return runnerPending
+	case corev1.PodSucceeded:
+		return runnerSucceeded
+	case corev1.PodFailed:
+		return runnerFailed
+	case corev1.PodUnknown:
+		return runnerUnknown
+
+	// With runners with neonvm-container-mgr, the pod phase itself won't be updated if just the
```
shayanh, line 814: What happens exactly if one the containers fail in the VM?

neonvm/runner/main.go
shayanh: I don't understand why we need the new arguments here. Can't we just set the initial limits once `container-mgr` is starting?

shayanh:
I added a few more comments.

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
shayanh: Looks good. It's a bit unusual since the common convention is to put `main.go` files under the `cmd` directory but I don't have a strong incline towards it.

shayanh:
I've been thinking it's better to move this into VMLet and kick off the development of VMLet with this. Because VMs' cgroup stuff perfectly fits with the responsibilities of the VMLet.

I spent some time thinking and figuring out the details, and my conclusion  is it's not the best decision to start VMLet with crictl. The main reason is that, with a partial setup of the VMLet, we have to expose an HTTP server on it to serve the incoming cgroup requests from the NeonVM controller. VMLet is deployed as a daemon set. Thus, to have a HTTP server, we need to use a NodePort service, which totally defeats the benefits of having VMLet (esp from a security perspective).

My main concern in this PR is the complexity associated with handling several containers inside the runner pod (#749), but I think at this point it's fine, and I don't see a better way to avoid it.

neonvm/runner/main.go
sharnoff: In theory, yeah. In practice, it seemed like it wasn't simple to make sure that we didn't get any weird transitory states — we'd have to update the runner before moving from Pending to Running, or e.g. if .status.cpus is nil, update the runner before fetching the info.

The PR in its present form (reverted nillable runner CPU; container-mgr has the new arg) doesn't need any of the meat of the reconcile logic to be updated (with the notable exception of #749). That's my primary motivation (because the existing logic is _so complex_).

neonvm/runner/containermgr.go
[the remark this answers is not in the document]
sharnoff: Update: was reverted in a479f9ba0a2fea6083a137da7da6af850fccf660.

Makefile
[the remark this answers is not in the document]
sharnoff: Now outdated. We build it here: https://github.com/neondatabase/autoscaling/pull/738/files#diff-b1be71828a4d4788ff25a785658b52c114a178585350bce6bd34963367c5ec70R28

neonvm/runner/container-mgr/crictl.go
sharnoff: Yeah, fair - updated in f1f564275337b517b657a23be44121af6748b080.

neonvm/controllers/virtualmachine_controller.go
sharnoff: If one of the containers fails (but not the other), the the pod will still be `Running`, but the VM _should_ be `Failing`.

The version of this function in this PR is now outdated; #749 has the the updated version, and should be merged before this.

sharnoff:
Rebased onto main. In the process, decided to drop the nillable vcpus commit and its revert.

shayanh (commented):
Looks good to me. There is only a small thing mentioned in this comment:
https://github.com/neondatabase/autoscaling/pull/738#issuecomment-1899677723

sharnoff:
> only a small thing mentioned in this comment

Ah right, good catch - thanks!

sharnoff:
Current status:

- [ ] Blocked on (approving and merging) #749
- [ ] Need to test on staging — as yet, haven't checked it works on EKS (although I expect it should be ok)

sharnoff:
Follow-up that was brought up in a meeting: We should try out using the client library directly, rather than via the CLI. May substantially reduce latency.

shayanh:
Related: I spent a couple of hours setting up autoscaling on a fresh Ubuntu bare metal machine and it didn't work out first due to our legacy cgroup stuff. Apparently, we don't use the right cgroup version in some cases, which doesn't allow the VMs to start. It's nice that we are moving away from manually reading and modifying cgroup files.

Issue #247, opened 2023-05-09
no observability into neonvm reconcile failures (needs metrics!)


sharnoff:
Problems like #243 would be easier to detect and debug if we had more observability into the neonvm controller, specifically around reconciliation failures.

Off-hand, I see two ways of doing this (but there are definitely more!)

1. Increment some counter when reconciling fails (i.e. global rate of failed reconciliation)
2. Globally track how many VMs are failing to reconcile, expose this as a gague
    - This is more work (and uses more memory, and has the "multiple sources of truth" problem), but is maybe closer to what we actually care about.

We can also do both options, because each option has cases that it will report but the other won't.

shayanh:
I think the `rate` of the first metrics is almost the second metrics. Isn't it?

sharnoff:
status: Second half is almost done, PR to come this week

PR #739, opened 2024-01-16
neonvm: Expose controllers default set of metrics

shayanh:
Kubebuilder provides these by default. More info [here](https://book.kubebuilder.io/reference/metrics-reference).

This provides the first metric discussed at #247.  I have tested this on a local Victoria Metrics deployment. Note that currently we're bypassing kube-builder's kube-rbac-proxy that protects controller metrics by default. We don't protect metrics in other services (agent, scheduler) either. If needed I can add necessary role bindings.


deploy/vmscrape.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     matchExpressions:
       - key: name
         operator: In
-        values: [autoscaler-agent, autoscale-scheduler]
+        values: [autoscaler-agent, autoscale-scheduler, neonvm-controller]
   namespaceSelector:
     matchNames:
       - kube-system
```
sharnoff, line 20: IIUC, we can't simply add `neonvm-controller` here for a couple reasons:

1. This only matches pods with the label `name=neonvm-controller`; the controller doesn't have that label
2. This only matches pods in the `kube-system` namespace (`neonvm-controller` is in `neonvm-system`).

Maybe have a separate `VMPodScrape` below this for `neonvm-system` that uses a different label?

sharnoff:
OTOH I don't know how to test victoria metrics locally. Maybe @lassizci would know?

lassizci:
> OTOH I don't know how to test victoria metrics locally. Maybe @lassizci would know?

Doesn't need much configuration, so perhaps helm-chart or kustomize: https://docs.victoriametrics.com/operator/setup.html

On the other hand there's a possibility to validate the manifest against CRD https://github.com/kubernetes-sigs/kubectl-validate#local-crds in case that's more appropriate.

deploy/vmscrape.yaml
shayanh: fixed in 6ea7a9c04ca5e5662a0ec4ae5c3d6776ae939b84

deploy/vmscrape.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: operator.victoriametrics.com/v1beta1
 kind: VMPodScrape
 metadata:
   name: autoscaling-components
-  namespace: monitoring
+  namespace: monitoring-system
```
sharnoff, line 6: What was the reason for this change? IIUC the metrics scrapes need to be in the `monitoring` namespace in order for the metrics to be routed to the right place.

deploy/vmscrape.yaml
```diff
@@ -18,3 +18,22 @@ spec:
...
+    - port: metrics
+      path: /metrics
+      interval: 60s
+      scrapeTimeout: 10s
+  selector:
+    matchLabels:
+      control-plane: controller
+  namespaceSelector:
+    matchNames:
+      - neonvm-system
```
sharnoff, line 39: missing newline

deploy/vmscrape.yaml
shayanh: Good catch. I named my local victoria metrics namespace `monitoring-system` and forgot to change it here.

sharnoff (approved):
lgtm!

double-checked namespace/label with
```
$ kubectl get pod -n neonvm-system -l control-plane=controller
NAME                                READY   STATUS    RESTARTS   AGE
neonvm-controller-f594ddf46-2jv89   2/2     Running   0          24h
```
and it looks right

PR #743, opened 2024-01-16
plugin: Remove ComputeUnit from config and PluginResponse

sharnoff:
... and every where else that it filters into.

This is part of the process of moving the ComputeUnit source of truth from the scheduler to the autoscaler-agent (#706). We've already redefined the autoscaler-agent's own config as its source of truth, so this is mostly just following through on the previous deprecation.


sharnoff:
After thinking some more, I think probably #653 should be merged + deployed before we merge this, alongside sending the compute unit from the autoscaler-agent. Otherwise we'd probably end up delaying deploy of #653.

shayanh (approved):
lgtm!

sharnoff:
@omrigan LMK if you plan to review; otherwise will merge in ~36h. This isn't urgent, so if you'd like, I'm happy to leave it waiting while you tackle other stuff.

Omrigan (approved):

Omrigan (approved):

Omrigan:
> @Omrigan LMK if you plan to review; otherwise will merge in ~36h. This isn't urgent, so if you'd like, I'm happy to leave it waiting while you tackle other stuff.

LGTM. A good net negative line count diff is always welcome :+1:

PR #744, opened 2024-01-17
agent: Send compute unit in requests to plugin

sharnoff:
Related to moving the compute unit source of truth from the scheduler plugin to the autoscaler-agent (see #706).

We also particularly want to have this so that the plugin doesn't approve resources that aren't an integer multiple of the respective value in the compute unit. If we didn't we could e.g. approve 1.01 CPUs, leaving the autoscaler-agent with 0.01 CPUs that it'll never use. In practice, that's quite unlikely to happen, but it's still worth avoiding potential issues there.

~~**NB: This PR builds on #653 and must not be merged before it.**~~

~~**Before merging:** Still need to check that we're able to run with old autoscaler-agent + new scheduler, with this PR.~~ Done

### DIFF
--- a/pkg/plugin/run.go
+++ b/pkg/plugin/run.go
@@ -13,7 +13,6 @@ import (
 	"github.com/tychoish/fun/srv"
 	"go.uber.org/zap"
 
-	vmapi "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
@@ -150,6 +149,13 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
 		return nil, 400, fmt.Errorf("nil metrics not supported for protocol version %v", req.ProtoVersion)
 	}
 
+	// Check that req.ComputeUnit == nil matches whether the protocol version supports it.
+	if req.ComputeUnit == nil && req.ProtoVersion.AgentSendsComputeUnit() {
+		return nil, 400, fmt.Errorf("computeUnit required for protocol version %v", req.ProtoVersion)
+	} else if req.ComputeUnit != nil && !req.ProtoVersion.AgentSendsComputeUnit() {
+		return nil, 400, fmt.Errorf("computeUnit field not supported for protocol version %v", req.ProtoVersion)
+	}
+
 	e.state.lock.Lock()
 	defer e.state.lock.Unlock()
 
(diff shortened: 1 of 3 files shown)


shayanh:
Is it possible for agent to request resources that aren't multiple of CU?

pkg/plugin/run.go
```diff
@@ -150,6 +149,13 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
 		return nil, 400, fmt.Errorf("nil metrics not supported for protocol version %v", req.ProtoVersion)
 	}
 
+	// Check that req.ComputeUnit == nil matches whether the protocol version supports it.
+	if req.ComputeUnit == nil && req.ProtoVersion.AgentSendsComputeUnit() {
+		return nil, 400, fmt.Errorf("computeUnit required for protocol version %v", req.ProtoVersion)
```
shayanh, line 154: We can use `http.StatusBadRequest` instead of 400 but better to be consistent in the function. So leave it for now.

pkg/plugin/run.go
```diff
@@ -163,12 +169,26 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
...
+		)
+	}
+
 	// If the request was actually sending a quantity of *memory slots*, rather than bytes, then
 	// multiply memory resources to make it match the
 	if !req.ProtoVersion.RepresentsMemoryAsBytes() {
 		req.Resources.Mem *= pod.vm.memSlotSize
 	}
 
+	computeUnit := e.state.conf.ComputeUnit
```
shayanh, line 187: I thought the plan is to remove `ComputeUnit` from scheduler's config.

shayanh (commented):
Looks good in general. Added some minor comments. Also in the commit message we should make it clear that what happens when agent's requested values are not multiple of CU. It's not the case that plugin rejects the request but it actually approves the requested values rounded-down.

pkg/plugin/run.go
sharnoff: That's the plan, yeah - see #743. But the intermediate states should be stable, in particular any release must fully support new scheduler + old autoscaler-agent. IMO it's probably too complicated to try to do that if we merge #653, this PR, _and_ #743, so I figured that PR should be delayed until the release after. But we _do_ want this PR in the release with #653 in order to be able to accurately use the compute unit when handling agent requests.

pkg/plugin/run.go
sharnoff: Long ago, I had this conversation with @﻿tychoish, and we ended up keeping the "common" numbers because they're easily recognizable (and IMO, end up being more readable because of it).

If you'd like, we can revisit this. I'm slight-negative on making that change, but ok to go with team consensus.

sharnoff:
> we should make it clear that what happens when agent's requested values are not multiple of CU

Yeah, ultimately this is kind of tricky to deal with, because the agent may be forced to make requests that aren't a multiple of CU if the VM's configured min/max aren't a multiple of CU. TBH there's not much we can do about that, but IMO it's not worth worrying about.
wdyt?

shayanh:
@sharnoff Makes sense. I meant we should document plugin's behaviour in this case.

pkg/plugin/run.go
shayanh: I see. I'm a fan of using constants myself. We can leave this for now and discuss it later.

shayanh (approved):

sharnoff:
Ah yep, sorry :sweat_smile: Done now.

Issue #741, opened 2024-01-16
autoscaler-agent periodic requests can become synchronized


sharnoff:
## Problem description / Motivation

Pretty much the title. As a result of this, the periodic requests we make (metrics, scheduler plugin) end up being synchronized after restart, resulting in uneven second-to-second CPU usage by the autoscaler-agent.

This is relatively low priority, tbh, but might be worth doing anyways, just so the logs distribution doesn't look weird and spiky after restart, e.g.:

![graph showing small spikes with a consistent baseline sharply transitioning to higher spikes with no baseline](https://github.com/neondatabase/autoscaling/assets/29154784/9a15e10f-e187-4dea-a0a5-44bd63fb709a)

## Feature idea(s) / DoD

autoscaler-agent's periodic requests should be randomly distributed

## Implementation ideas

On `agent.Runner` startup, wait for a random delay before making the first metrics request.

For periodic scheduler plugin requests, we could randomize the period itself for each `agent.Runner`

---

```[tasklist]
# Tasks
- [ ] #745
- [ ] #746
```

sharnoff:
Sure- here's e.g. autoscaler-agent log volume across all instances in us-east-2 after rollout earlier this week ([link](https://neonprod.grafana.net/explore?schemaVersion=1&panes=%7B%22wm6%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22sum%28count_over_time%28%7Bcontainer%3D%5C%22autoscaler-agent%5C%22,%20cluster%3D%5C%22prod-us-east-2-delta%5C%22%7D%20%5B$__interval%5D%29%29%22,%22queryType%22:%22range%22,%22refId%22:%22A%22%7D%5D,%22range%22:%7B%22from%22:%221705428600000%22,%22to%22:%221705429500000%22%7D%7D%7D&orgId=1))

![graph from 18:10 to 18:25 showing relatively steady line at 1K switching around 18:17:30 to a much spikier load quickly oscillating between 0.5K and 2-3K](https://github.com/neondatabase/autoscaling/assets/29154784/148f8cde-c258-4fe8-8059-253ada615f92)

Finer grained requests data is a little more annoying because the metrics aren't high enough resolution, but it shouldn't be too tricky to write a logs query to fetch the info

PR #745, opened 2024-01-17
agent: Add small random amount to plugin request tick

sharnoff:
Part of #741, see there for more info. Basically, we generally want to avoid letting requests become synchronized across VMs, and this is one part of doing that.

### DIFF
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -190,8 +190,8 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 
 	execLogger := logger.Named("exec")
 
-	// Add a small random amount to core.Config.PluginRequestTick so that periodic requests tend to
-	// become distribted randomly over time.
+	// Subtract a small random amount from core.Config.PluginRequestTick so that periodic requests
+	// tend to become distribted randomly over time.
 	pluginRequestJitter := util.NewTimeRange(time.Millisecond, 0, 100).Random()
 
 	coreExecLogger := execLogger.Named("core")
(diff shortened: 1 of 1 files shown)


pkg/agent/runner.go
```diff
@@ -190,14 +190,18 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
+
 	coreExecLogger := execLogger.Named("core")
 	executorCore := executor.NewExecutorCore(coreExecLogger, getVmInfo(), executor.Config{
 		OnNextActions: r.global.metrics.runnerNextActions.Inc,
 		Core: core.Config{
 			ComputeUnit:                        r.global.config.Scaling.ComputeUnit,
 			DefaultScalingConfig:               r.global.config.Scaling.DefaultConfig,
 			NeonVMRetryWait:                    time.Second * time.Duration(r.global.config.Scaling.RetryFailedRequestSeconds),
-			PluginRequestTick:                  time.Second * time.Duration(r.global.config.Scheduler.RequestAtLeastEverySeconds),
+			PluginRequestTick:                  time.Second*time.Duration(r.global.config.Scheduler.RequestAtLeastEverySeconds) + pluginRequestJitter,
```
Omrigan, line 204: nit: maybe subtract here? Because `RequestAtLeastEverySeconds = 1` might not give *at least every second*, but *every 1.1 second*.

Also, do we ideally want to randomize every request delay, not just at the start?

Omrigan (approved):

pkg/agent/runner.go
sharnoff: Ah, good catch :)

My thinking for modifying `PluginRequestTick` itself (rather than waiting at the start) is twofold:

1. Unlike fetching metrics, we _really do_ want to let the plugin know as soon as possible, because delay forces it to act on uncertain data (this probably isn't a _huge_ issue, but maybe it could be, if a ton of VMs are created all at once)
2. There may still be ways for the requests for each VM to become synchronized after startup, although none come to mind at the moment. (I guess in the future, with neondatabase/cloud#6701 this would be more likely, but for now not as much).

sharnoff: Switched to subtraction in bc53fe566f722877ad0e4531aaa58076512782d5

shayanh (approved):

Issue #733, opened 2024-01-11
Neonvm controller codebase improvements


shayanh:
## Motivation

The current implementation of VirtualMachineController and VirtualMachineMigrationController are mainly implemented in one single function. Each of these reconcile functions are about 500 lines long, which makes it a bit hard to read and navigate through the codebase. We can refactor the implementation into smaller function organized in the right set of files.

Smaller issues:
- The events we emit have redundant information. We can omit the virtual machine's name on the message. For example: https://github.com/neondatabase/autoscaling/blob/a23780d8d9c338fbe0e2ea3c123f62357803fb2d/neonvm/controllers/virtualmachine_controller.go#L293-L296
- Organize the readme file. It seems some parts are repeated both in the "Getting Started" and the "Local development" sections.

sharnoff:
The redundant information in events was fixed by #1344.

For the issues with the reconcile function - @Omrigan do you know if we already have an issue for those? (I vaguely recall something like that, but don't want to close this one without linking it)

PR #749, opened 2024-01-19
neonvm: Use container statuses, not pod phase, to trigger restart

sharnoff:
Extracted from #738, which adds a second container to the runner pods. Because of that second container, if only one container exits, the pod will still have `.status.phase = Running`, so we need to proactively notice that one of the containers has stopped and propagate that status to the VM itself.

This also introduces some funky logic around how we handle restarts: Because the `Succeeded` and `Failed` phases no longer imply that QMEU itself has stopped, we need to explicitly wait until either the pod is gone or the neonvm-runner container has stopped; otherwise we could end up with >1 instance of the VM running at a time.

---

Tested locally as part of #738 and validated that it has the intended effect and works as expected when I `kubectl exec <pod> -c <container> -- kill 1`.

NB: Some follow-up changes in #754 building on this. See there for details.

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -429,8 +429,8 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 			return err
 		}
 		// runner pod found, check phase
-		switch vmRunner.Status.Phase {
-		case corev1.PodRunning:
+		switch runnerContainerStatus(vmRunner) {
+		case runnerRunning:
 			virtualmachine.Status.PodIP = vmRunner.Status.PodIP
 			virtualmachine.Status.Phase = vmv1.VmRunning
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
@@ -438,21 +438,21 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) created successfully", virtualmachine.Status.PodName, virtualmachine.Name)})
-		case corev1.PodSucceeded:
+		case runnerSucceeded:
 			virtualmachine.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeAvailableVirtualMachine,
 					Status:  metav1.ConditionFalse,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) succeeded", virtualmachine.Status.PodName, virtualmachine.Name)})
-		case corev1.PodFailed:
+		case runnerFailed:
 			virtualmachine.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) failed", virtualmachine.Status.PodName, virtualmachine.Name)})
-		case corev1.PodUnknown:
+		case runnerUnknown:
 			virtualmachine.Status.Phase = vmv1.VmPending
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeAvailableVirtualMachine,
@@ -489,8 +489,8 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		}
 
 		// runner pod found, check/update phase now
-		switch vmRunner.Status.Phase {
-		case corev1.PodRunning:
+		switch runnerContainerStatus(vmRunner) {
+		case runnerRunning:
 			// update status by IP of runner pod
 			virtualmachine.Status.PodIP = vmRunner.Status.PodIP
 			// update phase
@@ -557,21 +557,21 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 				virtualmachine.Status.Phase = vmv1.VmScaling
 			}
 
-		case corev1.PodSucceeded:
+		case runnerSucceeded:
 			virtualmachine.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeAvailableVirtualMachine,
 					Status:  metav1.ConditionFalse,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) succeeded", virtualmachine.Status.PodName, virtualmachine.Name)})
-		case corev1.PodFailed:
+		case runnerFailed:
 			virtualmachine.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) failed", virtualmachine.Status.PodName, virtualmachine.Name)})
-		case corev1.PodUnknown:
+		case runnerUnknown:
 			virtualmachine.Status.Phase = vmv1.VmPending
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeAvailableVirtualMachine,
@@ -609,8 +609,8 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		}
 
 		// runner pod found, check that it's still up:
-		switch vmRunner.Status.Phase {
-		case corev1.PodSucceeded:
+		switch runnerContainerStatus(vmRunner) {
+		case runnerSucceeded:
 			virtualmachine.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&virtualmachine.Status.Conditions,
 				metav1.Condition{Type: typeAvailableVirtualMachine,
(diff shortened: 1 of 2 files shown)


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+		return runnerUnknown
+
+	// With runners with neonvm-container-mgr, the pod phase itself won't be updated if just the
+	// neonvm-runner container exits or completes.
+	// So we need to manually check through the container statuses to check whether they're ok.
+	case corev1.PodRunning:
+		completed := false
+
+		for _, stat := range pod.Status.ContainerStatuses {
+			if stat.State.Terminated != nil {
```
Omrigan, line 818: Wait, this checks all containers, while from the function name I'd expect it to check for `runner` state. Which one is correct?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
 		}
 
 	default:
-		// do nothing
+		panic(fmt.Errorf("unknown pod phase: %q", pod.Status.Phase))
 	}
+}
 
-	return nil
+func runnerContainerStopped(pod *corev1.Pod) bool {
```
Omrigan, line 839: from the function name, i'd expect it to be something like `runnerContainerStatus(pod) != runnerRunning`

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
-			// get runner to delete
-			vmRunner := &corev1.Pod{}
-			err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
-			if err == nil {
-				// delete current runner
-				if err := r.deleteRunnerPodIfEnabled(ctx, virtualmachine, vmRunner); err != nil {
-					return err
-				}
-			} else if !apierrors.IsNotFound(err) {
+		// Always delete runner pod. Otherwise, we could end up with one container succeeded/failed
```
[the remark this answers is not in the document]
Omrigan: Perhaps somewhere exists a state diagram which would explain?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -609,24 +609,24 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 		}
 
 		// runner pod found, check that it's still up:
-		switch vmRunner.Status.Phase {
-		case corev1.PodSucceeded:
+		switch runnerContainerStatus(vmRunner) {
```
Omrigan, line 612: This function is in dire need of refactoring. 500+ lines and nested switch cases make it hard to follow, and e.g. to understand which `case`/`default` applies to which `switch`.

shayanh: see #733

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
+
+const (
+	runnerUnknown   runnerStatus = "Unknown"
+	runnerPending   runnerStatus = "Pending"
+	runnerRunning   runnerStatus = "Running"
+	runnerFailed    runnerStatus = "Failed"
+	runnerSucceeded runnerStatus = "Succeeded"
+)
+
+func runnerContainerStatus(pod *corev1.Pod) runnerStatus {
```
shayanh, line 800: Is there a way to declare in k8s if one of the containers in the pod fails, then the entire pod also should fail?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
-			if err == nil {
-				// delete current runner
-				if err := r.deleteRunnerPodIfEnabled(ctx, virtualmachine, vmRunner); err != nil {
-					return err
-				}
-			} else if !apierrors.IsNotFound(err) {
+		// Always delete runner pod. Otherwise, we could end up with one container succeeded/failed
+		// but the other one still running (meaning that the pod still ends up Running).
+		vmRunner := &corev1.Pod{}
+		err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
```
shayanh, line 750: When dealing with state machines like this, it's easier to track down what's going on where each case is handled separately. Now it's kinda hard to tell what we do in case of `vmv1.RestartPolicyAlways` and `vmv1.RestartPolicyAlways`. I suggest handling each case separately as before.

neonvm/controllers/virtualmachine_controller.go
shayanh: Also this function doesn't return runner container status exactly, which is misleading.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -744,55 +744,109 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
 			}
-			// delete runner only when VM failed
-			if found && virtualmachine.Status.Phase == vmv1.VmFailed {
-				// delete current runner
-				if err := r.deleteRunnerPodIfEnabled(ctx, virtualmachine, vmRunner); err != nil {
-					return err
+
+			// TODO for RestartPolicyNever: implement TTL or do nothing
+		}
+	default:
```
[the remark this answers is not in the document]
sharnoff: Honestly not sure; don't want to remove it in case it's load-bearing :sweat_smile:
Probably we'll drop it as part of refactoring all this.

neonvm/controllers/virtualmachine_controller.go
sharnoff: > Is there a way

I spent a while looking and couldn't find any way to do this. K8s 1.29 introduced ["sidecar containers"](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) as an explicit feature, which _do_ allow this.

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Yeah; it's a bit confusing - I'm not sure there's a better way to do it though.

I also don't think there's an existing state diagram. I'll try to give an explanation here, for this part of the state diagram. Hopefully we can make this clearer both in code and in documentation as part of #733.

---

## Before this PR:

| Runner pod status | VM status |
|---|---|
| Overlay IP not acquired _or_ mid-restart | `""` |
| Pod doesn't exist | `"Pending"` |
| `"Pending"` | `"Pending"` |
| `"Running"` | `"Running"`, `"Scaling"`, `"PreMigrating"`, `"Migrating"` |
| `"Succeeded"` | `"Succeeded"` |
| `"Failed"` | `"Failed"` |

Note that the _Pod_ is "Succeeded" if all containers exit successfully, and "Failed" if all containers exit and at least one failed. This means that if there is at least one container still running (like `container-mgr`), the pod will continue to be in the "Running" phase.

Also note: When we say "Pod doesn't exist", technically the controller (before this PR) does not wait for the pod to be fully deleted when restarting. By the time the VM gets to "Succeeded" or "Failed" (prompting restart), we already know that there's nothing running in the pod, so we can just _start_ deleting the pod and don't need to wait for it to finish:

https://github.com/neondatabase/autoscaling/blob/a7943db25bf5d97eaf4771b46e10a3d97ed4b3ea/neonvm/controllers/virtualmachine_controller.go#L748-L764

(nb: `virtualmachine.Cleanup()` sets `.Status.PodName = ""`, among other things)

## After this PR:

| Runner pod status | VM status |
|---|---|
| Overlay IP not acquired | `""` |
| Pod doesn't exist | `"Pending"` (or `"Succeeded"` / `"Failed"` on exit, if no restart) |
| `"Pending"` | `"Pending"` |
| `"Running"` (and no containers exited) | `"Running"`, `"Scaling"`, `"PreMigrating"`, `"Migrating"` |
| `"Running"` (neonvm-runner exited successfully) | `"Succeeded"` |
| `"Running"` (other container exited successfully) | `"Failed"` |
| `"Running"` (≥1 container failed) | `"Failed"` |
| `"Failed"` | `"Failed"` (but pod should be immediately terminated, to ensure other containers stop) | 
| `"Succeeded"` | `"Succeeded"` (but pod should be immediately terminated, to ensure other containers stop) |

neonvm/controllers/virtualmachine_controller.go
sharnoff: Updated the function with a comment describing its semantics (eb8368ed5e3cbfb29162f4c9997eec3d39744769). Is that more consistent?

neonvm/controllers/virtualmachine_controller.go
sharnoff: Changed the name to `runnerStatus` (1da6e055e8c6541325cbf59b5679608bb9dca11c), also to help differentiate from `runnerContainerStopped`. wdyt?

neonvm/controllers/virtualmachine_controller.go
sharnoff: Fair enough, reintroduced that explicit handling but kept the deduplication for the restarting logic (5c19ad6601ccedc2e2d1e32418b816dd1e005acd). wdyt?

shayanh: Looks good

neonvm/controllers/virtualmachine_controller.go
Omrigan: Yes, makes sense

neonvm/controllers/virtualmachine_controller.go
Omrigan: Thanks, it's clear now.

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
Omrigan: That's a good table. I'd consider putting it to https://github.com/neondatabase/autoscaling/blob/main/neonvm/README.md

Omrigan (approved):

shayanh (approved):

PR #751, opened 2024-01-21
lint: add actionlint

bayandin:
This PR adds [`actionlint`](https://github.com/rhysd/actionlint), a linter for GitHub Workflow files.
It also checks inlined bash scripts with [`shellcheck`](https://www.shellcheck.net/).

I've disabled a couple of noisy `shellcheck` rules (`SC2046`, `SC2086`), which in most cases produce benign warnings.
The action itself copied from `neondatabase/neon` (https://github.com/neondatabase/neon/pull/5265).

Run it locally (you should have both `actionlint`, and `shellcheck` installed):
```bash
SHELLCHECK_OPTS=--exclude=SC2046,SC2086 actionlint
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -96,3 +96,17 @@ jobs:
         with:
           check_hidden: true
           skip: go.sum,*.patch # '*.patch' references cluster-autoscaler/ca.patch, but somehow skipping directly doesn't work...
+
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: reviewdog/action-actionlint@54e8faeff91fe07595b1ad3cfdc6aee0157bf51b # v1.40.0
+        env:
+          # SC2046 - Quote this to prevent word splitting.                - https://www.shellcheck.net/wiki/SC2046
+          # SC2086 - Double quote to prevent globbing and word splitting. - https://www.shellcheck.net/wiki/SC2086
+          SHELLCHECK_OPTS: --exclude=SC2046,SC2086
+        with:
+          fail_on_error: true
+          filter_mode: nofilter
+          level: error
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -40,7 +44,7 @@ jobs:
           go mod tidy
       - name: check diff
         run: |
-          if ! (test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"); then
+          if ! test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```
bayandin, line 47:
```suggestion
          if test -n "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```

.github/workflows/lint.yml
```diff
@@ -61,7 +65,7 @@ jobs:
           make manifests
       - name: check diff
         run: |
-          if ! (test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"); then
+          if ! test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```
bayandin, line 68:
```suggestion
          if test -n "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```

.github/workflows/lint.yml
```diff
@@ -79,7 +83,7 @@ jobs:
           make generate
       - name: check diff
         run: |
-          if ! (test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"); then
+          if ! test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```
bayandin, line 86:
```suggestion
          if test -n "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```

bayandin (commented):
Should we use `test -n` instead of `! test -z`?

sharnoff (approved):
lgtm! thanks!

Honestly not sure whether `test -n` is better than `! test -z` (`-n` is simpler, but it feels like `-z` is more common / well-known). IMO, feel free to go with whichever you'd prefer.

PR #753, opened 2024-01-22
neonvm-controller: Fix overwriting runner version

sharnoff:
Noticed while working on #738. In short, because the runner API version was part of `labelsForVirtualMachine`, any update to the runner version would be updated for *all* VM pods, not just new one.

This is (probably) not an issue in prod right now, but could be an issue for future changes. ~~Right now, #738 would trigger this bug, but remain unaffected (it doesn't have version-dependent logic *currently*).~~

This PR fixes the behavior by adding the runner API version as an explicit argument to `labelsForVirtualMachine` and ignoring the label in `updatePodMetadataIfNecessary`.

---

Notes for review: Planning to merge 48h after 1 approval, or 24h after 2 approvals. Want to give people a chance to take a look, but overall change is not too complex.

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -837,10 +837,14 @@ func updatePodMetadataIfNecessary(ctx context.Context, c client.Client, vm *vmv1
 		ignoreExtra map[string]bool // use bool here so `if ignoreExtra[key] { ... }` works
 	}{
 		{
-			metaField:   "labels",
-			expected:    labelsForVirtualMachine(vm),
-			actual:      runnerPod.Labels,
-			ignoreExtra: map[string]bool{},
+			metaField: "labels",
+			expected:  labelsForVirtualMachine(vm, nil), // don't include runner version
+			actual:    runnerPod.Labels,
+			ignoreExtra: map[string]bool{
+				// Don't override the runner pod version - we need to keep it around without
+				// changing it; otherwise it's not useful!
+				vmv1.RunnerPodVersionLabel: true,
+			},
 		},
 		{
 			metaField: "annotations",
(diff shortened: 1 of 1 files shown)


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1014,15 +1018,17 @@ func sshSecretSpec(virtualmachine *vmv1.VirtualMachine) (*corev1.Secret, error)
 
 // labelsForVirtualMachine returns the labels for selecting the resources
 // More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
-func labelsForVirtualMachine(virtualmachine *vmv1.VirtualMachine) map[string]string {
+func labelsForVirtualMachine(virtualmachine *vmv1.VirtualMachine, runnerVersion *api.RunnerProtoVersion) map[string]string {
```
Omrigan, line 1021: Why does it have to be a pointer?

Omrigan: Ah, because you check for nil

Omrigan (approved):

neonvm/controllers/virtualmachine_controller.go
sharnoff: There's other ways to do it (originally I was going to make it a bool and set the version within the function), but the reasoning is because there's two slightly different requirements:

1. in `podSpec`, when we're generating the pod to create, we want the labels to include the runner API version
2. in `updatePodMetadataIfNecessary`, we want the 'desired' labels to *not* include the runner API version, so that the existing ignoring logic respects that

the pointer solution is kind of annoying though, so I'm open to other ideas if you've got something :)

(the reason I didn't go with the bool approach is because boolean args often make readability hard)

shayanh (approved):

PR #757, opened 2024-01-22
neonvm: Custom reconciler metrics

### LINKED ISSUE #247: no observability into neonvm reconcile failures (needs metrics!)

shayanh:
Providing custom metrics for reconciler objects.

- `reconcile_failing_objects` represents the number of objects that are failing to reconcile for each specific controller.

Fixes #247 (along with #739).

### DIFF
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -1,0 +1,64 @@
+package controllers
+
+import (
+	"context"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type ReconcilerMetrics struct {
+	failing *prometheus.GaugeVec
+}
+
+func MakeReconcilerMetrics() ReconcilerMetrics {
+	m := ReconcilerMetrics{
+		failing: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "failing",
+				Help: "number of failing objects",
+			},
+			[]string{"controller"},
+		),
+	}
+	metrics.Registry.MustRegister(m.failing)
+	return m
+}
+
+type wrappedReconciler struct {
+	ControllerName string
+	Reconciler     reconcile.Reconciler
+	Metrics        ReconcilerMetrics
+
+	lock    sync.Mutex
+	failing map[client.ObjectKey]struct{}
+}
+
+func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) reconcile.Reconciler {
+	return &wrappedReconciler{
+		ControllerName: cntrlName,
+		Reconciler:     reconciler,
+		Metrics:        rm,
+		lock:           sync.Mutex{},
+		failing:        make(map[client.ObjectKey]struct{}),
+	}
+}
+
+func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	res, err := d.Reconciler.Reconcile(ctx, req)
+
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	if err != nil {
+		d.failing[req.NamespacedName] = struct{}{}
+	} else {
+		delete(d.failing, req.NamespacedName)
+	}
+	d.Metrics.failing.WithLabelValues(d.ControllerName).Set(float64(len(d.failing)))
+
+	return res, err
+}
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -53,9 +53,8 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	res, err := d.Reconciler.Reconcile(ctx, req)
 
 	// This part is executed sequentially since we acquire a mutex lock. It
-	// should be quite fast since a mutex lock/unlock + 1 memory write takes
-	// ~60ns. Having 8 reconile workers, we wait at most 8*60=480ns to acquire
-	// the lock. I (@shayanh) preferred to go with the simplest implementation
+	// should be quite fast since a mutex lock/unlock + 2 memory writes takes less
+	// than 100ns. I (@shayanh) preferred to go with the simplest implementation
 	// as of now. For a more performant solution, if needed, we can switch to an
 	// async approach.
 	d.lock.Lock()
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -19,8 +19,8 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
 	m := ReconcilerMetrics{
 		failing: prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
-				Name: "failing",
-				Help: "number of failing objects",
+				Name: "reconcile_failing_objects",
+				Help: "Number of objects that are failing to reconcile for each specific controller",
 			},
 			[]string{"controller"},
 		),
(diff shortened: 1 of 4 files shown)


neonvm/controllers/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	res, err := d.Reconciler.Reconcile(ctx, req)
+
+	// This part is executed sequentially since we acquire a mutex lock. It
+	// should be quite fast since a mutex lock/unlock + 1 memory write takes
+	// ~60ns. Having 8 reconile workers, we wait at most 8*60=480ns to acquire
+	// the lock. I (@shayanh) preferred to go with the simplest implementation
+	// as of now. For a more performant solution, if needed, we can switch to an
+	// async approach.
```
sharnoff, line 60: fwiw, I think it's more than "1 memory write" — hashmap access plus whatever `WithLabelValues(..).Set(..)` does is probably non-trivial.
But yeah, as long as we're not holding a mutex while doing IO we should be good :)

neonvm/controllers/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	if err != nil {
+		d.failing[req.NamespacedName] = struct{}{}
```
sharnoff, line 64: Is it possible for an object to be deleted while still failing? (maybe with `kubectl delete --force`?)

If not, it'd be good to have a brief explainer why. If so, any ideas how we can avoid false positives in the failing count?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -74,6 +74,8 @@ type VirtualMachineReconciler struct {
 	client.Client
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
+
+	Metrics ReconcilerMetrics `exhaustruct:"optional"`
```
sharnoff, line 78: TIL there's exhaustruct struct tags - nice! :D

sharnoff: Is there a reason why it's needed here and not for `VirtualMachineMigrationReconciler` ?

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -666,11 +668,14 @@ func (r *VirtualMachineMigrationReconciler) doFinalizerOperationsForVirtualMachi
...
 // desirable state on the cluster
 func (r *VirtualMachineMigrationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	cntrlName := "virtualmachinemigration"
+	reconciler := WithMetrics(r, r.Metrics, cntrlName)
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&vmv1.VirtualMachineMigration{}).
 		Owns(&corev1.Pod{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 8}).
-		Complete(r)
+		Named(cntrlName).
```
sharnoff, line 677: not a blocker: what's the impact of `Named` here?

neonvm/controllers/virtualmachine_controller.go
shayanh: It's because of tests. There is a test that we have initialized a `VirtualMachineReconciler`.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: Sets the name of the controller. Its used in metrics particularly.

https://github.com/kubernetes-sigs/controller-runtime/blob/b88ed7a3602b85b6cfd0acfe9c25033b978bdb83/pkg/builder/controller.go#L210-L214

neonvm/controllers/metrics.go
shayanh: I forgot about `WithLabelValues(..).Set(..)` but it's an `atomic.StoreUint64`.

neonvm/controllers/metrics.go
shayanh: I think if an object is deleted then it still goes through the `Reconcile` function but we treat it differently.

https://github.com/neondatabase/autoscaling/blob/ac66e2c7b11e7a8fd209f300a208de5d2003585a/neonvm/controllers/virtualmachine_controller.go#L140-L158

sharnoff: Is that still guaranteed for `--force` ?

neonvm/controllers/virtualmachinemigration_controller.go
sharnoff: Ah fair enough - so this is more a continuation of #739?

neonvm/controllers/metrics.go
sharnoff: Hm, I appreciate the concern for efficiency. In this case, I'd recommend dropping the exact numbers — the metrics handling isn't trivial¹ and IMO it's not worth it to actually collect exact timings. Either way though, I'd be shocked if manipulating the metrics was a bottleneck, given the number of http requests we make while reconciling :sweat_smile:

---

¹ For example, in the case where the metric already exists, we still end up calling:

- [`(*GaugeVec).GetMetricWithLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/gauge.go#L182-L188)
  - [`(*MetricVec).GetMetricWithLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L201-L208)
    - [`(*MetricVec).hashLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L235-L256) (
      - [`validateLabelValues`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/labels.go#L68-L84) (should be relatively cheap)
      - `(*MetricVec).hashAdd{,Byte}` (internally [`hashAdd`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/fnv.go#L29-L42)) (computes FNV64a hash of the label value(s) - realistically, pretty cheap, but non-zero)
    - [`(*metricMap).getOrCreateMetricWithLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L474-L493) (does RWMutex lock/unlock)
      - [`(*metricMap).getMetricWithHashAndLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L522-L532) (does hashmap lookup)
        - [`findMetricWithlabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L550-L559) (iterates through hashmap bucket - should be n=1)
          - [`matchLabelValues()`](https://github.com/prometheus/client_golang/blob/v1.14.0/prometheus/vec.go#L574-L593) (somewhat complex control flow, but should be pretty cheap for 1 label value, not too long)

[^examples]: For example, in the case where the metric already exists, we still end up transitively calling `(*GaugeVec).GetMetricWithLabelValues()` → `

shayanh: You're right. The set is the `atomic.StoreUint64`. I already removed the numbers from the comment.

neonvm/controllers/virtualmachinemigration_controller.go
shayanh: Right, I want to expose the metrics in the same format at kubebuilder does.

neonvm/controllers/metrics.go
shayanh: Yes, `kubectl --force` doesn't wait for confirmation. It's a client-side behaviour. More info: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_delete/

sharnoff: Tried it out locally and I see that `--force` still respects the finalizer; fair enough

sharnoff (approved):

PR #758, opened 2024-01-23
Add eBPF for network monitoring based on containermgr

save-buffer:


sharnoff:
Heads-up, I rebased #738 and made some changes that introduced a couple conflicts. `git rebase --onto origin/sharnoff/neonvm-runner-crictl HEAD^^` should _mostly_ do the trick.

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -455,6 +466,37 @@ type VirtualMachine struct {
 	Status VirtualMachineStatus `json:"status,omitempty"`
 }
 
+func (vm *VirtualMachine) GetNetworkUsage(ctx context.Context, timeout time.Duration) (*VirtualMachineNetworkUsage, error) {
```
sharnoff, line 469: Still thinking we need some way to avoid trying to fetch network usage for old VMs that don't support it. Happy to implement that if it'd be helpful.

neonvm/runner/container-mgr/main.go
```diff
@@ -139,14 +143,201 @@ func main() {
...
+		asm.StoreXAdd(asm.R0, asm.R6, asm.DWord),
+		asm.Mov.Imm(asm.R0, 1).WithSymbol("exit"),
+		asm.Return(),
+	)
+
+	//nolint:exhaustruct // This has a whole bunch of optional configuration options that we don't care about
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Name:         "neon_network_monitor",
+		Type:         ebpf.CGroupSKB,
+		License:      "GPL",
```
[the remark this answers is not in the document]
save-buffer: Ah yeah, I changed it to GPL  for the printing thing because I thik I originally had "Apache 2.0" and it was rejecting it. Changed it to "Apache-2.0" and I think it works now?

neonvm/runner/container-mgr/crictl.go
```diff
@@ -110,6 +110,22 @@ type crictlContainerResourcesCPU struct {
 	Shares uint64 `json:"shares"`
 }
 
+type crictlPodInspect struct {
```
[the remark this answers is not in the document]
save-buffer: Switched

neonvm/runner/container-mgr/main.go
```diff
@@ -139,14 +143,201 @@ func main() {
 		logger.Fatal("could not set initial runner container CPU", zap.Error(err))
 	}
 
+	var podInspect crictlPodInspect
+	if err := crictl.run(logger, &podInspect, "inspectp", criPodID); err != nil {
```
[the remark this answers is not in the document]
save-buffer: Got it, changed it

pkg/agent/billing/billing.go
```diff
@@ -130,7 +143,30 @@ func RunBillingMetricsCollector(
...
+	endpointID := vm.Annotations[api.AnnotationBillingEndpointID]
+	metricsChan <- vmMetricsKV{
+		key: metricsKey{
+			uid:        vm.UID,
+			endpointID: endpointID,
+		},
+		value: vmMetricsInstant{
+			cpu:          *vm.Status.CPUs,
+			ingressBytes: byteCounts.IngressBytes,
+			egressBytes:  byteCounts.EgressBytes,
```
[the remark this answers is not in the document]
save-buffer: Good point, for this it's the absolute value and it gets subtracted off later. Added a comment

pkg/agent/billing/billing.go
```diff
@@ -130,7 +143,30 @@ func RunBillingMetricsCollector(
 	}
 }
 
-func (s *metricsState) collect(logger *zap.Logger, store VMStoreForNode, metrics PromMetrics) {
+func collectMetricsForVM(ctx context.Context, logger *zap.Logger, vm *vmapi.VirtualMachine, metricsChan chan vmMetricsKV) {
+	byteCounts, err := vm.GetNetworkUsage(ctx, 500*time.Millisecond)
+	if err != nil {
+		logger.Error("Failed to collect network usage", util.VMNameFields(vm), zap.Error(err))
```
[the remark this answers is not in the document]
save-buffer: What ever's easiest. Where do metrics currently get exported?

neonvm/apis/neonvm/v1/virtualmachine_types.go
save-buffer: Won't the requests just fail and it'll be fine? But yeah if there's an easy way to do it that would be good

pkg/agent/billing/billing.go
```diff
@@ -157,20 +198,23 @@ func (s *metricsState) collect(logger *zap.Logger, store VMStoreForNode, metrics
 		if !vm.Status.Phase.IsAlive() || vm.Status.CPUs == nil {
 			continue
 		}
+		go collectMetricsForVM(ctx, logger, vm, metricsChan)
+		metricsToCollect += 1
+	}
+
+	for i := 0; i < metricsToCollect; i++ {
+		kv := <-metricsChan
```
sharnoff, line 206: Directly receiving from a channel like this is scary. If we have a logic error, the program might deadlock.

Let's (a) introduce a timeout _within_ calls to `collect` (maybe e.g. half of the collection period?) and (b) select over whether that context/timeout expires here, while waiting to receive.

neonvm/runner/container-mgr/crictl.go
```diff
@@ -110,6 +118,22 @@ type CrictlContainerResourcesCPU struct {
 	Shares uint64 `json:"shares"`
 }
 
+type CrictlPodInspect struct {
```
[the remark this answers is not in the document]
save-buffer: Added

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1400,6 +1400,16 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret, conf
 				containerMgr := corev1.Container{
 					Image: image,
 					Name:  "neonvm-container-mgr",
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &[]bool{true}[0],
```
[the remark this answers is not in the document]
save-buffer: Yes, I wasn't able to get any combination of capabilities to work without privileged (see https://github.com/kubernetes/kubernetes/issues/121190)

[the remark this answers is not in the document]
sharnoff: Huh, we already had rw cgroups access without privileged before, if that's the problem. (we manually mounted the entire `/sys/fs/cgroup`).

I have some ideas here, will try that out today.

pkg/agent/billing/billing.go
save-buffer: Added a select based on the context's timeout.

pkg/agent/billing/billing.go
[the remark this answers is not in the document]
save-buffer: Just added some metrics collection, please take a look

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Tried it out. Didn't work - now realizing eBPF access requires you to be acting as root on the host (i.e. privileged), because otherwise it's too easy to break out of the container.

Weird that CAP_BPF doesn't work though...

If we set `Privileged: true`, the other capabilities are not needed.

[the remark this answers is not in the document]
save-buffer: Got rid of the capabilities

neonvm/apis/neonvm/v1/virtualmachine_types.go
save-buffer: Added a flag to the config, and I tested that its absence disables the network monitoring stuff

neonvm/runner/container-mgr/main.go
```diff
@@ -139,14 +143,201 @@ func main() {
...
+}
+
+// Attaches a BPF program to the given cgroup which checks each packet to see if it comes
+// from/is going to the external network. These counts are exposed through a map which
+// can be read.
+func attachBPF(logger *zap.Logger, cgroupPath string) ([]link.Link, *ebpf.Map, error) {
+	const (
+		MAP_KEY_SIZE     = 4
+		MAP_VAL_SIZE     = 8
+		cgroupMountPoint = "/sys/fs/cgroup"
```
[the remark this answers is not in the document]
save-buffer: From inside the container-mgr, I get the following
```
/ # cat /proc/self/cgroup
0::/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-pod3c6b0cdf_bfa6_47ce_ae7d_a94af20f903d.slice/cri-containerd-0329f98223d9d3df357a286d2817314401777bf3e9e5432f675accf4c5f259c4.scope
```

So I don't think that'll tell us what the cgroup mount point is right?

neonvm/runner/container-mgr/main.go
```diff
@@ -184,14 +388,48 @@ func (s *cpuServerState) listenForCPUChanges(ctx context.Context, logger *zap.Lo
...
+		logger.Error("could not marshal byte counts", zap.Error(err))
+		w.WriteHeader(500)
+		return
+	}
+
+	w.Header().Add("Content-Type", "application/json")
+	if _, err := w.Write(body); err != nil {
+		logger.Error("could not write body", zap.Error(err))
+		w.WriteHeader(500)
+		return
```
sharnoff, line 430: I _think_ writing the body preemptively writes the header already (which I guess makes sense if you think about the structure of an HTTP request/response).

Elsewhere, we just ignore the error — if writing the response fails, there's not much we can do.

pkg/agent/billing/billing.go
```diff
@@ -130,7 +161,46 @@ func RunBillingMetricsCollector(
 	}
 }
 
-func (s *metricsState) collect(logger *zap.Logger, store VMStoreForNode, metrics PromMetrics) {
+func collectMetricsForVM(ctx context.Context, logger *zap.Logger, vm *vmapi.VirtualMachine, metricsChan chan vmMetricsKV, promMetrics *PromMetrics) {
```
sharnoff, line 164: formatting nit: split into multiple lines?

pkg/agent/billing/billing.go
```diff
@@ -130,7 +161,46 @@ func RunBillingMetricsCollector(
...
+	byteCounts, err := vm.GetNetworkUsage(ctx, 500*time.Millisecond)
```
sharnoff, line 165: If the VM doesn't have the relevant flag enabled, let's just skip trying to fetch network usage?

pkg/agent/billing/billing.go
```diff
@@ -130,7 +161,46 @@ func RunBillingMetricsCollector(
...
+	if err != nil {
+		logger.Error("Failed to collect network usage", util.VMNameFields(vm), zap.Error(err))
+
+		var rootErr string
+		//nolint:errorlint // This is ok because vmapi guarantees the error types
+		switch err.(type) {
+		case vmapi.NetworkUsageEndpointNotFoundError:
+			rootErr = "Endpoint not found"
+		case vmapi.NetworkUsageTimeoutError:
+			rootErr = "Timeout"
```
sharnoff, line 175: I'd probably use the same categories of errors as billing sending uses - JSON (although in this case it's unmarshaling, not marshaling) and HTTP code.

The intent behind those is that _generally_ other errors are pretty easy to make sense of if you only look at the root, but HTTP status code / JSON unmarshaling are hard to make sense of from the metrics.

pkg/agent/billing/billing.go
```diff
@@ -110,7 +123,16 @@ func RunBillingMetricsCollector(
...
-	state.collect(logger, store, metrics)
+	func() {
+		ctx, cancel := context.WithTimeout(backgroundCtx, time.Second*time.Duration(conf.CollectEverySeconds/2))
+		defer cancel()
+		state.collect(
+			ctx,
+			logger,
+			store,
+			metrics,
+		)
```
sharnoff, line 134: collapse into a single line?

pkg/agent/billing/billing.go
```diff
@@ -120,7 +142,16 @@ func RunBillingMetricsCollector(
...
-			state.collect(logger, store, metrics)
+			func() {
+				ctx, cancel := context.WithTimeout(backgroundCtx, time.Second*time.Duration(conf.CollectEverySeconds/2))
+				defer cancel()
+				state.collect(
+					ctx,
+					logger,
+					store,
+					metrics,
+				)
```
sharnoff, line 153: same line collapse here

pkg/agent/billing/billing.go
```diff
@@ -287,6 +371,26 @@ func (s *metricsState) drainEnqueue(logger *zap.Logger, conf *Config, hostname s
...
+		countInBatch += 1
+		queue.enqueue(logAddedEvent(logger, billing.Enrich(now, hostname, countInBatch, batchSize, &billing.IncrementalEvent{
+			MetricName:     conf.EgressBytesMetricName,
+			Type:           "", // set by billing.Enrich
+			IdempotencyKey: "", // set by billing.Enrich
+			EndpointID:     key.endpointID,
+			StartTime:      s.pushWindowStart,
+			StopTime:       now,
+			Value:          int(history.totalEgressBytes),
+		})))
```
sharnoff, line 393: Another piece of disabling code if it's not used: Can we avoid emitting events if the VM has network usage tracking disabled? (I think "are the metrics zero" might be an ok-enough proxy). Reasoning is that it allows us to use dummy values for the `MetricName` in the config, meaning that we can deploy without waiting on billing team.

nb: I guess it requires more complex around `batchSize` :)

sharnoff (commented):
Handful of small things. I think I got through everything though - not expecting unrelated follow-up requests.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1399,11 +1399,20 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret, conf
 				containerMgr := corev1.Container{
 					Image: image,
 					Name:  "neonvm-container-mgr",
-					Command: []string{
-						"container-mgr",
-						"-port", strconv.Itoa(int(virtualmachine.Spec.RunnerPort)),
-						"-init-milli-cpu", strconv.Itoa(int(*virtualmachine.Spec.Guest.CPUs.Use)),
+					SecurityContext: &corev1.SecurityContext{
```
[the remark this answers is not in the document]
save-buffer: Moved

neonvm/runner/container-mgr/crictl.go
```diff
@@ -87,6 +87,14 @@ func (c *Crictl) Inspect(logger *zap.Logger, containerID string) (*CrictlContain
 	return &container, nil
 }
 
+func (c *Crictl) InspectPod(logger *zap.Logger, podID string) (*CrictlPodInspect, error) {
+	var pod CrictlPodInspect
+	if err := c.run(logger, &pod, "inspectp", podID); err != nil {
+		return nil, err
+	}
+	return &pod, nil
+}
```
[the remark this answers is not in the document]
save-buffer: Moved

neonvm/runner/container-mgr/main.go
```diff
@@ -139,14 +145,205 @@ func main() {
...
+		defer counters.Close()
+	}
+
 	srvState := cpuServerState{
 		podID:        criPodID,
 		containerID:  criRunnerContainerID,
 		lastMilliCPU: atomic.Uint32{},
 	}
 	srvState.lastMilliCPU.Store(uint32(initMilliCPU))
+	srvState.listenForHTTPRequests(context.TODO(), logger, crictl, int32(httpPort), counters)
```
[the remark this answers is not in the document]
save-buffer: Added a newline

pkg/agent/billing/billing.go
save-buffer: Done

pkg/agent/billing/billing.go
save-buffer: Done

pkg/agent/billing/billing.go
save-buffer: Done

pkg/agent/billing/billing.go
save-buffer: Added - just had to loop over the history twice to compute the batch size.

pkg/agent/billing/billing.go
save-buffer: Added

neonvm/runner/container-mgr/main.go
save-buffer: Sounds good, I switched it to how handleCPUCurrent does it

pkg/agent/billing/billing.go
save-buffer: Switched it

Omrigan:
@save-buffer remind me please, this is blocked on two things: https://github.com/neondatabase/cloud/issues/11655 and availability of cgroups v2 in the Amazon Machine Image, right?

save-buffer:
Yes that's correct

Omrigan:
I'll close it then for the time being.

PR #767, opened 2024-01-26
CI: Fix e2e-tests failure from empty $SHA on main

sharnoff:
ref https://github.com/neondatabase/autoscaling/actions/runs/7673566092

`${{ github.event.pull_request }}` doesn't exist for 'push' events :facepalm:


sharnoff:
Going to merge this without review so main is (hopefully) happy. Happy to revert if there's issues.

PR #768, opened 2024-01-29
runner: don't log i/o timeout

Omrigan:
Last minute logging change in
  6addbdc (runner: pass logs through virtio-serial (#724), 2024-01-24)

Resulted in flooding logs with the following error message:
  "msg":"failed to read from log serial",
  "error":"read unix @->/vm/log.sock: i/o timeout",

Filtering for os.ErrDeadlineExceeded before logging fixes it.

Related to neondatabase/cloud#8602.


sharnoff (approved):
Noticed this as well yesterday - thanks!

shayanh (approved):

Issue #759, opened 2024-01-24
Epic: VM startup latency metrics & SLOs


sharnoff:
## Motivation

Startup latency has downstream effects that easily become visible to users in the form of start_compute latency.

We should minimize this where possible — and we need observability and alerting to track progress and make sure we don't regress.

## DoD

The time between VirtualMachine object creation and QEMU bootup complete is accurately represented by a metric exposed by the neonvm-controller, with SLOs and alerting based on it.

## Implementation ideas

With #754, we could add a histogram metric that we add to exactly when the VM goes to the `Running` phase with `.Status.Restarts == 0` — we'd use the difference between `time.Now()` and `.Metadata.CreationTimestamp`. This would _probably_ be good enough, but may miss certain kinds of metrics.

## Tasks

```[tasklist]

```

## Other related tasks, Epics, and links

- #594

shayanh:
I'm planning to add the following metrics to the NeonVM controller:
1. `vm_creation_to_runner_creation_duration`: VM creation timestamp to runner pod creation timestamp
2. `runner_creation_to_vm_running_duration`: Runner pod creation timestamp to the moment when NeonVM controller changes VM's status to Running from Pending.

Note that the second metric is not quite precise due to #771.

sharnoff:
status: Had a PR implementing this, was blocked on a PR by @sharnoff, but now that's merged this can continue. Plan to merge this week

sharnoff:
status: Metrics are merged & on staging. There's a [notion doc](https://www.notion.so/neondatabase/Autoscaling-Alerts-4f793e2822114465a10b4cf7daeaa57d) for planning what alerts we should have. Waiting on deploy -> prod to get a sense of numbers for the SLOs.

shayanh:
Done. Has been discussed here in [Autocaling alerting Notion doc](https://www.notion.so/neondatabase/Autoscaling-Alerts-4f793e2822114465a10b4cf7daeaa57d).

sharnoff:
@shayanh should this issue be closed?

shayanh:
Yes, I close it.

PR #774, opened 2024-01-30
neonvm: VM startup metrics

shayanh:
Implemented VM startup latency metrics.

The following metrics have been added:
1. `vm_creation_to_runner_creation_duration_seconds`: VM creation timestamp to runner pod creation timestamp
2. `vm_runner_creation_to_vm_running_duration_seconds`: Runner pod creation timestamp to the moment when NeonVM controller changes VM's status to Running from Pending.
3. `vm_creation_to_vm_running`: VM creation timestamp to the moment when NeonVM controller change VM's status to Running from Pending.

Related to #759


neonvm/controllers/metrics.go
```diff
@@ -24,8 +32,32 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
...
+				Name:    "vm_creation_to_vm_running_duration_seconds",
+				Help:    "Time duration from VirtualMachine.CreationTimeStamp to the moment when VirtualMachine.Status.Phase becomes Running",
+				Buckets: buckets,
+			},
+		),
 	}
 	metrics.Registry.MustRegister(m.failing)
+	metrics.Registry.MustRegister(m.vmCreationToRunnerCreationTime)
+	metrics.Registry.MustRegister(m.runnerCreationToVMRunningTime)
+	metrics.Registry.MustRegister(m.vmCreationToVMRunningTime)
```
sharnoff, line 60: I'd recommend using `util.RegisterMetric` here so that registering can be done at the same time as initializing (avoids the possibility of issues like #787)

neonvm/controllers/metrics.go
```diff
@@ -24,8 +32,32 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
...
+		vmCreationToRunnerCreationTime: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "vm_creation_to_runner_creation_duration_seconds",
+				Help:    "Time duration from VirtualMachine.CreationTimestamp to runner Pod.CreationTimestamp",
+				Buckets: buckets,
+			},
+		),
+		runnerCreationToVMRunningTime: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "runner_creation_to_vm_running_duration_seconds",
```
sharnoff, line 44: nit: _maybe_ prepend `vm_` to the start of this name, so that these have a common prefix?
wdyt?

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -444,6 +448,15 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
...
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) created successfully", virtualmachine.Status.PodName, virtualmachine.Name)})
+			{
+				now := time.Now()
+				d := now.Sub(vmRunner.CreationTimestamp.Time)
+				r.Metrics.runnerCreationToVMRunningTime.Observe(d.Seconds())
+				if !virtualmachine.HasRestarted() {
+					d := now.Sub(virtualmachine.CreationTimestamp.Time)
+					r.Metrics.vmCreationToVMRunningTime.Observe(d.Seconds())
+				}
+			}
```
sharnoff, line 459: Only things I'd note are:
1. This block could use a comment noting what we're doing (short one-line sorta thing)
2. IMO it'd be good to log the duration here, so that we can _in theory_ dig into the logs to figure out why metrics are weird. I'm imagining something like "we keep seeing the occasional VM that takes +inf time here [due to a bug in the state machine], how can we debug it?"

sharnoff:
realized I didn't leave a general comment for my review: Looks good, just a couple things.

neonvm/controllers/metrics.go
shayanh: fixed in c7654bb64d275c857d62acb8332b58c76a203f34

neonvm/controllers/virtualmachine_controller.go
shayanh: Makes sense. Fixed in c7654bb64d275c857d62acb8332b58c76a203f34

neonvm/controllers/metrics.go
shayanh: I tried to use that but the types are incompatible and had to type casting in runtime. See this: https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/metrics/registry.go#L30

sharnoff: Ugh, ok yeah fair. Maybe worthwhile to change `util.RegisterMetric` so that it takes `prometheus.Registerer` rather than `*prometheus.Registry`? I _think_ that'd allow it to work here

shayanh: Cool, I didn't know about the `prometheus.Registerer`. Actually, it's better to use the interface in the `util.RegisterMetric`. I'll send a separate PR for that.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -430,6 +430,10 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 				msg = fmt.Sprintf("%s, SSH Secret %s", msg, sshSecret.Name)
 			}
 			r.Recorder.Event(virtualmachine, "Normal", "Created", msg)
+			if !virtualmachine.HasRestarted() {
+				d := pod.CreationTimestamp.Time.Sub(virtualmachine.CreationTimestamp.Time)
```
Omrigan, line 434: I am slightly concerned about is that different `CreationTimestamp's can be recorded by different nodes in the cluster, thus this value can be inconsistent. I hope it is not noticeable on our scale.

Omrigan (approved):
LGTM. The only thing.

neonvm/controllers/virtualmachine_controller.go
shayanh: That's a valid concern. These values are created by k8s API server and we should have clock sync on them (infra team should know the exact details here).

neonvm/controllers/metrics.go
shayanh: Opened and merged #792

sharnoff (approved):

PR #776, opened 2024-01-31
neonvm: Tune QEMU disk cache settings

### LINKED ISSUE #775: Tune QEMU disk caching

Currently, when we mount disks to QEMU they have `cache=none`. This means that various kinds of filesystem -> disk flushes from the guest are converted to direct IO on the host for safety guarantees, but those guarantees are just generating extra work — we don't persist disks across restarts.

If we use `cache=unsafe`, we skip those extra flush operations, but we _also_ start relying on the host's page cache, which leaves us open to significant degradation if/when the host's memory fills up with page cache usage.

So _probably_ we'd want a mix of the two, which may be difficult.

## Prior work

Previously, we set `cache=unsafe`. We then had to revert it because it was was using up a lot of cache in the host, which eventually caused hard-to-attribute catastrophic results with the host stuck doing IO trying to evict things from the page cache.

Original PR enabling it was #628, issues were discussed in [this thread](https://neondb.slack.com/archives/C03F5SM1N02/p1702571133926869?thread_ts=1702558673.445669), revert PR was #694, coordinated in [this thread](https://neondb.slack.com/archives/C03TN5G758R/p1702585801270379?thread_ts=1702577770.752879).

At the time, we were using k8s nodes backed by EBS, which had a very small IOPS limit. We're now using real SSDs, so IOPS saturation should also be less of a concern.

## Feature idea(s) / DoD

Copy-pasted from [this slack comment](https://neondb.slack.com/archives/C03TN5G758R/p1706638929242789?thread_ts=1706637266.617099&cid=C03TN5G758R):

> * qemu should emulate all disk writes synchronously using O_DIRECT to the FS that hosts the qcow2.
> * It should implement artificial IOPS limit.
> * For performance improvements, it should drop the disk flush operations from the guest.

## Implementation ideas

See [this slack comment](https://neondb.slack.com/archives/C03TN5G758R/p1706666403509689?thread_ts=1706637266.617099&cid=C03TN5G758R):

> I think this is _maybe_ possible with the following block device settings:
> * `cache.writeback = off` (forces writes to be synchronous)
> * `cache.direct = on` (forces using direct IO)
> * `cache.no-flush = on` (ignores disk flush operations)
I think we can specify these directly in the `-drive` options, but if not, we can switch to `-blockdev` + `-device` (which `-drive` is just an abbreviation for).

sharnoff:
Resolves #775, refer there for more information.

Figured out how to implement the thing while writing the issue, figured I might as well open this, if only for discussion.

### DIFF
--- a/neonvm/config/default-vxlan/manager_config_patch.yaml
+++ b/neonvm/config/default-vxlan/manager_config_patch.yaml
@@ -24,9 +24,9 @@ spec:
         # FIXME: Rebase & move this to the appropriate place after merging #779.
         #
         # See #775 and its links.
-        # * cache.writeback=on  - don't set O_DSYNC (don't flush every write)
-        # * cache.direct=on     - forces using direct IO (don't abuse host's page cache!)
-        # * cache.no-flush=on   - ignores disk flush operations (not needed; our disks are ephemeral)
+        # * cache.writeback=on - don't set O_DSYNC (don't flush every write)
+        # * cache.direct=on    - forces using direct IO (don't abuse host's page cache!)
+        # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
         env:
         - name: NAD_IPAM_NAME
--- a/neonvm/config/default-vxlan/manager_config_patch.yaml
+++ b/neonvm/config/default-vxlan/manager_config_patch.yaml
@@ -24,10 +24,10 @@ spec:
         # FIXME: Rebase & move this to the appropriate place after merging #779.
         #
         # See #775 and its links.
-        # * cache.writeback=off - forces writes to be synchronous (avoid hidden caching)
+        # * cache.writeback=on  - don't set O_DSYNC (don't flush every write)
         # * cache.direct=on     - forces using direct IO (don't abuse host's page cache!)
         # * cache.no-flush=on   - ignores disk flush operations (not needed; our disks are ephemeral)
-        - "--qemu-disk-cache-settings=cache.writeback=off,cache.direct=on,cache.no-flush=on"
+        - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
         env:
         - name: NAD_IPAM_NAME
           value: $(NAD_IPAM_NAME)
--- a/neonvm/config/default-vxlan/manager_config_patch.yaml
+++ b/neonvm/config/default-vxlan/manager_config_patch.yaml
@@ -25,7 +25,7 @@ spec:
         #
         # See #775 and its links.
         # * cache.writeback=on - don't set O_DSYNC (don't flush every write)
-        # * cache.direct=on    - forces using direct IO (don't abuse host's page cache!)
+        # * cache.direct=on    - use O_DIRECT (don't abuse host's page cache!)
         # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
         env:
--- a/neonvm/config/default-vxlan/manager_config_patch.yaml
+++ b/neonvm/config/default-vxlan/manager_config_patch.yaml
@@ -21,6 +21,13 @@ spec:
         - "--zap-time-encoding=iso8601"
         - "--zap-log-level=info"
         - "--zap-stacktrace-level=panic"
+        # FIXME: Rebase & move this to the appropriate place after merging #779.
+        #
+        # See #775 and its links.
+        # * cache.writeback=off - forces writes to be synchronous (avoid hidden caching)
+        # * cache.direct=on     - forces using direct IO (don't abuse host's page cache!)
+        # * cache.no-flush=on   - ignores disk flush operations (not needed; our disks are ephemeral)
+        - "--qemu-disk-cache-settings=cache.writeback=off,cache.direct=on,cache.no-flush=on"
         env:
         - name: NAD_IPAM_NAME
           value: $(NAD_IPAM_NAME)
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -61,6 +61,12 @@ const (
 	sshAuthorizedKeysDiskPath   = "/vm/images/ssh-authorized-keys.iso"
 	sshAuthorizedKeysMountPoint = "/vm/ssh"
 
+	// See #775 and its links.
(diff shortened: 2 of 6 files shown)


neonvm/runner/main.go
```diff
@@ -61,6 +61,12 @@ const (
 	sshAuthorizedKeysDiskPath   = "/vm/images/ssh-authorized-keys.iso"
 	sshAuthorizedKeysMountPoint = "/vm/ssh"
 
+	// See #775 and its links.
+	// * cache.writeback=off - forces writes to be synchronous (avoid hidden caching)
```
shayanh, line 65: Having `cache.writeback=off` means qemu makes sure to flush the data to disk on write requests. My understanding is this is like calling fsync on every write, which would have a major performance impact. Since we don't care about persisting data on host we can go with `cache.writeback=on`.

Source: https://www.qemu.org/docs/master/system/qemu-manpage.html

shayanh (commented):
Thanks for this! I didn't know it's possible to have finer-grain control over qemu cache settings. I left a comment about writeback setting.

I suggest doing some benchmarks and investigate the impact of this before going to prod.

neonvm/runner/main.go
sharnoff: Ah right, good catch! Honestly I'm not sure whether `cache.writeback=off` should have any effect if `cache.direct` is used... Because `cache.direct=on` forces all writes to use direct IO (rather than via the page cache) and `cache.writeback=on` seems to say that it's more about returning early and doing flushes asynchronously.

Given that, I'm happy to set `cache.writeback=on` — on the off-chance that it does have _some_ effect even if `cache.direct=on`, I'd rather go with the one that is more efficient by default.

I'll also update this PR so that we can configure QEMU cache settings with an arg to `neonvm-controller` that propagates to `neonvm-runner` — that way, we can easily test this out on staging / pre-prod without needing to recompile each time.

shayanh: > Ah right, good catch! Honestly I'm not sure whether cache.writeback=off should have any effect if cache.direct is used... Because cache.direct=on forces all writes (and reads I geuss?) to use direct IO (rather than via the page cache) and cache.writeback=on seems to say that it's more about returning early and doing flushes asynchronously.

I think having `cache.direct=on` doesn't mean all writes would go to the disk, but it means we only use guest's memory and avoid host's page cache. Here is the definition from the QEMU manpage:

> The host page cache can be avoided with `cache.direct=on`. This will attempt to do disk IO directly to the guest’s memory. QEMU may still perform an internal copy of the data.

---
> I'll also update this PR so that we can configure QEMU cache settings with an arg to neonvm-controller that propagates to neonvm-runner — that way, we can easily test this out on staging / pre-prod without needing to recompile each time.

Sounds good.

sharnoff: Ah ok, so what I can find is:

- `cache.direct=on` means setting `O_DIRECT` when opening the backing file
- `cache.writeback=off` means setting `O_DSYNC` when opening the backing file (i.e. every write flushes, like you mentioned)

Some of this info is collected from wiki pages of projects built on QEMU, but I _think_ these are the semantics.

So, given that, I'd say yeah we definitely should have `cache.writeback=off`

shayanh: When the disks are ephemeral, why do we want to call fsync on every write? It makes every write much slower.

sharnoff: Oh sorry, typo. Meant to say `writeback=on` :sweat_smile:

neonvm/config/default-vxlan/manager_config_patch.yaml
```diff
@@ -21,6 +21,13 @@ spec:
         - "--zap-time-encoding=iso8601"
         - "--zap-log-level=info"
         - "--zap-stacktrace-level=panic"
+        # FIXME: Rebase & move this to the appropriate place after merging #779.
+        #
+        # See #775 and its links.
+        # * cache.writeback=off - forces writes to be synchronous (avoid hidden caching)
+        # * cache.direct=on     - forces using direct IO (don't abuse host's page cache!)
+        # * cache.no-flush=on   - ignores disk flush operations (not needed; our disks are ephemeral)
+        - "--qemu-disk-cache-settings=cache.writeback=off,cache.direct=on,cache.no-flush=on"
```
problame, line 30: why are the CLIs below using different defaults than this here?

shayanh:
@sharnoff are we going to merge this?

sharnoff:
Thanks for the reminder. I think I need to fix the defaults to set writeback=on, but otherwise we should be good to merge. _Probably_ want to rollout to production with cache=none still, but then region-by-region enable the fine-tuning.

neonvm/config/default-vxlan/manager_config_patch.yaml
sharnoff: late reply, but: not quite sure what you mean

shayanh (approved):

Omrigan (approved):
LGTM, some minor things.

neonvm/config/default-vxlan/manager_config_patch.yaml
sharnoff: Ah I see yeah, right now they have that default because (a) it's simpler, and (b) IMO it's a safer fallback than this untested concoction of flags :sweat_smile:

neonvm/runner/main.go
```diff
@@ -488,11 +488,13 @@ func main() {
 	var kernelPath string
 	var appendKernelCmdline string
 	var skipCgroupManagement bool
+	var diskCacheSettings string
 	flag.StringVar(&vmSpecDump, "vmspec", vmSpecDump, "Base64 encoded VirtualMachine json specification")
 	flag.StringVar(&vmStatusDump, "vmstatus", vmStatusDump, "Base64 encoded VirtualMachine json status")
 	flag.StringVar(&kernelPath, "kernelpath", defaultKernelPath, "Override path for kernel to use")
 	flag.StringVar(&appendKernelCmdline, "appendKernelCmdline", "", "Additional kernel command line arguments")
 	flag.BoolVar(&skipCgroupManagement, "skip-cgroup-management", false, "Don't try to manage CPU (use if running alongside container-mgr)")
+	flag.StringVar(&diskCacheSettings, "qemu-disk-cache-settings", "cache=none", "Cache settings to add to -drive args for VM disks")
```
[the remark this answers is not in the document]
sharnoff: Mostly the idea was:

1. I want this to be easy to roll out region-by-region (otherwise would need to change CRD defaults, or cplane)
2. I don't think it _needs_ to be different for each VM; otherwise old VMs restarting would keep old cache settings, which imo is probably counterproductive

But yeah, storing this in the spec _would_ mean we'd only need to change the CRD and runner.

WDYT?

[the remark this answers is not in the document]
Omrigan: Makes sense.

PR #773, opened 2024-01-30
neonvm-controller: Make max concurrency configurable via CLI

sharnoff:
We noticed a couple prod regions are close to all capacity used w.r.t. the existing concurrency limits.

This PR intends to provide a low-risk way to allow tweaking the precise limit more quickly.

ref https://neondb.slack.com/archives/C03TN5G758R/p1706641824349959?thread_ts=1706160071.213319
ref https://neondb.slack.com/archives/C03TN5G758R/p1706646482943109

PR #779, opened 2024-01-31
neonvm-controller: Standardize / clean up kustomize configuration

sharnoff:
While working on #773, I hit some behavior that was different from what I expected.

This PR makes some of the kustomize stuff around the neonvm-controller cleaner (IMO).

The commits are logically separate, and may be easier to review as such. I'll detail them in order below.

---

5349f6acea0ded22b315a2d84575987c519637a0

> mv neonvm/config/common/controller/{manager => deployment}.yaml
>
> This has no effect on the rendered manifests.

---

83751247a30bd8e9627abd6a8752da0c98928209

> neonvm: Move manager_config_patch.yaml into manager deployment
>
> This has no effect on the rendered manifests (but it would have made it nicer to implement #773).

---

c1cc0f1c97f1d5cfbd3ea448183c32c6639c5aee

> neonvm: Remove controller metrics auth proxy
>
> This wasn't being used (not scraping metrics), and then when we started scraping metrics, we just exposed the port directly.
>
> Maybe in the future we'll want to make these metrics more restrictive, but for now there's bigger fish to fry.

Diff to `neonvm.yaml` from this commit:

```diff
3729,3744d3728
<       affinity:
<         nodeAffinity:
<           requiredDuringSchedulingIgnoredDuringExecution:
<             nodeSelectorTerms:
<             - matchExpressions:
<               - key: kubernetes.io/arch
<                 operator: In
<                 values:
<                 - amd64
<                 - arm64
<                 - ppc64le
<                 - s390x
<               - key: kubernetes.io/os
<                 operator: In
<                 values:
<                 - linux
3803,3825d3786
<       - args:
<         - --secure-listen-address=0.0.0.0:8443
<         - --upstream=http://127.0.0.1:8080/
<         - --logtostderr=true
<         - --v=0
<         image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
<         name: kube-rbac-proxy
<         ports:
<         - containerPort: 8443
<           name: https
<           protocol: TCP
<         resources:
<           limits:
<             cpu: 500m
<             memory: 128Mi
<           requests:
<             cpu: 5m
<             memory: 64Mi
<         securityContext:
<           allowPrivilegeEscalation: false
<           capabilities:
<             drop:
<             - ALL
```

---

We _may_ want to consider getting rid of `config/default` and only use `config/default-vxlan`. AFAICT it isn't used anywhere and hasn't been for some time.


shayanh (approved):
These manifests are all generated by kubebuilder and I totally agree that they are confusing. I spent a day figuring them out the first time I had to work on neonvm.

shayanh:
> We may want to consider getting rid of config/default and only use config/default-vxlan. AFAICT it isn't used anywhere and hasn't been for some time.

+1

sharnoff:
Opened #812 - looking back at this now, I think _maybe_ it'd make sense to combine the two (especially looking at 33496328e9718d3c7ea5a99ad5ab31aabbd34705 here vs 1521e908c468b6a1966e1bc39ef827b228b410e1 there).

WDYT @shayanh?

shayanh:
@sharnoff sounds good to me.

PR #781, opened 2024-01-31
plugin: Make auto- live migration opt-in

### LINKED ISSUE #756: scheduler plugin triggered migration should be opt-in

When we enable automatic live migration in prod, we'll require enabling the overlay IP addresses on all existing VMs. This is not currently enabled, so enabling migration will either (a) have to wait until all VMs use the overlay IP addresses, or (b) be enabled gradually, on a case-by-case basis.

IMO the second option is better, because it allows a gradual rollout.

Making scheduler-triggered live migration opt-in also allows us to enable it in the config (or make it always-enabled; see also #134), which would also have more of the code always in use.

## DoD

Scheduler-triggered live migration should be enabled individually for each VM, with the default that it's not enabled.

## Implementation ideas

Probably a new `autoscaling.neon.tech/...` annotation?

sharnoff:
Resolves #756. Copying from that issue:

> When we enable automatic live migration in prod, we'll require enabling the overlay IP addresses on all existing VMs. This is not currently enabled, so enabling migration will either (a) have to wait until all VMs use the overlay IP addresses, or (b) be enabled
> gradually, on a case-by-case basis.

Now that it's opt-in, we can *also* set the global config setting to default to true. Still worth keeping for now, at the very least as a "break glass in case of emergency" button.

---

**Notes for review:** This is a lot of commits. Every commit except the last one is just laying the groundwork for this PR, so I figured I'd include it here.

Happy to split those out into a separate PR if that'd be easier.

I intend to merge this PR via rebase, so the commits history is preserved (and so blame becomes easier).

TBH not sure the easiest way to review this. I'd _probably_ recommend going commit-by-commit.

### DIFF
--- a/deploy/scheduler/config_map.yaml
+++ b/deploy/scheduler/config_map.yaml
@@ -39,6 +39,6 @@ data:
         "timeoutSeconds": 5
       },
       "migrationDeletionRetrySeconds": 5,
-      "doMigration": false,
+      "doMigration": true,
       "randomizeScores": true
     }
--- a/pkg/api/vminfo.go
+++ b/pkg/api/vminfo.go
@@ -19,6 +19,7 @@ import (
 )
 
 const (
+	LabelEnableAutoMigration      = "autoscaling.neon.tech/auto-migration-enabled"
 	LabelTestingOnlyAlwaysMigrate = "autoscaling.neon.tech/testing-only-always-migrate"
 	LabelEnableAutoscaling        = "autoscaling.neon.tech/enabled"
 	AnnotationAutoscalingBounds   = "autoscaling.neon.tech/bounds"
@@ -35,6 +36,12 @@ func hasTrueLabel(obj metav1.ObjectMetaAccessor, labelName string) bool {
 // HasAutoscalingEnabled returns true iff the object has the label that enables autoscaling
 func HasAutoscalingEnabled(obj metav1.ObjectMetaAccessor) bool {
 	return hasTrueLabel(obj, LabelEnableAutoscaling)
+}
+
+// HasAutoMigrationEnabled returns true iff the object has the label that enables "automatic"
+// scheduler-triggered migration, and it's set to "true"
+func HasAutoMigrationEnabled(obj metav1.ObjectMetaAccessor) bool {
+	return hasTrueLabel(obj, LabelEnableAutoMigration)
 }
 
 func HasAlwaysMigrateLabel(obj metav1.ObjectMetaAccessor) bool {
@@ -110,9 +117,13 @@ func NewVmMemInfo(memSlots vmapi.MemorySlots, memSlotSize resource.Quantity) (*V
 //
 // This is separate from the bounds information stored in VmInfo (even though that's also derived
 // from annotations), because VmConfig is meant to store values that either qualitatively change the
-// handling for a VM (e.g., AlwaysMigrate) or are expected to largely be the same for most VMs
+// handling for a VM (e.g., AutoMigrationEnabled) or are expected to largely be the same for most VMs
 // (e.g., ScalingConfig).
 type VmConfig struct {
+	// AutoMigrationEnabled indicates to the scheduler plugin that it's allowed to trigger migration
+	// for this VM. This defaults to false because otherwise we might disrupt VMs that don't have
+	// adequate networking support to preserve connections across live migration.
+	AutoMigrationEnabled bool `json:"autoMigrationEnabled"`
 	// AlwaysMigrate is a test-only debugging flag that, if present in the VM's labels, will always
 	// prompt it to migrate, regardless of whether the VM actually *needs* to.
 	AlwaysMigrate  bool           `json:"alwaysMigrate"`
@@ -189,6 +200,7 @@ func extractVmInfoGeneric(
 		return nil, fmt.Errorf("Error extracting memory info: %w", err)
 	}
 
+	autoMigrationEnabled := HasAutoMigrationEnabled(obj)
 	scalingEnabled := HasAutoscalingEnabled(obj)
 	alwaysMigrate := HasAlwaysMigrateLabel(obj)
 
@@ -198,9 +210,10 @@ func extractVmInfoGeneric(
 		Cpu:       *cpuInfo,
 		Mem:       *memInfo,
 		Config: VmConfig{
-			AlwaysMigrate:  alwaysMigrate,
-			ScalingEnabled: scalingEnabled,
-			ScalingConfig:  nil, // set below, maybe
+			AutoMigrationEnabled: autoMigrationEnabled,
+			AlwaysMigrate:        alwaysMigrate,
+			ScalingEnabled:       scalingEnabled,
+			ScalingConfig:        nil, // set below, maybe
 		},
 	}
 
@@ -406,13 +419,13 @@ func (cfg VmConfig) Format(state fmt.State, verb rune) {
 	switch {
 	case verb == 'v' && state.Flag('#'):
 		state.Write([]byte(fmt.Sprintf(
-			"api.VmConfig{AlwaysMigrate:%t, ScalingEnabled:%t, ScalingConfig:%#v}",
-			cfg.AlwaysMigrate, cfg.ScalingEnabled, cfg.ScalingConfig,
+			"api.VmConfig{AutoMigrationEnabled:%t, AlwaysMigrate:%t, ScalingEnabled:%t, ScalingConfig:%#v}",
+			cfg.AutoMigrationEnabled, cfg.AlwaysMigrate, cfg.ScalingEnabled, cfg.ScalingConfig,
 		)))
 	default:
 		state.Write([]byte(fmt.Sprintf(
-			"{AlwaysMigrate:%t ScalingEnabled:%t ScalingConfig:%+v}",
-			cfg.AlwaysMigrate, cfg.ScalingEnabled, cfg.ScalingConfig,
+			"{AutoMigrationEnabled:%t AlwaysMigrate:%t ScalingEnabled:%t ScalingConfig:%+v}",
+			cfg.AutoMigrationEnabled, cfg.AlwaysMigrate, cfg.ScalingEnabled, cfg.ScalingConfig,
 		)))
 	}
 }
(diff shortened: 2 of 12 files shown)


pkg/plugin/queue.go
```diff
@@ -30,6 +30,7 @@ func (mq migrationQueue) isNextInQueue(vm *vmPodState) bool {
 func (mq *migrationQueue) removeIfPresent(vm *vmPodState) {
 	if vm.MqIndex != -1 {
 		_ = heap.Remove(mq, vm.MqIndex)
+		vm.MqIndex = -1
```
shayanh, line 33: What does this mean?

pkg/plugin/run.go
```diff
@@ -359,11 +357,13 @@ func (e *AutoscaleEnforcer) updateMetricsAndCheckMustMigrate(
 	node *nodeState,
 	metrics *api.Metrics,
 ) bool {
-	// This pod should migrate if (a) we're looking for migrations and (b) it's next up in the
-	// priority queue. We will give it a chance later to veto if the metrics have changed too much
+	// This pod should migrate if (a) it's allowed to migrate, (b) we're looking for migrations and
```
shayanh, line 360: What does (b) mean?

pkg/api/vminfo.go
```diff
@@ -69,6 +76,10 @@ type VmMemInfo struct {
 }
 
 type VmConfig struct {
+	// AutoMigrationEnabled indicates to the scheduler plugin that it's allowed to trigger migration
+	// for this VM. This defaults to false because otherwise we might disrupt VMs that don't have
+	// adequate networking support to preserve connections across live migration.
+	AutoMigrationEnabled bool `json:"autoMigrationEnabled"`
```
shayanh, line 82: Where do we set the default value for `AutoMigrationEnabled`?

shayanh (commented):
Nicely done! A few comments and questions.

pkg/api/vminfo.go
sharnoff: Default is defined by how this value is set:

https://github.com/neondatabase/autoscaling/blob/aced6c486f370d989514e7b3a6e9ddd15543ee39/pkg/api/vminfo.go#L172

That calls `hasTrueLabel` which returns true iff the label is present and equals `"true"`.

pkg/plugin/run.go
sharnoff: I clarified the language. It now reads:

> This pod should migrate if (a) it's allowed to migrate, (b) node resource usage is high enough that we should migrate \*something\*, and (c) it's next up in the priority queue. We will give it a chance later to veto if the metrics have changed too much.

LMK if that makes more sense.

pkg/api/vminfo.go
```diff
@@ -68,6 +68,12 @@ type VmMemInfo struct {
 	SlotSize Bytes `json:"slotSize"`
 }
 
+type VmConfig struct {
```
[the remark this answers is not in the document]
shayanh: Also I think it would be good to clarify info vs config in the comments. Description of `VmInfo` and `VmConfig` seem quite similar right now.

pkg/api/vminfo.go
```diff
@@ -351,3 +362,19 @@ func (mem VmMemInfo) Format(state fmt.State, verb rune) {
 		state.Write([]byte(fmt.Sprintf("{Min:%d Max:%d Use:%d SlotSize:%v}", mem.Min, mem.Max, mem.Use, mem.SlotSize)))
 	}
 }
+
+func (cfg VmConfig) Format(state fmt.State, verb rune) {
+	// same-ish style as for VmInfo, differing slightly from default repr.
```
Omrigan, line 395: How is it different? I see comment above talking about `resource.Quantity`. But if `VmConfig` doesn't have any `resource.Quantity`, maybe we can have the default formatting here?

pkg/api/vminfo.go
```diff
@@ -82,6 +82,14 @@ type VmConfig struct {
 	ScalingConfig  *ScalingConfig `json:"scalingConfig,omitempty"`
 }
 
+// Equals returns true if the two VmConfig structs are deep-equal to each other.
+func (c VmConfig) Equals(other VmConfig) bool {
```
Omrigan, line 97: Maybe `DeepEqual` [1] or serialize to json and compare jsons? There comparators aren't very pretty + easy to forget when adding new fields.


[1]: https://pkg.go.dev/reflect#DeepEqual

deploy/scheduler/config_map.yaml
```diff
@@ -38,6 +38,6 @@ data:
         "timeoutSeconds": 5
       },
       "migrationDeletionRetrySeconds": 5,
-      "doMigration": false,
+      "doMigration": true,
```
Omrigan, line 41: > Now that it's opt-in, we can *also* set the global config setting to
default to true. Still worth keeping for now, at the very least as a
"break glass in case of emergency" button.


Does the commit message relates to this change?

Omrigan (approved):
Looks good! A couple of questions and suggestions.

deploy/scheduler/config_map.yaml
sharnoff: Yep, that's what that refers to

pkg/api/vminfo.go
sharnoff: I mostly wanted to avoid that if possible, because we fetch and compare the `VmInfo` every time we receive an Update event, and I thought it'd be too inefficient to use something like that.

Maybe `reflect.DeepEqual` is the right call though... I'll pull some stats from production to see what the actual event rate is.

pkg/api/vminfo.go
```diff
@@ -75,6 +75,8 @@ type VmMemInfo struct {
 // values that either qualitatively change the handling for a VM *or* ar eexpected to largely be the
 // same for most VMs (e.g., ScalingConfig).
 type VmConfig struct {
+	// AlwaysMigrate is a test-only debugging flag that, if present in the VM's labels, will always
```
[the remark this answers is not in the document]
sharnoff: I _think_ I added this here, copying over from `pkg/plugin/state.go`:

https://github.com/neondatabase/autoscaling/pull/781/files/357d01495ad989af7799be8dbae167e5f5d1c4cb..67b2ad1edd6473e4d75487280f7620fb6a8779e8#diff-f2f1d4fbaf99622ddc26f2c717dd4b99b478b7d1e191037323ebfe12b44a51a7L190-L192

pkg/api/vminfo.go
sharnoff: The main difference from the default representation is that for `%v` we're still including the field names. I don't know if it's honestly necessary though tbh

deploy/scheduler/config_map.yaml
Omrigan: I'd rewrite can -> do and mention a name of a field to be explicit. Like this:

> Also set the global doMigration to default to true. 
> Still worth keeping for now, at the very least as a
> "break glass in case of emergency" button.

pkg/api/vminfo.go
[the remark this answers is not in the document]
Omrigan: Just surprising to see a field itself and a comment to be added in separate commits.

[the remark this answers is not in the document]
sharnoff: Yeah, that's fair.

(also, now that you've pointed it out, I realized there was a typo in the original comment :sweat_smile:)

deploy/scheduler/config_map.yaml
sharnoff: done

pkg/api/vminfo.go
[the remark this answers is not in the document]
sharnoff: Fair, makes sense. Reworded again:

https://github.com/neondatabase/autoscaling/blob/06b0c1261e53e1765ecc0414e45c6f280df1d7f4/pkg/api/vminfo.go#L115-L121

WDYT?

pkg/api/vminfo.go
```diff
@@ -68,6 +75,32 @@ type VmMemInfo struct {
 	SlotSize Bytes `json:"slotSize"`
 }
 
+// VmConfig stores the autoscaling-specific "extra" configuration derived from labels and
+// annotations on the VM object.
+//
+// While annotations may set the autoscaling bounds for a VM as well, VmConfig is meant more to store
+// values that either qualitatively change the handling for a VM *or* ar eexpected to largely be the
```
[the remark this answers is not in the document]
sharnoff: Fixed in rewording

pkg/api/vminfo.go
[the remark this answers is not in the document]
sharnoff: Fixed the typo; resolving.

pkg/api/vminfo.go
sharnoff: Stats from prod: Seems like us-west-2 peaks the highest, at about 3k pod add/update events per minute. I don't know what the second-to-second events load is like, but presumably it's similarly spiky to what we see from neonvm-controller (albeit with some flattening because a spike of VM starts has many subsequent pod events).

All that said, I think we can get away with `reflect.DeepEqual`.

sharnoff: Switched to `reflect.DeepEqual`, resolving

sharnoff:
CI failed 4 times in a row, timing out at deploying neonvm, even though this was the diff: https://github.com/neondatabase/autoscaling/compare/6837e7a77fad29ffc59ec08f805fe66a6566fa58..8da420f4fdd3874f414f4e9f1d2a025ed9c9e37a

Going to try reverting to see if it still fails (or if it's somehow that change)

sharnoff:
yep, seems unrelated

pkg/api/vminfo.go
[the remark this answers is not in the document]
Omrigan: Sounds good :+1:

Omrigan (approved):

Omrigan:
> yep, seems unrelated

@sharnoff and no idea why it is flaky?

sharnoff:
Yeah, no clue why it was failing. It was being consistently assigned the same runner, so I'd guess something up with that.

But I can see you retried it & it works now, so :shrug: :sweat_smile:

Issue #747, opened 2024-01-17
Merge auxillary containers to reduce pressure on container runtime


Omrigan:
https://github.com/neondatabase/autoscaling/issues/711#issuecomment-1893637974 suggests there is no system in how long every container takes to be created/started. 

This can be explained by the fact that `containerd` has to create and run each container. In the peaks it might be overwhelmed, thus sporadic queues might occur.  

The idea is to reduce the number of containers. In the best case we need only one - main container with QEMU.

sharnoff:
A couple notes:

1. IIRC there's three reasons we require init containers:
    1. To make image loading easier (see `init-rootdisk`, or [when using a custom kernel] `init-kernel`)
    2. To perform privileged operations without giving `neonvm-runner` those privileges (see `neonvm-runner`)
    3. For setting up iptables rules externally (cplane injects an extra init container to do this, because (a) it's not natively supported by neonvm, and (b) using an init container means the process is the same for k8s-pod computes)
2. Given this, there's several reasons in practice we may need to keep _at least_ one init container. But I'm generally +1 to reducing the number of init containers where we're able to.

Omrigan:
As the result of discussions with @sharnoff and @ololobus I intend to do the following changes:
1. Merge `init-rootdisk` and `sysctl` containers (#769)
2. Move iptables configuration logic from `compute-init` to `neonvm-runner`, thus dropping `compute-init`
3. Get rid of the `init` container by:
  a. Merging compute and `runner` images, so there is no need to copy disk
  b. Configuring parent network namespace, so that sysctl is not needed

sharnoff:
status: Two PRs are ready (#769 and #782), need review on those. Planning to open a follow-up to #782 to refactor neonvm-runner.

sharnoff:
status: #769 and #782 still pending, reviewed & need to reply. #790 opened and should help as well, WIP.
Previously blocked on deploy of startup metrics; they aren't yet on prod, but should be this week.

Omrigan:
#769 is in production since yesterday, and the graph suggests 1s reduction it timings (at least for the fastest ones): 

![screenshot-2024-02-23_02-19-30_504387060](https://github.com/neondatabase/autoscaling/assets/2663743/0a3b57d4-ac71-4e45-9cc3-76123755a98c)

#782 is merged, but the enablement on cplane side is WIP: https://github.com/neondatabase/cloud/pull/10630

#790 is still WIP.

PR #782, opened 2024-02-01
WIP: add InitScript to spec

Omrigan:
Part of #747

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -480,6 +480,39 @@ func checkDevTun() bool {
 	return mode&os.ModeCharDevice == os.ModeCharDevice
 }
 
+func runInitScript(script []byte, logger *zap.Logger) error {
+	if len(script) == 0 {
+		return nil
+	}
+
+	// creates a tmp file with the script content
+	tmpFile, err := os.CreateTemp(os.TempDir(), "init-script-")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpFile.Name()) // clean up
+
+	if _, err := tmpFile.Write(script); err != nil {
+		return err
+	}
+
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+
+	if err := os.Chmod(tmpFile.Name(), 0755); err != nil {
+		return err
+	}
+
+	logger.Info("running init script", zap.String("path", tmpFile.Name()))
+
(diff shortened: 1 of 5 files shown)


neonvm/runner/main.go
```diff
@@ -480,6 +480,39 @@ func checkDevTun() bool {
 	return mode&os.ModeCharDevice == os.ModeCharDevice
 }
 
+func runInitScript(script string, logger *zap.Logger) error {
```
sharnoff, line 483:
```suggestion
func runInitScript(logger *zap.Logger, script string) error {
```

style nit: Generally logger, context, things like that are at the beginning of the args

neonvm/runner/main.go
```diff
@@ -480,6 +480,39 @@ func checkDevTun() bool {
...
+		return err
+	}
+
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+
+	if err := os.Chmod(tmpFile.Name(), 0755); err != nil {
+		return err
+	}
```
sharnoff, line 505: out of curiosity, why the chmod here?

sharnoff (commented):
Looks good - just a couple small things.

neonvm/runner/main.go
Omrigan: That's `chmod +x`, to run the script

neonvm/runner/main.go
Omrigan: Fixed. 

In the Go teams I worked before it was usually:
 - context is the first argument
 - logger is the last argument

But I am happy with whatever as long as it is consistent.

neonvm/runner/main.go
sharnoff: Hm, I thought we didn't need that if we're not directly executing it (instead we're `/bin/sh <script name>`)

neonvm/runner/main.go
sharnoff: huh, interesting! yeah, in this codebase we have this pattern I guess :sweat_smile:

neonvm/runner/main.go
Omrigan: Ok, I guess I did it out of habit.

Deleted it.

sharnoff (approved):

shayanh (approved):

Issue #13, opened 2023-01-11
Use "proper" web framework for HTTP servers


sharnoff:
See: https://github.com/neondatabase/autoscaling/pull/8#discussion_r1065054595

There's a few places we have HTTP servers:

* pkg/plugin/run.go: `(*AutoscaleEnforcer) runPermitHandler`
* ~~pkg/agent/serve.go: `startInformantServer`~~
* ~~cmd/informant/main.go: `main`~~
* pkg/plugin/dumpstate.go: `(*AutoscaleEnforcer) startDumpStateServer`
* pkg/agent/dumpstate.go: `(*agentState) StartDumpStateServer`

(time of writing: 2023-01-10, commit 65d1528f46682086345cb28920204436e7600be1)
(updated: 2023-10-12, commit 166c8d01fa12c1d8eac2a586403bd83648349d73)

Ideally we'd have something less home-grown here, to reduce custom code & leverage existing wisdom.

sharnoff:
This is now just the scheduler plugin, because the vm-informant was removed in #442, and the agent's HTTP server (which only existed to handle connections from the informant) was removed in #506.

stradig:
We are okay with the current state, the code is very readable.

PR #785, opened 2024-02-02
plugin: Handle panics from agent requests

### LINKED ISSUE #760: plugin panic handling (convert panics to handled errors)
Part of neondatabase/company_projects#187.

We have a handful of places in the scheduler plugin where panics are not adequately handled. They should be handled properly (e.g., converted to HTTP 500s) or converted to fallible errors.

sharnoff:
Closes #760.

AFAIK this hasn't been an issue in the past, but as we're trying to improve reliability, it's good to get this out of the way before it becomes an issue.

Note that this PR is quite minimal - expanding the existing tech debt we have around how the scheduler plugin handles HTTP requests. It's probably ok *enough* for now. I don't expect we'll be making too many changes to it in the near future. See also: #13.

Tested locally by forcing it to panic on every request:

```diff
diff --git a/pkg/plugin/run.go b/pkg/plugin/run.go index 007554a..6da7728 100644
--- a/pkg/plugin/run.go
+++ b/pkg/plugin/run.go
@@ -262,8 +262,10 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
 		}
 	}

-	pod.vm.mostRecentComputeUnit = &e.state.conf.ComputeUnit
-	return &resp, 200, nil
+	panic(errors.New("test panic!")) +
+	// pod.vm.mostRecentComputeUnit = &e.state.conf.ComputeUnit
+	// return &resp, 200, nil }

 // getComputeUnitForResponse tries to return compute unit that the agent supports
```

The change appears to work as intended.

### DIFF
--- a/pkg/plugin/run.go
+++ b/pkg/plugin/run.go
@@ -34,9 +34,22 @@ const (
 func (e *AutoscaleEnforcer) startPermitHandler(ctx context.Context, logger *zap.Logger) error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		logger := logger // copy locally, so that we can add fields and refer to it in defers
+
 		var finalStatus int
 		defer func() {
 			e.metrics.resourceRequests.WithLabelValues(r.RemoteAddr, strconv.Itoa(finalStatus)).Inc()
+		}()
+
+		// Catch any potential panics and report them as 500s
+		defer func() {
+			if err := recover(); err != nil {
+				msg := "request handler panicked"
+				logger.Error(msg, zap.String("error", fmt.Sprint(err)))
+				finalStatus = 500
+				w.WriteHeader(finalStatus)
+				_, _ = w.Write([]byte(msg))
+			}
 		}()
 
 		if r.Method != "POST" {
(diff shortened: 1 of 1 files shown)


pkg/plugin/run.go
```diff
@@ -34,11 +34,24 @@ const (
...
 		}()
 
+		// Catch any potential panics and report them as 500s
+		defer func() {
+			if err := recover(); err != nil {
+				msg := "request handler panicked"
+				logger.Error(msg, zap.String("error", fmt.Sprint(err)))
+				finalStatus = 500
+				w.WriteHeader(finalStatus)
+				_, _ = w.Write([]byte(msg))
```
Omrigan, line 51: how about sending not just `msg` here, but  `msg: err`?

Omrigan (approved):

pkg/plugin/run.go
sharnoff: Hm, yeah. My impression was that it was better from a security perspective to not include the full error message, but tbh maybe it doesn't matter.
wdyt?

Omrigan: Hmm, I see your concern. Although this is our internal API, so risk is not high.   

But I guess, it can be left as is, it is easy to grep for logs. Maybe make the message more distinct, like "permit handler panicked" or else.

@areyou1or0 do you have an opinion on whether we should return the full error message, from catch-all exception handler?

areyou1or0: @Omrigan somehow I didn't get the notification for the mention, just seeing this now. 
Yes, it's best practice to not reveal non-customised error message as this can disclose the details of underlying technology. Low severity but valid.

PR #792, opened 2024-02-06
util: use prometheus.Registerer in RegisterMetrics

shayanh:
Using the interface allows to use other implementations as well.


sharnoff (approved):
lgtm. For posterity, this was previously mentioned here: https://github.com/neondatabase/autoscaling/pull/774#discussion_r1479089868

PR #793, opened 2024-02-07
neonvm-controller: Fix field name typo in webhook validation

sharnoff:
Noticed while reviewing #782.

Maybe we want to go through the list more thoroughly, I think there's a handful of fields missing.


shayanh (approved):
good catch!

Issue #794, opened 2024-02-07
Bug: NeonVM controller creates more than one pod for a VM


shayanh:
## Environment

Local

## Steps to reproduce

```console
$ kubectl apply -f vm-deploy.yaml
virtualmachine.vm.neon.tech/postgres14-disk-test created

$ kubectl get pods
NAME                                 READY   STATUS    RESTARTS   AGE
postgres14-disk-test-g5qvs           2/2     Running   0          2m26s

$ kubectl delete pod postgres14-disk-test-g5qvs
pod "postgres14-disk-test-g5qvs" deleted

$ kubectl get pods
NAME                                 READY   STATUS    RESTARTS   AGE
postgres14-disk-test-rgv6t           2/2     Running   0          28s
postgres14-disk-test-jcxsd           2/2     Running   0          28s
```

## Expected result

Seeing one pod after deleting the first one.

## Actual result

Two pods are running for the same VM.

## Why this happens?

Running two reconcile jobs for the same VM might lead to this situation. This is a classic race condition. This problem has always has been there but now it's visible since we have increased the number of reconcile workers.

shayanh:
I investigated this issue and discussed earlier today with the autoscaling team. Previously I was wrong about having two concurrent reconcile jobs. Controller runtime guarantees that there will be at most one reconcile job running for each object in a single controller ([more info](https://openkruise.io/blog/learning-concurrent-reconciling/)).

Here is what causes the problem:
1. Reconcile job 1: Observes the runner pod doesn't exist (since we killed it earlier). Reconciler logs `Restarting VM runner pod` and cleans up the VM object.
2. Reconcile job 2: Creates a new pod and updates VM's status with the new `PodName`.
3. Reconcile job 3: Reads stale data from k8s API server. Note that this stale object doesn't have a `.Status.PodName`. So reconciler creates a pod and tries to update the VM status object with the new `PodName`. Here we get to the point where we realized our retry logic for updating VM status is flawed. Here is what happens in the retry logic:
    - We get a conflict error and we update the `virtualmachine` object with the value we received from API server.
    - We update the VM status on API server with the latest value we got from API server, the recently updated `virtualmachine` object. This is almost equivalent to assigning a variable to itself and ignoring the changes we wanted to actually make.
https://github.com/neondatabase/autoscaling/blob/52a883bfa5cb20e95c2c8f0bc2e913326c6f36a6/neonvm/controllers/virtualmachine_controller.go#L175-L194

[...]

PR #796, opened 2024-02-08
neonvm-controller: Don't retry update on conflict

sharnoff:
The previous logic would:

1. Try to update the status
2. If we got a conflict, overwrite the current VirtualMachine object equal to the state on the API server
3. Then, retry with the udpated object (without changing anything!)

This is basically doing extra work for nothing!

We discussed potentially changing this to overwrite the status on conflict, but this could result in issues when operating on stale data (e.g., overwriting .status.podName back to "" after it was already created).

ref https://www.notion.so/neondatabase/Autoscaling-Team-Internal-Sync-179c7d597dbb4fe5b565d9c482d4d166
ref https://neondb.slack.com/archives/C03TN5G758R/p1707414998235459

### DIFF
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -168,32 +168,13 @@ func (r *VirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	// update status after reconcile loop (try 10 times)
-	statusNow := statusBefore.DeepCopy()
-	statusNew := virtualmachine.Status.DeepCopy()
-	try := 1
-	for try < 10 {
-		if !DeepEqual(statusNow, statusNew) {
-			// update VirtualMachine status
-			// log.Info("DEBUG", "StatusNow", statusNow, "StatusNew", statusNew, "attempt", try)
-			if err := r.Status().Update(ctx, &virtualmachine); err != nil {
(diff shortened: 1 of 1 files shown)


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -168,32 +168,13 @@ func (r *VirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-					// re-get statusNow from current state
-					statusNow = virtualmachine.Status.DeepCopy()
-					continue
-				}
-				log.Error(err, "Failed to update VirtualMachine status after reconcile loop",
-					"virtualmachine", virtualmachine.Name)
-				return ctrl.Result{}, err
-			}
+	// If the status changed, try to update the object
+	if !DeepEqual(statusBefore, virtualmachine.Status) {
```
sharnoff, line 172: _Maybe_ we want to always try to update the status just so that we get an error if there would have been a conflict. Not sure.

shayanh (approved):

neonvm/controllers/virtualmachine_controller.go
shayanh: Let's avoid doing that. It'll add some unnecessary load on API server.

shayanh:
More context on what is wrong with the old logic: https://github.com/neondatabase/autoscaling/issues/794#issuecomment-1934834774

PR #797, opened 2024-02-08
neonvm: Update VM's .Status.PodName immediately on API server

### LINKED ISSUE #794: Bug: NeonVM controller creates more than one pod for a VM

shayanh:
+ Added e2e test for runner pod

Fixes #794


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -321,6 +321,11 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 	// Generate runner pod name
 	if len(virtualmachine.Status.PodName) == 0 {
 		virtualmachine.Status.PodName = names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", virtualmachine.Name))
+		// Update the .Status.PodName on API Server to avoid creating multiple pods for a single VM
+		// See https://github.com/neondatabase/autoscaling/issues/794 for the context
+		if err := r.Status().Update(ctx, virtualmachine); err != nil {
+			return fmt.Errorf("Failed to update .Status.PodName: %w", err)
+		}
 	}
```
sharnoff, line 329: For later: When I realized the pod name was set _here_ (and not inside of the phase switch statement), I was quite surprised. We might want to move it into the handling for Pending or something.

Maybe not necessary until #733.

tests/e2e/runner-pod/01-assert.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 60
+commands:
+  - script: kubectl get pods -n "$NAMESPACE" --no-headers=true -l vm.neon.tech/name=example | [ $(wc -l) -eq 1 ]
```
sharnoff, line 5: This _may_ fail due to the PRESERVE_RUNNER_PODS env var in CI (which in turn sets the `nodelete` build tag) — basically, in order to keep VM runner pod logs available in CI, we don't delete the runner pod.

sharnoff (approved):
LGTM. If the test fails because we intentionally don't delete runner pods in CI, I'm happy to merge without the test.

tests/e2e/runner-pod/01-assert.yaml
shayanh: I see, but still there should be only one running pod, right? I can query only the pods that are running.

sharnoff: For now, yes. With container-mgr enabled, no: the original pod will remain "running" because even though the neonvm-runner container will be killed, the the container-mgr container will still be running.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -321,6 +321,11 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
 	// Generate runner pod name
 	if len(virtualmachine.Status.PodName) == 0 {
 		virtualmachine.Status.PodName = names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", virtualmachine.Name))
+		// Update the .Status.PodName on API Server to avoid creating multiple pods for a single VM
```
Omrigan, line 324: nit: it can also update `.Status.SSHSecretName`. So technically we update not the particular field, but the entire `virtualmachine` object. Maybe the wording here and in the error message can be improved.

tests/e2e/runner-pod/01-assert.yaml
Omrigan: I thought with crictl PR merged container-mgr will also be killed when runner dies, no?

tests/e2e/runner-pod/00-assert.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: kuttl.dev/v1beta1
```
Omrigan, line 1: Suggestion: maybe have a readme in test directory explaining what the test does? It is non-trivial how this tests .Status.PodName

Omrigan (approved):

tests/e2e/runner-pod/01-assert.yaml
sharnoff: Basically with crictl enabled, we would have to make a choice between deleting the pod (to kill container-mgr) or leaving it running. In CI, we want to keep the runner pods around to be able to fetch their logs, which means we can't delete the pod (and therefore can't kill container-mgr).

Omrigan: Would killing container-mgr (and not the whole pod) be a soultion. Thus pod is preserved when we have `nodelete`?

sharnoff: I checked this when working on the original PR and AFAICT there isn't a clean way to kill a container within a pod — but, I'd love to find a solution :sweat_smile:

shayanh: I made it work with deleting the entire runner pod (kubectl delete). Works with both crictl and `--skip-delete` enabled.

neonvm/controllers/virtualmachine_controller.go
shayanh: I agree, let's leave it once we get to #733.

tests/e2e/runner-pod/00-assert.yaml
shayanh: This test is broader than just testing `.Status.PodName`. We test our basic guarantees around runner pod.

tests/e2e/runner-pod/01-assert.yaml
shayanh: For deleting a specific container container in a pod, you can do this: `kubectl exec <PodName> --container <ContainerName> -- kill 1`.

tests/e2e/runner-pod/00-assert.yaml
Omrigan: Maybe add it as a separate commit then? For better blame.

But the suggestion still stands :laughing:

PR #798, opened 2024-02-08
neonvm-controller: Add reconcile metrics debug endpoint

sharnoff:
Similar in spirit to the "dump state" endpoints exposed by the autsocaler-agent and scheduler plugin.

This one is on port 7778 (+1 from the pprof server on 7777).

The state can be fetched with:

```sh
kubectl port-forward -n neonvm-system pod/<pod-name> 7778:7778 &
curl http://localhost:7778/
```

---

Part of the reason for this change is that the metrics reported by #757 indicate there's a VM in prod that has been failing to reconcile, even though the logs don't show per-VM error rates consistent with that. So hopefully this should allow us to debug issues like this in the future.

### DIFF
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -90,8 +90,10 @@ type ReconcileSnapshot struct {
 	Failing []string `json:"failing"`
 }
 
-// WithMetrics wraps a given Reconciler with metrics capabilities, also returning a function that
-// produces a snapshot of the metrics, for easier inspection.
+// WithMetrics wraps a given Reconciler with metrics capabilities.
+//
+// The returned reconciler also provides a way to get a snapshot of the state of ongoing reconciles,
+// to see the data backing the metrics.
 func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) ReconcilerWithMetrics {
 	return &wrappedReconciler{
 		Reconciler:     reconciler,
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -68,7 +68,7 @@ type wrappedReconciler struct {
 	failing map[client.ObjectKey]struct{}
 }
 
-// ReocncileSnapshot provides a glimpse into the current state of ongoing reconciles
+// ReconcileSnapshot provides a glimpse into the current state of ongoing reconciles
 //
 // This type is (transitively) returned by the controller's "dump state" HTTP endpoint, and exists
 // to allow us to get deeper information on the metrics - we can't expose information for every
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -72,13 +72,13 @@ type wrappedReconciler struct {
 //
 // This type is (transitively) returned by the controller's "dump state" HTTP endpoint, and exists
 // to allow us to get deeper information on the metrics - we can't expose information for every
-// VirtualMachine into the metrics (it'd be too high cardinality), but we *can* make it availalbe
+// VirtualMachine into the metrics (it'd be too high cardinality), but we *can* make it available
 // when requested.
 type ReconcileSnapshot struct {
 	// ControllerName is the name of the controller: virtualmachine or virtualmachinemigration.
 	ControllerName string
 
-	// Failing is the list of objects currenly failing to reconcile
+	// Failing is the list of objects currently failing to reconcile
 	Failing []string
 }
 
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -97,15 +97,13 @@ func WithMetrics(
 	rm ReconcilerMetrics,
 	cntrlName string,
 ) ReconcilerWithMetrics {
-	r := &wrappedReconciler{
+	return &wrappedReconciler{
 		Reconciler:     reconciler,
 		Metrics:        rm,
 		ControllerName: cntrlName,
 		lock:           sync.Mutex{},
 		failing:        make(map[client.ObjectKey]struct{}),
 	}
-
-	return r
 }
 
 func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -82,7 +82,8 @@ type ReconcileSnapshot struct {
 	Failing []string
 }
 
-// WithMetrics wraps a given Reconciler with metrics capabilities.
+// WithMetrics wraps a given Reconciler with metrics capabilities, also returning a function that
+// produces a snapshot of the metrics, for easier inspection.
 func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) (reconcile.Reconciler, func() ReconcileSnapshot) {
 	r := &wrappedReconciler{
 		Reconciler:     reconciler,
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -68,6 +68,14 @@ type wrappedReconciler struct {
 	failing map[client.ObjectKey]struct{}
 }
 
+// ReconcilerWithMetrics is a Reconciler produced by WithMetrics that can return a snapshot of the
+// state backing the metrics.
+type ReconcilerWithMetrics interface {
+	reconcile.Reconciler
+
+	Snapshot() ReconcileSnapshot
+}
+
 // ReconcileSnapshot provides a glimpse into the current state of ongoing reconciles
 //
 // This type is (transitively) returned by the controller's "dump state" HTTP endpoint, and exists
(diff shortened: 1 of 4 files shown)


neonvm/controllers/metrics.go
```diff
@@ -68,15 +68,46 @@ type wrappedReconciler struct {
...
+// when requested.
+type ReconcileSnapshot struct {
+	// ControllerName is the name of the controller: virtualmachine or virtualmachinemigration.
+	ControllerName string
+
+	// Failing is the list of objects currently failing to reconcile
+	Failing []string
+}
+
 // WithMetrics wraps a given Reconciler with metrics capabilities.
```
shayanh, line 85: Good to update the comment here mentioning we return a snapshot function as well.

shayanh (commented):
lgtm in general. Let me what you think about the `SetupWithManager` comment.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -1681,15 +1681,16 @@ func podSpec(virtualmachine *vmv1.VirtualMachine, sshSecret *corev1.Secret, conf
 // SetupWithManager sets up the controller with the Manager.
 // Note that the Runner Pod will be also watched in order to ensure its
 // desirable state on the cluster
-func (r *VirtualMachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *VirtualMachineReconciler) SetupWithManager(mgr ctrl.Manager) (func() ReconcileSnapshot, error) {
```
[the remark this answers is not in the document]
sharnoff: I agree, had the same thought while writing it. Also not sure any better ideas — maybe split the "create the reconciler" and "setup the reconciler" steps, so that wrapping with metrics would be done by the caller? (although maybe it's iffy w.r.t. how the config is used)

neonvm/controllers/metrics.go
sharnoff: done, in bee7db9e3097aa0588d6d6d4203e25d7adbd2bb6

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
shayanh: I think it would be the best approach. I thought about it when I was writing `WithMetrics` but didn't do it bc calling `WithMetrics` in the `SetupWithManager` was looking fine.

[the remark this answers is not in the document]
sharnoff: Ok - in the interest of getting this PR in quickly, I'll do that in a follow-up PR? Can also do that here. WDYT?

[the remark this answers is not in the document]
shayanh: Fine with either.

neonvm/controllers/metrics.go
```diff
@@ -68,15 +68,47 @@ type wrappedReconciler struct {
...
-// WithMetrics wraps a given Reconciler with metrics capabilities.
-func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) reconcile.Reconciler {
-	return &wrappedReconciler{
+// ReocncileSnapshot provides a glimpse into the current state of ongoing reconciles
+//
+// This type is (transitively) returned by the controller's "dump state" HTTP endpoint, and exists
+// to allow us to get deeper information on the metrics - we can't expose information for every
+// VirtualMachine into the metrics (it'd be too high cardinality), but we *can* make it available
+// when requested.
+type ReconcileSnapshot struct {
```
Omrigan, line 77: It is a good idea to have `json` tags on a struct. Allows to rename fields, linters can catch when unexported fields have json tags, etc

I don't remember which particular problem I have faced, but I remember to always have json tags now :)

neonvm/controllers/metrics.go
```diff
@@ -68,15 +68,47 @@ type wrappedReconciler struct {
...
+func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) (reconcile.Reconciler, func() ReconcileSnapshot) {
+	r := &wrappedReconciler{
 		Reconciler:     reconciler,
 		Metrics:        rm,
 		ControllerName: cntrlName,
 		lock:           sync.Mutex{},
 		failing:        make(map[client.ObjectKey]struct{}),
 	}
+
+	snapshot := func() ReconcileSnapshot {
```
Omrigan, line 96: Suggestion: How about instead of having a freestanding callback, instead define a custom interface `Reconciler`, which will have `reconcile.Reconciler` embedded, but also have a method `Snapshot() ReconcileSnapshot`, which will do what this callback does? 

Reasoning:
1. No need to pass callback through the chain of calls (I think)
2. Closures make me uneasy :laughing: because: when you have a function, you don't understand which object it get's it data from (unclear dataflow). Feels like we can usually have a normal struct with an attached method instead, which is more clear IMO.
 
And also I worked with closures in an asynchronous code in C++, where it is a recipe for use-after-free. I guess in Go it will always hold a correct link to the object.

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
Omrigan: My suggestion below of ditching the callback also works :laughing:

neonvm/main.go
```diff
@@ -157,27 +159,32 @@ func main() {
 		MaxConcurrentReconciles: concurrencyLimit,
 	}
 
-	if err = (&controllers.VirtualMachineReconciler{
+	virtualmachineReconciler := &controllers.VirtualMachineReconciler{
```
Omrigan, line 162: nit: We have to start writing `vm` instead of `virtualmachine` in code at some point. I know the full name is all over the codebase and it will look inconsistent. But it is as if this file has not many of those, so maybe a good place to start.

Omrigan (approved):
Nothing serious, mostly thoughts out loud about style, etc. 

If we need to merge it ASAP, some stuff can be fixed pretty quickly and then PR can be merged.

neonvm/main.go
```diff
@@ -223,3 +236,48 @@ func checkIfRunningInK3sCluster(cfg *rest.Config) (bool, error) {
...
+		ctx, cancel := context.WithCancel(ctx)
+
+		go func() {
+			<-ctx.Done()
+			_ = server.Shutdown(context.TODO())
+		}()
+
+		err := server.ListenAndServe()
+		cancel() // avoid leaking the goroutine. It's *probably* fine, but good practice.
+		return err
```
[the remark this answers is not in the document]
sharnoff: Hm, now that you point it out though, I should have actually just `defer cancel()` up where the context is created, and then it's `return server.ListenAndServe()`.

+1 to your frustrations - I think this is typically the argument behind small standard libraries.

neonvm/controllers/metrics.go
sharnoff: Makes sense, yeah, will add that.

neonvm/controllers/metrics.go
```diff
@@ -68,15 +68,47 @@ type wrappedReconciler struct {
...
+	// ControllerName is the name of the controller: virtualmachine or virtualmachinemigration.
+	ControllerName string
+
+	// Failing is the list of objects currently failing to reconcile
+	Failing []string
+}
+
+// WithMetrics wraps a given Reconciler with metrics capabilities, also returning a function that
+// produces a snapshot of the metrics, for easier inspection.
+func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlName string) (reconcile.Reconciler, func() ReconcileSnapshot) {
```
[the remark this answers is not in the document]
sharnoff: ah yup, thanks :)

neonvm/main.go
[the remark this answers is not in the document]
sharnoff: fixed in 8f6d3aad3995c1958b49fd18a3332e1359103fd1

neonvm/controllers/metrics.go
sharnoff: done

neonvm/controllers/metrics.go
sharnoff: That's a good point, I'll go that route

neonvm/main.go
sharnoff: Fair enough - handled that as part of 23148bdd35e7b7e3fba12d756bd47176c2c8cd71

neonvm/controllers/virtualmachine_controller.go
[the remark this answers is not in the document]
sharnoff: Ended up going with the "ditch the callback" approach :)

shayanh (approved):

neonvm/controllers/metrics.go
```diff
@@ -68,15 +68,44 @@ type wrappedReconciler struct {
...
+}
+
+// WithMetrics wraps a given Reconciler with metrics capabilities, also returning a function that
+// produces a snapshot of the metrics, for easier inspection.
+func WithMetrics(
+	reconciler reconcile.Reconciler,
+	rm ReconcilerMetrics,
+	cntrlName string,
+) ReconcilerWithMetrics {
+	r := &wrappedReconciler{
```
Omrigan, line 100: nit: you can `return` on this line already, diff should be more compact.

sharnoff: ah right, good catch! thanks :)

PR #799, opened 2024-02-09
tests: Add e2e test for VM SSH

shayanh:
Making sure once ssh is enabled, we create the SSH secret and can ssh into the
VM from the runner pod.


sharnoff (approved):
lgtm!

PR #803, opened 2024-02-12
neonvm-controller: Enable container-mgr by default

sharnoff:
We should have this for dev at least. We'll probably need to be careful not to prematurely enable it staging/prod, but that should be straightforward enough.


shayanh (approved):
lgtm. Seems we got a lot of problems with old way of cgroup handling. I suggest creating an issue to tracking the slowness of crictl.

PR #809, opened 2024-02-14
neonvm, agent: VM restart metrics

### LINKED ISSUE #449: Feature: Improve observability around VM restarts
First issue with the new feature template :)

There's broadly two areas of motivation here:

First: We don't have any high-level observability into the _rate of restarts and kernel panics, cluster-wide_. Sometimes we get hints of this, because failure to restart on kernel panic (will be fixed by #448) and our ongoing communications means they show up in the metrics for _other_ systems, but broadly our observability here currently comes from customer support tickets.

Second: It's hard to know if/when a _particular_ VM restarted (or had a kernel panic). Currently we have to dig through the logs, and that only _sometimes_ gives us the relevant information (c.f. [[slack link]](https://neondb.slack.com/archives/C04USJQNLD6/p1690491659926209), where a VM's logs suddenly cut out, without any shutdown or things of that nature). This is too much effort to be scalable.

## Feature idea(s) / DoD

We should have a dashboard that shows:

1. Per k8s cluster, total rate of restarts / kernel panics across all VMs; and
2. per-VM, when it's restarted / kernel panicked

And it would be _good_ to tell what the cause of a restart was, although that's significantly more difficult, and _should_ already be available via logs / k8s events. It may also be worthwhile to have a "restart count" field in the VM object's spec (although that could make idempotency hard).

There's also these _other_ types of restart to consider:

* QEMU process exits unexpectedly / killed
* Pod OOMkilled / failed / etc.

AFAIK we haven't yet seen them, but... maybe we haven't seen them because we don't have observability there?

## Implementation ideas

Related to / in addition to #448, we could use the technique outlined in answers to [this stackexchange question](https://unix.stackexchange.com/questions/443017/can-i-make-qemu-exit-with-failure-on-kernel-panic)[^stackexchange-question] to get QEMU to exit when there's a kernel panic. It seems like either way it's hard to distinguish between "normal shutdown" and "kernel panic" — maybe that's fine? The same logic in the runner to handle "QEMU exited when we weren't expecting it to" could also handle "QEMU got killed, etc". The main challenge as I see it is how to actually expose that metric — probably: how to get it to the controller? If we expose it in the pod, we'll miss kernel panics if the VM is shut down before the metrics are scraped.

In the VM controller, we also want to track the times we create new runners when the (old) runner pod becomes e.g. "succeeded" or "failed" without the VM being deleted in some metric. It could be the same as from the runner pod(s), or different.

We should also carefully consider whether we want these metrics to be per-pod or global. Or, maybe we have both, so that aggregate queries are efficient, but it's possible to look at the history for a particular VM. See also: https://neondb.slack.com/archives/C03F5SM1N02/p1690668679492609?thread_ts=1690668200.116389&cid=C03F5SM1N02

[^stackexchange-question]: NB: One of the answers says that pvpanic is only available on x86. I'm not sure that's still true: https://bugzilla.redhat.com/show_bug.cgi?id=1961178 (see "fixed in version: kernel-5.13.0...") — the question is then whether QEMU supports it.

shayanh:
- Cluster-wide metric in the controller
- Per-VM metric in the autoscaler agent

Resolves #449


pkg/agent/prommetrics.go
```diff
@@ -299,19 +300,35 @@ func makePerVMMetrics() (PerVMMetrics, *prometheus.Registry) {
...
-	return prometheus.Labels{
+	labels := prometheus.Labels{
 		"vm_namespace": namespace,
 		"vm_name":      vmName,
 		"endpoint_id":  endpointID,
 		"project_id":   projectID,
-		"value":        string(valueType),
 	}
+	valueTypeStr := string(valueType)
+	if len(valueType) > 0 {
```
sharnoff, line 328: do we actually need to explicitly convert to string here? I would have assumed Go allows `len(..)` for any string-like, but... maybe not.

pkg/agent/watch.go
```diff
@@ -317,6 +317,21 @@ func makeVMMemMetrics(vm *vmapi.VirtualMachine) []vmMetric {
 	return metrics
 }
 
+func makeVMRestartCountMetrics(vm *vmapi.VirtualMachine) []vmMetric {
```
sharnoff, line 320: wdyt about just returning a single `vmMetric` here? we'd have to wrap it in a `[]vmMetric` inside of `updateVMMetrics` but otherwise would be simpler imo

(as I was reading through the other functions, I was confused why we were iterating over a single restart count, then realized this returned a slice)

sharnoff (commented):
lgtm- left a couple comments, otherwise looks good

pkg/agent/prommetrics.go
shayanh: The casting on the line above is unnecessary. Fixed it.

pkg/agent/watch.go
shayanh: I returned a slice to keep the code similar to other metrics in all places and avoid handling whether the metric exist (since `.status.restartCount` may be nil). But I agree this is a trade-off. wdyt?

sharnoff: Hm, yeah fair enough — I forgot `.status.restartCount` can be nil (_maybe_ we'd want to use a value of zero when it's nil? wdyt?). Either way ok by me

sharnoff (approved):

pkg/agent/watch.go
Omrigan: One way around this would be call it `makeVMRestartMetrics`, without the `Count` word. This way it suggests there can be more metrics in the future e.g. `TotalRestartTimeSeconds`

Also can add smth like 
```
// Only one metric for now
```

pkg/agent/prommetrics.go
```diff
@@ -299,19 +300,34 @@ func makePerVMMetrics() (PerVMMetrics, *prometheus.Registry) {
 				"value",        // vmResourceValue: min, spec_use, status_use, max
 			},
 		)),
+		restartCount: util.RegisterMetric(reg, prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "autoscaling_vm_restart_count",
+				Help: "Number of times that the VM has restarted",
+			},
+			[]string{
+				"vm_namespace", // .metadata.namespace
```
Omrigan, line 309: I think this should be reused with other metrics. 

But can be left for the next time someone touches this code.

Omrigan (approved):

pkg/agent/prommetrics.go
shayanh: I know we repeated label names for each metrics, although restart count's labels are slightly different. On the other hand, having the labels explicitly defined here is nice and straightforward. Anyway I'd rather not touch them on this PR and leave the potential improvements for the future.

PR #811, opened 2024-02-15
runner: log forward: wait for QEMU startup and prevent panic

### LINKED ISSUE #791: runner: delay virtio log forwarding until QEMU is up and running

When QEMU fails to start, log forwarding fails with the following message in log:

```
2024-02-06T17:40:05.47029304Z stderr F {"level":"info","ts":1707241205.470249,"logger":"neonvm-runner","caller":"runner/main.go:733","msg":"calling cgexec","args":["-g","cpu:/neonvm-qemu-postgres14-disk-test-d69bf","qemu-system-x86_64","-runas","qemu","-machine","q35","-nographic","-no-reboot","-nodefaults","-only-migratable","-audiodev","none,id=noaudio","-serial","pty","-serial","stdio","-msg","timestamp=on","-qmp","tcp:0.0.0.0:20183,server,wait=off","-qmp","tcp:0.0.0.0:20184,server,wait=off","-qmp","unix:/vm/qmp-sigterm.sock,server,wait=off","-device","virtio-serial","-chardev","socket,path=/vm/log.sock,server=on,wait=off,id=log","-device","virtserialport,chardev=log,name=tech.neon.log.0","-drive","id=rootdisk,file=/vm/images/rootdisk.qcow2,if=virtio,media=disk,index=0,cache=none","-drive","id=runtime,file=/vm/images/runtime.iso,if=virtio,media=cdrom,readonly=on,cache=none","-drive","id=ssh-authorized-keys,file=/vm/images/ssh-authorized-keys.iso,if=virtio,media=cdrom,cache=none","-enable-kvm","-cpu","max","-smp","cpus=1,maxcpus=2,sockets=1,cores=2,threads=1","-m","size=1073741824b,slots=4,maxmem=5368709120b","-netdev","tap,id=default,ifname=tap-def,script=no,downscript=no,vhost=on","-device","virtio-net-pci,netdev=default,mac=ca:71:dc:25:ff:39","-kernel","/vm/kernel/vmlinuz","-append","panic=-1 init=/neonvm/bin/init memhp_default_state=online_movable console=ttyS1 loglevel=7 root=/dev/vda rw"]}
```

## DoD

There is no message in log if QEMU just failed to start

## Implementation ideas

Delay log forwarding until QEMU properly runs

## Misc

CC: @shayanh

Omrigan:
Previously, if QEMU failed to startup, the log output contained either:
1. Panic, because of `reader==nil`
2. `"error":"dial unix /vm/log.sock: connect: no such file or directory","stacktrace":"main.forwardLogs.func1\n\t/workspace/neonvm/runner/main.go:897\nmain.forwardLogs\n\t/workspace/neonvm/runner/main.go:921"`

Which was confusing. Not it will have:

`{"level":"warn","ts":1707993260.548491,"logger":"neonvm-runner","caller":"runner/main.go:897","msg":"QEMU shut down too soon to start forwarding logs"}`

Panic was found by the Discord user @ido6668.

Fixes #791

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -892,7 +892,7 @@ func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) {
 	// Wait a bit to reduce the chance we attempt dialing before
 	// QEMU is started
 	select {
-	case <-time.After(500 * time.Millisecond):
+	case <-time.After(200 * time.Millisecond):
 	case <-ctx.Done():
 		logger.Warn("QEMU shut down too soon to start forwarding logs")
 	}
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -925,7 +925,9 @@ func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) {
 			if conn != nil {
 				conn.Close()
 			}
-			_ = drainLogsReader(reader, logger)
+			if reader != nil {
+				_ = drainLogsReader(reader, logger)
+			}
 			return
 		case <-time.After(b.Duration()):
 		}
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -881,11 +881,20 @@ func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) {
 	delay := 3 * time.Second
 	var conn net.Conn
 	var reader *bufio.Reader
+
 	b := &backoff.Backoff{
 		Min:    100 * time.Millisecond,
 		Max:    delay,
 		Factor: 2,
 		Jitter: true,
+	}
+
+	// Wait a bit to reduce the chance we attempt dialing before
+	// QEMU is started
+	select {
+	case <-time.After(500 * time.Millisecond):
+	case <-ctx.Done():
+		logger.Warn("QEMU shut down too soon to start forwarding logs")
 	}
 
 	for {
--- a/neonvm/config/default-vxlan/manager_config_patch.yaml
+++ b/neonvm/config/default-vxlan/manager_config_patch.yaml
@@ -17,6 +17,7 @@ spec:
         - "--metrics-bind-address=:8080"
         - "--leader-elect"
         - "--concurrency-limit=128"
+        - "--enable-container-mgr"
         - "--zap-devel=false"
         - "--zap-time-encoding=iso8601"
         - "--zap-log-level=info"


Omrigan:
Warning! Based on #803, otherwise I couldn't test it.

neonvm/runner/main.go
```diff
@@ -881,13 +881,22 @@ func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) {
...
 		Min:    100 * time.Millisecond,
 		Max:    delay,
 		Factor: 2,
 		Jitter: true,
 	}
 
+	// Wait a bit to reduce the chance we attempt dialing before
+	// QEMU is started
+	select {
+	case <-time.After(500 * time.Millisecond):
```
sharnoff, line 895: Adding sleeps seems a bit hacky imo, especially one as long as 0.5s. Can we get away with not having this?

neonvm/runner/main.go
```diff
@@ -925,7 +934,9 @@ func forwardLogs(ctx context.Context, logger *zap.Logger, wg *sync.WaitGroup) {
 			if conn != nil {
 				conn.Close()
 			}
-			_ = drainLogsReader(reader, logger)
+			if reader != nil {
+				_ = drainLogsReader(reader, logger)
+			}
```
sharnoff, line 939: This could cause us to drop logs if the VM shuts down very quickly after starting up, right?

Can we try dialing the socket if `reader == nil` to possibly recover?

neonvm/runner/main.go
Omrigan: Yep, it is a bit hacky. Probably the idiomatic solution would be to have inotify, but then this sleep is not critical, since we have a backoff below anyway. So I decided not to bother. 

I can shorten it to 200ms, feels like enough for QEMU to create the socket. WDYT?

neonvm/runner/main.go
sharnoff: oh, QEMU removes the socket when it's done? or is it that you only can receive from the socket when there's someone else writing

neonvm/runner/main.go
sharnoff: IIUC without the wait it _will_ work, but it'll have unrelated errors, right? 200ms should be fine - boot takes 0.4s in prod I think? As long as we aren't risking delaying compute_ctl startup by blocking its log sending, we should be good I think

neonvm/runner/main.go
Omrigan: > QEMU removes the socket when it's done?

Yes. Socket is a "part" of virtio-serial device, so it's lifecycle is tied to it.

neonvm/runner/main.go
Omrigan: > IIUC without the wait it will work, but it'll have unrelated errors, right? 

Yes

> 200ms should be fine - boot takes 0.4s in prod I think? 

Yes

PR #820, opened 2024-02-21
use user-independent cluster name

skyzh:
revert (unexpected?) changes introduced in https://github.com/neondatabase/autoscaling/pull/809


sharnoff (approved):
good catch, thanks!

PR #822, opened 2024-02-21
runner: mount raw block device

skyzh:
This pull request adds support for mounting a raw block device to the VM for future local file cache implementation.

* Raw block device does not have a filesystem UUID, and therefore we cannot use `blkid` to find the disk. We attached the disk with `serial=<ID>`.
* Due to (probably) a bug of QEMU, we cannot directly append `serial=<ID>` in the `-drive` field. It can only be specified in `-device`. And it doesn't boot if we only change one drive to use `-device virtio`. We have to change all attached drives to use `-drive` + `-device`.
* `blkid` does not look at serial number. `udevadm info --query=all` will only print the serial number after `udevadm trigger`, but it doesn't work at boot. Therefore, we have to use `udevadm -a` to walk through all attributes and find the serial number.
* If the serial number in block device information matches the disk ID, we will create a soft link to `mountPath`.

https://github.com/neondatabase/neon/issues/6919


neonvm/runner/main.go
```diff
@@ -277,11 +281,19 @@ func createISO9660runtime(
 				mounts = append(mounts, fmt.Sprintf(`/neonvm/bin/mount %s $(/neonvm/bin/blkid -L %s) %s`, opts, disk.Name, disk.MountPath))
 				// Note: chmod must be after mount, otherwise it gets overwritten by mount.
 				mounts = append(mounts, fmt.Sprintf(`/neonvm/bin/chmod 0777 %s`, disk.MountPath))
+				diskCnt += 1
+			case disk.RawDisk != nil:
+				device := fmt.Sprintf("/dev/vd%s", string(rune(int('a')+diskCnt)))
```
sharnoff, line 286: This seems fragile (smells like multiple sources of truth). Any ideas for ways around it?

neonvm/runner/main.go
```diff
@@ -703,16 +728,31 @@ func main() {
...
+			extraOptions = "," + extraOptions
+		}
+		if deviceExtraOptions != "" && deviceExtraOptions[0] != ',' {
+			deviceExtraOptions = "," + deviceExtraOptions
+		}
+		qemuCmd = append(qemuCmd, "-drive", fmt.Sprintf("id=%s,file=%s,if=none,index=%d%s", id, file, *diskIdxCnt, extraOptions))
+		qemuCmd = append(qemuCmd, "-device", fmt.Sprintf("virtio-blk,drive=%s%s", id, deviceExtraOptions))
+		*diskIdxCnt += 1
+		return qemuCmd
+	}
```
sharnoff, line 744: Would be nice to have this as a standalone function if possible (perhaps: return only the args that would be added, and then `qemuCmd = append(qemuCmd, qemuDiskArgs(..)...)` ?)

sharnoff: It'd also be great to have this as a separate commit (and then merge via rebase onto main), or as a separate PR and have this one based on it. That way, it's clear which changes are refactoring(ish) vs adding the feature.

sharnoff (commented):
partial review; focusing on the larger stuff first.

neonvm/runner/main.go
skyzh: Yep it's true, but I could not find a good way of doing an 1v1 mapping between serial number and `vd*`. The way it works is that we first guess where the raw block device would be, i.e., vde, and then use the utility tools to verify if it matches the serial number.

neonvm/runner/main.go
skyzh: sounds good I'll make it in a separate commit.

skyzh:
TODO: add test cases, and ready for review :)

neonvm/runner/main.go
Omrigan: How about with smth like `lsblk -o "serial,name" -p`? Or perhaps look it up how `lsblk` does it and reimplement the same way.

neonvm/runner/main.go
```diff
@@ -558,6 +583,22 @@ func runInitScript(logger *zap.Logger, script string) error {
 	return nil
 }
 
+func qemuDiskArgs(id string, file string, extraOptions string, deviceExtraOptions string, diskIdxCnt *int) []string {
+	if extraOptions != "" && extraOptions[0] != ',' {
+		extraOptions = "," + extraOptions
```
Omrigan, line 588: if pass it as a map - will it be prettier? 

Or use builder here.

Omrigan:
> Due to (probably) a bug of QEMU, we cannot directly append serial=<ID> in the -drive field. It can only be specified in -device. And it doesn't boot if we only change one drive to use -device virtio. We have to change all attached drives to use -drive + -device.

Perhaps this is intentional. From `man qemu-system-x86_64`:

```
       -drive option[,option[,option[,...]]]
              Define  a new drive. This includes creating a block driver node (the backend) as well
              as a guest device, and is mostly a shortcut for defining the corresponding  -blockdev
              and -device options.
```

So the idea is `-blockdev` is responsible for the disk "backend" and `-device` is for disk "frontend". Since `serial` is a virtualized property, it is define on a "frontend" side.

neonvm/runner/main.go
```diff
@@ -558,6 +583,22 @@ func runInitScript(logger *zap.Logger, script string) error {
...
+	}
+	if deviceExtraOptions != "" && deviceExtraOptions[0] != ',' {
+		deviceExtraOptions = "," + deviceExtraOptions
+	}
+
+	args := []string{}
+	args = append(args, "-drive", fmt.Sprintf("id=%s,file=%s,if=none,index=%d%s", id, file, *diskIdxCnt, extraOptions))
```
Omrigan, line 595: maybe, we shoud use `-blockdev` here, if we have `-device` also? 

See also https://github.com/neondatabase/autoscaling/pull/822#issuecomment-1986842294

skyzh: i'll have a try!

skyzh:
> How about with smth like lsblk -o "serial,name" -p? Or perhaps look it up how lsblk does it and reimplement the same way.

@Omrigan unfortunately, lsblk does not show the serial number or name for the raw block device :( could be a bug with lsblk.

neonvm/runner/main.go
```diff
@@ -277,11 +281,19 @@ func createISO9660runtime(
 				mounts = append(mounts, fmt.Sprintf(`/neonvm/bin/mount %s $(/neonvm/bin/blkid -L %s) %s`, opts, disk.Name, disk.MountPath))
 				// Note: chmod must be after mount, otherwise it gets overwritten by mount.
 				mounts = append(mounts, fmt.Sprintf(`/neonvm/bin/chmod 0777 %s`, disk.MountPath))
+				diskCnt += 1
+			case disk.RawDisk != nil:
+				device := fmt.Sprintf("/dev/vd%s", string(rune(int('a')+diskCnt)))
+				// If the serial matches the record, soft-link it to the mountPath.
+				mounts = append(mounts, fmt.Sprintf(`(/neonvm/bin/udevadm info --query=all --name=%s -a | grep 'ATTR{serial}==%q') && /neonvm/bin/ln -s %s %s && /neonvm/bin/chmod 0777 %s`, device, disk.Name, device, disk.MountPath, disk.MountPath))
```
shayanh, line 288: nit: I suggest moving the `fmt.Sprintf` to a separate line to make it easier to read.

neonvm/runner/main.go
```diff
@@ -558,6 +583,22 @@ func runInitScript(logger *zap.Logger, script string) error {
...
+	if extraOptions != "" && extraOptions[0] != ',' {
+		extraOptions = "," + extraOptions
+	}
+	if deviceExtraOptions != "" && deviceExtraOptions[0] != ',' {
+		deviceExtraOptions = "," + deviceExtraOptions
+	}
+
+	args := []string{}
+	args = append(args, "-drive", fmt.Sprintf("id=%s,file=%s,if=none,index=%d%s", id, file, *diskIdxCnt, extraOptions))
+	args = append(args, "-device", fmt.Sprintf("virtio-blk,drive=%s%s", id, deviceExtraOptions))
```
shayanh, line 596: Why are we adding a device here? I thought it might be necessary only for block devices but here we add it for all the drives. Previously we only had `-drive` for disks.

neonvm/runner/main.go
shayanh, line 595: I wonder what is the effect of `index` since we didn't use it everywhere before.

neonvm/runner/main.go
```diff
@@ -558,6 +583,22 @@ func runInitScript(logger *zap.Logger, script string) error {
 	return nil
 }
 
+func qemuDiskArgs(id string, file string, extraOptions string, deviceExtraOptions string, diskIdxCnt *int) []string {
```
shayanh, line 586: Nit: having an int pointer and mutating it doesn't look good. Also, increasing the `diskIdxCnt` seems out of scope for this function. To avoid this you can define a closure in main where it calls `qemuDiskArgs` and then increases the counter, which is locally defined there.

neonvm/runner/main.go
skyzh: if we don't do it for all devices, it does not boot unfortunately 🤣

neonvm/runner/main.go
skyzh: added it for an extra layer of stability of assigning `vdX`, should be fine if we don't have that.

neonvm/runner/main.go
skyzh: sounds good :)

skyzh:
Put this pull request on hold until we hit the disk space limit. This pull request only makes sense if the LFC take a significant amount of space and we need to reclaim the space very quickly. Also need to address the page cache problem -- it seems that raw block devices do not take the full page cache.

Omrigan:
Discussed with @skyzh, closing for the time being.

PR #823, opened 2024-02-21
neonvm: increase neonvm-controller replicas

### LINKED ISSUE #762: neonvm-controller replicas + leadership election
Optional for neondatabase/company_projects#187.

During a previous incident where a k8s node became unavailable, it took somewhere around 5-10 minutes for the neonvm-controller instance to be recreated.

This means that a single node failure caused an outage for the entire region for 5 minutes. We can do better.

## Feature idea(s) / DoD

neonvm-controller should have >1 replica, using leadership election.

We should carefully consider how we want to handle releases — handling by interleaving versions may have unexpected results. E.g. do we want to scale down to 1 replica, update, scale back up? Or is it mostly fine.

shayanh:
Increase neonvm-controller replicas to have highly available deployment for neonvm controller.

We already have leader-election enabled for neonvm-controller: https://github.com/neondatabase/autoscaling/blob/9beedbd2624fc04dcfe4efd600ebc7dd771c355f/neonvm/config/common/controller/manager.yaml#L60

Because of leader election at any point in time, at most one controller can run the reconcile loop. No matter if we're in the middle of a deployment, rollback, or any other case. Increasing the number of replicas doesn't break this safety property. 

Leader election is enabled by default in kube-builder projects and turning it off is unsafe in our case (even with a single replica) since during deployments or rollbacks two replicas may run the reconcile loop.

## Caveats

Kubernetes implements leader-election in the client side (client-go package) using leases and doesn't implement fencing. As a result in case of clock skew, we might end up having two leaders running in the cluster. This is an interesting design choice. We have to make sure we have clock sync in our k8s nodes.

More about client-go leader election: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection

Resolves #762

### DIFF
--- a/neonvm/config/common/controller/manager.yaml
+++ b/neonvm/config/common/controller/manager.yaml
@@ -16,7 +16,7 @@ spec:
   selector:
     matchLabels:
       control-plane: controller
-  replicas: 1
+  replicas: 3
   template:
     metadata:
       annotations:


sharnoff (approved):
great summary! lgtm

Omrigan (approved):

PR #812, opened 2024-02-16
neonvm: Remove `config/default`, flatten `config/*`

sharnoff:
WIP, adding commit-by-commit so I can get CI on each commit

ref https://github.com/neondatabase/autoscaling/pull/779#issuecomment-1947245145

Issue #778, opened 2024-01-31
Remove or resolve "TODO(user)" comments


sharnoff:
We have a lot of these "TODO(user)" comments in `./neonvm` from the original code generated by kube scaffold (IIRC). Right now, these only serve to make the code harder to read.

See also #719.

PR #824, opened 2024-02-21
neonvm: Cleanup kustomize

sharnoff:
This PR combines #779 and #812, leaving out removing the controller metrics proxy - that's left for a follow-up PR (this one is focused on _not_ changing the yaml).

Brief summary of changes:

1. Rename `neonvm/config/common/controller/{manager => deployment}.yaml`
2. Remove `neonvm/config/default/`
    - This is not currently used. AFAIK it was never seriously used, but I'm not 100% sure.
3. Collapse `neonvm/config/common/*` and `neonvm/config/default-vxlan/*` into `neonvm/config/`
    - Basically, after removing `neonvm/config/default`, there's only one target to produce, and so there's no need to have a separation between the "common" components vs the ones specific to default-vxlan
4. Merge kustomize patches into source yamls
    - Previously these were needed because of the `common`/`default[-vxlan]` separation, or were generated as part of skaffold — they make things much more complex though, and merging them back into the objects they change should make things simpler.

Supersedes #779.
Supersedes #812.

---

Notes for review: _Probably_ easiest to go commit-by-commit? 

---

Plan for merging: will rename commits to include the PR number, and then merge via rebasing onto main.


neonvm/config/kustomization.yaml
```diff
@@ -13,41 +13,23 @@ namePrefix: neonvm-
...
-# Protect the /metrics endpoint by putting it behind auth.
-# If you want your controller to expose the /metrics
-# endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
-
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
-# crd/kustomization.yaml
-- manager_webhook_patch.yaml
-
-# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
```
Omrigan, line 43: No need to keep comments here, above and below?

Omrigan (approved):

neonvm/config/kustomization.yaml
sharnoff: My thought was no, mostly because these comments (similar to #778) are directed at the code author for customizing the output of skaffold

shayanh (approved):

Issue #827, opened 2024-02-22
neonvm-controller should catch reconcile panics


sharnoff:
## Problem description / Motivation

Currently if a call to `Reconcile()` panics, it takes down the entire controller, which can result in extended unavailability. This is way far more fragile than it needs to be.

- ref https://neondb.slack.com/archives/C06KXBQCUJE/p1708526697192119
- ref https://app.incident.io/neondb/incidents/151?tab=follow-ups

## Feature idea(s) / DoD

If a call to `Reconcile()` panics, that should be converted into an error and returned to the controller manager.

## Implementation ideas

See e.g. how the autoscaler-agent catches `Runner` panics.

PR #828, opened 2024-02-22
neonvm: Refactor+deduplicate controller setup

sharnoff:
WIP- thinking of this because of #827. Maybe we want to actually unify the reconciler types?

### DIFF
--- a/neonvm/controllers/common.go
+++ b/neonvm/controllers/common.go
@@ -1,0 +1,62 @@
+package controllers
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


sharnoff:
cc @Omrigan, @mikhail-sakhnov anything useful here, do you think? or should just close

Omrigan:
> cc @Omrigan, @mikhail-sakhnov anything useful here, do you think? or should just close

Let's just close, maybe look into it during #832

PR #829, opened 2024-02-22
neonvm-controller: Catch Reconcile() panics

### LINKED ISSUE #827: neonvm-controller should catch reconcile panics

sharnoff:
In theory I think this could be done with some kind of middleware, similarly to our metrics collection. In practice, it's quite small, and probably more effort than it's worth to do the abstraction.

Resolves #827.

---

~~Haven't tested it yet, need to try that locally to check it works.~~

### DIFF
--- a/neonvm/controllers/catch_panic.go
+++ b/neonvm/controllers/catch_panic.go
@@ -1,0 +1,33 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"runtime/debug"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type catchPanicReconciler struct {
+	inner reconcile.Reconciler
+}
+
+func withCatchPanic(r reconcile.Reconciler) reconcile.Reconciler {
+	return &catchPanicReconciler{inner: r}
+}
+
+func (r *catchPanicReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	log := log.FromContext(ctx) // TODO: is this valid?
+
+	defer func() {
+		if v := recover(); v != nil {
+			err = fmt.Errorf("Reconcile panicked: %v", v)
+			log.Error(err, "stack", string(debug.Stack()))
+		}
+	}()
+
+	result, err = r.inner.Reconcile(ctx, req)
+	return
+}
--- a/neonvm/controllers/catch_panic.go
+++ b/neonvm/controllers/catch_panic.go
@@ -19,7 +19,7 @@ func withCatchPanic(r reconcile.Reconciler) reconcile.Reconciler {
 }
 
 func (r *catchPanicReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
-	log := log.FromContext(ctx) // TODO: is this valid?
+	log := log.FromContext(ctx)
 
 	defer func() {
 		if v := recover(); v != nil {
--- a/neonvm/controllers/virtualmachinemigration_controller.go
+++ b/neonvm/controllers/virtualmachinemigration_controller.go
@@ -89,13 +89,6 @@ type VirtualMachineMigrationReconciler struct {
 func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, err error) {
 	log := log.FromContext(ctx)
 
-	defer func() {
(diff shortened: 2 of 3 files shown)


shayanh:
I'm more on the side of middleware approach and we already have a similar one.

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -147,7 +155,7 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 
 	// Fetch the corresponding VirtualMachine instance
 	vm := new(vmv1.VirtualMachine)
-	err := r.Get(ctx, types.NamespacedName{Name: migration.Spec.VmName, Namespace: migration.Namespace}, vm)
+	err = r.Get(ctx, types.NamespacedName{Name: migration.Spec.VmName, Namespace: migration.Namespace}, vm)
```
Omrigan, line 158: This case is I think the reason why having named return parameters is less than ideal. `err` is here is not necessarily returned, so the semantics and effects of this assignment can be misleading.

neonvm/controllers/virtualmachinemigration_controller.go
```diff
@@ -85,9 +86,16 @@ type VirtualMachineMigrationReconciler struct {
 // - About Operator Pattern: https://kubernetes.io/docs/concepts/extend-kubernetes/operator/
 // - About Controllers: https://kubernetes.io/docs/concepts/architecture/controller/
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
-func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, err error) {
```
Omrigan, line 89: I've expressed this opinion before: I don't think it adds to readability.

Omrigan (approved):
> In theory I think this could be done with some kind of middleware, similarly to our metrics collection. In practice, it's quite small, and probably more effort than it's worth to do the abstraction.

I think we should try to get it to prod ASAP. So the fastest approach :)

sharnoff:
Went with the middleware approach, via new file `neonvm/controllers/catch_panic.go`.
I think that satisfies both of your concerns?

neonvm/controllers/catch_panic.go
```diff
@@ -0,0 +1,33 @@
...
+type catchPanicReconciler struct {
+	inner reconcile.Reconciler
+}
+
+func withCatchPanic(r reconcile.Reconciler) reconcile.Reconciler {
+	return &catchPanicReconciler{inner: r}
+}
+
+func (r *catchPanicReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	log := log.FromContext(ctx) // TODO: is this valid?
```
shayanh, line 22: Not sure what is the answer to this question, but `log.FromContext` returns a dummy logger that does nothing if ctx is nil or it doesn't have a logger. Since Reconcile is being called by the controller manager its context *should have* a valid logger.

shayanh (approved):
lgtm!

Omrigan:
lgtm!

Omrigan (approved):

sharnoff:
Tested locally by forcing the controller to panic on every reconcile when the VM is running:

```diff
diff --git a/neonvm/controllers/virtualmachine_controller.go b/neonvm/controllers/virtualmachine_controller.go
index 0c4ae66..1868b6c 100644
--- a/neonvm/controllers/virtualmachine_controller.go
+++ b/neonvm/controllers/virtualmachine_controller.go
@@ -470,6 +470,10 @@ func (r *VirtualMachineReconciler) doReconcile(ctx context.Context, virtualmachi
             // do nothing
         }
     case vmv1.VmRunning:
+        if true {
+            panic("test panic!")
+        }
+
         // Check if the runner pod exists
         vmRunner := &corev1.Pod{}
         err := r.Get(ctx, types.NamespacedName{Name: virtualmachine.Status.PodName, Namespace: virtualmachine.Namespace}, vmRunner)
```

PR #836, opened 2024-02-27
neonvm: Remove support for pre-v0.7.0 neonvm-runner versions

### LINKED ISSUE #710: neonvm should remove support for pre-cgroups neonvm-runner version

In neonvm-controller, changes to the reconcile logic are complicated by handling the possibility of VMs with runner pods that don't support fractional CPU (motivating example: #702). In practice, these changes are difficult, and we have no validation that things continue to work for such old VMs.

So: better to just remove support entirely. I've checked our clusters, and it seems there's none of those old runners (which makes sense - the original PR (#172) was deployed to prod [on 2023-05-11](https://github.com/neondatabase/autoscaling/commit/c36fb8424b613d64ecfef0fc0937ce95fb579196)).

<details>
<summary> for reference, the command I checked clusters with </summary>

```sh
kubectl --context=$cluster get pod -l vm.neon.tech/name -A -o json \
    | jq -r '.items[] | select(.metadata.labels["vm.neon.tech/runner-version"] == null) | "\(.metadata.namespace)/\(.metadata.name)"'
```

</details>

## Feature idea(s) / DoD

Logic inside the neonvm-controller's reconcile loop should be free to assume that fractional CPU via cgroups is supported.

Some kind of visible error should occur for VMs with runner pods that are too old, the error should not automatically trigger restart, _and_ it should not prevent deleting the VM.

## Implementation ideas

I _think_ it's acceptable to have scaling disabled — so in `doReconcile` if we find that a runner pod doesn't have the minimum version _and_ the VM is in a "normal" phase (running/scaling/etc), we can just log an error (or create an error event) and return. IIUC, that should leave deletion logic intact _and_ just disable scaling - and killing the runner pod should resolve any issues, if something comes up.

sharnoff:
In short: Handling for supporting pre-["fractional CPU"](https://github.com/neondatabase/autoscaling/pull/172) neonvm-runner versions introduces complexity via subtle logic that's hard to check in practice, because we haven't run any versions this old in a long time.

Resolves #710.


neonvm/controllers/virtualmachine_controller.go
```diff
@@ -235,20 +235,6 @@ func (r *VirtualMachineReconciler) doFinalizerOperationsForVirtualMachine(ctx co
...
-	if !ok {
-		return false
-	}
-
-	uintVal, err := strconv.ParseUint(val, 10, 32)
-	if err != nil {
-		return false
-	}
-
-	return api.RunnerProtoVersion(uintVal).SupportsCgroupFractionalCPU()
```
Omrigan, line 249: does it make sense to do either of those?
1. Keep this line as an assert
2. Delete SupportsCgroupFractionalCPU function

pkg/api/VERSIONING.md
```diff
@@ -95,7 +95,7 @@ number.
 Note: Components v0.6.0 and below did not have a versioned protocol between the controller and the runner.
 | Release | controller | runner |
 |---------|------------|--------|
-| _Current_ | 0 - 1 | 1 |
```
Omrigan, line 98: Side note: the versioning scheme here is a little bit not intuitive, and I was thinking whether it would simplify things if we would just have "the autoscaling protocol", meaning, would use the same versioning scheme for all components protocols.

Omrigan:
I would replace "old" in commit subj with the concrete version number which was made incompatible

Omrigan (approved):

shayanh (approved):

pkg/api/VERSIONING.md
sharnoff: I'm not sure tbh.

From my perspective, the primary reason to have a versioned protocol is to make releasing & rollback more clear - especially when things are already off the happy path. Having a single versioning scheme would reduce information.

At the same time, I think the information it currently gives us is insufficient and sometimes misleading by omission. For example, the changes from your PR #865 weren't reflected in any versioning here, but _did_ have compatibility implications. I think our lack of existing coverage for more complex relationships like that is a strong reason to switch to a unified versioning scheme.

Could you open an issue for it?

neonvm/controllers/virtualmachine_controller.go
sharnoff: Makes sense, yeah. Added some assertions, basically. LMK what you think.

pkg/api/VERSIONING.md
Omrigan: Opened https://github.com/neondatabase/autoscaling/issues/888

sharnoff: thanks!

Omrigan:
> Remove support for pre-v1 neonvm-runner versions

I'd write pre-[v0.7.0](https://github.com/neondatabase/autoscaling/releases/tag/v0.7.0). 

We definitely don't have computes alive for 11 months, do we?

Omrigan (approved):

sharnoff:
> I'd write pre-v0.7.0

Ah yeah, good point

> We definitely don't have computes alive for 11 months, do we?

nope — see here: https://github.com/neondatabase/autoscaling/issues/710#issuecomment-1967478953

PR #845, opened 2024-03-05
neonvm: Make .status.restartCount not a pointer

sharnoff:
While drawing the state machine we used for RestartCount as part of an incident summary, I realized we actually didn't even need the field to be a pointer. :facepalm:

Coudln't connect to dockerhub, so couldn't run `make generate` - tried to do that manually. We'll see if it works.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -393,7 +393,7 @@ type VirtualMachineStatus struct {
 	Phase VmPhase `json:"phase,omitempty"`
 	// Number of times the VM runner pod has been recreated
 	// +optional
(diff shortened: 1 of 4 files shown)


shayanh (approved):
Nice!

Side-note: It's good to write a few sentences somewhere (probably in the `virtualmachine_types.go`) describing what should be pointer in a k8s object API.

Omrigan (approved):
> Couldn't connect to dockerhub, so couldn't run make generate - tried to do that manually. We'll see if it works.

Better to do it still. Would be unfortunate to get a change of e.g. single space when we run make generate next time.

sharnoff:
> Better to do it still. Would be unfortunate to get a change of e.g. single space when we run make generate next time.

Yup - we actually have a lint to check generated files :wink:

(+ checked locally and it's ok)

Issue #851, opened 2024-03-09
Bug: scheduler's critically delayed event processing has positive feedback loop


sharnoff:
## Environment

Prod (occurred twice recently)

## Steps to reproduce

Hard to reproduce locally - requires odd circumstances under a lot of load. AFAICT we've only seen it occur at startup — probably because that's when there's the most stuff going on.

The general idea of the triggering behavior is this:

1. For some reason, event processing in the scheduler gets delayed (whether that's because of weird behavior at startup, or some other failure mode)
2. _Some_ VM pod start events are delayed enough that the VMs are deleted before the events are handled
3. While handling those events:
    1. We don't see the VM in the local VM store (because it doesn't exist)
    2. Thinking the store is out of date, we `Relist()` — this can take ~2s
    3. After relisting, the VM is still not in the store, so we return error
4. Because relisting takes _so long_, more events get delayed, so we handle more pod start events after the VM was deleted, so we cause _even more_ delays

Note that this is _also_ because we only handle a single event at a time, so waiting for 2s handling one event holds up the entire queue.

(Originally, that's because otherwise we'd have to be careful to avoid out of order start/stop events - there's ways around this, though).

## Other logs, links

- [initial discussion in slack](https://neondb.slack.com/archives/C03TN5G758R/p1709923277096179)
- [logs link](https://neonprod.grafana.net/explore?schemaVersion=1&panes=%7B%22bdw%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D%5C%22prod-us-west-2-eta%5C%22,namespace%3D%5C%22kube-system%5C%22,pod%3D%5C%22autoscale-scheduler-84696b4d94-2kwdd%5C%22%7D%20%7C%3D%20%5C%22autoscale-scheduler.handlers%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%221709922300000%22,%22to%22:%221709922600000%22%7D%7D%7D&orgId=1)

## Tasks

```[tasklist]
- [ ] #853
- [ ] #854
- [ ] #863
```

[...]

PR #854, opened 2024-03-09
plugin: Use multiple event handler threads

sharnoff:
Each handler has its own queue, and we distribute events by hashing the name of the object we're sending through, so that pod start and stop event ordering is preserved by going in through the same queue.

This will help reduce the impact of any future issues where the event queue gets backed up.

Part of #851.

---

Like #853, I'll probably put this on staging for testing over the weekend, so we can (optimistically) do a hotfix early next week.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -212,7 +212,7 @@ func makeAutoscaleEnforcerPlugin(
 
 	for i := 0; i < config.EventQueueWorkers; i += 1 {
 		// copy the loop variable to avoid it escaping pre Go 1.22
-		go func(idx int) {
+		go func(ctx context.Context, idx int) {
 			for {
 				callback, err := queueSet.wait(ctx, idx) // NB: wait pulls from the front of the queue
 				if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/plugin/plugin.go
```diff
@@ -205,17 +210,20 @@ func makeAutoscaleEnforcerPlugin(
...
-	go func() {
-		for {
-			callback, err := queue.Wait(ctx) // NB: Wait pulls from the front of the queue
-			if err != nil {
-				logger.Info("Stopped waiting on pod/VM queue", zap.Error(err))
-				break
-			}
+	for i := 0; i < config.EventQueueWorkers; i += 1 {
+		// copy the loop variable to avoid it escaping pre Go 1.22
+		go func(idx int) {
```
shayanh, line 215: nit: it's good to add ctx as a argument for this goroutine so it would be clear that we respect the context. At first, I thought we have no control over this goroutine's lifetime.

shayanh (approved):
lgtm!

pkg/plugin/plugin.go
sharnoff: fair enough, makes sense - done

PR #855, opened 2024-03-11
agent: track more liveness in vm-stuck metrics

### LINKED ISSUE #770: Track more liveness in "autoscaling stuck"

Currently, "autoscaling stuck" metrics and logs use the following definition:

> An autoscaling-enabled VM is "stuck" if there has not been a successful health check response for the last 20s.

Current implementation is [here](https://github.com/neondatabase/autoscaling/blob/a7943db25bf5d97eaf4771b46e10a3d97ed4b3ea/pkg/agent/globalstate.go#L511-L528).

This misses various other ways that autoscaling may currently be failing for a particular VM, some of which we've seen in prod (e.g. due to missing a pod start event, the scheduler doesn't know about a particular VM and always returns 404 to the autoscaler-agent).

## Feature idea(s) / DoD

Some other types of "stuckness" we should look at including:

- Requests to the scheduler plugin are failing
- Requests to update the VM object are failing
- _Other_ requests to the vm-monitor are failing

And possibly also (although more difficult):

- Scheduler plugin consistently denying desired upscaling
- vm-monitor consistently denying desired downscaling (currently sometimes expected in practice)

See also #594. Tracking the delay between VM object change and VM status change is currently blocked on #592.

shayanh:
Now a VM is considered stuck if:
- Health checks to vm-monitor are failing for the configured period of time
- There are `n` failed or denied requests to the scheduler over the last `t` seconds
- There are `n` failed or denied requests to the vm-monitor over the last `t` seconds
- There are `n` failed requests to the NeonVM over the last `t` seconds

I have defined the values `n` and `t` for each component in the config map. Comment out if you think the chosen values could be better.

Part of a #770


pkg/agent/execbridge.go
```diff
@@ -74,6 +97,14 @@ func (iface *execNeonVMInterface) Request(ctx context.Context, logger *zap.Logge
 	if err != nil {
 		return fmt.Errorf("Error making VM patch request: %w", err)
 	}
+	iface.runner.status.update(iface.runner.global, func(ps podStatus) podStatus {
+		if err == nil {
+			ps.UnsuccessfulNeonVMCommCnt = 0
+		} else {
+			ps.UnsuccessfulNeonVMCommCnt++
```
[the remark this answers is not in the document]
shayanh: Thanks for pointing this out! Somehow I missed it during testing, now it's fixed!

pkg/agent/globalstate.go
```diff
@@ -466,6 +475,22 @@ func (s *lockedPodStatus) update(global *agentState, with func(podStatus) podSta
 	newStatus := with(s.podStatus)
 	now := time.Now()
 
+	isStuck := func() bool {
```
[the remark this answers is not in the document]
shayanh: Makes sense, done.

pkg/agent/execbridge.go
```diff
@@ -51,6 +51,29 @@ func (iface *execPluginInterface) Request(
...
+		}
+		if lastPermit != nil {
+			if resp.Permit == *lastPermit {
+				return false
+			}
+		}
+		return true
+	}()
+	iface.runner.status.update(iface.runner.global, func(ps podStatus) podStatus {
+		if successful {
```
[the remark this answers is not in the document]
shayanh: I didn't exactly understand your proposed solution. I see how this could be problematic in languages like C++ but afaik this is a common pattern in Go.

Omrigan (approved):

pkg/agent/execbridge.go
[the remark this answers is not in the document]
shayanh: I see. I rather tend to keep the shorter version, which is more concise and consistent with the rest of code.

shayanh:
We had a discussion yesterday with @sharnoff about this. This implementation has a problem that a VM maybe marked stuck incorrectly. This scenario could happen when agent changes its decision after one unsuccessful try.

We discussed two solutions:
1. Mark VM as stuck based on request failure rate over time.
2. Having an observer that monitors VM's object and ensures its status values converge to agent's desired values.

I decided to go with the first approach mainly because of simplicity. Second approach gives us some nice end-to-end observability, however, it's hard to done right. If needed we can implement this separately later.

shayanh:
@sharnoff @Omrigan Update this PR with what I have described in [this comment](https://github.com/neondatabase/autoscaling/pull/855#issuecomment-2008015033).

pkg/agent/globalstate.go
```diff
@@ -153,8 +153,11 @@ func (s *agentState) handleVMEventAdded(
 			state:              "", // Explicitly set state to empty so that the initial state update does no decrement
 			stateUpdatedAt:     now,
 
-			startTime:                 now,
-			lastSuccessfulMonitorComm: nil,
+			startTime:                     now,
+			lastSuccessfulMonitorComm:     nil,
+			failedMonitorRequestCounter:   util.NewRecentCounter(time.Duration(s.config.Monitor.MaxFailedRequestRate.IntervalSeconds) * time.Second),
+			failedNeonVMRequestCounter:    util.NewRecentCounter(time.Duration(s.config.NeonVM.MaxFailedRequestRate.IntervalSeconds) * time.Second),
+			failedSchedulerRequestCounter: util.NewRecentCounter(time.Duration(s.config.Scheduler.MaxFailedRequestRate.IntervalSeconds) * time.Second),
```
sharnoff, line 160: I guess this is a case where #35 would help (specifically: so we don't need to do the verbose conversion from integer seconds to `time.Duration`)

pkg/agent/config.go
```diff
@@ -16,10 +16,16 @@ type Config struct {
 	Metrics   MetricsConfig    `json:"metrics"`
 	Scheduler SchedulerConfig  `json:"scheduler"`
 	Monitor   MonitorConfig    `json:"monitor"`
+	NeonVM    NeonVMConfig     `json:"neonvm"`
 	Billing   *billing.Config  `json:"billing,omitempty"`
 	DumpState *DumpStateConfig `json:"dumpState"`
 }
 
+type RateThresholdConfig struct {
```
sharnoff, line 24: Validation of the added configs should be added to `(*config).validate()` below

pkg/agent/globalstate.go
```diff
@@ -507,6 +517,23 @@ func (s *lockedPodStatus) update(global *agentState, with func(podStatus) podSta
...
+	if s.failedMonitorRequestCounter.Get() > global.config.Monitor.MaxFailedRequestRate.Threshold {
+		return true
+	}
+	if s.failedSchedulerRequestCounter.Get() > global.config.Scheduler.MaxFailedRequestRate.Threshold {
+		return true
+	}
+	if s.failedNeonVMRequestCounter.Get() > global.config.NeonVM.MaxFailedRequestRate.Threshold {
+		return true
+	}
+	return false
```
sharnoff, line 533: IMO this should just be a single large statement, something like
```go
return s.monitorStuckAt(global.config).Before(now) ||
    s.failedMonitorRequestCounter.Get() > global.config.Monitor.MaxFailedRequestRate.Threshold ||
    ...
```


wdyt? my thinking is that visually, it should be easier to parse, and it's syntactically more clear

pkg/agent/globalstate.go
```diff
@@ -480,7 +490,7 @@ func (s *lockedPodStatus) update(global *agentState, with func(podStatus) podSta
 		case podStatusExitPanicked:
 			newState = runnerMetricStatePanicked
 		}
-	} else if newStatus.monitorStuckAt(global.config).Before(now) {
+	} else if newStatus.isStuck(global, now) {
```
sharnoff, line 493: I _think_ this should still be sound, but it makes `periodicallyRefreshState` a little weirder. I think ultimately it's probably the right call, but let's discuss - I'll book some time on your calendar.

pkg/agent/globalstate.go
sharnoff: Alternatively, it'd be good to include information in the logs (and maybe metrics?) about _why_ the VM is stuck - changing `isStuck` to return `[]string` would make a lot of sense, I think.

pkg/util/recentcnt.go
```diff
@@ -0,0 +1,40 @@
...
+	for ; i < len(rc.timestamps); i++ {
+		if rc.timestamps[i].After(checkpoint) {
+			break
+		}
+	}
+	rc.timestamps = rc.timestamps[i:]
+}
+
+// Inc increments the counter and adds the current timestamp to the list of timestamps.
+func (rc *RecentCounter) Inc() {
```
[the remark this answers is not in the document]
Omrigan: Would need to check the time is non-decreasing.

Also, can do 
`
func (rc *RecentCounter) Inc() {
    rc.inc(time.Now())
}
`

and test the private interface.

pkg/agent/globalstate.go
Omrigan: > to return []string

Or a binary mask.

pkg/agent/execbridge.go
```diff
@@ -150,6 +177,13 @@ func (h *execMonitorHandle) Downscale(
 		h.runner.recordResourceChange(current, target, h.runner.global.metrics.monitorApprovedChange)
 	}
 
+	h.runner.status.update(h.runner.global, func(ps podStatus) podStatus {
```
Omrigan, line 180: You can call this in an `else` for an above `if` and get rid of an `if` inside. 

Same for other status updates.

Omrigan (changes requested):
Implementation changed, so removing approval for now.

pkg/util/recentcnt.go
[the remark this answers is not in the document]
sharnoff: > Would need to check the time is non-decreasing

IIRC (not 100% sure) Go guarantees this, even when the system clock is not monotonic.

pkg/agent/execbridge.go
```diff
@@ -51,6 +51,27 @@ func (iface *execPluginInterface) Request(
...
+			return true
+		}
+		if lastPermit != nil {
+			if resp.Permit == *lastPermit {
+				return false
+			}
+		}
+		return true
+	}()
+	iface.runner.status.update(iface.runner.global, func(ps podStatus) podStatus {
```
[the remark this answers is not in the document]
shayanh: > Also, what's the motivation behind this logic? To handle a concurrent upscale request?

We want to check that the request hasn't failed (`err == nil`) and hasn't been denied (scheduler approved a non-zero upscale, that is not necessarily equal to `target`).

Regarding the control flow, I think the verbose version is easier to understand and less error-prune. I'll add some commentary to make it more clear.

pkg/agent/config.go
```diff
@@ -110,6 +119,16 @@ type SchedulerConfig struct {
 	RetryDeniedUpscaleSeconds uint `json:"retryDeniedUpscaleSeconds"`
 	// RequestPort defines the port to access the scheduler's ✨special✨ API with
 	RequestPort uint16 `json:"requestPort"`
+	// MaxFailedRequestRate defines the maximum rate of failed scheduler requests, above which
+	// a VM is considered stuck.
+	MaxFailedRequestRate RateThresholdConfig `json:"maxFailedRequestRate"`
+}
+
+// NeonVMConfig defines a few parameters for NeonVM requests
+type NeonVMConfig struct {
```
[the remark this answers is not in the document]
shayanh: Fixed in f4f192472a7c61a23ae41c02fb4f513526d0ec2f

pkg/agent/execbridge.go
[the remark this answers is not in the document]
shayanh: See 017813bb45eeaa016ece4f47e7a8be1f8d4abebd

pkg/util/recentcnt_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	time.Sleep(2 * time.Second)
+	rc.Inc()
+	assert.Equal(t, uint(3), rc.Get())
+
+	time.Sleep(3 * time.Second)
+	assert.Equal(t, uint(1), rc.Get())
+
+	time.Sleep(2 * time.Second)
+	assert.Equal(t, uint(0), rc.Get())
```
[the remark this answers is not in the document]
shayanh: I agree that this is not the best impl. I'm not sure what do you mean by using a fixed timestamp, but I can mock the system's clock and use a dummy clock in the tests.

pkg/util/recentcnt.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func NewRecentCounter(interval time.Duration) *RecentCounter {
+	return &RecentCounter{
+		interval:   interval,
+		timestamps: make([]time.Time, 0),
+	}
+}
+
+func (rc *RecentCounter) cleanup(t time.Time) {
```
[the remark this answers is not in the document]
shayanh: sounds good, see f1aa527e0fca92b5011e5676f25ec89b90c38207

pkg/util/recentcnt.go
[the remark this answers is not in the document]
Omrigan: > IIRC (not 100% sure) Go guarantees this

It won't be guaranteed if time is accepted through an argument.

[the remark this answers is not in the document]
shayanh: > IIRC (not 100% sure) Go guarantees this, even when the system clock is not monotonic.

Go does provide monotonic clocks. See the [time package docs](https://pkg.go.dev/time#hdr-Monotonic_Clocks).

> It may make sense to have Inc() and Get() also take the current time as input. More testable imo, and would have consistency with cleanup()

With this design, it's caller's responsibility to provide monotonic timestamps to `Inc` and `Get` functions. I don't think this is a good idea, since it's fragile and easy to break.

@Omrigan's idea is nice, however, looks a bit hacky.

pkg/agent/config.go
shayanh: fixed in 26f1567d463e630c4da6f9d10d29b843101901db

pkg/agent/execbridge.go
shayanh: Fixed in a2f0bcf6a60a92aa8c53cd6b52749a76f7bcbe90

pkg/util/recentcnt.go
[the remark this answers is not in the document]
shayanh: Addressed here 20a66f353704170fae0f1aa760905f4ce3674a1e.

pkg/agent/globalstate.go
shayanh: Added reason here 20a66f353704170fae0f1aa760905f4ce3674a1e.

pkg/agent/globalstate.go
shayanh: See this 20a66f353704170fae0f1aa760905f4ce3674a1e.

shayanh: And a bug fix: 51b2459d44f69418dde70800c615a7f0d6419b52

pkg/agent/execbridge.go
```diff
@@ -51,6 +51,25 @@ func (iface *execPluginInterface) Request(
...
+		if err != nil { // request is failed
+			return false
+		}
+		if resp.Permit == target { // request is fully approved by the scheduler
+			return true
+		}
+		if lastPermit != nil && *lastPermit != resp.Permit { // request is partially approved by the scheduler
+			return true
+		}
+		return false // scheduler denied the request
```
sharnoff, line 64: Why is partial approval a "success" but full approval is not? is it something along the lines of "partial approval amounts are monotonically increasing, so it still tracks liveness but has fewer false-positives" ?

pkg/agent/config.go
```diff
@@ -12,14 +12,22 @@ import (
...
 	Scheduler SchedulerConfig  `json:"scheduler"`
 	Monitor   MonitorConfig    `json:"monitor"`
+	NeonVM    NeonVMConfig     `json:"neonvm"`
 	Billing   *billing.Config  `json:"billing,omitempty"`
 	DumpState *DumpStateConfig `json:"dumpState"`
 }
 
+type RateThresholdConfig struct {
+	IntervalSeconds uint `json:"intervalSeconds"`
+	Threshold       uint `json:"threshold"`
```
sharnoff, line 28: I noticed there's validation for `IntervalSeconds` but not `Threshold`, is that intentional?

pkg/agent/globalstate.go
```diff
@@ -432,7 +440,10 @@ type podStatusDump struct {
 	EndState          *podStatusEndState  `json:"endState"`
 	PreviousEndStates []podStatusEndState `json:"previousEndStates"`
 
-	LastSuccessfulMonitorComm *time.Time `json:"lastSuccessfulMonitorComm"`
+	LastSuccessfulMonitorComm     *time.Time `json:"lastSuccessfulMonitorComm"`
+	FailedMonitorRequestCounter   uint       `json:"failedMonitorRequestCounter"`
+	FailedNeonVMRequestCounter    uint       `json:"failedNeonVMRequestCounter"`
+	FailedSchedulerRequestCounter uint       `json:"failedSchedulerRequestCounter"`
```
sharnoff, line 446: I wonder if these should have more information — the general philosophy behind these "state dumps" is that it exposes as much internal state as possible, so it might make sense to include the raw timestamps here instead of or in addition to the count.

WDYT?

pkg/util/recentcnt_test.go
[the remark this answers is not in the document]
sharnoff: Yeah, I was a bit vague - I meant roughly what you ended up implementing :)

pkg/agent/globalstate.go
sharnoff: > bug fix

hah, I think I've made that mistake before :)

sharnoff (approved):
nice! approval pending comments (mine & others)

pkg/agent/execbridge.go
Omrigan: Full approval is also a success, no?

pkg/agent/globalstate.go
```diff
@@ -507,6 +518,23 @@ func (s *lockedPodStatus) update(global *agentState, with func(podStatus) podSta
 	s.podStatus = newStatus
 }
 
+func (s podStatus) isStuck(global *agentState, now time.Time) (bool, []string) {
+	var reasons []string
+	if s.monitorStuckAt(global.config).Before(now) {
+		reasons = append(reasons, "monitor health check failed")
+	}
+	if s.failedMonitorRequestCounter.Get() > global.config.Monitor.MaxFailedRequestRate.Threshold {
```
Omrigan, line 526: What if Threshold is equal to zero? 

Probably the check should be skipped?

pkg/agent/globalstate.go
Omrigan: I don't think this info  will be very useful: we don't store the VM names near the timestamps.

pkg/agent/config.go
Omrigan: I'd expect 0 to be a valid threshold, meaning infinity.

pkg/agent/execbridge.go
shayanh: The if statement above cover the full approval case, which defines it as a success.

pkg/agent/config.go
shayanh: It doesn't mean infinity. It means `> 0`.

pkg/agent/globalstate.go
shayanh: I agree with @Omrigan. We can add timestamps and VMs names both but I don't see it's going to be helpful. I suggest keeping it simple for now.

pkg/agent/globalstate.go
shayanh: It makes sense for the threshold to be zero. If means `> 0`.

Omrigan (approved):

PR #857, opened 2024-03-13
neonvm: Add neonvm-runner image loader

sharnoff:
When rolling out an update to neonvm, we currently don't pre-load the neonvm-runner images. This has a couple effects:

1. For a brief period, new VM runner pods are additionally blocked on pulling the runner image (which can take a while); and
2. If there's issues with pulling the runner image, we only detect that once we're actually attempting to use it.

This PR adds a new daemonset to just keep the neonvm-runner image around on each node. Before rolling out a neonvm update, we can update the runner image in the daemonset to make sure that the new image is loaded on all the nodes.

Part of neondatabase/cloud#11232

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,6 +256,7 @@ render-manifests: $(RENDERED) kustomize
 	# Prepare:
 	cd neonvm/config/controller && $(KUSTOMIZE) edit set image controller=$(IMG_CONTROLLER) && $(KUSTOMIZE) edit add annotation buildtime:$(BUILDTS) --force
 	cd neonvm/config/vxlan-controller && $(KUSTOMIZE) edit set image vxlan-controller=$(IMG_VXLAN_CONTROLLER) && $(KUSTOMIZE) edit add annotation buildtime:$(BUILDTS) --force
+	cd neonvm/runner-image-loader && $(KUSTOMIZE) edit set image runner=$(IMG_RUNNER) && $(KUSTOMIZE) edit add annotation buildtime:$(BUILDTS) --force
 	cd deploy/scheduler && $(KUSTOMIZE) edit set image autoscale-scheduler=$(IMG_SCHEDULER) && $(KUSTOMIZE) edit add annotation buildtime:$(BUILDTS) --force
 	cd deploy/agent && $(KUSTOMIZE) edit set image autoscaler-agent=$(IMG_AUTOSCALER_AGENT) && $(KUSTOMIZE) edit add annotation buildtime:$(BUILDTS) --force
 	# Build:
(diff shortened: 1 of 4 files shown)


shayanh (approved):
Nice!

shayanh:
I saw that the CI has failed because of a namespace error but I think it should a small issue.

neonvm/runner-image-loader/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+      name: neonvm-runner-image-loader
+  template:
+    metadata:
+      labels:
+        name: neonvm-runner-image-loader
+    spec:
+      containers:
+      - image: runner:dev
+        name: neonvm-runner-loader
+        command: ["sh", "-c", "echo 'image loaded and container started' && sleep 100d"]
```
Omrigan, line 18: What happens after 100d? Container exits and is restarted by the DaemonSet?

neonvm/runner-image-loader/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
...
+  selector:
+    matchLabels:
+      name: neonvm-runner-image-loader
+  template:
+    metadata:
+      labels:
+        name: neonvm-runner-image-loader
+    spec:
+      containers:
+      - image: runner:dev
```
Omrigan, line 16: Do we have any other images in our system? I know `compute-init` is now replaced by initScript :smiley: 

`crictl` container uses the same image? Also, `runner` is not customizable my cplane, only `rootdisk` image, correct?

Makefile
```diff
@@ -296,6 +302,8 @@ deploy: check-local-context docker-build load-images manifests render-manifests
 	$(KUBECTL) -n kube-system rollout status daemonset kube-multus-ds
 	$(KUBECTL) apply -f $(RENDERED)/whereabouts.yaml
 	$(KUBECTL) -n kube-system rollout status daemonset whereabouts
+	$(KUBECTL) apply -f $(RENDERED)/neonvm-runner-image-loader.yaml
```
Omrigan, line 305: There also should be similar changes to the `deploy.sh` script we now have. Should it be stored in the `main` branch also?

Can we somehow universalize scripts for local deployments to `k3d`/`kind`, setup in GitHub workflow and deployment to dev/prod?

Omrigan (approved):

neonvm/runner-image-loader/daemonset.yaml
sharnoff: That's what would happen, yeah - if we somehow didn't deploy for 100d :sweat_smile:

neonvm/runner-image-loader/daemonset.yaml
sharnoff: > Do we have any other images in our system?

In terms of per-compute images, we still have the actual compute image itself. And there's the various images that are only a few per cluster (autoscaler-agent, scheduler, neonvm-controller, neonvm-vxlan-controller, cplane, proxy - and maybe more?)

> `crictl` container uses the same image?

Yup

> Also, `runner` is not customizable by cplane, only `rootdisk` image, correct?

_Technically_ `runner` is customizable (VM spec supports it), but that's not currently hooked up anywhere. `rootdisk` image customization is fully hooked up though.

Makefile
sharnoff: > There also should be similar changes to the `deploy.sh` script we now have

Yup, see neondatabase/cloud#11232.

> Should it be stored in the `main` branch also?

Not yet, I think. Happy to elaborate offline.

> Can we somehow universalize scripts for local deployments to k3d/kind, setup in GitHub workflow and deployment to dev/prod?

We could, yeah. Right now, `deploy.sh` has enough interaction and secondary checks that it'd be annoying for local setup or CI — and we _could_ conditionally deploy without checking, I'm hesitant to make the script more complicated than it needs to be (especially if we're adding a feature that allows bypassing safety checks).

PR #865, opened 2024-03-19
plugin: construct VmInfo from pod, not vm

### LINKED ISSUE #863: VM info should be available on runner pod, so scheduler doesn't need to watch VMs

There's a lot of complication in the scheduler and our `Watch` implementation that comes from needing to `Relist()` to synchronize state (see #851). It'd be nice if we could _just_ look at a single object source, rather than trying to link two of them together.

## Feature idea(s) / DoD

Similar to the `vm.neon.tech/usage` annotation, all of the information extracted and used in `api.VmInfo` should be available from annotations & labels on the _pod_, without consulting the VirtualMachine object.

## Implementation ideas

Probably can do something similar to `vm.neon.tech/usage`.

Omrigan:
Simplify scheduler plugin so it no longer watches for VMs, avoiding issues for state synchronization. 

Note for the review: can be reviewed per-commit.

Supersedes #853 

Fixes #863


sharnoff:
Nice. As it's currently implemented, I'm concerned that this will add dependence from neonvm onto the rest of autoscaling that previously didn't exist (i.e. before this PR, any amount of misuse of "autoscaling" [via annotations, labels, whatever] will not affect the VM other than autoscaling not working).

What do you think about separately exposing only the additional information that would be required to get `api.VmInfo` from the annotations? (note: various autoscaling settings will already be present, because labels & annotations are propagated from the VM -> pod)
i.e. add annotation(s) with the data for spec CPUs and memory slots (and beyond that, I don't _think_ there's anything else?)

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -124,6 +137,14 @@ type VirtualMachineSpec struct {
 	EnableSSH *bool `json:"enableSSH,omitempty"`
 }
 
+func (spec *VirtualMachineSpec) Resources() *VirtualMachineResources {
```
sharnoff, line 140: out of curiosity: why return a pointer instead of the object itself?

pkg/api/vminfo.go
```diff
@@ -138,42 +138,57 @@ func (vm VmInfo) NamespacedName() util.NamespacedName {
...
 
 func ExtractVmInfo(logger *zap.Logger, vm *vmapi.VirtualMachine) (*VmInfo, error) {
-	cpuInfo, err := NewVmCpuInfo(vm.Spec.Guest.CPUs)
+	logger = logger.With(util.VMNameFields(vm))
+	return extractVmInfoGeneric(
+		logger,
+		vm.Name,
+		vm,
+		vm.Spec.Resources(),
+	)
```
sharnoff, line 148: should this be collapsed into a single line?

pkg/plugin/plugin.go
```diff
@@ -285,51 +284,11 @@ func (e *AutoscaleEnforcer) Name() string {
 //
 // This function returns nil, nil if the pod is not associated with a NeonVM virtual machine.
 func (e *AutoscaleEnforcer) getVmInfo(logger *zap.Logger, pod *corev1.Pod, action string) (*api.VmInfo, error) {
-	vmName := util.TryPodOwnerVirtualMachine(pod)
-	if vmName == nil {
-		return nil, nil
-	}
```
sharnoff, line 291: Do we still need this check here?

Asking because of the comment above the function:

> This function returns nil, nil if the pod is not associated with a NeonVM virtual machine.

pkg/plugin/plugin.go
```diff
@@ -285,51 +284,11 @@ func (e *AutoscaleEnforcer) Name() string {
...
-
-	vmInfo, err := api.ExtractVmInfo(logger, vm)
+	vmInfo, err := api.ExtractVmInfoFromPod(logger, pod)
 	if err != nil {
 		e.handle.EventRecorder().Eventf(
-			vm,              // regarding
-			pod,             // related
+			pod, // related
+			nil,
```
sharnoff, line 291:
```suggestion
			pod, // regarding
			nil, // related
```

(names in the comment are duplicated from the names of the args to `Eventf`)

sharnoff (commented):
awesome! super clean. A few questions & small things, but otherwise lgtm :)

neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan: I usually write code in a way where all manipulation with structs is done through pointers. This is to not think whether struct is small enough to copy it around without performance impact. Or maybe this is java experience talking deep inside me.

I've changed it here, maybe I'll get used to it eventually.

pkg/plugin/plugin.go
Omrigan: fixed

pkg/plugin/plugin.go
Omrigan: brought it back.

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Haha, fair enough, makes sense :)

sharnoff (approved):
nice! lgtm. Last note: I see you renamed first two commits to include the PR number. If you merge via rebase, the last commit will also need that (it's not added automatically :disappointed:)

PR #866, opened 2024-03-19
replace hardcoded image names with corresponding variables

idrawone:
All the image names have corresponding variables defined at the beginning of the Makefile, but despite this, some hardcoded image names are still being used. This simple patch replaces those hardcoded image names with the corresponding variables for better maintenance.


sharnoff:
Thanks for the PR! IIRC, these are intentional — in order to render the manifests, we modify the files to temporarily set the images to the desired values (which are frequently not the default), then set them back to the "default" ones once rendering is done.

LMK if there's something I'm missing

PR #870, opened 2024-03-21
plugin: Add pod desired AZ to plugin method metrics

### LINKED ISSUE #861: scheduler plugin's failed scheduling metrics should include preferred AZ of the pod, if available

From the current metrics related to scheduling failures (`PostFilter` and `Unreserve` method calls), it's hard to tell the scope of any impact. Being able to see the affected AZs would significantly help with after-the-fact analysis and in-the-moment triage.

## Feature idea(s) / DoD

Relevant scheduler metrics have a label for the availability zone. We _may_ also want to consider node and node group labels. We may also want to survey the other metrics we expose, to see if they'd benefit from these labels as well.

sharnoff:
- Adds `util.PodPreferredAZIfPresent` to determine the AZ
- Adds the `desired_availability_zone` label to the plugin method metrics
- Passes the `*corev1.Pod` through to `IncMethodCall` and `IncFailIfNotSuccess` to inclulde the AZ

Closes #861.

### DIFF
--- a/pkg/util/k8s.go
+++ b/pkg/util/k8s.go
@@ -30,6 +30,52 @@ func PodStartedBefore(p, q *corev1.Pod) bool {
 	return p.Status.StartTime.Before(q.Status.StartTime)
 }
 
+// TryPodPreferredAZ returns the desired availability zone of the Pod, if it has one
+func TryPodPreferredAZ(pod *corev1.Pod) string {
+	if pod.Spec.Affinity == nil || pod.Spec.Affinity.NodeAffinity == nil {
+		return ""
+	}
+
+	affinity := pod.Spec.Affinity.NodeAffinity
+
+	azForTerm := func(term corev1.NodeSelectorTerm) string {
+		for _, expr := range term.MatchExpressions {
+			isAZ := expr.Key == "topology.kubernetes.io/zone" &&
+				expr.Operator == corev1.NodeSelectorOpIn &&
+				len(expr.Values) == 1
+			if isAZ {
+				return expr.Values[0]
+			}
+		}
+
+		return ""
+	}
+
+	// First, check required affinities for AZ:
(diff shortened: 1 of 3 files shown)


pkg/util/k8s.go
```diff
@@ -30,6 +30,47 @@ func PodStartedBefore(p, q *corev1.Pod) bool {
 	return p.Status.StartTime.Before(q.Status.StartTime)
 }
 
+// TryPodPreferredAZ returns the desired availability zone of the Pod, if it has one
+func TryPodPreferredAZ(pod *corev1.Pod) string {
```
shayanh, line 34: The naming here that starts with "try" is a little bit confusing. If you want to emphasize that the preferred AZ might not exist, might it would be better to rename to `PodPreferredAZIfPresent`.

shayanh (approved):
lgtm!

pkg/util/k8s.go
sharnoff: Fair enough, makes sense - done

pkg/util/k8s.go
```diff
@@ -30,6 +30,47 @@ func PodStartedBefore(p, q *corev1.Pod) bool {
...
 
+// PodPreferredAZIfPresent returns the desired availability zone of the Pod, if it has one
+func PodPreferredAZIfPresent(pod *corev1.Pod) string {
+	if pod.Spec.Affinity == nil || pod.Spec.Affinity.NodeAffinity == nil {
+		return ""
+	}
+
+	affinity := pod.Spec.Affinity.NodeAffinity
+
+	azForTerm := func(term corev1.NodeSelectorTerm) string {
```
Omrigan, line 41: Can this function be standalone? If so, I'd prefer that way, it's more natural.

Omrigan (approved):

pkg/util/k8s.go
sharnoff: Fair enough, done

PR #875, opened 2024-03-22
plugin: eventqueue metrics

### LINKED ISSUE #862: scheduler plugin should have metrics for time events spend waiting in the queue

We don't have this internal queue in the scheduler that's critical to correct functioning, but no observability into whether the queue is actually healthy.

## Feature idea(s) / DoD

Scheduler plugin should have metrics exposed for queue wait duration, maybe similar to how neonvm-controller does it.

shayanh:
Scheduler plugin has a set of queue for events (eventqueue) it receives from k8s API server. This PR adds the following set of metrics related to  eventqueue:
- `autoscaling_plugin_eventqueue_depth`: Current depth of eventqueue.
- `autoscaling_plugin_eventqueue_adds_total`: Total number of events added to eventqueue.
- `autoscaling_plugin_eventqueue_duration_seconds`: How long in seconds an item stays in eventqueue before being processed.

Inspired by [controller-runtime workqueue metrics](https://book.kubebuilder.io/reference/metrics-reference).

This is a small PR but might be easier to review it commit by commit.

Fixes #862.

### DIFF
--- a/pkg/plugin/eventqueueset.go
+++ b/pkg/plugin/eventqueueset.go
@@ -1,0 +1,34 @@
+package plugin
+
+import (
+	"context"
+	"hash/fnv"
+
+	"github.com/tychoish/fun/pubsub"
+)
+
+type eventQueueSet[T any] struct {
+	queues []*pubsub.Queue[T]
+}
+
+func newEventQueueSet[T any](size int) eventQueueSet[T] {
+	queues := make([]*pubsub.Queue[T], size)
+	for i := 0; i < size; i += 1 {
+		queues[i] = pubsub.NewUnlimitedQueue[T]()
+	}
+	return eventQueueSet[T]{queues: queues}
+}
+
+func (s eventQueueSet[T]) enqueue(key string, item T) error {
+	hasher := fnv.New64()
+	// nb: Hash guarantees that Write never returns an error
+	_, _ = hasher.Write([]byte(key))
+	hash := hasher.Sum64()
+
+	idx := int(hash % uint64(len(s.queues)))
+	return s.queues[idx].Add(item)
+}
+
+func (s eventQueueSet[T]) wait(ctx context.Context, idx int) (T, error) {
+	return s.queues[idx].Wait(ctx)
+}
--- a/pkg/plugin/eventqueueset.go
+++ b/pkg/plugin/eventqueueset.go
@@ -3,20 +3,30 @@ package plugin
 import (
 	"context"
 	"hash/fnv"
+	"time"
 
 	"github.com/tychoish/fun/pubsub"
 )
 
-type eventQueueSet[T any] struct {
-	queues []*pubsub.Queue[T]
+type queueItem[T any] struct {
+	item    T
+	addTime time.Time
 }
 
-func newEventQueueSet[T any](size int) eventQueueSet[T] {
-	queues := make([]*pubsub.Queue[T], size)
+type eventQueueSet[T any] struct {
+	queues  []*pubsub.Queue[queueItem[T]]
+	metrics PromMetrics
+}
+
+func newEventQueueSet[T any](size int, metrics PromMetrics) eventQueueSet[T] {
+	queues := make([]*pubsub.Queue[queueItem[T]], size)
 	for i := 0; i < size; i += 1 {
-		queues[i] = pubsub.NewUnlimitedQueue[T]()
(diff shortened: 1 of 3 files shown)


pkg/plugin/eventqueueset.go
```diff
@@ -26,9 +36,21 @@ func (s eventQueueSet[T]) enqueue(key string, item T) error {
 	hash := hasher.Sum64()
 
 	idx := int(hash % uint64(len(s.queues)))
-	return s.queues[idx].Add(item)
+
+	s.metrics.eventQueueDepth.Inc()
```
Omrigan, line 40: We have multiple queues here, and there is a potential situation where only one of those gets very long if the problem is caused by one pod/node.  We won't see this if we only have a single metric for all the queues.

Wdyt about exposing 1 metric per queue or a histogram? This increases cardinality, so it is up for debate.

shayanh: I see your point, but and the same time I'm not sure if that's really needed. We have like 64 workers and I'm not sure visualizing them all would be useful at this point. I suggest starting off with this simpler version and then add metrics for each worker if needed.

sharnoff: I think a reasonable middle ground might be to include the worker number in the logs but keep the metrics aggregated. That way, if we see issues, it's still possible to assess the distribution of impact, but we don't end up creating thousands of time series :sweat_smile:

shayanh: We already include pod names in the logs, so if there is a problem with a specific group of pods, it is already visible in the logs. I don't see how logging the queue index could help here. To troubleshot any issue, we have to look for the pod name, not the queue index. Plus, if a particular queue is acting slow, the p99 of latency metrics would rise so we can get notified and see the logs.

sharnoff: The idea behind logging the queue index is that it becomes possible to determine the cause of widespread impact. Idea being that this:

> if a particular queue is acting slow, the p99 of latency metrics would rise so we can get notified and see the logs

works as a start, but doesn't confirm anything because it's very difficult to see "oh, all these events with high latency came through the same queue"

Basically: while the problem may already be visible, the cause may not.

pkg/plugin/eventqueueset.go
```diff
@@ -0,0 +1,56 @@
...
+		addTime: time.Now(),
+	}
+	return s.queues[idx].Add(queueItem)
+}
+
+func (s eventQueueSet[T]) wait(ctx context.Context, idx int) (T, error) {
+	queueItem, err := s.queues[idx].Wait(ctx)
+
+	s.metrics.eventQueueDepth.Dec()
+	s.metrics.eventQueueLatency.Observe(float64(time.Since(queueItem.addTime).Seconds()))
```
sharnoff, line 53: This is only valid if err != nil

shayanh: Right. I wanted to capture the latency even if we got an error, but I forget we don't get the right item in this case. Fixed in 129dd3529d0c53c0531a8b9f84020525e8befe47.

sharnoff: Same thing with decreasing the event queue depth, I think?

shayanh: Fixed.

pkg/plugin/eventqueueset.go
shayanh: > works as a start, but doesn't confirm anything because it's very difficult to see "oh, all these events with high latency came through the same queue"

My point is knowing that all high latency event came through the same queue doesn't help us with debugging. Because whenever we see eventqueue has a high latency, we would look for pods/VMs  with errors in the logs. Knowing the associated queue index here doesn't give us any useful information.

I'm just trying to be conservative with what we log to keep the logs concise and useful, otherwise, it's fine to log the queue index here.

Omrigan (approved):

sharnoff (approved):
I adjusted some wording on the metrics descriptions (see 0a8076ab0ed55a20b922e96e8be78173f58ae39e), but other than that lgtm.

Rebased onto main + squashed + removed the `Signed-off-by` comments from the commits (assuming those were auto-added? feel free to re-add if not) + reworded the commits (prepending `plugin:`) + added the metric names to the commit description.

Looks like the commit dates were clobbered as well, but it seemed like too much effort to restore those.

@Omrigan lmk what you think; if no objections feel free to merge - I think either squash or rebase is fine here.

PR #877, opened 2024-03-25
Update README.md

Omrigan:


sharnoff:
WDYT about moving this info to notion?

PR #887, opened 2024-04-01
neonvm: Support resizeable swap via sized mkswap

### LINKED ISSUE #886: Support resizeable swap

For neondatabase/neon#7239, we'd like to resize swap when compute_ctl receives the compute spec. We'd still like to keep swap on a separate disk, like the current implementation.

The current neonvm swap implementation mounts an entire disk as swap (see #801 for more). This cannot easily be resized by the guest.

## Feature idea(s) / DoD

Swap enabled with `.spec.guest.settings.swap` can be resized from within the VM.

## Implementation ideas

To decrease the size of swap from within the VM, we have a few options:

1. Use a swapfile
2. Mount the swap as a resizeable partition (and then to resize, we `swapoff`, shrink the partition, `mkswap`, and `swapon`)
3. Mount individual disks for swap "chunks" so they can be individually turned off or on

Swapfiles apparently have a performance overhead compared to a "raw" swap disk. We *could* go with the "chunks" method, especially if we wanted to control swap amounts at runtime — but this is quite complex, and AFAIK we don't really need this (at least, not right now).

---

With the partition approach, there's separate issues around initialization — basically: how do we `mkswap` the partition _inside_ the file? I _think_ this is possible by getting the partition offset within the file and `dd`'ing a separate `mkswap`'d file into place, but tbh I'm not entirely sure. It _may_ also be possible with GNU `mkswap` (which supports the `-o`/`--offset` flag), but there doesn't appear to be an easy way to get GNU `mkswap` into alpine, which is neonvm-runner's base image.

We could also initialize from within the VM as well.

sharnoff:
This commit adds support for resizeable swap in NeonVM, with a breaking change to the CRD (changed type of `.guest.settings.swap`) and changes to vm-builder and neonvm-runner to provide the necessary support.

<details>
<summary>Original PR description — now outdated</summary>

<br>

This consists of an opt-in flag to mount the swap as a GPT partition, as well as a convenience script inside the VM to resize the partition and re-enable swap.

## Detailed summary of changes

1. Change type of `.guest.settings.swap` from `resource.Quantity` to a struct, with fields: `size` (`resource.Quantity`), `shrinkable`, `skipSwapon`.

   The new type also accepts the same syntax as `resource.Quantity`, and acts as if only the size was specified. This is temporary, and exists only for backwards compatibility.

2. If `shrinkable = true`, neonvm-runner instead creates a disk with GPT and a single swap partition. It will also include `/neonvm/runtime/resize-swap.sh` as the underlying implementation of `/neonvm/bin/resize-swap`.

   `resize-swap.sh` doesn't depend on any runtime information, but because the swap disk is supplied externally and set up via `/neonvm/runtime/mounts.sh`, my view is that it makes more sense for the management script to be injected by neonvm-runner as well (instead of being added by vm-builder).

3. If `skipSwapon = true`, `/neonvm/runtime/mounts.sh` will not include the typical `swapon`. This is meant for cases where we don't need swap to be available until we later call `/neonvm/bin/resize-swap`, like what's described in neondatabase/neon#7239.

4. Because `resize-swap.sh` requires `parted` for editing the partitions and fetching associated info, we have to add that in vm-builder now. It has some shared libraries (check out `ldd /usr/sbin/parted`), and won't start without them, so those have been added to `/neonvm/lib`.

5. This commit also includes added tests for the various swap parameterizations. These are, briefly:

   - non-resizeable: Check that it's the expected size
   - resizeable: Check that it's initially the expected size, and we can downscale it.
   - resizeable, with `skipSwapon`: Check that it's initially 0, and we can resize it to the desired size (same as above).

   I'm not sure if we want to support non-resizeable swap long-term, or if `skipSwapon` is really necessary, but in the meantime we should definitely have tests.

Closes #886.

See also #801, which initially added swap support.

---

</details>

This consists of changing neonvm-runner to inject a new resize-swap.sh script in /neonvm/runtime, changing vm-builder to add a new resize-swap wrapper in /neonvm/bin, and adding a new flag to skip enabling the swap on startup (for when it'll always be resized later).

## Detailed summary of changes

1. Change type of .guest.settings.swap from resource.Quantity to a struct, with fields: size (resource.Quantity), skipSwapon (bool).

   The new type also accepts the same syntax as resource.Quantity, and acts as if only the size was specified. This is temporary, and exists only for backwards compatibility.

2. If skipSwapon = true, /neonvm/runtime/mounts.sh will not include the typical swapon. This is meant for cases where we don't need swap to be available until we later call /neonvm/bin/resize-swap, like what's described in neondatabase#neon#7239.

3. This commit also includes added tests for swap. These are, briefly:

   - normal: Check that it's initially the expected size, and we can shrink it.
   - with skipSwapon: Check that it's initially 0, and we can resize it to the desired size.

   I'm not sure if skipSwapon is really necessary, but in the meantime
   we should definitely have tests.

We may also, in the future, want to change the underlying implementation of swap resizing (e.g. instead of swapoff-mkswap-swapon, we could offer control at runtime with many smaller swap regions that could be individually enabled or disabled). Those ideas can be discussed separately, if/when they are needed.

Closes #886.

See also #801, which initially added swap support.


neonvm/tools/vm-builder/files/resize-swap.sh
```diff
@@ -0,0 +1,33 @@
...
+elif [ "$#" -eq 1 ]; then
+    size="$1"
+    once=no
+else
+    echo "expected 1 or 2 args, got $#" >&2
+    echo "$USAGE" >&2
+    exit 1
+fi
+
+if [ ! -f /neonvm/runtime/resize-swap.sh ]; then
```
Omrigan, line 24: Why the need for two `resize-swap.sh`?

neonvm/runner/main.go
```diff
@@ -299,6 +301,58 @@ func createISO9660runtime(
...
+			//     /dev/vdd:20971520s:virtblk:512:512:gpt:Virtio Block Device:;
+			//     1:4096s:20969471s:20965376s:linux-swap(v1):primary:;
+			//
+			// So here, we're just running it through the standard unix commands to get the integer
+			// values of the fields we want. (note: 'unit s' means output is measured in sectors)
+			`partinfo="$(parted -m "$swapdisk" unit s print)"`,
+			// in the example above: 512
+			`sector_size="$(echo "$partinfo" | tail -n+2 | head -1 | cut -d: -f4 | grep -oE '[[:digit:]]+')"`,
+			// in the example above: 4096
+			`part_start="$(echo "$partinfo" | tail -n1 | cut -d: -f2 | grep -oE '[[:digit:]]+')"`,
```
Omrigan, line 333: I think this is too complicated of a logic for bash script.

neonvm/runner/main.go
```diff
@@ -359,16 +413,126 @@ func calcDirUsage(dirPath string) (int64, error) {
...
+		}
+		swapRaw, err := os.OpenFile("swap.raw", os.O_RDWR, 0)
+		if err != nil {
+			return fmt.Errorf("failed to open swap.raw: %w", err)
+		}
+		if _, err := swapRaw.Seek(int64(partStart), 0); err != nil {
+			return fmt.Errorf("failed to seek to %d in swap.raw: %w", partStart, err)
+		}
+
+		if _, err := io.CopyN(swapRaw, swapfile, int64(pagesize)); err != nil {
```
Omrigan, line 528: This feels fragile

Omrigan (commented):
Looked through, looks quite complicated.

Are you sure we can't just use the swapfile residing on a main disk? This would drop the need for this patch (and, actually, #801) and the entire thing would be replaced by a `fallocate` and `mkswap` in compute_ctl. Right?

neonvm/runner/main.go
sharnoff: How so? we're basically just getting the Nth fields of a couple lines of output.

Do you have a suggestion for what it should be written in instead?

see also: similar (but not the same) logic implemented in go: https://github.com/neondatabase/autoscaling/blob/eb6fee94c93f57c83c7785f6791113f9d84cc6ad/neonvm/runner/main.go#L462-L484

neonvm/runner/main.go
sharnoff: How do you mean? AFAICT this is consistent with the format for linux swap — although I haven't found documentation for that directly, and instead got this from the implementation of `mkswap`.

Would it help to add some sources justifying this?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -196,9 +196,60 @@ type GuestSettings struct {
...
+	Size resource.Quantity `json:"size"`
+	// Shrinkable changes the swap to sit inside a GPT partition of the mounted disk, so that it can
+	// be resized from within the guest.
+	//
+	// Resizing can be done with the provided program at '/neonvm/bin/resize-swap'. It requires
+	// superuser permissions to perform the resizing. Please note that this DISABLES ALL SWAP before
+	// enabling it at the new size.
+	//
+	// +optional
+	Shrinkable *bool `json:"shrinkable,omitempty"`
```
Omrigan, line 216: Is there a real need to make it not shrinkable?

Omrigan (commented):
Couple of ideas how to make the implementation simpler. Happy to discuss it verbally.

neonvm/runner/main.go
```diff
@@ -359,16 +413,126 @@ func calcDirUsage(dirPath string) (int64, error) {
...
+
+	if swapInfo.Shrinkable == nil || !*swapInfo.Shrinkable {
+		// If we're not making the swap shrinkable, it can be relatively straightforward:
+		// just use the entire virtual disk as swap.
+		if err := execFg("mkswap", "-L", diskName, "swap.raw"); err != nil {
+			return err
+		}
+	} else {
+		// In order to allow shrinking the swap available to a VM without incurring the overhead of swap
+		// files, we create a partition *inside* the qcow2 file.
```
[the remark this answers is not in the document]
sharnoff: Oh, yup good point.

sharnoff:
Discussed with @Omrigan, going to make the following changes:

1. Always make swap resizeable
2. Use `resize-swap.sh` during vminit if SkipSwapon is not true
3. Make the "raw" disk slightly bigger than we need, so that we can specify exact sizes and don't need to parse the output of `parted`. This is _potentially_ more fragile but should be _much_ simpler and easier to understand.

neonvm/runner/main.go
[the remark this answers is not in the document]
sharnoff: 3fe2c287cf46a99a61be454ef9dd42d0fbcbe066

neonvm/runner/main.go
```diff
@@ -299,6 +301,58 @@ func createISO9660runtime(
...
+			`part_start="$(echo "$partinfo" | tail -n1 | cut -d: -f2 | grep -oE '[[:digit:]]+')"`,
+			`new_part_end=$(( part_start + ( new_size / sector_size ) - 1 ))`,
+			// Actually resize partition
+			//
+			// NB: if we don't give 'yes' on stdin with '---pretend-input-tty', it'll refuse to
+			// resize because of the chance of data loss. This is a swap partition that we've
+			// disabled beforehand, so it's ok for us to do this; no data will be lost.
+			//
+			// For more, see: https://unix.stackexchange.com/questions/190317/gnu-parted-resizepart-in-script
+			`echo yes | parted "$swapdisk" ---pretend-input-tty resizepart 1 "$new_part_end"s`,
```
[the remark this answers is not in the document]
sharnoff: With GNU mkswap yes, but not with busybox unfortunately.

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Made it always shrinkable with e77479e9f055aaa0f0a196d48e86be9cd0040aa3

sharnoff:
Ok, looked into this more, and... a couple things:

First: If we want to skip parsing `parted` output in `resize-swap.sh`, we need to know the new end position of the partition. This is equal to `start_position + partition_size - 1 sector`. For some reason. Now, inside the VM it _is_ possible to know (and specify) the sector size, because the disk is emulated — see e.g. here https://www.mail-archive.com/qemu-discuss@nongnu.org/msg06618.html.

But that doesn't help us _outside_ the VM, because there `parted` will use the sector size from the host when creating the GPT table.

Now, in the host we can _in theory_ just set the partition to whatever size we want and get away with it by resizing in the guest, but AFAICT we'll still be stuck with sector sizes from the host, and so we'd still need that information _somehow_. (I wouldn't feel comfortable assuming it's always 512.)

So maybe this is all fine and what we should do is actually still parse the output of `parted` on the host but then use that information in `resize-swap.sh` instead of _also_ running parted there. If we do that though, the plumbing is tricky - currently, `resize-swap.sh` is populated by `createISO9660runtime`, which runs in parallel with swap creation (part of `buildQEMUCmd`).

---

Second: If we _do_ still want to run `parted` in the guest, using `resize-swap.sh` in `/neonvm/runtime/mounts.sh` is a little trickier — if we directly used the size specified in the VM spec, we'd need to make sure that the partition is a little bit bigger. This is also entirely possible, but... yet another thing.


---

So, to recap. AFAICT, our options are:

1. Keep as it is — swap available in the VM is ~3MiB less than what the spec says, and is enabled with plain `swapon`.
2. Use `resize-swap.sh`, but keep parsing `parted` output — swap partition gets some extra space (doesn't matter how much, as long as it's >3MiB) and we `resize-swap.sh` on startup, which ends up parsing `parted` output to determine sector size. Maybe partition start location is fixed. 
3. Parse `parted` on the host to pass sector size and partition start to the guest, so _it_ doesn't need to parse `parted` output.
    - This seems _probably_ ok to me? I'd like to get a +1 on this to agree this should be sound.
4. Instead of parsing `parted` on the host, we just assume that sector size is always 512 bytes.
    - As mentioned above, want to avoid this if possible. But technically it _is_ an option.
5. Instead of resizing a GPT partition, we just make a swapfile in the guest.
    - As mentioned elsewhere, want to avoid this if possible. But technically it is also an option.
6. Instead of resizing a GPT partition, we just split the swap into N chunks, each 1GiB in size, so they can be individually turned on and off to resize.
    - This would be quite complicated, and probably result in an even more complex `resize-swap.sh`, _but_ we don't have to deal with partitions, and we get to resize again over time :shrug:

Also, because `mounts.sh` and `resize-swap.sh` are both injected, we can also start with one solution and change to another later as a follow-up. Switching to/from (6) is probably the hardest, but switching between any of them is entirely feasible.

LMK what you think @Omrigan.

Omrigan:
Option 3 sounds reasonable to me. 

Have you looked into something more convenient than `parted` output? Like https://pkg.go.dev/github.com/diskfs/go-diskfs@v1.4.0/partition/gpt  or sfdisk.

neonvm/tools/vm-builder/files/resize-swap.sh
Omrigan: > The short version is that one is the public interface, guaranteed to be stable, and the other is the implementation, which may change with new neonvm-runner versions.

So `/neonvm/bin/resize-swap` is stable, but `/neonvm/runtime/resize-swap.sh` is implementation-defined?

But anyway we control both scripts, so if the need to extend capability arises, we can change it in a backward-compatible way (or introduce new script). WDYT?

sharnoff:
> Have you looked into something more convenient than `parted` output?

I tried `sgdisk` but found it was quite slow (took ~1s to write a partition).

After looking at the Go package, it seems to still require the user to specify the sector size, so we'd need some way to get that. But otherwise it looks close to what we'd need.

After your comment, I tried `sfdisk` — it calls `sync(2)` after writing the partition table, which... is not necessary in this case, and on my machine takes ~0.3s. No clue what it'd take in prod, but... it's entirely possible we'd have a lot of activity on the kubernetes nodes' root filesystems, which IIUC is what'd be syncing here. Given the timing difference (`parted` is practically instant), I think we should keep with `parted` for configuring (at least in the host), but use `sfdisk` to get the sector size, because yeah, its output is much better.

(it's quite unfortunate actually — `sfdisk`'s input is _also_ much better than `parted`... if only it had a way to skip the `sync()` call)

IIUC `sync()` _should_ be quick in the guest, when referring to a device, so maybe we could use `sfdisk` there. For now, I'm inclined to just stick to configuring with `parted` in both the guest and the host.

cicdteam:
@sharnoff 

Have you tried something like [go-diskfs](https://github.com/diskfs/go-diskfs)?

sharnoff:
I looked into it. One of the difficulties in this PR is that we need to know the sector size when resizing the swap partition inside the guest (`parted` because the "end" position is actually `start + size - sector size`, and `sfdisk` because it uses units of sectors).

AFAICT go-diskfs doesn't provide a way to fetch the recommended sector size, but maybe I missed it.

sharnoff:
todo:

1. merge main into this branch
2. update PR description (probably will keep the old one under a `<details>` because the PR did change a fair amount)

neonvm/tools/vm-builder/files/resize-swap.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo 'not found: swap may not be enabled' >&2
+    exit 1
+fi
+
+/neonvm/bin/sh /neonvm/runtime/resize-swap.sh "$size"
+if [ "$once" = 'yes' ]; then
```
Omrigan, line 30: Why the need for this?

neonvm/runner/main.go
```diff
@@ -359,19 +417,183 @@ func calcDirUsage(dirPath string) (int64, error) {
...
+	// Now that we have sector size, communicate that info to createISO9660runtime, which needs it
+	// to create resize-swap.sh:
+	swapParamsResult <- swapParams{
+		sectorSize:            *sfdiskOut.PartitionTable.SectorSize,
+		partitionStartInBytes: swapStartLocation,
 	}
 
-	if err := execFg(QEMU_IMG_BIN, "create", "-q", "-f", "raw", "swap.raw", fmt.Sprintf("%d", diskSize.Value())); err != nil {
+	// Create the swap. We'll always resize it from within the guest, so we just need it to provide
+	// the label for blkid. Technically we only need to copy the first page (that's where the header
```
Omrigan, line 560: Did you consider using PARTLABEL instead of LABEL?

This would mean using `sudo blkid -t PARTLABEL=main  -o device` to match and `parted name NUMBER NAME` to rename. But no need for mkswap on the host.

neonvm/runner/Dockerfile
```diff
@@ -54,6 +54,8 @@ RUN apk add --no-cache \
     e2fsprogs \
     qemu-system-x86_64 \
     qemu-img \
+    parted \
+	sfdisk \
```
Omrigan, line 58: Alignment issues?

Omrigan (commented):
Getting closer

neonvm/tools/vm-builder/files/resize-swap.sh
sharnoff: In short, so that if compute_ctl restarts it won't disable and re-enable swap — should help in edge cases if e.g. compute_ctl or postgres gets OOM killed, for example.

neonvm/runner/main.go
```diff
@@ -359,19 +417,183 @@ func calcDirUsage(dirPath string) (int64, error) {
 	return size, nil
 }
 
-func createSwap(diskName string, diskPath string, diskSize *resource.Quantity) error {
-	if diskSize == nil {
-		return errors.New("diskSize should be specified")
+const (
+	swapDiskClusterSize string = "2M"
+	// Note: we use an offset of 2MiB here in order to match the cluster_size=2M that we give to
```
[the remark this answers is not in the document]
sharnoff: qcow2 does caching based on subdividing the disk into clusters (size determined by cluster_size here). I haven't done any testing on this, but I _expect_ that aligning the partition start to equal the cluster size is a safe bet, but probably doesn't make much difference in practice (because it's already going to be 1 MiB-aligned). Mostly just there because we have to pick some number, and it might as well equal the cluster size.

Details on qcow2 caching here if you're curious: https://github.com/qemu/qemu/blob/master/docs/qcow2-cache.txt

neonvm/runner/Dockerfile
sharnoff: argh, spaces vs tabs. Thanks.

neonvm/runner/main.go
sharnoff: Implementation changed.

neonvm/runner/main.go
sharnoff: Implementation changed

neonvm/tools/vm-builder/files/resize-swap.sh
Omrigan: I'd rather keep this logic on the `compute_ctl` side. From our point of view we can always do the swap resize, so it is up to the caller to decide whether to do it once or multiple times.

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -196,9 +196,60 @@ type GuestSettings struct {
...
+	// SkipSwapon instructs the VM to *not* run swapon for the swap on startup.
+	//
+	// This is intended to be used in cases where you will *always* resize the swap post-startup,
+	// and don't need it available before that resizing.
+	//
+	// +optional
+	SkipSwapon *bool `json:"skipSwapon,omitempty"`
```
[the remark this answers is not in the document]
Omrigan: Or, suggest `compute_ctl` to always call `resize-swap`. I understand this breaks the current contract of the swap being available by default, but maybe it is fine if we expect `compte_ctl` to start calling `resize-swap` before prod anyway?

neonvm/runner/main.go
```diff
@@ -359,19 +417,145 @@ func calcDirUsage(dirPath string) (int64, error) {
...
+	// qemu-img convert when creating the qcow2 image.
+	// Does it really matter? probably not? But it takes only an additional 1MiB from the
+	// partition, so IMO it's worth the peace of mind.
```
Omrigan, line 425:
```suggestion
	// Align the swap start location with a start of a QCOW2 cluster (2M). 
	// Doesn't matter much, but might improve cache efficiency. 
```

neonvm/runner/main.go
```diff
@@ -359,19 +417,145 @@ func calcDirUsage(dirPath string) (int64, error) {
...
+	swapStartLocationMiB int64 = 2
+	swapAlignMiB         int64 = 2
```
Omrigan, line 427: nit: I'd do this https://go.dev/play/p/-rLyXg4_tWV

neonvm/runner/main.go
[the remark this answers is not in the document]
Omrigan: Thx for explaining, suggested a comment change

neonvm/tools/vm-builder/files/resize-swap.sh
sharnoff: Hm... The thing that makes that tricky is that compute_ctl _generally_ acts as if it's only run once, even though that's not the case. To avoid hitting this, we'd have to either (a) check whether the swap amount is already what we want (easier if we always `DisableSwapon`, so then it's just "is it zero"), or (b) create some file after resizing swap, to record that it's taken place.

Those both feel kind of hacky to me, but maybe they're better than supporting `--once` here? WDYT?

neonvm/runner/main.go
sharnoff: Fair enough, yeah - done

neonvm/apis/neonvm/v1/virtualmachine_types.go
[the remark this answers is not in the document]
sharnoff: So, currently the implementation calls `resize-swap` in `/neonvm/runtime/mounts.sh` iff `!SkipSwapon`. I think I'd like to keep it like that. The added cost of supporting this is small, and breaking existing contracts (especially if they're being used, even if not in prod) is always risky w.r.t. developer productivity. Switching to always skipping swapon would mean that there's no zero-downtime way to change from using the previous swap implementation to the new one (well... I guess technically you could check for the existence of `/neonvm/runtime/resize-swap.sh`, after first updating the compute image, but... :grimacing:)

Also, I think the most reasonable behavior would be to have swap on by default if you ask for it, so keeping `SkipSwapon` appropriately makes the weirder thing (mounting a swap disk but not turning it on) more verbose.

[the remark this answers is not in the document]
Omrigan: Today I learned we might rollout swap before we merge this, so backwards compatible makes sense.
Thanks for explaining!

neonvm/tools/vm-builder/files/resize-swap.sh
Omrigan: I am ok with keeping it as-is, but I'd consider it tech debt if compute_ctl can't determine whether it has restarted or not. Feels like it might have implications in other parts.

But, on the other hand it is also ok if we provide indempotent execution environment for the compute_ctl, if we do it consistently.

neonvm/tools/vm-builder/files/resize-swap.sh
sharnoff: We could in theory, yeah. In practice, maintaining backwards compatibility is hard IMO — I think the contract is more explicit if it's very clearly split across components.

And also, we only currently inject into `/neonvm/runtime` and not `/neonvm/bin`, so either compute_ctl would have to know about `/neonvm/runtime`, which feels like relying on implementation details IMO, or we set up _something_ to link between them (which is relatively close to what we have anyways). And it's easier for that to be a separate script than a symlink, because inserting a new executable file into `/neonvm/runtime` requires fiddling with the file permissions, which we currently get to ignore.

Does that make sense? wdyt?

neonvm/tools/vm-builder/files/resize-swap.sh
sharnoff: > but I'd consider it tech debt if compute_ctl can't determine whether it has restarted or not. Feels like it might have implications in other parts

Yeah 100% - this has caught people by surprise previously. It's kind of an open question what the desired behavior should be though... e.g. if the policy is "restart the VM when compute_ctl exits", then if compute_ctl gets OOM-killed that'll restart the entire VM

> But, on the other hand it is also ok if we provide indempotent execution environment for the compute_ctl, if we do it consistently.

Could you expand on that?

sharnoff:
Just realized busybox mkswap _does_ have a way to specify the size :facepalm:

For a certain size `$size` in bytes, you can just:

```sh
mkswap <path> $(( size / 1024 ))
```

Oh to live in a world where busybox has complete documentation...

---

Anyways I think a lot of this PR can be ripped out and replaced with the previous version - basically exclusively making the changes inside the VM... :disappointed: At least we caught it now, I guess.

I'll try to get that done, so we can do final review on a (hopefully) much smaller PR.

sharnoff:
Ok, opened #899 — it borrows a lot from this PR, but I felt like the approach was different enough that it should be its own PR.

I'm happy to merge it into this PR or merge that one directly into main, whichever you think makes the most sense @Omrigan.

neonvm/tools/vm-builder/files/resize-swap.sh
Omrigan: > Does that make sense?

TBH, still not fully. I don't see the problem neither with runner injecting into `/neonvm/bin`, not with teaching `compute_ctl` to call `/neonvm/runtime/resize-swap`. We have only one implementation of neonvm-runner, so the notion of "implementation-defined" doesn't make much sense. I do acknowledge one file is tied to the `neonvm-runner` version, and the other to the compute image version, but I don't like this distinction, and don't think it should dictate any design choices which have no practical implications.

Probably you see some abstraction layer which I don't see.

Also, if we keep both scripts one should be renamed, e.g.  `/neonvm/runtime/resize-swap-internal.sh`. Every time it takes me some time to figure out which script calls which.

neonvm/tools/vm-builder/files/resize-swap.sh
Omrigan: > > but I'd consider it tech debt if compute_ctl can't determine whether it has restarted or not. Feels like it might have implications in other parts
> 
> Yeah 100% - this has caught people by surprise previously. It's kind of an open question what the desired behavior should be though... e.g. if the policy is "restart the VM when compute_ctl exits", then if compute_ctl gets OOM-killed that'll restart the entire VM

Makes sense, why don't we have that policy yet?

> 
> > But, on the other hand it is also ok if we provide indempotent execution environment for the compute_ctl, if we do it consistently.
> 
> Could you expand on that?

I meant explicitly acknowledging the ability of `compute_ctl` to restart and continue further execution as normal is also a valid contract. Seems redundant though.

Omrigan:
@sharnoff Can you merge #899? It still shares a lot of code and if we started the review here, easier to finish it here IMO

Omrigan (approved):
Pending resolution of comments

PR #890, opened 2024-04-03
agent,billing: Prep for supporting multiple billing clients

sharnoff:
3 loosely-related changes in preparation for neondatabase/cloud#9825. I figured I'd open this as a separate PR so we can reduce the number of changes when we eventually enable the second client. There's definitely other refactoring required as well, but I figured I'd submit these as they're relatively straightforward.

I'd probably recommend reviewing commit-by-commit.

I plan to rebase-and-merge this PR.

cc @Daniel-ef for review on the config changes described in the commit message of a7f3d8714a542885a3192b0f1348a1875b4ea29a — does that seem about right?


pkg/billing/client.go
```diff
@@ -8,27 +8,40 @@ import (
...
-	hostname string
+	URL   string
+	httpc *http.Client
 }
 
-func NewClient(url string, c *http.Client) Client {
-	hostname, err := os.Hostname()
+var hostname atomic.Pointer[string]
+
+func GetHostname() string {
```
Omrigan, line 24: You can also do `init()` function, there will be no need for an atomic

https://www.digitalocean.com/community/tutorials/understanding-init-in-go

pkg/agent/billing/billing.go
```diff
@@ -18,14 +19,26 @@ import (
...
-	PushRequestTimeoutSeconds uint   `json:"pushRequestTimeoutSeconds"`
-	MaxBatchSize              uint   `json:"maxBatchSize"`
+	Clients                ClientsConfig `json:"clients"`
+	CPUMetricName          string        `json:"cpuMetricName"`
+	ActiveTimeMetricName   string        `json:"activeTimeMetricName"`
+	CollectEverySeconds    uint          `json:"collectEverySeconds"`
+	AccumulateEverySeconds uint          `json:"accumulateEverySeconds"`
+}
+
+type ClientsConfig struct {
```
Omrigan, line 29: I'd use a single `ClientConfig` with a type field. If we will need more fields for the S3, that's fine, those fields will be empty for HTTP. This will allow to have multiple clients of the same type.

But the current implementation is also fine.

pkg/billing/client.go
sharnoff: Ohhh right, yeah fair enough. _Generally_ I seem to remember init is frowned upon, but maybe it's ok here :smile:

pkg/agent/billing/billing.go
sharnoff: Yeah, I wanted to do that, but felt like dealing with the JSON unmarshaling and validation was going to be more annoying that way.

pkg/billing/client.go
sharnoff: Switched to init(). I did a brief search and it _does_ seem to be widely frowned upon, but I think this one is simple enough that we can get away with it.

Omrigan: If you plan to rebase, you need to squash that commit elsewhere

Omrigan (approved):
Approved pending `init()`commit squashed.

pkg/billing/client.go
sharnoff: Oh right, good catch! thanks!

PR #891, opened 2024-04-04
plugin: don't fail in Reserve() 

### LINKED ISSUE #869: Change plugin's `Reserve()` method so it cannot fail

This seems counterintuitive, but: If there's issues where the scheduler plugin's internal state does not match the cluster (either because it's delayed, or consistent due to some bug), we can end up performing the correction action at the `Filter` step (i.e. allowing a Pod onto a node) but subsequently incorrectly rejecting it at the `Reserve` step.

When this kind of inconsistency happens, it's often made more severe by the fact that the scheduler framework does not take reserve failures into account when retrying scheduling something — if `Filter` and `Reserve` disagree, the Pod can get stuck repeatedly failing to be scheduled onto the same node, even if there's room elsewhere in the cluster.

In practice, we find we're much more likely to have `Reserve` failures because of bugs in our scheduler plugin (false positives) rather than racy resource acquisition (true positives).

## Feature idea(s) / DoD

Change `Reserve` so that it cannot reject the pod.

## Implementation ideas

This should be as simple as changing the value of a boolean passed to `(*AutoscaleEnforcer).reserveResources()`

This must be discussed internally before implementing — we should make sure that we can continue to detect issues that would have previously caused scheduling failures.

Omrigan:
The only reason why it could fail in Reserve step, while not failing in
Filter step is due to an inconsitency, and we don't want to make any
decisions based on an inconsistent state.

Also, count cases where it would've failed before. This boosts observability even if it Reserve actually won't deny the
request

Fixes #869

### DIFF
--- a/pkg/plugin/prommetrics.go
+++ b/pkg/plugin/prommetrics.go
@@ -25,6 +25,7 @@ type PromMetrics struct {
 	migrationDeletions    *prometheus.CounterVec
 	migrationCreateFails  prometheus.Counter
 	migrationDeleteFails  *prometheus.CounterVec
+	reserveShouldDeny     *prometheus.CounterVec
 }
 
 func (p *AutoscaleEnforcer) makePrometheusRegistry() *prometheus.Registry {
@@ -108,6 +109,13 @@ func (p *AutoscaleEnforcer) makePrometheusRegistry() *prometheus.Registry {
 			},
 			[]string{"phase"},
 		)),
+		reserveShouldDeny: util.RegisterMetric(reg, prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "autoscaling_plugin_reserve_should_deny_total",
+				Help: "Number of times the plugin should deny a reservation",
+			},
+			[]string{"availability_zone", "node", "node_group"},
+		)),
 	}
 
 	return reg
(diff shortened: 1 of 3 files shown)


pkg/plugin/state.go
```diff
@@ -641,6 +641,11 @@ func (e *AutoscaleEnforcer) reserveResources(
 	}
 
 	shouldDeny := add.VCPU > node.remainingReservableCPU() || add.Mem > node.remainingReservableMem()
+
+	if shouldDeny {
+		e.metrics.ReserveShouldDeny(pod, node)
+	}
```
sharnoff, line 647: We might want to make this more specific to just `(*AutoscaleEnforcer).Reserve()` (the plugin method), but for now I think it's ok.

Highest priority thing looking at this was making sure that we log something when we increment this metric, which we do.

pkg/plugin/prommetrics.go
```diff
@@ -124,3 +132,7 @@ func (m *PromMetrics) IncFailIfNotSuccess(method string, pod *corev1.Pod, ignore
 
 	m.pluginCallFails.WithLabelValues(method, util.PodPreferredAZIfPresent(pod), strconv.FormatBool(ignored), status.Code().String())
 }
+
+func (m *PromMetrics) ReserveShouldDeny(pod *corev1.Pod, node *nodeState) {
```
sharnoff, line 136:
```suggestion
func (m *PromMetrics) IncReserveShouldDeny(pod *corev1.Pod, node *nodeState) {
```

To match other methods, making it more clear that this increments the metric

sharnoff (approved):
nice!

PR #892, opened 2024-04-04
agent: Refactor metrics definitions

sharnoff:
Mostly just extracted from #737, thought it would be useful for https://github.com/neondatabase/autoscaling/pull/878#issuecomment-2033387609, and in general nice to have.

We can also rebase #750 on top of this now.

---

Changes are separated into three commits, best reviewed separately. I think they're _probably_ best merged via rebase, but I also don't want to spam the repo history :sweat_smile:

LMK what you think.


Omrigan (approved):
Do I understand correctly, `api.Metrics` will be a subset of `core.Metrics` -> e.g. LFC metrics will be collected by agent, but won't be sent over to scheduler? If so, I'd clarify it this way in a commit message.

Otherwise LGTM.

> best merged via rebase 

I am all for it.

sharnoff:
Updated the message for the second commit, now includes:

> This commit keeps api.Metrics for the second purpose, but makes a copy
> of the same type as core.Metrics so they're no longer linked.
> So, going forward, api.Metrics will be a subset of core.Metrics, because
> the scheduler only uses 1-minute load average, whereas the
> autoscaler-agent uses both 1-minute load average and memory usage.

LMK what you think

sharnoff:
Merging, discussed via DM w/ @Omrigan

PR #894, opened 2024-04-05
Sync versioning in VERSIONING.md and plugin

sharnoff:
Looked at rebasing #750 and realized:

1. The scheduler plugin's range of accepted versions in the code was wider than documented in VERSIONING.md
2. v0.27.0 was not present.

This commit fixes both of those.

To be clear: This commit DISALLOWS VERISONS <v3.0. This should have been done in #743 but must have been missed then.

---

This sort of thing is perhaps good motivation for #888.


Omrigan (approved):
CI fails after .md changes, huh

sharnoff:
Yeah, same thing as the other day, I think - https://github.com/neondatabase/autoscaling/pull/781#issuecomment-2030793190

Even the same CI runner, actually.

PR #895, opened 2024-04-06
agent: Support fetching LFC metrics (but don't use them yet)

sharnoff:
Probably best to review each commit separately.

Most of the commits here are just repositioning to make the final commit, which adds LFC metrics collection, as simple as possible.

~~Builds on #892, must not be merged before then.~~

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -86,9 +86,6 @@ type ScalingConfig struct {
 type MetricsConfig struct {
 	// Port is the port that VMs are expected to provide metrics on
 	Port uint16 `json:"port"`
-	// LoadMetricPrefix is the prefix at the beginning of the load metrics that we use. For
-	// node_exporter, this is "node_", and for vector it's "host_"
-	LoadMetricPrefix string `json:"loadMetricPrefix"`
 	// RequestTimeoutSeconds gives the timeout duration, in seconds, for metrics requests
 	RequestTimeoutSeconds uint `json:"requestTimeoutSeconds"`
 	// SecondsBetweenRequests sets the number of seconds to wait between metrics requests
@@ -186,7 +183,6 @@ func (c *Config) validate() error {
 	erc.Whenf(ec, c.DumpState != nil && c.DumpState.Port == 0, zeroTmpl, ".dumpState.port")
 	erc.Whenf(ec, c.DumpState != nil && c.DumpState.TimeoutSeconds == 0, zeroTmpl, ".dumpState.timeoutSeconds")
 	erc.Whenf(ec, c.Metrics.Port == 0, zeroTmpl, ".metrics.port")
-	erc.Whenf(ec, c.Metrics.LoadMetricPrefix == "", emptyTmpl, ".metrics.loadMetricPrefix")
 	erc.Whenf(ec, c.Metrics.RequestTimeoutSeconds == 0, zeroTmpl, ".metrics.requestTimeoutSeconds")
 	erc.Whenf(ec, c.Metrics.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.secondsBetweenRequests")
 	erc.Whenf(ec, c.Scaling.ComputeUnit.VCPU == 0, zeroTmpl, ".scaling.computeUnit.vCPUs")
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -85,6 +85,7 @@ type ScalingConfig struct {
 // MetricsConfig defines a few parameters for metrics requests to the VM
 type MetricsConfig struct {
 	System MetricsSourceConfig `json:"system"`
+	LFC    MetricsSourceConfig `json:"lfc"`
 }
 
 type MetricsSourceConfig struct {
@@ -195,6 +196,7 @@ func (c *Config) validate() error {
 		erc.Whenf(ec, cfg.SecondsBetweenRequests == 0, zeroTmpl, fmt.Sprintf(".metrics.%s.secondsBetweenRequests", key))
 	}
 	validateMetricsConfig(c.Metrics.System, "system")
+	validateMetricsConfig(c.Metrics.LFC, "lfc")
 	erc.Whenf(ec, c.Scaling.ComputeUnit.VCPU == 0, zeroTmpl, ".scaling.computeUnit.vCPUs")
 	erc.Whenf(ec, c.Scaling.ComputeUnit.Mem == 0, zeroTmpl, ".scaling.computeUnit.mem")
 	erc.Whenf(ec, c.NeonVM.RequestTimeoutSeconds == 0, zeroTmpl, ".scaling.requestTimeoutSeconds")
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -84,7 +84,13 @@ type ScalingConfig struct {
 
 // MetricsConfig defines a few parameters for metrics requests to the VM
 type MetricsConfig struct {
-	// Port is the port that VMs are expected to provide metrics on
+	System MetricsSourceConfig `json:"system"`
+}
+
+type MetricsSourceConfig struct {
+	// Port is the port that VMs are expected to provide the metrics on
+	//
+	// For system metrics, vm-builder installs vector (from vector.dev) to expose them on port 9100.
 	Port uint16 `json:"port"`
 	// RequestTimeoutSeconds gives the timeout duration, in seconds, for metrics requests
 	RequestTimeoutSeconds uint `json:"requestTimeoutSeconds"`
@@ -182,9 +188,13 @@ func (c *Config) validate() error {
 	}
 	erc.Whenf(ec, c.DumpState != nil && c.DumpState.Port == 0, zeroTmpl, ".dumpState.port")
 	erc.Whenf(ec, c.DumpState != nil && c.DumpState.TimeoutSeconds == 0, zeroTmpl, ".dumpState.timeoutSeconds")
-	erc.Whenf(ec, c.Metrics.Port == 0, zeroTmpl, ".metrics.port")
-	erc.Whenf(ec, c.Metrics.RequestTimeoutSeconds == 0, zeroTmpl, ".metrics.requestTimeoutSeconds")
-	erc.Whenf(ec, c.Metrics.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.secondsBetweenRequests")
+
+	validateMetricsConfig := func(cfg MetricsSourceConfig, key string) {
+		erc.Whenf(ec, cfg.Port == 0, zeroTmpl, fmt.Sprintf(".metrics.%s.port", key))
+		erc.Whenf(ec, cfg.RequestTimeoutSeconds == 0, zeroTmpl, fmt.Sprintf(".metrics.%s.requestTimeoutSeconds", key))
+		erc.Whenf(ec, cfg.SecondsBetweenRequests == 0, zeroTmpl, fmt.Sprintf(".metrics.%s.secondsBetweenRequests", key))
+	}
+	validateMetricsConfig(c.Metrics.System, "system")
 	erc.Whenf(ec, c.Scaling.ComputeUnit.VCPU == 0, zeroTmpl, ".scaling.computeUnit.vCPUs")
 	erc.Whenf(ec, c.Scaling.ComputeUnit.Mem == 0, zeroTmpl, ".scaling.computeUnit.mem")
 	erc.Whenf(ec, c.NeonVM.RequestTimeoutSeconds == 0, zeroTmpl, ".scaling.requestTimeoutSeconds")
--- a/pkg/agent/core/metrics.go
+++ b/pkg/agent/core/metrics.go
@@ -26,6 +26,14 @@ func (m SystemMetrics) ToAPI() api.Metrics {
 	}
 }
 
+type LFCMetrics struct {
+	CacheHitsTotal   float64
+	CacheMissesTotal float64
+	CacheWritesTotal float64
+
+	ApproximateWorkingSetSizeTotal float64 // approximate_working_set_size
+}
+
 // FromPrometheus represents metric types that can be parsed from prometheus output.
 type FromPrometheus interface {
 	fromPrometheus(map[string]*promtypes.MetricFamily) error
@@ -85,6 +93,37 @@ func (m *SystemMetrics) fromPrometheus(mfs map[string]*promtypes.MetricFamily) e
 		LoadAverage1Min: load1,
 		// Add an extra 100 MiB to account for kernel memory usage
 		MemoryUsageBytes: memTotal - memAvailable + 100*(1<<20),
+	}
+
+	if err := ec.Resolve(); err != nil {
+		return err
+	}
+
+	*m = tmp
+	return nil
+}
+
+// fromPrometheus implements FromPrometheus, so LFCMetrics can be used with ParseMetrics.
+func (m *LFCMetrics) fromPrometheus(mfs map[string]*promtypes.MetricFamily) error {
+	ec := &erc.Collector{}
+
+	getFloat := func(metricName string) float64 {
+		if mf := mfs[metricName]; mf != nil {
+			f, err := extractFloatGauge(mf)
+			ec.Add(err) // does nothing if err == nil
+			return f
+		} else {
+			ec.Add(missingMetric(metricName))
+			return 0
+		}
+	}
+
+	tmp := LFCMetrics{
+		CacheHitsTotal:   getFloat("lfc_hits"),
+		CacheMissesTotal: getFloat("lfc_misses"),
+		CacheWritesTotal: getFloat("lfc_writes"),
+
+		ApproximateWorkingSetSizeTotal: getFloat("lfc_approximate_working_set_size"),
 	}
 
 	if err := ec.Resolve(); err != nil {
--- a/pkg/agent/core/metrics.go
+++ b/pkg/agent/core/metrics.go
@@ -4,78 +4,93 @@ package core
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
+	"io"
+
+	promtypes "github.com/prometheus/client_model/go"
+	promfmt "github.com/prometheus/common/expfmt"
+	"github.com/tychoish/fun/erc"
 
 	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
 type Metrics struct {
-	LoadAverage1Min  float32
-	MemoryUsageBytes float32
+	LoadAverage1Min  float64
+	MemoryUsageBytes float64
 }
 
 func (m Metrics) ToAPI() api.Metrics {
 	return api.Metrics{
-		LoadAverage1Min:  m.LoadAverage1Min,
+		LoadAverage1Min:  float32(m.LoadAverage1Min),
 		LoadAverage5Min:  nil,
 		MemoryUsageBytes: nil,
 	}
 }
 
-// ReadMetrics generates Metrics from vector.dev's host metrics output, or returns error on failure
-//
-// This function could be more efficient, but realistically it doesn't matter. The size of the
-// output from node_exporter/vector is so small anyways.
-func ReadMetrics(nodeExporterOutput []byte, loadPrefix string) (m Metrics, err error) {
-	lines := strings.Split(string(nodeExporterOutput), "\n")
+// FromPrometheus represents metric types that can be parsed from prometheus output.
+type FromPrometheus interface {
+	fromPrometheus(map[string]*promtypes.MetricFamily) error
+}
 
-	getField := func(linePrefix, dontMatch string) (float32, error) {
-		var line string
-		for _, l := range lines {
-			if strings.HasPrefix(l, linePrefix) && (len(dontMatch) == 0 || !strings.HasPrefix(l, dontMatch)) {
-				line = l
-				break
-			}
-		}
-		if line == "" {
-			return 0, fmt.Errorf("No line in metrics output starting with %q", linePrefix)
-		}
-
-		fields := strings.Fields(line)
-		if len(fields) < 2 {
-			return 0, fmt.Errorf(
-				"Expected >= 2 fields in metrics output for %q. Got %v",
-				linePrefix, len(fields),
-			)
-		}
-
-		v, err := strconv.ParseFloat(fields[1], 32)
-		if err != nil {
-			return 0, fmt.Errorf(
-				"Error parsing %q as float for line starting with %q: %w",
-				fields[1], linePrefix, err,
-			)
-		}
-		return float32(v), nil
-	}
-
-	m.LoadAverage1Min, err = getField(loadPrefix+"load1", loadPrefix+"load15")
+// ParseMetrics reads the prometheus text-format content, parses it, and uses M's implementation of
+// FromPrometheus to populate it before returning.
+func ParseMetrics(content io.Reader, metrics FromPrometheus) error {
+	var parser promfmt.TextParser
+	mfs, err := parser.TextToMetricFamilies(content)
 	if err != nil {
-		return
+		return fmt.Errorf("failed to parse content as prometheus text format: %w", err)
 	}
 
-	availableMem, err := getField(loadPrefix+"memory_available_bytes", "")
-	if err != nil {
-		return
-	}
-	totalMem, err := getField(loadPrefix+"memory_total_bytes", "")
-	if err != nil {
-		return
+	if err := metrics.fromPrometheus(mfs); err != nil {
+		return fmt.Errorf("failed to extract metrics: %w", err)
 	}
 
-	// Add an extra 100 MiB to account for kernel memory usage
-	m.MemoryUsageBytes = totalMem - availableMem + 100*(1<<20)
+	return nil
+}
 
-	return
+func extractFloatGauge(mf *promtypes.MetricFamily) (float64, error) {
+	if mf.GetType() != promtypes.MetricType_GAUGE {
+		return 0, fmt.Errorf("wrong metric type: expected %s but got %s", promtypes.MetricType_GAUGE, mf.GetType())
+	} else if len(mf.Metric) != 1 {
+		return 0, fmt.Errorf("expected 1 metric, found %d", len(mf.Metric))
+	}
+
+	return mf.Metric[0].GetGauge().GetValue(), nil
+}
+
+// Helper function to return an error for a missing metric
+func missingMetric(name string) error {
+	return fmt.Errorf("missing expected metric %s", name)
+}
+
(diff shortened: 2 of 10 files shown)


pkg/api/vminfo.go
```diff
@@ -337,29 +337,77 @@ func (b ResourceBounds) validate(ec *erc.Collector, path string, memSlotSize *re
...
+// defaults but all others remain the same.
+//
+// overrides may be nil; if so, this method just returns defaults.
+func (defaults ScalingConfig) WithOverrides(overrides *ScalingConfig) ScalingConfig {
+	if overrides == nil {
+		return defaults
+	}
+
+	if overrides.LoadAverageFractionTarget != nil {
+		defaults.LoadAverageFractionTarget = &[]float64{*overrides.LoadAverageFractionTarget}[0] // slice trick to shallow copy
```
Omrigan, line 374: Did you mean deep copy?

Also, WDYT about https://go.dev/play/p/wDUnmOqCSqC ?

pkg/agent/runner.go
```diff
@@ -257,12 +257,30 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
 			func(metrics *core.SystemMetrics, withLock func()) {
 				ecwc.Updater().UpdateSystemMetrics(*metrics, withLock)
 			},
 		)
 	})
+	r.spawnBackgroundWorker(ctx, logger, "get LFC metrics", func(c context.Context, l *zap.Logger) {
+		getMetricsLoop(
+			r,
+			c,
+			l,
```
Omrigan, line 271: ctx2, logger2?

Otherwise, `r, c, l` is confusing IMO

pkg/agent/runner.go
```diff
@@ -346,17 +354,24 @@ func (r *Runner) spawnBackgroundWorker(ctx context.Context, logger *zap.Logger,
 // getMetricsLoop repeatedly attempts to fetch metrics from the VM
 //
 // Every time metrics are successfully fetched, the value is recorded with newMetrics.
-func (r *Runner) getMetricsLoop(
+func getMetricsLoop[M core.FromPrometheus[C], C any](
+	r *Runner,
 	ctx context.Context,
 	logger *zap.Logger,
-	newMetrics func(metrics core.Metrics, withLock func()),
+	kind string,
```
Omrigan, line 379: Can you have it as a part of an interface on C? 

Also, C is metrics names, right, maybe should be `Names`?

pkg/agent/runner.go
```diff
@@ -346,17 +354,24 @@ func (r *Runner) spawnBackgroundWorker(ctx context.Context, logger *zap.Logger,
...
+	emptyMetrics func() M,
```
Omrigan, line 380: Can you create empty M, and drop this arg?

Omrigan (commented):
This takes a little bit of time to review for me, as I am not used to this many generics usages. Usually, I would advocate against it in Go, relaying on dynamic dispatch/code duplication. But since we have Rust in the picture, maybe I should get used to it.

LMK if you come up the a way to make it less generic

skyzh:
Discussion from the 1-1 meeting: scraping sql_exporter from the autoscaling-agent is the right way to go. I'll add a separate sql_exporter solely for autoscaling metrics. For now, we can just scrape the current one with a reasonable interval.

pkg/api/vminfo.go
sharnoff: Hm... control plane uses https://github.com/samber/lo's [`lo.ToPtr`](https://pkg.go.dev/github.com/samber/lo#ToPtr) to fulfill this purpose - see e.g. here:

https://github.com/neondatabase/cloud/blob/2b0562e85d4586ad62a8c7af84fbc0af647391f0/goapp/internal/operationsv2/provisioner/provisioner_k8s_common.go#L289-L293

We also define a similar helper function in `pkg/agent/core/state_test.go`:

https://github.com/neondatabase/autoscaling/blob/4d83eb6efe45b08ea8d8a1c93d69013a1a23b549/pkg/agent/core/state_test.go#L234-L236

I'm hesitant to introduce a temporary function for just this location, but I'd be happy to open a follow-up PR to tackle this repo-wide ?

WDYT?

pkg/agent/runner.go
sharnoff: That's fair. All of the other functions in that area use the same naming — I'm hesitant to change these without changing those as well. The reason they're like that currently is because I thought shadowing would be confusing, and it'd be too verbose to write out `ctx2` and `logger2` in all of the usages.

What do you think?

pkg/agent/core/metrics.go
```diff
@@ -118,3 +147,39 @@ func (m *SystemMetrics) fromPrometheus(names SystemMetricNames, mfs map[string]*
 	}
 	return nil
 }
+
+// fromPrometheus implements FromPrometheus, so LFCMetrics can be used with ParseMetrics.
+//
+//nolint:unused // https://github.com/dominikh/go-tools/issues/1294
+func (m *LFCMetrics) fromPrometheus(names LFCMetricNames, mfs map[string]*promtypes.MetricFamily) error {
```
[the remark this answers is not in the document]
sharnoff: Makes sense, done. See (0af52ef29d80836bfbb698e4bc449e86b6cbe550) and (fe8737b19026effd9829abab18e5e8c94b299f6a).

pkg/agent/runner.go
sharnoff: > Can you have it as a part of an interface on C?

Not quite sure what you're referring to here — is it `kind`?

> Also, C is metrics names, right, maybe should be `Names`?

I was originally thinking we'd want some per-VM configuration for metrics for feature-gating new behavior, but upon reflection, I think implementing that properly is actually more effort than it's worth.

At the same time... I think the semantics are clearer if the interface defines it as "config" for metrics parsing, and then something external provides that "config" - even if it's always just the metric names in practice.

For now, I've added some comments to `FromPrometheus` (f47097a984e94f8b704a86f01d9c987e2dd1acd7). LMK what you think.

pkg/agent/runner.go
sharnoff: Unfortunately `M` is basically always a pointer, which makes this difficult.

We _can_ solve the issue by further convoluting the generics — maybe it's worthwhile, maybe it's not. That would look something like:

```diff
diff --git a/pkg/agent/runner.go b/pkg/agent/runner.go
index 52cecb0..5216a3c 100644
--- a/pkg/agent/runner.go
+++ b/pkg/agent/runner.go
@@ -256,7 +256,6 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 			c,
 			l,
 			"system",
-			func() *core.SystemMetrics { return new(core.SystemMetrics) },
 			func() bool { return true },
 			r.global.config.Metrics.System,
 			func(metrics *core.SystemMetrics, withLock func()) {
@@ -270,7 +269,6 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 			c,
 			l,
 			"LFC",
-			func() *core.LFCMetrics { return new(core.LFCMetrics) },
 			func() bool {
 				scalingConfig := r.global.config.Scaling.DefaultConfig.WithOverrides(getVmInfo().Config.ScalingConfig)
 				return *scalingConfig.EnableLFCMetrics // guaranteed non-nil as a required field.
@@ -369,15 +367,19 @@ func (r *Runner) spawnBackgroundWorker(ctx context.Context, logger *zap.Logger,
 	}()
 }
 
+type metricsPointer[C any, V any] interface {
+	core.FromPrometheus[C]
+	~*V
+}
+
 // getMetricsLoop repeatedly attempts to fetch metrics from the VM
 //
 // Every time metrics are successfully fetched, the value is recorded with newMetrics.
-func getMetricsLoop[M core.FromPrometheus[C], C any](
+func getMetricsLoop[M metricsPointer[C, V], C any, V any](
 	r *Runner,
 	ctx context.Context,
 	logger *zap.Logger,
 	kind string,
-	emptyMetrics func() M,
 	isActive func() bool,
 	config MetricsSourceConfig[C],
 	newMetrics func(metrics M, withLock func()),
@@ -418,14 +420,15 @@ func getMetricsLoop[M core.FromPrometheus[C], C any](
 			}
 			lastActive = true
 
-			metrics := emptyMetrics() // dependency-injected because M is usually a pointer, and otherwise the generics get too annoying.
-			err := doMetricsRequest[M, C](r, ctx, logger, timeout, metrics, config.Names)
+			var metrics V
+
+			err := doMetricsRequest[M, C](r, ctx, logger, timeout, M(&metrics), config.Names)
 			if err != nil {
 				logger.Error("Error making metrics request", zap.Error(err))
 				goto next
 			}
 
-			newMetrics(metrics, func() {
+			newMetrics(M(&metrics), func() {
 				logger.Info("Updated metrics", zap.Any("metrics", metrics))
 			})
 		}
```

The shenanigans at the bottom with repeated `M(&metrics)` _could_ get cleaner with https://github.com/samber/lo, because we could instead construct an `M` in a single line :sweat_smile:

pkg/api/vminfo.go
Omrigan: > en a follow-up PR to tackle this repo-wide ?

Sure, when you have the time.

Omrigan (commented):
I think mostly nits are left, we should discuss them verbally

sharnoff:
Discussion with @Omrigan : It's fine to just hard-code the metric names, and that makes it a lot simpler. Some discussion about whether to encapsulate the callbacks for metrics collection into a single struct with `func(...)` fields or an interface.

pkg/agent/runner.go
```diff
@@ -250,10 +250,36 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
+			"LFC",
+			func() *core.LFCMetrics { return new(core.LFCMetrics) },
+			func() bool {
+				scalingConfig := r.global.config.Scaling.DefaultConfig.WithOverrides(getVmInfo().Config.ScalingConfig)
+				return *scalingConfig.EnableLFCMetrics // guaranteed non-nil as a required field.
+			},
+			r.global.config.Metrics.LFC,
+			func(metrics *core.LFCMetrics, withLock func()) {
+				ecwc.Updater().UpdateLFCMetrics(*metrics, withLock)
+			},
```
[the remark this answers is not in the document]
Omrigan: Thought about this a bit, and I guess creating another type would require passing some dependencies to it, which will make it more verbose. 

So defining a struct and passing callbacks through it, so that they are named, and we don't have to guess what :question:  `func() bool { return true }` :question:  does.

pkg/api/vminfo.go
sharnoff: PR here: #946

pkg/api/vminfo.go
```diff
@@ -337,29 +337,77 @@ func (b ResourceBounds) validate(ec *erc.Collector, path string, memSlotSize *re
...
+func (c *ScalingConfig) ValidateOverrides() error {
+	return c.validate(false)
+}
+
+func (c *ScalingConfig) validate(requireAllRequiredFields bool) error {
 	ec := &erc.Collector{}
 
-	// Check c.loadAverageFractionTarget is between 0 and 2. We don't
-	// *strictly* need the upper
+	// Check c.loadAverageFractionTarget is between 0 and 2. We don't *strictly* need the upper
```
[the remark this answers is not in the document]
sharnoff: Fixed in #948 bba199cbc7fa33499285f17effefa8ba1ec30946

pkg/agent/config.go
```diff
@@ -82,13 +83,21 @@ type ScalingConfig struct {
...
+	// Port is the port that VMs are expected to provide the metrics on
+	//
+	// For system metrics, vm-builder installs vector (from vector.dev) to expose them on port 9100.
 	Port uint16 `json:"port"`
-	// LoadMetricPrefix is the prefix at the beginning of the load metrics that we use. For
-	// node_exporter, this is "node_", and for vector it's "host_"
-	LoadMetricPrefix string `json:"loadMetricPrefix"`
+	// Names gives the names of the metrics representing the values we want to fetch
+	//
+	// These are defined externally to avoid having weird constants in the code.
```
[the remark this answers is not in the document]
sharnoff: Resolved this with [`4b06d09`..`0bb2b59`](https://github.com/neondatabase/autoscaling/compare/4b06d09cc3545f5f386cf1d8fe7bd90165ff7a75..0bb2b59f8326a70c6438a1623f98d69ae4c4f34e), as discussed :)

pkg/agent/runner.go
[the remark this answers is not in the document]
sharnoff: Implemented this in 5dc98a95675fcb6573278834a688b123eda89149

pkg/agent/config.go
```diff
@@ -82,13 +83,21 @@ type ScalingConfig struct {
...
 
-// MetricsConfig defines a few parameters for metrics requests to the VM
 type MetricsConfig struct {
-	// Port is the port that VMs are expected to provide metrics on
+	System MetricsSourceConfig[core.SystemMetricNames] `json:"system"`
+	LFC    MetricsSourceConfig[core.LFCMetricNames]    `json:"lfc"`
+}
+
+// MetricsConfig defines a few parameters for metrics requests to the VM
+type MetricsSourceConfig[N any] struct {
```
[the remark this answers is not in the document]
sharnoff: I think this is no longer needed now that metric names have been moved into the code?

Resolving; feel free to reopen if I misjudged

pkg/agent/config.go
```diff
@@ -173,10 +182,14 @@ func (c *Config) validate() error {
...
-	erc.Whenf(ec, c.Metrics.RequestTimeoutSeconds == 0, zeroTmpl, ".metrics.requestTimeoutSeconds")
-	erc.Whenf(ec, c.Metrics.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.secondsBetweenRequests")
+	erc.Whenf(ec, c.Metrics.System.Port == 0, zeroTmpl, ".metrics.system.port")
+	ec.Add(c.Metrics.System.Names.Validate())
+	erc.Whenf(ec, c.Metrics.System.RequestTimeoutSeconds == 0, zeroTmpl, ".metrics.system.requestTimeoutSeconds")
+	erc.Whenf(ec, c.Metrics.System.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.system.secondsBetweenRequests")
+	erc.Whenf(ec, c.Metrics.LFC.Port == 0, zeroTmpl, ".metrics.lfc.port")
+	ec.Add(c.Metrics.LFC.Names.Validate())
+	erc.Whenf(ec, c.Metrics.LFC.RequestTimeoutSeconds == 0, zeroTmpl, ".metrics.lfc.requestTimeoutSeconds")
+	erc.Whenf(ec, c.Metrics.LFC.SecondsBetweenRequests == 0, zeroTmpl, ".metrics.lfc.secondsBetweenRequests")
```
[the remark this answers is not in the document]
sharnoff: Yep - implemented in 86cb9f4a277a5357f95920a82d25b074eef3e60a and c1ae8d3b1ed5c2badcf0d3873b939b6be70137a9 (kept separate so that they can be properly squashed later).

pkg/agent/runner.go
sharnoff: I think 5dc98a95675fcb6573278834a688b123eda89149 satisfies this (in addition to the changes from [`4b06d09`..`0bb2b59`](https://github.com/neondatabase/autoscaling/compare/4b06d09cc3545f5f386cf1d8fe7bd90165ff7a75..0bb2b59f8326a70c6438a1623f98d69ae4c4f34e) ?

sharnoff:
New changes:

1. Move metric names into the code (rather than config) and remove the generics it required. Diff here: [`4b06d09`..`0bb2b59`](https://github.com/neondatabase/autoscaling/compare/4b06d09cc3545f5f386cf1d8fe7bd90165ff7a75..0bb2b59f8326a70c6438a1623f98d69ae4c4f34e).
2. Fixed a bug where we'd always use the "system" metrics port, even for LFC metrics: da48a4cd3f89ef8eaceaf620e7e6faff781aef77 (NOTE: this was from reading the code; I haven't yet tested it)
3. Switch from unnamed callbacks to a new `metricsMgr[M]` struct to contain them and the `kind` string: 5dc98a95675fcb6573278834a688b123eda89149
4. Switch from the `emptyMetrics()` constructor callback to a generics-based approach: 031cc2d13a74e4c232122053404b6119cae04f47
    - cc @Omrigan particularly want your feedback on this one. I could go either way on this -- left it as a separate commit so it's easy to revert.
5. Deduplicate metrics config parsing: 86cb9f4a277a5357f95920a82d25b074eef3e60a + c1ae8d3b1ed5c2badcf0d3873b939b6be70137a9

pkg/api/vminfo.go
```diff
@@ -337,29 +337,92 @@ func (b ResourceBounds) validate(ec *erc.Collector, path string, memSlotSize *re
...
+
+// ValidateOverrides checks that the ScalingConfig is safe to use to override preexisting settings.
+//
+// This is less strict than ValidateDefaults, because with ValidateOverrides even required fields
+// are optional.
+func (c *ScalingConfig) ValidateOverrides() error {
+	return c.validate(false)
+}
+
+func (c *ScalingConfig) validate(requireAllRequiredFields bool) error {
```
Omrigan, line 403:
```suggestion
func (c *ScalingConfig) validate(requireAll bool) error {
```

pkg/api/vminfo.go
```diff
@@ -337,29 +337,92 @@ func (b ResourceBounds) validate(ec *erc.Collector, path string, memSlotSize *re
...
+	}
+
+	if overrides.LoadAverageFractionTarget != nil {
+		defaults.LoadAverageFractionTarget = &[]float64{*overrides.LoadAverageFractionTarget}[0] // slice trick to shallow copy
+	}
+	if overrides.MemoryUsageFractionTarget != nil {
+		defaults.MemoryUsageFractionTarget = &[]float64{*overrides.MemoryUsageFractionTarget}[0]
+	}
+	if overrides.EnableLFCMetrics != nil {
+		defaults.EnableLFCMetrics = &[]bool{*overrides.EnableLFCMetrics}[0]
```
Omrigan, line 380: Shall we use `lo.ToPtr`?

pkg/agent/runner.go
```diff
@@ -343,20 +371,56 @@ func (r *Runner) spawnBackgroundWorker(ctx context.Context, logger *zap.Logger,
...
+	// newMetrics is a callback to update the internal state with new values for these metrics.
+	newMetrics func(metrics M, withLock func())
+}
+
+// metricsPtr wraps core.FromPrometheus with the requirement that the interface be implemented on a
+// pointer type.
+//
+// We use this in getMetricsLoop to construct new values of type M that point to a zero value
+// (rather than being nil pointers themselves).
+type metricsPtr[V any] interface {
```
Omrigan, line 393: > cc @Omrigan particularly want your feedback on this one. I could go either way on this -- left it as a separate commit so it's easy to revert.

Yeah, I'd revert it. It was educational - I learned `~` is a thing in golang :laughing:  But for real-life usage it'd stick to the trivial approach.

pkg/agent/runner.go
```diff
@@ -365,17 +429,28 @@ func (r *Runner) getMetricsLoop(
...
-		})
+			metrics := M(new(V)) // construct M as a pointer to the zero value. This is what V is for.
+			err := doMetricsRequest(r, ctx, logger, metrics, config)
+			if err != nil {
+				logger.Error("Error making metrics request", zap.Error(err))
+				goto next
+			}
+
+			mgr.newMetrics(metrics, func() {
+				logger.Info("Updated metrics", zap.Any("metrics", metrics))
```
Omrigan, line 451: I know this is a prior change, but why this callback exists, and why is named `withLock`? Wouldn't just adding this `logger.Info` wherever `withLock` is called serve the same purpose, only simpler?

pkg/agent/core/metrics.go
```diff
@@ -4,78 +4,143 @@ package core
...
+			ec.Add(missingMetric(metricName))
+			return 0
 		}
-		return float32(v), nil
 	}
 
-	m.LoadAverage1Min, err = getField(loadPrefix+"load1", loadPrefix+"load15")
-	if err != nil {
-		return
+	names := struct {
```
Omrigan, line 91: I'd just have 4 constants, but up to you.

pkg/agent/core/metrics.go
```diff
@@ -4,78 +4,143 @@ package core
...
-				line = l
-				break
-			}
-		}
-		if line == "" {
-			return 0, fmt.Errorf("No line in metrics output starting with %q", linePrefix)
-		}
+// It is parameterized by the config 'C' passed to the parsing function, to allow things like metric
+// names to be defined externally, rather than compile-time constants.
+type FromPrometheus interface {
```
Omrigan, line 41: WDYT about naming it `Metrics`, and using as a type parameter constraint instead of any? You mention "metric types" in the comment anyway, and this would help readability.

Also, two bottom lines are no longer true.

pkg/agent/runner.go
sharnoff: Haha fair enough- reverted in 60f835b5fc21bc867d8b3a466f1896a22d44f6dc

pkg/agent/core/metrics.go
sharnoff: > Also, two bottom lines are no longer true.

Ah good catch, fixed in 662613d170363fa9cf240e5a0225e0c2f1c09674.

> and using as a type parameter constraint instead of any

Not quite sure I follow, could you elaborate?

pkg/agent/runner.go
```diff
@@ -343,20 +371,56 @@ func (r *Runner) spawnBackgroundWorker(ctx context.Context, logger *zap.Logger,
...
+	// It's either "system" or "LFC".
+	kind string
+
+	// isActive returns whether these metrics should currently be collected for the VM.
+	//
+	// For example, with LFC metrics, we return false if they are not enabled for the VM.
+	isActive func() bool
+
+	// newMetrics is a callback to update the internal state with new values for these metrics.
+	newMetrics func(metrics M, withLock func())
```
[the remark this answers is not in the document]
sharnoff: Fair, yeah. Fixed in cb14dafe0dcefdb61bec02e365c9626af81b2cfa + b9e3ebd1ea51a7d3730b965d3ecd77c0318776f4 (split up to make squashing easier)

pkg/agent/runner.go
sharnoff: Added this as a new commit (to be kept separate after rebasing) in efdda7c819bbfe1a5178b884d9803438b5cebba1.

pkg/api/vminfo.go
sharnoff: Ah yup, meant to do this - thanks for the reminder :) Implemented in #948 f9ed5d27408b77231d1e77d6338c6f248f939970

pkg/api/vminfo.go
sharnoff: Fair enough. Implemented in #948 d77b97c5103c968bcee08496e8746daf9ae59027

sharnoff:
Continuing with the changelog approach, if only because it helps me manage the rebasing:

1. Reverted 031cc2d13a74e4c232122053404b6119cae04f47 ("tmp: Switch from emptyMetrics() callback to generics"): 60f835b5fc21bc867d8b3a466f1896a22d44f6dc
2. Squashed everything back down: [`60f835b`..`666e11f`](https://github.com/neondatabase/autoscaling/compare/60f835b5fc21bc867d8b3a466f1896a22d44f6dc..666e11f02bae311c0ac57dbe129e42f1487262db)
3. Rebased onto main: 666e11f02bae311c0ac57dbe129e42f1487262db -> 34afd60dc06f5db3551ae1893054716a65cf5b2e
4. Renamed `newMetrics` -> `updateMetrics`: cb14dafe0dcefdb61bec02e365c9626af81b2cfa+b9e3ebd1ea51a7d3730b965d3ecd77c0318776f4
5. Fix outdated comment on `FromPrometheus`: 662613d170363fa9cf240e5a0225e0c2f1c09674
6. Renamed the `c`/`l` context/logger variables to `ctx2`/`logger2`: efdda7c819bbfe1a5178b884d9803438b5cebba1

Plus a couple more changes to #948 based on review here.

pkg/agent/runner.go
sharnoff: I think I did it at the time because it felt like a cleaner separation of concerns -- the caller is doing the logging anyways, so the executor stuff only needs to be concerned with doing the updating. 

Looking at it now, I'm not sure either way, and hesitant to change it. Happy to discuss more though.

pkg/agent/core/metrics.go
sharnoff: Reworked this so we're the strings are directly passed to `getFloat`, similar to how LFC metrics parsing is now implemented, see 66b92893a804fc0f17b84f7c67358d9610573cb1.

pkg/agent/runner.go
Omrigan: > it's easier if that doesn't also have logging

It doesn't have logging, but if have to pass the callback, so not much profit :upside_down_face: 

> I'm not sure either way, and hesitant to change it.

No problem, that's was before the current change.

Omrigan (approved):
LGTM, pending comments and conflicts resolution.

sharnoff:
Final changes:

1. Squash previous commits. Diff here (empty): [`66b9289`..`20bd2ba`](https://github.com/neondatabase/autoscaling/compare/66b92893a804fc0f17b84f7c67358d9610573cb1..20bd2ba55a40f0873971bb19d19d4ba8b9631704)
2. Rebase onto main (dropping #948 from this branch). Diff here: [`20bd2ba`..`fb2f910`](https://github.com/neondatabase/autoscaling/compare/20bd2ba55a40f0873971bb19d19d4ba8b9631704..fb2f9104ad9220f8d685f80d99554a1a480050bc)
3. Fix compile error from rebase: 689e18d0d66b935f9c0ffba6cae5df2e4f8f8306
4. Change `metricsMgr[M any]` -> `metricsMgr[M core.FromPrometheus]`: d0991a9d2447ff1c2a37e0cf2241422df51dea52

Waiting on some issues with latest release. Will hold off on merging for now to avoid accidentally adding much larger changes alongside a small fix.

pkg/agent/core/metrics.go
sharnoff: Hm, that's fair. In rust, it's generally an anti-pattern to restrict generic types more than they need to be, but I'm not sure if it's the same in Go.

I like the idea of using it to signal purpose. Changed from `M any` to `M FromPrometheus` in d0991a9d2447ff1c2a37e0cf2241422df51dea52

PR #898, opened 2024-04-09
agent: push metrics to S3

Omrigan:
Add s3 billing destination. 

Part of a neondatabase/cloud#9825


pkg/billing/client.go
```diff
@@ -35,16 +35,135 @@ func GetHostname() string {
...
+}
+
+type S3Client struct {
+	bucket, prefixInBucket string
+	client                 *s3.Client
+	now                    func() time.Time
+}
+
+func NewS3Client(
+	bucket, region, prefixInBucket, endpoint string,
```
sharnoff, line 92: style nit: separate args should have separate types — IMO this is an anti-pattern that makes the code harder to read and changes more likely to interfere with each other.

pkg/billing/client.go
```diff
@@ -35,16 +35,135 @@ func GetHostname() string {
...
+
+	return nil
+}
+
+func (c HTTPClient) LogFields() zap.Field {
+	return zap.String("url", c.URL)
+}
+
+type S3Client struct {
+	bucket, prefixInBucket string
```
sharnoff, line 86: same style nit here

pkg/billing/client.go
sharnoff: At the same time:

This looks like a symptom of the underlying issue that there's a bunch of string args in a row with little protection against providing them in the wrong order.

Maybe `pkg/billing` should expose a `S3ClientConfig` that contains these as fields? It could even be directly embedded in `pkg/agent/billing.S3ClientConfig` and passed in, so the args don't need to be rewritten in multiple places.

pkg/billing/client.go
```diff
@@ -35,16 +35,135 @@ func GetHostname() string {
...
+	return S3Client{
+		bucket:         bucket,
+		prefixInBucket: prefixInBucket,
+		client:         client,
+		now:            now,
+	}, nil
+}
+
+func (c S3Client) key() string {
+	// Example: year=2021/month=01/day=26/hh:mm:ssZ_{autoscaler_agent_id}.ndjson.gz
```
sharnoff, line 115: nit: the example says `autoscaler_agent_id` but below, we use a random UUID(ish)

sharnoff (commented):
Initial review, mostly nits. Haven't reviewed the python. There's also some subtlety around client errors that I need to dig more deeply into.

I think it's also missing s3 client config validation `pkg/agent/config.go`.

In general, looks quite good! Feels sparse on comments — I'm left without any hints for building my mental context.

pkg/agent/billing/billing.go
```diff
@@ -87,17 +93,34 @@ func RunBillingMetricsCollector(
...
 			name:   "http",
 			config: c.BaseClientConfig,
 		})
 	}
-
-	logger := parentLogger.Named("billing")
+	if c := conf.Clients.S3; c != nil {
+		client, err := billing.NewS3Client(c.S3ClientConfig, time.Now)
+		if err != nil {
+			logger.Panic("Failed to create S3 client", zap.Error(err))
```
sharnoff, line 110: This panic makes me a little nervous — I'm not sure how a panic here would be handled, given that it's a separate goroutine. I _think_ it'd take down the entire process?

What do you think about moving the `go RunBillingMetricsCollector` from `pkg/agent/entrypoint.go` into here, with a separate `StartBillingMetricsCollector()` that synchronously sets up the clients, so it can return an error if that fails?

pkg/agent/config.go
```diff
@@ -163,14 +163,27 @@ func (c *Config) validate() error {
 		zeroTmpl  = "field %q cannot be zero"
 	)
 
+	validateBaseBillingConfig := func(cfg *billing.BaseClientConfig, key string) {
+		erc.Whenf(ec, cfg.PushEverySeconds == 0, zeroTmpl, key+".pushEverySeconds")
+		erc.Whenf(ec, cfg.PushRequestTimeoutSeconds == 0, zeroTmpl, ".pushRequestTimeoutSeconds")
+		erc.Whenf(ec, cfg.MaxBatchSize == 0, zeroTmpl, ".maxBatchSize")
```
sharnoff, line 169: 1. IIRC it's generally better to use `fmt.Sprint[f]` than string addition, because the fmt methods use an internal buffer pool whereas plain addition doesn't
2. Looks like the second two are missing the `key+`?

Otherwise, I think this is a clean way to do it, within the existing configuration mess we have :sweat_smile:, +1

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
...
+	Bucket         string `json:"bucket"`
+	Region         string `json:"region"`
+	PrefixInBucket string `json:"prefixInBucket"`
+	Endpoint       string `json:"endpoint"`
+}
+
+type S3Client struct {
+	cfg    S3ClientConfig
+	client *s3.Client
+	now    func() time.Time
```
sharnoff, line 95: Does `now()` need to be dependency-injected? If no, can we get rid of it?

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
...
+}
+
+func NewS3Client(cfg S3ClientConfig, now func() time.Time) (S3Client, error) {
+	s3Config, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion(cfg.Region))
```
sharnoff, line 99: Any idea what the context is used for here? Generally good to avoid allowing things to run indefinitely.

sharnoff, line 99: WDYT about importing `github.com/aws/aws-sdk-go-v2/config` as e.g. `awsconfig` ? I found it surprising that s3-specific stuff was available just by `config`

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
...
+	return S3Client{
+		cfg:    cfg,
+		client: client,
+		now:    now,
+	}, nil
+}
+
+func (c S3Client) key() string {
+	// Example: year=2021/month=01/day=26/hh:mm:ssZ_{uuid}.ndjson.gz
+	now := c.now()
```
sharnoff, line 119: When reading `(*S3Client).Send()` it wasn't clear to me that it uses the current time in generating the key. WDYT about generating the time in `Send` and passing it in here?

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
...
+
+func (c s3LogFields) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddString("bucket", c.cfg.Bucket)
+	enc.AddString("prefixInBucket", c.cfg.PrefixInBucket)
+	return nil
+}
+
+func (c S3Client) LogFields() zap.Field {
+	return zap.Inline(s3LogFields{c})
+}
```
sharnoff, line 142: Maybe https://pkg.go.dev/go.uber.org/zap@v1.27.0/zapcore#ObjectMarshalerFunc could make this simpler? (i.e. implementing it all inside `LogFields()` instead of introducing a one-off type for ti)

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
...
+
+func (c S3Client) Send(ctx context.Context, payload []byte, traceID TraceID) error {
+	key := c.key()
+	buf := bytes.Buffer{}
+
+	gzW := gzip.NewWriter(&buf)
+	_, err := gzW.Write(payload)
+	_ = gzW.Close()
```
sharnoff, line 150: this would be a good place for a couple comments. Questions I have reading this:

1. Why do we need to gzip the payload? (_maybe_ I can intuit this if I've read `key()` and happened to notice the `.gz` at the end, but it's certainly not obvious)
2. Why do we ignore the error from `gzW.Close()`? (and actually, _can_ we get an error there, if we're just writing to a buffer?)
3. Typically people `defer thing.Close()` - why aren't we doing that with `gzW.Close()`?


And also: Why is this happening inside `(*S3Client).Send()` where we're double-buffering? Would it make more sense for `billing.Send` to do the gzip encoding at the same time as marshaling the JSON? (_can_ we do that?)

pkg/billing/client.go
```diff
@@ -35,16 +35,138 @@ func GetHostname() string {
 	return hostname
 }
 
-func NewClient(url string, c *http.Client) Client {
-	return Client{URL: fmt.Sprintf("%s/usage_events", url), httpc: c}
+type Client interface {
+	Send(ctx context.Context, payload []byte, traceID TraceID) error
```
sharnoff, line 39: IIUC: Generally the `Send` methods are not intended to be used externally, and exist only so that `billing.Send` can be generic.

WDYT about making it private? (and adding a comment noting that it's just the implementation-specific part of `Send[E]()`)

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+import subprocess
+import json
+import os
+import sys
+import boto3
+import time
+
+
+def update_agent_biling_clients(billing):
```
sharnoff, line 11: Could use a comment that answers the following:

1. What does this function do?
2. What is `billing`?
3. What communication is made in order to accomplish this?

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+            "configmap",
+            "autoscaler-agent-config",
+            "-n",
+            "kube-system",
+            "-o",
+            "json",
+        ]
+    )
+    config = json.loads(config)
+    print(config)
```
sharnoff, line 25: Leftover from debugging? or intentional

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+    agent_config_json = config["data"]["config.json"]
+    agent_config = json.loads(agent_config_json)
+    agent_config["billing"] = billing
+    agent_config_json = json.dumps(agent_config)
+    config["data"]["config.json"] = agent_config_json
```
sharnoff, line 30:
```suggestion
    # Extract existing config JSON
    agent_config_json = config["data"]["config.json"]
    agent_config = json.loads(agent_config_json)
    # Set the billing config and write back into the ConfigMap
    agent_config["billing"] = billing
    agent_config_json = json.dumps(agent_config)
    config["data"]["config.json"] = agent_config_json
```

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+import subprocess
+import json
+import os
+import sys
+import boto3
+import time
+
+
+def update_agent_biling_clients(billing):
+    config = subprocess.check_output(
```
sharnoff, line 12: My python's a little rusty — will `subprocess.check_output` raise an exception if the command exits with a non-zero exit code?

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+            "autoscaler-agent-config",
+            "-n",
+            "kube-system",
+            "--type",
+            "merge",
+            "-p",
+            json.dumps(config),
+        ]
+    )
+    assert result == 0
```
sharnoff, line 46: `assert` only gets run in debug mode — should this instead be an explicit check & `raise Exception(...)` ?

(and I guess similar things for asserts lower down)

tests/e2e/s3/utils.py
sharnoff, line 11: From reading `agent_remove_s3`, seems like this function is mis-named - IIUC it currently updates all of the billing config, but it's called `update_agent_billing_clients`. IMO just updating the billing clients will allow the calling code to be simpler and more readable

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+def agent_restart():
+    os.system("kubectl rollout restart daemonset autoscaler-agent -n kube-system")
+
+
+def s3_create_bucket(local_endpoint):
+    s3 = boto3.client("s3", endpoint_url=local_endpoint)
+    s3.create_bucket(Bucket="metrics")
+
+
+def s3_check_file(local_endpoint):
```
sharnoff, line 93: Could use a comment explaining

1. What this does
2. What it checks

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+        [
+            "idempotency_key",
+            "metric",
+            "type",
+            "endpoint_id",
+            "start_time",
+            "stop_time",
+            "value",
+        ]
+    ), events[0].keys()
```
sharnoff, line 148: IMO: we're unlikely to mess up the JSON fields. We _may_ be likely to mess up the values we send - it'd be good to check those (at least, make sure that the start/stop times are self-consistent, and the values make sense for those start/stop times)

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+                    "%s:9000" % self.port,
+                    "-n",
+                    self.namespace,
+                ]
+            )
+
+            time.sleep(1)
+
+            if self.process.poll() is None:
+                break
```
sharnoff, line 172: What does this logic do?

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+            self.port += 1
```
sharnoff, line 174: This seems... sketchy? Why are we iterating through a bunch of ports? IMO: Better to just pick a 5-digit random number.

Would there be any chance of leaking `kubectl port-forward` processes like this?

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,219 @@
...
+    forward = KubctlForward()
+
+    action = sys.argv[1]
+    if action == "step2_init":
+        step2_init()
+    elif action == "step2_assert":
+        step2_assert()
+    elif action == "restore":
+        agent_remove_s3()
+        agent_restart()
```
sharnoff, line 219: WDYT about creating a `def main()` at the top of the file, and then calling it from here? No clue if that's idiomatic, but I often find it's helpful because it declares the relationships up front

tests/e2e/s3/02-init.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - script: |
+      set -eux
+      export AWS_ACCESS_KEY_ID=minio
+      export AWS_SECRET_ACCESS_KEY=minio123
+      kubectl set env daemonset autoscaler-agent -n kube-system AWS_ACCESS_KEY_ID=minio
+      kubectl set env daemonset autoscaler-agent -n kube-system AWS_SECRET_ACCESS_KEY=minio123
+      
+      ../.venv/bin/python3 utils.py step2_init        
+      ../.venv/bin/python3 utils.py step2_assert
+      ../.venv/bin/python3 utils.py restore
```
sharnoff, line 14: Is there a reason not to have `step2_assert` in a separate kuttl assertion step? I assumed that would naturally give the autoscaler-agent more time to generate the billing events.

sharnoff (commented):
More thorough review now :sweat_smile:

---

re: subtle error handling — if the error is not a `JSONError` or `UnexpectedStatusCodeError`, we use the "root" error as a label in the metrics. See here:

https://github.com/neondatabase/autoscaling/blob/4d83eb6efe45b08ea8d8a1c93d69013a1a23b549/pkg/agent/billing/send.go#L147-L157

If we're not careful, that can result in timeseries explosion.

I haven't checked if the S3 API will return us weird base errors. What I would recommend is that for now, we use some catch-all error type until we've seen some examples, and then maybe switch to using the base error in metrics at some later date. (i.e. for now, any error from the S3 client should show up in the metrics like "S3 error", or something)

pkg/billing/client.go
Omrigan: Don't know here, and the function's own docstring uses it with TODO().

But overall, yes, I am all for using context whereever possible, if the API provides this option. Here we actually have the background context, so fixed in 13c5934 :)

pkg/billing/client.go
Omrigan: >  it wasn't clear to me that it uses the current time in generating the key

Maybe that's alright? 

I've renamed the function to generateKey (80dcd8e), thus on the callsite should be more clear key is indeed generated. And what exactly is used to generate it can be understood by looking at the body.

pkg/billing/client.go
Omrigan: Rewrote a bit:

1. Added link
2. Rewrote to handle it. Although, probably it is impossible in practice, but less 
3. Added comment

> Why is this happening inside (*S3Client).Send() where we're double-buffering? 

Because json is common for http and s3, but gzip is specific to S3. Also, typical json.Marshal outputs `[]byte`, so it is not practical to pass `io.Reader`.

Anyway, I hope billing jsons won't be the thing we OOM because of

pkg/billing/client.go
Omrigan: Indeed it does, thanks! Fixed in d6659f1

pkg/agent/config.go
Omrigan: Fixed, thx.

I've actually deliberated for quite some time how to best change this code. The conclusion I came to is: ideally, each object should have it's own `Validate()`, but this will make printing the full key path in the error message harder, so I left it for future.

pkg/billing/client.go
sharnoff: Hm, in that case I'd recommend either adding the unit tests in this PR, or leaving `now` out for now.

pkg/agent/billing/billing.go
```diff
@@ -81,23 +87,40 @@ type vmMetricsSeconds struct {
 }
 
 func RunBillingMetricsCollector(
-	backgroundCtx context.Context,
+	ctx context.Context,
```
sharnoff, line 90: Curious about this rename- why `backgroundCtx` -> `ctx` ?

I see that this commit (13c5934add9d24cc72d9615b722ebb3d7d4925dc) changes `billing.NewS3Client` to use the context, but IIUC the semantics associated with the context haven't changed

pkg/billing/client.go
sharnoff: Makes sense, I think that's a good solution

pkg/billing/client.go
```diff
@@ -35,16 +35,140 @@ func GetHostname() string {
...
+	)
+	return fmt.Sprintf("%s/%s", c.cfg.PrefixInBucket, filename)
+}
+
+func (c S3Client) LogFields() zap.Field {
+	return zap.Inline(zapcore.ObjectMarshalerFunc(func(enc zapcore.ObjectEncoder) error {
+		enc.AddString("bucket", c.cfg.Bucket)
+		enc.AddString("prefixInBucket", c.cfg.PrefixInBucket)
+		enc.AddString("region", c.cfg.Region)
+		enc.AddString("endpoint", c.cfg.Endpoint)
```
sharnoff, line 135: Noticed that the change to move this inline (d6659f1e8f6a840bcba06379d4cf111a1a336555) added these two fields (`region` and `endpoint`) - was that intentional? if so, feel free to ignore :)

pkg/billing/client.go
sharnoff: Nice!

pkg/billing/client.go
sharnoff: Hm... fwiw, you _can_ construct JSON via the `json.Encoder` interface, which takes an `io.Writer`: https://pkg.go.dev/encoding/json#Encoder

I think the control flow might be quite annoying if we fully went this way though...

> Anyway, I hope billing jsons won't be the thing we OOM because of

Haha yeah, exactly.

Anyways, lgtm, fair enough.

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,269 @@
...
+                "service/minio-service",
+                "%s:9000" % self.port,
+                "-n",
+                self.namespace,
+            ]
+        )
+
+        # Check if the process is still running after 1 second.  
+        # If the port is already in use, it would have exited.
+        time.sleep(1)
```
sharnoff, line 217: Anecdotally, I've seen port-forwarding take a while to set up — I'd be worried about this `sleep(1)` introducing some flakiness to our e2e tests.

If you haven't seen any issues with it though, then rate of occurrence is probably low enough that it's ok for now.

tests/e2e/s3/utils.py
sharnoff: Why is updating the other fields required? Is it just that otherwise the test will take too long?

(if so: maybe it'd be preferable to keep the same settings we use in staging/prod, even if the test takes longer — wdyt?)

tests/e2e/s3/utils.py
sharnoff: Hm, I think it's ok enough for now — but a good follow-up would be more exhaustive checking of all of the contents. Basically thinking that we should check that every value we receive makes sense (i.e. timestamps line up right, metric values make sense for the timestamps).

WDYT?

tests/e2e/s3/utils.py
```diff
@@ -0,0 +1,269 @@
...
+    elif action == "step2_assert":
+        step2_assert()
+    elif action == "restore":
+        agent_remove_s3()
+        agent_restart()
+    else:
+        raise Exception("Unknown action: %s" % action)
+
+if __name__ == "__main__":
+    main()
```
sharnoff, line 269: NB: missing newline at the end of the file

tests/e2e/s3/02-init.yaml
```diff
@@ -0,0 +1,21 @@
...
+      kubectl set env daemonset autoscaler-agent -n kube-system AWS_ACCESS_KEY_ID=minio
+      kubectl set env daemonset autoscaler-agent -n kube-system AWS_SECRET_ACCESS_KEY=minio123
+      
+      ../.venv/bin/python3 utils.py step2_init
+
+      # The reason we put the assert here as well is that a separate
+      # TestAssert step has different semantics:
+      #  1. It has a different home directory (root of the repo)
+      #  2. If the assert fails, the the assert step keeps retrying
+      #     until it succeeds or the test times out
```
sharnoff, line 18: Maybe I'm missing something, but _presumably_ we'll be waiting for billing data to come in at some point, right? Wouldn't we want to retry that until the data is available?

tests/e2e/s3/utils.py
Omrigan: I've added a schema validation in 3a2a87c

tests/e2e/s3/utils.py
Omrigan: >  Is it just that otherwise the test will take too long?

Yes, that was my reasoning.

> even if the test takes longer — wdyt?

I think, it is important for developer productivity to have tests which run as fast as possible. I don't see a case where keeping the default period values helps us catch something we wouldn't have caught otherwise.

sharnoff: The idea behind keeping the default period would be that we can test the values are actually what we expect. There's rounding that happens, so in this case the correctness is impacted by what period is used.

At the same time: if we're not testing it the actual values, we don't get any benefit from using the default period. Happy to leave that for a follow-up.

tests/e2e/s3/utils.py
sharnoff: Nice :)

pkg/agent/billing/billing.go
sharnoff: It was `backgroundCtx` because it's only used so that the billing subsystem can be notified when the autoscaler-agent is going to exit soon (so we can send any events that are waiting around).

Its direct usage is in the collector thread:

https://github.com/neondatabase/autoscaling/blob/3a2a87ce3609448023cb2421136fb84bcf67ec9a/pkg/agent/billing/billing.go#L187-L188

which in turn causes `(eventSender).senderLoop` to finish up:

https://github.com/neondatabase/autoscaling/blob/3a2a87ce3609448023cb2421136fb84bcf67ec9a/pkg/agent/billing/send.go#L60-L62

Looking at it now though, `senderLoop` should probably just take a context...

pkg/agent/billing/billing.go
sharnoff: panics are orthogonal to the errgroup stuff right? Either way, we can discuss next week

tests/e2e/s3/02-init.yaml
sharnoff: Hm, I see... fair enough. I guess any changes here can be left for a follow-up where we check correctness of the values we receive.

pkg/billing/client.go
```diff
@@ -35,16 +35,150 @@ func GetHostname() string {
...
+
+func (e S3Error) Unwrap() error {
+	return e.Err
+}
+
+func NewS3Client(ctx context.Context, cfg S3ClientConfig) (S3Client, error) {
+	s3Config, err := awsconfig.LoadDefaultConfig(ctx, awsconfig.WithRegion(cfg.Region))
+
+	if err != nil {
+		return S3Client{}, S3Error{Err: err} //nolint:exhaustruct // error is returned
```
sharnoff, line 113: 1. the error message for `S3Error` mentions making the request, presumably this shouldn't be wrapped?
2. May be worthwhile to just return `*S3Client` so you can return `nil`. Up to you

Omrigan: Fixed in a6a3d3b

sharnoff (approved):
nice!

PR #899, opened 2024-04-10
neonvm: Support resizeable swap via sized mkswap

### LINKED ISSUE #886: Support resizeable swap

sharnoff:
Resizeable swap, take 2. Supersedes #887.

This commit adds support for resizeable swap in NeonVM, with a breaking change to the CRD (changed type of .guest.settings.swap) and changes to vm-builder and neonvm-runner to provide the necessary support.

This consists of changing neonvm-runner to inject a new resize-swap.sh script in /neonvm/runtime, changing vm-builder to add a new resize-swap wrapper in /neonvm/bin, and adding a new flag to skip enabling the swap on startup (for when it'll always be resized later).

## Detailed summary of changes

1. Change type of .guest.settings.swap from resource.Quantity to a struct, with fields: size (resource.Quantity), skipSwapon (bool).

   The new type also accepts the same syntax as resource.Quantity, and acts as if only the size was specified. This is temporary, and exists only for backwards compatibility.

2. If skipSwapon = true, /neonvm/runtime/mounts.sh will not include the typical swpaon. THis is meant for cases where we don't need swap to be available until we later call /neonvm/bin/resize-swap, like what's described in neondatabase#neon#7239.

3. This commit also includes added tests for swap. These are, briefly:

   - normal: Check that it's initially the expected size, and we can shrink it.
   - with skipSwapon: Check that it's initially 0, and we can resize it to the desired size.

   I'm not sure if skipSwapon is really necessary, but in the meantime
   we should definitely have tests.

We may also, in the future, want to change the underyling implementation of swap resizing (e.g. instead of swapoff-mkswap-swapon, we could offer control at runtime with many smaller swap region sthat could be individually enabled or disabled). Those ideas can be discussed separately, if/when they are needed.

Closes #886.

See also #801, which initially added swap support.


sharnoff:
Changes have been merged into #887, will continue there.

Issue #840, opened 2024-03-01
Remove scheduler plugin "buffer" resources


sharnoff:
## Problem description / Motivation

In order to prevent accidental overcommitting, on startup the scheduler plugin has a measure of "uncertainty" for each VM's usage that is resolved only when the autoscaler-agent makes a request to the scheduler plugin to inform it of its intentions.

This has two issues:

1. Forced to choose between "unavailable" and "inaccurate", we have chosen "unavailable". In practice this is much worse / higher risk than inaccuracies.
2. There's a short (~5s) period of unavailability immediately after startup. When we add replicas & leader election for the scheduler, this unavailability could be frequent enough to cause liveness issues for the scheduler (i.e. it may be unable to schedule for extended periods of time)

## Feature idea(s) / DoD

Scheduler plugin scheduling uncertainty should not cause unavailability

## Implementation ideas

Instead of keeping "buffer" resources, we should just _entirely_ remove it, and be willing to make inaccurate scheduling decisions. Worst-case scenario is that we accidentally overcommit by a little bit — in practice, real resource usage in our clusters is much lower than reserved resources, so we have wiggle room.

sharnoff:
It's an interesting idea... I think _in theory_ we could do this, but in practice:

* Delaying scheduling via the _plugin_ is risky, because Pods may take a long time to be rescheduled if we fail them (and in general, we'd rather make sub-optimal scheduling decisions than completely block scheduling).
* Delaying scheduling by waiting to return from plugin creation (like in #904) would not work, I think — IIRC the scheduler is only marked "ready" once everything's running, and the autoscaler-agents only communicate with it once it's "ready", so the plugin initialization would be waiting for requests that never come
  * That said, we could make modifications to the conditions for communicating, or the readiness checks, and maybe this would work

For this issue, I'm currently thinking the easiest way forward is to move the autoscaler-agent <-> scheduler plugin protocol into annotations on the VMs, which would:

1. Guarantee that all information is available just by reading the cluster state (so: no waiting for comms — basically, if we do it right, "buffer" would always be zero because everything is known)
2. Make replicas / leader election simple, because the autoscaler-agents wouldn't be communicating with any particular scheduler instance (see #841)

What do you think?

Omrigan:
> For this issue, I'm currently thinking the easiest way forward is to move the autoscaler-agent <-> scheduler plugin protocol into annotations on the VMs, which would:
> 
[...]

Issue #517, opened 2023-09-04
Epic: Enable overcommit


vadim2404:
## Motivation

Look at the image:

| Resource | Requested | Used |
| --------- | ------------ | ------ |
| CPU | 121 | 6 |
| RAM | 457 | 53 |
 
![image](https://github.com/neondatabase/autoscaling/assets/1708025/a1cf273a-42ec-4951-b033-f4c1c81928f8)

Let's have at least 2x overcommit (because of the maximum number of pods that can be up and running)

## Implementation ideas

TODO

## Tasks

```[tasklist]
- [ ] ...
- [ ] List tasks as they're created for this Epic
```

## Other related tasks, Epics, and links

-

sharnoff:
Here's some design ideas I've considered. I believe the 3rd solution is probably the best one.

## 1. Implement exclusively within k8s scheduler plugin

**Idea:** Change the scheduler plugin so that it actually allows resource usage from 

**Good parts:**
- Only requires changes in one component; other components don't need to know about it

**Bad parts:**
- If usage from regular pods fills up a node, the base scheduler + cluster-autoscaler will still interpret that correctly.
- May have strange consequences with cluster-autoscaler being overly willing to scale up, because its internal scheduler simulation is less permissive than our plugin
- Setting would be global, increasing risk of volatility during rollout or revert

## 2. Implement within neonvm-controller/neonvm-runner

**Idea:** When a VM is set to a certain amount of memory and/or CPU, we _actually_ set QEMU to use some fixed multiple of that

**Good parts:**
- Has no effect on scheduler or cluster-autoscaler

**Bad parts:**
- Regular pods unaffected
- Counterintuitive (e.g. "why does setting it to 3 CPUs actually give it 6?")
- Will have knock-on effects on autoscaler-agent's scaling; it must be made aware of this global overcommit factor (even though it's not possible to atomically deploy changes to multiple components)
- Requires restart or migration of all VMs in order to get updated neonvm-runner pods
- Setting would be global, increasing risk of volatility during rollout or revert

## 3. Implement via "overcommit" factor per VM

**Idea:** Add a new VM setting determining an "overcommit" factor that both the scheduler and cluster-autoscaler respect. Kind of a combination of ideas 1 and 2, but per-VM rather than global.

**Good parts:**
- Rollout is gradual, does not require restarting any VMs

**Bad parts:**
- Blocked on #590
- Regular pods unaffected
[...]

Issue #873, opened 2024-03-22
Get rid of VM watcher in plugin


Omrigan:
Followup of #863

Need to move the logic into pod watcher. Also get rid of `Relist()` API

Omrigan:
I thought a bit how this would fit with neondatabase/cloud#11730. Perhaps, it should be the other way around: scheduler should operate on VMs only, so that after we get rid of pods, scheduler can still schedule.

@sharnoff WDYT?

sharnoff:
This will be closed by #1163 / #995 ("stateless scheduler")

sharnoff:
Closing, as #1163 was merged.

PR #904, opened 2024-04-15
plugin: Replace readClusterState with existing watch events

sharnoff:
... and a few other commits tossed in for fun.

Probably best to review commit-by-commit.

The thing I'm most worried about is that this potentially may change buffer handling.

_Might_ be a good idea to go for #840 first, but I suspect that might also have its own issues.

This is a precursor to #517, to minimize the number of places we need to change resource handling.


sharnoff:
> Do you mean these watch events? 

Broadly, watch events _like_ those - but it only relies on the Pod watch events, I think. (and specifically, this PR waits for only the initial Pod "add" events to be completely processed)

> I am trying to understand how this relates to #873.

Right, yeah fair. This PR should be ok - we only care about the Pod watch events. (unless I missed something - please feel free to double-check me no that)

pkg/plugin/state.go
```diff
@@ -728,16 +738,33 @@ func (e *AutoscaleEnforcer) reserveResources(
 	}
 	newNodeReservedCPU := node.cpu.Reserved + ps.cpu.Reserved
 	newNodeReservedMem := node.mem.Reserved + ps.mem.Reserved
+	newNodeBufferCPU := node.cpu.Buffer + ps.cpu.Buffer
```
Omrigan, line 741: This looks like `node.*.Buffer` is only increasing. Or am I missing something?

pkg/plugin/state.go
```diff
@@ -686,20 +686,30 @@ func (e *AutoscaleEnforcer) reserveResources(
...
-			Buffer:           0,
+			Reserved:         vmInfo.Max().Mem,
+			Buffer:           util.SaturatingSub(vmInfo.Max().Mem, vmInfo.Using().Mem),
 			CapacityPressure: 0,
 			Min:              vmInfo.Min().Mem,
 			Max:              vmInfo.Max().Mem,
 		}
+
+		migrating := util.TryPodOwnerVirtualMachineMigration(pod) != nil
+		if !vmInfo.Config.ScalingEnabled || migrating {
```
Omrigan, line 707: When Buffer values are zeroed if ScalingEnabled=true? Also, these two lines had a comment before, maybe bring it along?

pkg/plugin/plugin.go
```diff
@@ -118,15 +122,50 @@ func makeAutoscaleEnforcerPlugin(
...
+	//
+	// 1. Using watch.WatchModeSync to force queueing events *before* returning from creating the
+	//    watcher (note: and therefore, before any start to be handled); and
+	// 2. For each event created from the initial List(), increment a counter to track the number of
+	//    these events, and decrement it as events are handled.
+	//
+	// The initial state building is complete when the counter reaches zero, at which point we close
+	// the channel that this function will wait on.
+
+	var initEventsCount atomic.Int32
```
Omrigan, line 140: Can you use [semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore#NewWeighted) here instead? 
 - Acquire(nil, n) when events are added
 - Release(1) when event is processed
 - Acquire(ctxWithTimeout, n) when waiting for events processing

pkg/plugin/plugin.go
```diff
@@ -118,15 +122,50 @@ func makeAutoscaleEnforcerPlugin(
...
 	nwc := nodeWatchCallbacks{
 		submitNodeDeletion: func(logger *zap.Logger, nodeName string) {
 			pushToQueue(logger, nodeName, func() { p.handleNodeDeletion(hlogger, nodeName) })
 		},
 	}
 	pwc := podWatchCallbacks{
-		submitStarted: func(logger *zap.Logger, pod *corev1.Pod) {
-			pushToQueue(logger, pod.Name, func() { p.handleStarted(hlogger, pod) })
+		submitStarted: func(logger *zap.Logger, pod *corev1.Pod, preexisting bool) {
+			incEventCountIfInit(preexisting)
```
Omrigan, line 164: Why passing bool instead of an if here?

pkg/plugin/config.go
```diff
@@ -30,6 +30,14 @@ type Config struct {
 	// event queue.
 	EventQueueWorkers int `json:"eventQueueWorkers"`
 
+	// StartupEventHandlingTimeoutSeconds gives the maximum duration, in seconds, that we are
+	// allowed to wait to finish handling all of the initial events generated by reading the cluster
+	// state on startup.
+	//
+	// If event processing takes longer than this time, then plugin creation will fail, and the
+	// scheduler pod will retry.
+	StartupEventHandlingTimeoutSeconds int `json:"startupEventHandlingTimeoutSeconds"`
```
Omrigan, line 39: What does this timeout protects us against? 

If we have a lot of events on the start, maybe it is better to anyway process all of them? Otherwise, next initialization attempt will fail as well and plugin cannot start up. 

We can also not fail plugin's pod, and just keep on with inaccurate scheduling.

Omrigan (commented):
Have some questions

pkg/plugin/config.go
sharnoff: The intent is that this should protect us against deadlocks, so that if event processing doesn't make enough process, we just restart.

pkg/plugin/state.go
sharnoff: During reserve, yeah, that's correct.

pkg/plugin/state.go
sharnoff: > When Buffer values are zeroed if ScalingEnabled=true?

Not sure what you're asking, could you rephrase?

> Also, these two lines had a comment before, maybe bring it along?

Ah good catch, thanks! Fixed in b3f2d70a764418b44ce502b81e496cf1b5084322

pkg/plugin/plugin.go
sharnoff: Seems simpler. Implemented it in caba29f33339c53e094935e66239ee375d01b26a but tbh I'm not sure if it's better. WDYT?

pkg/plugin/plugin.go
sharnoff: Initially it was because I thought there'd be more than one callback that would be involved here, and that would keep callback functions small.

It's now clear that's not necessary though, so moved the if here in 564ec88212cc6571e3620737a1d48450850b2174

pkg/plugin/config.go
Omrigan: And how does it help? In case of a restart it will just keep trying to process the same events over and over again. 

But, if the event processing is not deadlocked, but just slow, it will take more time, but eventually the situation gets fixed.

Maybe instead of a restart it is better to fire an alert?

pkg/plugin/plugin.go
Omrigan: If the `incEventCount` is left as a semaphore wrapper anyway, maybe instead revert to your old implementation, but wrap all in a struct synchronization primitive?

The reason I suggested semaphore is that I am wary of using raw channels and atomics in the code, where it feels like it should be a synchronization primitive.

pkg/plugin/state.go
Omrigan: > > When Buffer values are zeroed if ScalingEnabled=true?
> 
> Not sure what you're asking, could you rephrase?

If ScalingEnabled=true, then buffer values are non-zero. How do they become zero? When we hear the report from the autoscaler-agent?

pkg/plugin/state.go
Omrigan: How do they decrease? I guess the answer is the same as for the above comment.

pkg/plugin/plugin.go
sharnoff: Made it a separate type in 521597e44affe70a50cccd1e0bf72d46fa500394 - LMK what you think.

pkg/plugin/state.go
sharnoff: Correct, yeah - when handling the first autoscaler-agent request.

pkg/plugin/config.go
sharnoff: I don't think we currently have a way for components to directly trigger an alert, at least not without metrics getting scraped. If the scheduler is down, someone will be paged either way - and we already have some alerts for if it's crashlooping.

I expect that restarting _may_ help against deadlocks because I _think_ it's more likely for us to have inconsistent deadlocks than it is for initial event processsing to just be really slow. But that's just a guess.

pkg/plugin/plugin.go
Omrigan: I imagined the implementation with initEventsCount hidden inside the new helper.

Also, you don't need incEventCount now

pkg/plugin/config.go
Omrigan: > I don't think we currently have a way for components to directly trigger an alert

I meant, expose a metric startup_event_handling_seconds, and configure an alert when it is >30s.

> I expect that restarting may help against deadlocks because I think it's more likely for us to have inconsistent deadlocks than it is for initial event processsing to just be really slow. But that's just a guess.

If we were to have an alert:
 - If it is a deadlock, then by logs it should be possible to see if event processing is progressing or not, and a scheduler can be restarted manually
 - If it is slow, then the issue improves on it's own. 

Also, deadlock means a bug in the code, and writing specific protections for bugs seems like a grey path (why not just fix the bug?). 

But I guess we don't actually expect any of the issues to be happening regularly, and even we observe too many events, the mitigation is easy -> bump the timeout. So, leaving this up to you at this point.

Omrigan (approved):

pkg/plugin/plugin.go
sharnoff: > I imagined the implementation with initEventsCount hidden inside the new helper.

Hm, it's hard to make that kind of abstraction safe to use (because the counter isn't monotonic) -- maybe it'd be fine here, but I'd rather keep the logic inlined than make an unsafe abstraction for just this purpose.

So in its current state, incEventCount or equivalent is still needed (we can just directly add to the atomic; don't know whether that's better)

pkg/plugin/config.go
sharnoff: > expose a metric startup_event_handling_seconds

The tricky thing with this kind of metric is that we want it to keep updating even while we're waiting. AFAICT that's non-trivial to implement, so will keep everything as is for now, but I'd be happy to revisit this at some point in the future.

sharnoff:
Realized that the necessity of debe3b22e89af228af757094cad22bb26f316b80 (from #936) is partially due to 25bbe43ce8163d0290cbfb210c3b1a2eb13cb866 here — it adds Buffer always, but there's a handful of circumstances where we don't actually need it (like when a new VM is added but doesn't go through our scheduler)

sharnoff:
Adapted debe3b22e89af228af757094cad22bb26f316b80 here as c8e250e4ebc0130e001cdd5f196af26ad209ffe7 to fix that issue. Impact should have been close to zero in practice, but wanted to fix it just to be safe. @Omrigan wdyt?

pkg/plugin/state.go
```diff
@@ -585,7 +585,18 @@ func (e *AutoscaleEnforcer) handleStarted(logger *zap.Logger, pod *corev1.Pod) {
 
 	logger.Info("Handling Pod start event")
 
-	_, _, _ = e.reserveResources(context.TODO(), logger, pod, "Pod started", false)
+	_, _, _ = e.reserveResources(context.TODO(), logger, pod, "Pod started", reserveOptions{
+		// pod already started, out of our control - we don't have a mechanism to deny it
+		allowDeny: false,
+		// this may be a preexisting VM. If so, we should include it in "buffer" as long it's
+		// supposed to be handled by us (otherwise, the "buffer" will never be resolved)
+		includeBuffer: pod.Spec.SchedulerName == e.state.conf.SchedulerName,
```
Omrigan, line 593: Why do we even have SchedulerName as a part of the spec, why do we ever want it to be different?

Omrigan (approved):

Omrigan:
Are you going to do rebase merge? If so, tmp commit needs to be squashed somewhere, maybe into middle commit.

pkg/plugin/state.go
sharnoff: Result of discussing this: it's part of the pod spec because neonvm is a logically separate dependency of "autoscaling", so the ability to run with/without the rest of autoscaling requires the ability to set the pod's scheduler name

PR #911, opened 2024-04-16
neonvm/api: Fix SwapInfo-related backwards compatibility issues

sharnoff:
In short, the PR switching swap to *SwapInfo (#887) added checks for backwards compatibility, but an older client operating with newer NeonVM could still break because after reconciling the controller would replace the *resource.Quantity for swap with a struct, and then because the type is different from expected, the older client would fail to fetch the VM if it tried to do that in the future.

_In addition to that_, the CRD change prevented creating VMs with the old, string-based swap field.

ref https://neondb.slack.com/archives/C033A2WE6BZ/p1713290569648829


Omrigan (approved):

sharnoff:
Closing, because this approach actually wasn't enough.

See #912 for more.

PR #912, opened 2024-04-17
neonvm: Make .guest.settings.swap *resource.Quantity, add swapV2

sharnoff:
There were a handful of issues with the approach from #887. It changed the type of `.guest.settings.swap`. This commit changes it back, and makes a new field for the new type.

Under certain circumstances, changing the type of the field *could have* worked, but even with with fully inter-compatible serializing (see #911) the CRD prevents a type's field from being either a string or an object, which is what's required for full compatibility.

We could have done this if there were no VMs using swap, but we unfortunately have a small number of VirtualMachines in production with swap enabled.

So we're left with the "hard" route - adding a new field with a different type (`swapV2`), special-casing object updates to allow switching from the old type to the new one (but only if the values are the same), and then *eventually* once all VirtualMachine objects are using the new field, switch the type of the old one and move back to the old field with the new type.

See https://neondb.slack.com/archives/C06SW383C79/p1713298689471319.
Part of neondatabase/cloud#12047.


neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -172,6 +187,26 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) error {
 		}
 	}
 
+	// validate swap changes by comparing the SwapInfo for each.
+	//
+	// If there's an error with the old object, but NOT an error with the new one, we'll allow the
+	// new one to proceed. This is to prevent any VirtualMachine objects getting stuck during
```
Omrigan, line 193: Can you clarify why the objects would get stuck if this logic wouldn't exist?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -196,9 +197,55 @@ type GuestSettings struct {
...
+	// SwapV2 is a temporary newer version of the Swap field.
+	//
+	// Eventually, after all VMs have moved from Swap to SwapV2, we can change the type of the Swap
+	// field to SwapInfo, move VMs from SwapV2 back to Swap, and then remove SwapV2.
+	//
+	// More information here: https://neondb.slack.com/archives/C06SW383C79/p1713298689471319
+	//
 	// +optional
-	Swap *SwapInfo `json:"swap,omitempty"`
+	SwapV2 *SwapInfo `json:"swapV2,omitempty"`
```
Omrigan, line 217: Can be named `SwapInfo`, not SwapV2, so we can keep it indefinitely

Omrigan (approved):
One alternative approach would be to just delete (or ignore) `Swap`, and add Swap2 as a usual field, meaning that for some time new computes would be started without swap if it is specified without swap. This is simpler and lower chance of issues, but won't allow us to test swap until cplane has support for the new format. I suppose that's not ideal, but I imagine we won't rollout the swap support until this changes to cplane anyway.

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Good idea. Renamed.

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
sharnoff: Done — tl;dr the webhook can reject potential object updates. If we refuse to accept changes because the _old_ state is bad, then it isn't possible to fix any objects if they end up broken.

sharnoff:
From discussion: Ideally we _could_ simply ignore/delete/replace `Swap`, but this isn't currently an option because we have a small number of computes in prod that are using it.

Omrigan (approved):

PR #917, opened 2024-04-21
Improvement template fixups

sharnoff:
Noticed while making some issues.

Not sure:

1. Whether to squash or rebase
2. If the commit subjects should be prefixed with `improvement template:` or something, like we have for everything else.

Curious for thoughts.


Omrigan:
> Noticed while making some issues.
> 
> Not sure:
> 
>     1. Whether to squash or rebase

I usually prefer rebase whenever possible 

> 
>     2. If the commit subjects should be prefixed with `improvement template:` or something, like we have for everything else.

`github:`?

> 
> 
> Curious for thoughts.

Omrigan (approved):

Issue #918, opened 2024-04-21
neonvm-controller reconcile failure metrics should differentiate k8s conflicts


sharnoff:
## Problem description / Motivation

When we see reconcile failures in the metrics, we currently have no way (other than via log aggregation) of determining whether they're due to kubernetes object update conflicts or something more severe.

Basically: We expect some base rate of update conflicts. We don't expect some base rate of other failures, but failing to distinguish these in the metrics makes it difficult to notice when something's wrong.

## Feature idea(s) / DoD

It should be possible to, from metrics alone, see the rate of reconcile failures over time, excluding update conflicts.

## Implementation ideas

A few options come to mind:

1. Completely ignore conflicts in the metrics
2. Use a separate metric for conflicts
3. Add a label within the existing reconcile failures metric

Issue #916, opened 2024-04-21
neonvm-controller logs don't show reconcile duration


sharnoff:
## Problem description / Motivation

Currently, we have various neonvm-controller metrics showing the duration of reconcile operations. However, if there's e.g. a sudden spike in long reconciles, we don't have a way to either (a) identify that a particular reconcile took a long time, or (b) filter the logs to look only at reconciles above a certain duration.

Lacking this ability is particularly impactful if a small number of reconcile operations are creating a disproportionate impact by running for _much_ longer than the median operation — the chances of finding the logs of an offending reconcile operation are low.

This came up in investigation here: https://neondb.slack.com/archives/C03TN5G758R/p1713734101217499?thread_ts=1713730859.492929

## Feature idea(s) / DoD

1. When looking at a reconcile operation, it should be possible to see how long it took
2. It should be possible to query to find the reconcile operations that took beyond a certain duration

Fetching e.g. _all_ the logs for sufficiently long reconcile operations is AFAICT not possible with our log querying service, because earlier log lines would need to know the final duration.

It would _also_ be nice if the final result of the reconcile operation were included in any log message mentioning the duration — that way, we could filter for reconcile operations with certain durations, matching certain error messages, for example.

## Implementation ideas

_Probably_ just need to track the time and insert/modify a log line somewhere?

Omrigan:
Made changes to "Overview" and "neonvm-controller" dashboards to separate errors vs conflicts. 

@sharnoff, please, take look.

sharnoff:
Looks good! Made a couple changes to the Overview dashboard; feel free to ping with questions.

To be clear, that closes #918 and is unrelated to this issue, right?

Omrigan:
> To be clear, that closes #918 and is unrelated to this issue, right?

Ah, right, thanks.

PR #920, opened 2024-04-25
neonvm: metrics for failing reconcile

Omrigan:
The need for this appeared during the investigation of an elevated level of reconciliation failures after neonvm-controllers were moved to an another node group.

Adds a single metric `reconcile_duration_seconds` with an `outcome` label

\+ relevant log messages

Part of a #916 and #918. Easier to do together because of conflicts.

### DIFF
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -19,6 +19,9 @@ type ReconcilerMetrics struct {
 	runnerCreationToVMRunningTime  prometheus.Histogram
 	vmCreationToVMRunningTime      prometheus.Histogram
 	vmRestartCounts                prometheus.Counter
+	reconcileDurationSuccessful    prometheus.Histogram
+	reconcileDurationFailure       prometheus.Histogram
+	statusUpdateFailureCount       prometheus.Counter
 }
 
 func MakeReconcilerMetrics() ReconcilerMetrics {
@@ -60,6 +63,26 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
 			prometheus.CounterOpts{
 				Name: "vm_restarts_count",
 				Help: "Total number of VM restarts across the cluster captured by VirtualMachine reconciler",
+			},
+		)),
+		reconcileDurationSuccessful: util.RegisterMetric(metrics.Registry, prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "reconcile_duration_successful_seconds",
+				Help:    "Time duration of successful reconciles",
+				Buckets: buckets,
+			},
+		)),
+		reconcileDurationFailure: util.RegisterMetric(metrics.Registry, prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "reconcile_duration_failure_seconds",
+				Help:    "Time duration of failed reconciles",
+				Buckets: buckets,
+			},
+		)),
+		statusUpdateFailureCount: util.RegisterMetric(metrics.Registry, prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Name: "status_update_failure_count",
+				Help: "Total number of failures to update the status of a VirtualMachine",
 			},
 		)),
 	}
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -3,12 +3,16 @@ package controllers
 import (
 	"context"
 	"sync"
+	"time"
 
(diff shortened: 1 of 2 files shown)


sharnoff:
I don't think this would resolve #918 — AFAIK most of the conflicts we get aren't actually at the final "update status" step, but earlier on in the reconcile operation. We'd need to inspect the error value we get.

Omrigan:
> I don't think this would resolve #918 — AFAIK most of the conflicts we get aren't actually at the final "update status" step, but earlier on in the reconcile operation. We'd need to inspect the error value we get.

Would 00e658d work?

neonvm/controllers/metrics.go
```diff
@@ -62,6 +70,32 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
...
+				Buckets: buckets,
+			},
+		)),
+		reconcileDurationFailure: util.RegisterMetric(metrics.Registry, prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "reconcile_duration_failure_seconds",
+				Help:    "Time duration of failed reconciles",
+				Buckets: buckets,
+			},
+		)),
```
sharnoff, line 86: Hm, thinking about what usage of this will look like - getting total average reconcile duration would be something like `(rate(reconcile_duration_successful_seconds_sum{}) + rate(reconcile_duration_failure_seconds_sum{})) / (rate(reconcile_duration_successful_seconds_count{}) + rate(reconcile_duration_failure_seconds_count{}))`

What do you think about making this a single metric and then differentiate via `outcome="..."` ? Could bundle in the status failures as well with `outcome="conflict"`...

neonvm/controllers/metrics.go
```diff
@@ -121,11 +154,22 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// async approach.
 	d.lock.Lock()
 	defer d.lock.Unlock()
+	outcome := SuccessOutcome
 	if err != nil {
 		d.failing[req.NamespacedName] = struct{}{}
```
sharnoff, line 159: WDYT about using separate maps / counts for failure vs conflict?

neonvm/controllers/metrics.go
```diff
@@ -121,11 +154,22 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			outcome = ConflictOutcome
+		} else {
+			outcome = FailureOutcome
+		}
+
+		log.Error(err, "Failed to reconcile VirtualMachine",
+			"duration", duration.String(), "outcome", outcome)
 	} else {
 		delete(d.failing, req.NamespacedName)
+		log.Info("Successful reconciliation", "duration", duration.String())
```
sharnoff, line 170: Would be good to check if Loki can filter on stringly-typed durations, or if it'd need to be a number. Otherwise, I'd suggest making this e.g. floating point number of seconds, rather than a pretty representation (unfortunately, because pretty representation would be much more readable)

neonvm/controllers/metrics.go
```diff
@@ -112,7 +141,11 @@ func WithMetrics(reconciler reconcile.Reconciler, rm ReconcilerMetrics, cntrlNam
 }
 
 func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	now := time.Now()
```
sharnoff, line 146: brief thought - wonder if it'd be worthwhile logging the start of the reconcile operation? may be too verbose... something to consider, perhaps for later.

sharnoff (approved):

neonvm/controllers/metrics.go
Omrigan: Yeah, it has e.g. `| duration > 10s`, and I think I used it already at some point.

sharnoff: Ah awesome :)

neonvm/controllers/metrics.go
Omrigan: Yeah, before we had 0 lines per reconciliation in logs, if no errors.  Let's have one for now.

neonvm/controllers/metrics.go
Omrigan: Created a follow-up https://github.com/neondatabase/autoscaling/issues/932

Issue #921, opened 2024-04-26
Non-functional codebase improvements


Omrigan:
## Problem description / Motivation

There are small-scale non-functional changes/tech-debt we occasionally want to make here and there, but they do not have enough priority to deserve a separate issue. Several example out of the top of my head:
 - Use `ptr[T](obj T) *T` everywhere
 - Don't panic, ever
 - Errorgroups for flow control everywhere
 - Outdated/unmaintained dependencies cleanup
 - Rename VirtualMachine to VM in neonvm-controller

Not sure it is the best way to track that, but I realized I keep a bunch of those in the heed and needed to dump them somewhere.

## Feature idea(s) / DoD

This issue doesn't have a DoD.

PR #922, opened 2024-04-26
util: add taskgroup

Omrigan:
taskgroup.Group is the better version of errgroup.Group with two changes:
1. Support for multerrors
2. Logging the errors, when they are returned

Part of #921


pkg/util/taskgroup/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Peter Kristensen
```
sharnoff, line 3: What's the original source for the included code?

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.uber.org/zap"
+)
+
+// Group manages goroutines and collect all the errors.
+// See https://pkg.go.dev/golang.org/x/sync/errgroup#group for more information
+type Group interface {
+	WithContext(ctx context.Context) context.Context
+	Wait() error
+	Go(name string, f func(logger *zap.Logger) error)
+}
```
sharnoff, line 20: Is there a reason for this to be an interface rather than just a public struct?

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,83 @@
...
+		cancel: nil, // Set separately by WithContext
+		logger: logger,
+		wg:     sync.WaitGroup{},
+
+		errMutex: sync.Mutex{},
+		err:      nil,
+	}
+}
+
+// WithContext returns a new Group with a associated Context.
```
sharnoff, line 44: > returns a new Group

typo? it modifies the `group` in place and returns the `Context`.

IMO the current behavior is a bit of a footgun: IMO a safer API would be to specify the context as an argument to `NewGroup`. Maybe also worthwhile to pass it through to the function passed to `(*group).Go()`

See also:

https://github.com/neondatabase/autoscaling/blob/e3b2cecb5ae0cc903785a6c8a27c60c4ba84f852/pkg/agent/runner.go#L302-L344

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,83 @@
...
+	if g.cancel != nil {
+		g.cancel()
+	}
+	return g.err
+}
+
+// Go calls the function in a new goroutine.
+// If a non-nil errors is returned, the context is canceled and
+// the error is collected using multierr and will be returned by Wait.
+func (g *group) Go(name string, f func(logger *zap.Logger) error) {
```
sharnoff, line 66: Might be good to catch panics here as well. WDYT?

pkg/util/taskgroup/LICENSE
Omrigan: Originally taken from https://github.com/ptxmac/multierrgroup, added the comment with it.

pkg/util/taskgroup/taskgroup.go
Omrigan: Implemented in 2c9ee71.

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,109 @@
...
+// See https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext
+func (g *group) WithContext(ctx context.Context) context.Context {
+	ctx, g.cancel = context.WithCancel(ctx)
+	return ctx
+}
+
+// WithPanicHandler sets a panic handler for the group.
+func (g *group) WithPanicHandler(f func(any)) {
+	g.panicHandler = f
+}
```
sharnoff, line 60: In the past, I've seen `WithFoo` methods typically return a new version of the callee, with "foo" added. Maybe these should be named more like `SetFoo` instead?

That said: Is there a typical use case for adding a context after construction? IMO the API would be cleaner if it's an optional arg to `NewGroup`, plus a new function to fetch the context for the `Group`

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,109 @@
...
+	return g.err
+}
+
+func (g *group) call(f func() error) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if g.panicHandler != nil {
+				g.panicHandler(r)
+			}
+			err = fmt.Errorf("panic: %v", r)
```
sharnoff, line 79: We should include a stack trace here. There's an implementation in `pkg/agent/runner.go` - if you like, `github.com/sharnoff/chord` also has some utils for this (but we also have no usage of it elsewhere, so... :shrug:)

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,109 @@
...
+// WithContext updates the current Group with a associated Context.
+// The context will be canceled if any goroutine returns an error.
+// See https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext
+func (g *group) WithContext(ctx context.Context) context.Context {
+	ctx, g.cancel = context.WithCancel(ctx)
+	return ctx
+}
+
+// WithPanicHandler sets a panic handler for the group.
+func (g *group) WithPanicHandler(f func(any)) {
```
sharnoff, line 58: Is `WithPanicHandler` currently used anywhere? (couldn't find anything, but maybe I missed it)

pkg/util/taskgroup/taskgroup.go
sharnoff: Resolving in favor of a newer comment

pkg/agent/billing/billing.go
```diff
@@ -174,15 +175,14 @@ func runBillingMetricsCollector(
 		case <-collectTicker.C:
 			logger.Info("Collecting billing state")
 			if store.Stopped() && ctx.Err() == nil {
-				err := errors.New("VM store stopped but background context is still live")
-				logger.Panic("Validation check failed", zap.Error(err))
+				return errors.New("VM store stopped but background context is still live")
```
sharnoff, line 178: IMO this should still be a panic. It's behavior that is never expected to occur that indicates a deeper issue somewhere else

pkg/agent/entrypoint.go
```diff
@@ -79,18 +75,32 @@ func (r MainRunner) Run(logger *zap.Logger, ctx context.Context) error {
...
+	mc, err := billing.NewMetricsCollector(ctx, logger, &r.Config.Billing)
+	if err != nil {
+		return fmt.Errorf("error starting billing metrics collector: %w", err)
+	}
 
-			logger.Error("vmEventQueue returned error", zap.Error(err))
-			return err
+	tg := taskgroup.NewGroup(logger)
+	ctx = tg.WithContext(ctx)
+	tg.Go("watch-metrics", func(logger *zap.Logger) error {
```
sharnoff, line 85: What does the "watch" in "watch-metrics" mean?

pkg/agent/entrypoint.go
```diff
@@ -79,18 +75,32 @@ func (r MainRunner) Run(logger *zap.Logger, ctx context.Context) error {
...
-	for {
-		event, err := vmEventQueue.Wait(ctx)
-		if err != nil {
-			if ctx.Err() != nil {
-				// treat context canceled as a "normal" exit (because it is)
-				return nil
-			}
+	mc, err := billing.NewMetricsCollector(ctx, logger, &r.Config.Billing)
+	if err != nil {
+		return fmt.Errorf("error starting billing metrics collector: %w", err)
```
sharnoff, line 80: This is no longer "starting" the billing metrics collector, right?

pkg/agent/entrypoint.go
Omrigan: Yeah, I guess collect metrics should be better.

pkg/agent/entrypoint.go
Omrigan: Yup, changed to creating.

pkg/agent/billing/billing.go
Omrigan: I am skeptical of ever using panic.

But what does this actually change, if the panic will be handled higher up the stack?

pkg/util/taskgroup/taskgroup.go
Omrigan: > IMO the API would be cleaner if it's an optional arg
Right, implemented in 836b1b0

This has and added advantage: no need to specify context if Background is the parent
And a disadvantage: the API is different from the one which is usually used around contexts: `ctx, ... := func1(ctx)`

pkg/util/taskgroup/taskgroup.go
sharnoff: Oh, interesting, fair enough haha

pkg/agent/billing/billing.go
sharnoff: My thinking is that "just" returning an error makes it seem like this issue is less severe, whereas a panic indicates that some invariant has been violated.

For this particular case, IIRC there is no expected outcome where the VM store stops while the autoscaler-agent is still running.

Thoughts?

pkg/util/taskgroup/taskgroup.go
sharnoff: Ah I see. Do you plan to port that to taskgroup soon after? (generally trying to avoid adding dead code -- alternatively, how much effort would it be to add this back when needed?)

My 2 cents - porting the autoscaler-agent's usage is probably a bit more complicated, and this work is not super high priority, so it's _probably_ a safer bet to keep it minimal for now.

WDYT?

pkg/agent/entrypoint.go
```diff
@@ -79,18 +75,31 @@ func (r MainRunner) Run(logger *zap.Logger, ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("error creating billing metrics collector: %w", err)
+	}
 
-			logger.Error("vmEventQueue returned error", zap.Error(err))
-			return err
+	tg := taskgroup.NewGroup(logger, taskgroup.WithParentContext(ctx))
+	tg.Go("collect-metrics", func(logger *zap.Logger) error {
+		return mc.Run(tg.Ctx(), logger, storeForNode, metrics)
+	})
```
sharnoff, line 86: A couple questions here:

1. Will this change the value of `.logger` in the billing-related log lines? (and if so, from what to what?)
2. Are the billing client goroutines now using taskgroup as well?

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+type GroupOption func(*group)
+
+// WithParentContext sets the parent context for the group.
+func WithParentContext(ctx context.Context) GroupOption {
+	return func(g *group) {
+		g.ctx, g.cancel = context.WithCancel(ctx)
+	}
+}
```
sharnoff, line 46: Very clean, nice :)

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,131 @@
...
+	return g.err
+}
+
+func (g *group) call(f func() error) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if g.panicHandler != nil {
+				g.panicHandler(r)
+			}
+			st := stack.GetStackTrace(nil, 1).String()
```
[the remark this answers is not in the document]
Omrigan: Fixed in 27467b4

pkg/util/taskgroup/taskgroup.go
Omrigan: > Do you plan to port that to taskgroup soon after? 

I didn't have speific plans.

> generally trying to avoid adding dead code

Generally, yes, but in this case it is a "general case" abstraction, so it's design doesn't have to strictly follow whether it is used in the codebase. Also, the change is tiny, shouldn't be in the way.

pkg/agent/entrypoint.go
Omrigan: 1. It will expand "logger" field in the json logline by more more segment - "collect-metrics"
2. You mean the onces which are run from the `MetricsCollector`? No, they are still `go ...()`.

pkg/agent/billing/billing.go
Omrigan: Ok, brought back the panic in 38c5512.

pkg/util/taskgroup/taskgroup.go
sharnoff: Fair enough, makes sense

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,132 @@
...
+	return g
+}
+
+// Ctx returns a context that will be canceled when the group is Waited.
+func (g *group) Ctx() context.Context {
+	return g.ctx
+}
+
+// WithPanicHandler sets a panic handler for the group.
+func (g *group) WithPanicHandler(f func(any)) {
```
sharnoff, line 78: Seems like this should also be a `GroupOption`?

pkg/util/taskgroup/taskgroup_test.go
```diff
@@ -0,0 +1,99 @@
...
+	select {
+	case <-g.Ctx().Done():
+		break
+	default:
+		t.Fatal("context should be done")
+	}
+}
+
+func setupLogsCapture() (*zap.Logger, *observer.ObservedLogs) {
+	core, logs := observer.New(zap.InfoLevel)
```
sharnoff, line 72: Oh nice, TIL :)

pkg/util/taskgroup/taskgroup_test.go
```diff
@@ -0,0 +1,99 @@
...
+	msg0 := logs.All()[0]
+	assert.Equal(t, "Task panicked", msg0.Message)
+	assert.Len(t, msg0.Context, 2)
+	assert.Equal(t, "payload", msg0.Context[0].Key)
+	assert.Equal(t, "panic message", msg0.Context[0].String)
+	assert.Equal(t, "stack", msg0.Context[1].Key)
+	stackTrace := msg0.Context[1].String
+	assert.True(t, strings.HasPrefix(stackTrace, "runtime.gopanic(...)\n"))
+	msg1 := logs.All()[1]
+	assert.Equal(t, "task task1 failed: panic: panic message", msg1.Message)
```
sharnoff, line 96: I found this quite hard to grok, could you add a high level comment explaining what we expect to see here?

sharnoff, line 96: Separately: It's easy to have off-by-one errors in generating stack traces. Can we also test the contents of the stack trace here?

If testing via strings is overly annoying, I could also imagine testing for equality of the stack frames array produced by `util/stack`

pkg/agent/entrypoint.go
sharnoff: So AFAICT then, the billing client log lines will have `logger` fields like `autoscaler-agent.collect-metrics.send-http` -- and collection will have `autoscaler-agent.collect-metrics.collect` ?

IMO:

1. They should still mention `billing` explicitly
2. Client `logger`s should not mention "collect-metrics" because that's a separate process
3. We shouldn't have duplicated `collect-metrics.collect`

pkg/util/taskgroup/taskgroup_test.go
Omrigan: There is an assert two lines above, checking the first frame. Is there a need to test more than one frames?

Omrigan: Take a look at 75dc6de, is it better?

pkg/util/taskgroup/taskgroup.go
Omrigan: Changed in 75dc6de

pkg/agent/entrypoint.go
Omrigan: Changed to `billing-metrics` in 954e2c6. Does this really matter in practice though, e.g. do you even look at the logger name?

sharnoff: I often use it for filtering -- e.g. to show all errors from `autoscaler-agent.billing.send-http`, or something. It's nice to keep those consistent.

pkg/util/taskgroup/taskgroup_test.go
sharnoff: Looks good, nice

sharnoff: Oh, good point — fair enough

pkg/util/taskgroup/taskgroup_test.go
```diff
@@ -0,0 +1,109 @@
...
+	// context, and one for any task failure.
+	assert.Equal(t, 2, logs.Len())
+
+	msg0 := logs.All()[0]
+	assert.Equal(t, "Task panicked", msg0.Message)
+	assert.Len(t, msg0.Context, 2)
+	assert.Equal(t, "payload", msg0.Context[0].Key)
+	assert.Equal(t, "panic message", msg0.Context[0].String)
+	assert.Equal(t, "stack", msg0.Context[1].Key)
+	stackTrace := msg0.Context[1].String
```
sharnoff, line 102:
```suggestion
	stackTrace := msg0.Context[1].String
	// test that the stack trace begins with gopanic(...) so we always start
	// the backtrace at the same place
```

(because I didn't realize this `strings.HasPrefix` was sufficient when I first read it :smile:)

pkg/agent/entrypoint.go
```diff
@@ -79,18 +75,31 @@ func (r MainRunner) Run(logger *zap.Logger, ctx context.Context) error {
...
-			}
+	mc, err := billing.NewMetricsCollector(ctx, logger, &r.Config.Billing)
+	if err != nil {
+		return fmt.Errorf("error creating billing metrics collector: %w", err)
+	}
 
-			logger.Error("vmEventQueue returned error", zap.Error(err))
-			return err
+	tg := taskgroup.NewGroup(logger, taskgroup.WithParentContext(ctx))
+	tg.Go("billing-metrics", func(logger *zap.Logger) error {
```
sharnoff, line 84: I _think_ actually just changing this to `"billing"` would resolve my [comments from earlier](https://github.com/neondatabase/autoscaling/pull/922#discussion_r1631329536) -- keeping the logger names the same.

Reading through it again (please check me on this!), AFAICT we would still end up with

`.billing.collect` as the main collection thread:

https://github.com/neondatabase/autoscaling/blob/954e2c64ed9ab3545f3327f702e2dff428bfb065/pkg/agent/billing/billing.go#L168-L169

and `.billing.send-$CLIENT` as the sending threads

https://github.com/neondatabase/autoscaling/blob/954e2c64ed9ab3545f3327f702e2dff428bfb065/pkg/agent/billing/billing.go#L165

---

It'd probably be good to move the sender threads to task group as well (especially so that we make sure we're actually waiting for them to finish before exiting!), but I'm happy to leave that for a follow-up.

sharnoff (approved):
Approved, pending a couple comments :)

(I know this one took quite a while; thanks for sticking through it!)

pkg/util/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}()
+	err = f()
+	return err
+}
+
+// Go calls the function in a new goroutine.
+// If a non-nil errors is returned, the context is canceled and
+// the error is collected using multierr and will be returned by Wait.
+func (g *group) Go(name string, f func(logger *zap.Logger) error) {
```
AlexanderYastrebov, line 113: It might be useful to pass context.Context (child of tg.Ctx()) instead of logger to f: logger could be extracted from context with a helper method and context could be used to cancel work in progress.

E.g. this
https://github.com/neondatabase/autoscaling/blob/6013d8b6e1b183c051ec9c22eba56b401f9fc6d3/pkg/agent/entrypoint.go#L97-L99
would become:  
```go
tg.Go("scalingevents-run", func(ctx context.Context) error {
    taskgroup.Logger(ctx).Debug("Logger could be easily extracted from context here or e.g. within called functions")
    return scalingReporter.Run(ctx)
})
```

PR #923, opened 2024-04-26
neonvm-runner-image-loader: Faster rollout

sharnoff:
Specifically, this PR sets `maxUnavailable = 3`, which allows updating up to 3 copies in parallel.

An issue we see during rollout is that it often takes 30+ seconds to create a new neonvm-runner-image-loader on a node, and doing this for many nodes takes an unreasonable amount of time. This improves that.

### DIFF
--- a/neonvm/runner-image-loader/daemonset.yaml
+++ b/neonvm/runner-image-loader/daemonset.yaml
@@ -6,6 +6,10 @@ spec:
   selector:
     matchLabels:
       name: neonvm-runner-image-loader
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 3
   template:
     metadata:
       labels:
--- a/neonvm/runner-image-loader/daemonset.yaml
+++ b/neonvm/runner-image-loader/daemonset.yaml
@@ -9,7 +9,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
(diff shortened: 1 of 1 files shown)


cicdteam:
@sharnoff 

why it hardcoded to 3 ? `maxUnavalable` and `maxSurge` could be percentage, like

```
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 100%
      maxSurge: 100%
```

if we have 10 pods then on rollout 10 old pods will be stopped and 10 new pods will be stared at the moment.

sharnoff:
> why it hardcoded to 3 ?

Yeah, I considered making it a percentage, but I'm not sure if doing a simultaneous image pull cluster-wide could potentially cause issues, so better to be safe.

If you think it's definitely ok, I'll bump it higher, if +1 from @Omrigan.

Omrigan:
> > why it hardcoded to 3 ?
> 
> Yeah, I considered making it a percentage, but I'm not sure if doing a simultaneous image pull cluster-wide could potentially cause issues, so better to be safe.
> 
> If you think it's definitely ok, I'll bump it higher, if +1 from @Omrigan.

I don't see the reason why 100% could failt

sharnoff:
Fair enough, went with 100%

sharnoff:
After trying it out, 100% was definitely the right move :)

PR #929, opened 2024-05-07
Upgrade k8s dependencies to 1.27

SergeyMelnikov:
Updating whole autoscaling to k8s version 1.27

In particular:
* Update all k8s dependency to 1.27.15
* latest k3d/k3s is 1.27.14
* latest kind is 1.27.13
* Update sigs.k8s.io/controller-runtime to 0.15.1
* Update Validate methods of the hooks to new interface, they need to return warnings, returning always nil for now.
* Update all other versions in Makefile
* Update cluster-autoscaler to be based on 1.27.8
* Update crictl to 1.27 ( https://neondb.slack.com/archives/C03TN5G758R/p1718632719461499 )

cc: https://github.com/neondatabase/cloud/issues/7922


Makefile
```diff
@@ -398,28 +398,34 @@ $(LOCALBIN):
...
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 # List of available versions: https://storage.googleapis.com/kubebuilder-tools
-ENVTEST_K8S_VERSION = 1.26.1
+ENVTEST_K8S_VERSION = 1.27.1
 
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
-CONTROLLER_TOOLS_VERSION ?= v0.11.4# k8s deps @ 1.26.1
-CODE_GENERATOR_VERSION ?= v0.26.15
+# k8s deps @ 1.27.1 https://github.com/kubernetes-sigs/controller-tools
+CONTROLLER_TOOLS_VERSION ?= v0.12.0
```
sharnoff, line 411: Should we use v0.12.1 instead?

cluster-autoscaler/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # NOTE: This must match CA's builder/Dockerfile:
 # https://github.com/kubernetes/autoscaler/blob/<GIT_TAG>/builder/Dockerfile
-FROM golang:1.19.7 AS builder
+FROM golang:1.20.12 AS builder
```
sharnoff, line 3: Was going to ask why 1.20, but then I saw my own comment :smile:

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -74,19 +75,19 @@ func (r *VirtualMachine) Default() {
 var _ webhook.Validator = &VirtualMachine{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
-func (r *VirtualMachine) ValidateCreate() error {
+func (r *VirtualMachine) ValidateCreate() (admission.Warnings, error) {
```
sharnoff, line 78: heads up, this is probably conflicting due to #971

neonvm/runner/Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN apk add --no-cache \
 WORKDIR /workspace
 # FIXME: There's non-overlapping version support for <1.27 and >=1.27.
 # We should carefully consider how we go about future-proofing this (or not).
-ENV VERSION="v1.26.1"
+ENV VERSION="v1.27.1"
```
sharnoff, line 37: We can remove the comment above this now, right?

sharnoff (commented):
I noticed we're using 1.27.13 in most places (1.27.14 for k3d image) -- 1.27.15 was released recently, should we use that instead?

Other than that, broadly looks good. Left a few comments :)

sharnoff:
Rebased onto #987 (which has a scheduler fix; more info [here](https://neondb.slack.com/archives/C0798QG4D0T)), which in turn is based on `v0.32.0` -- so that it can be put onto staging and we can check everything's happy.

Compiler's unhappy right now though - will do the fixes & post updates in slack.

sharnoff:
Rebased onto main; this one's been proving somewhat difficult to debug, actually... cc @Omrigan

Omrigan:
@SergeyMelnikov I fixed the tests!

The reason it was failing is that controller-runtime>v0.15.0 now requires to pass `WithStatusSubresource` when creating fake client.

See https://github.com/kubernetes-sigs/controller-runtime/commit/0bff6fc8af438727a2a9e5c8a2050fdad25952b4

neonvm/runner/Dockerfile
Omrigan: Maybe we remove the configuration option to enable `crictl`, so that it won't accidentally enabled by someone (e.g. by taking a release yaml without autoscaling-deploy's patches), since we expect it not to work with 1.27+?

Omrigan (approved):

Issue #450, opened 2023-07-31
Bug: Kernel panic on memory plug


vadim2404:
## Other logs, links

- ...

vadim2404:
@knizhnik created a branch with in-memory cache neondatabase/neon#4891

vadim2404:
Should we merge mmap version of the cache and remove our tmpfs hacks?

sharnoff:
Yes, once reviewed & passing tests, etc. I'm not familiar enough with C to review it.

sharnoff:
This has been mitigated by moving the file cache to disk, but the underlying kernel issue is still present, so: leaving this issue open, but reducing priority.

sharnoff:
We recently updated the kernel (#638), and should check whether this issue is now fixed because of it.

PR #931, opened 2024-05-09
runner: allow coredump collection

### LINKED ISSUE #784: Bug: no coredumps generated if Postgres segfaults
## Environment

Initially spotted on Staging and Production, haven't tried to reproduce it locally.

## Steps to reproduce

- Make Postgres segfault. You can find an example in [0]

## Expected result

We have a logic in `compute_ctl` that checks core dumps and prints stack traces:
https://github.com/neondatabase/neon/blob/350865392cd5bb38eb2a7ff6f45b36f11ac8a911/compute_tools/src/compute.rs#L1000-L1010
 
I expect, segfaulted postgres to produce a core dump, and `compute_ctl` prints the stack trace from it.

## Actual result

- Postgres doesn't produce core dumps
- The reason for it might be in neonVM setup, we expect (as mentioned in the `compute_ctl` comment, and how it's configured for k8s pods):
```
/proc/sys/kernel/core_pattern -> core
/proc/sys/kernel/core_uses_pid -> 1
ulimit -c -> unlimited
```

But in neonVM, we have:
```
/proc/sys/kernel/core_pattern -> core
/proc/sys/kernel/core_uses_pid -> 0
ulimit -c -> 0
```

## Other logs, links

- [0] https://neondb.slack.com/archives/C03QLRH7PPD/p1706797854242749?thread_ts=1706797255.300329&cid=C03QLRH7PPD
- [1] https://neondb.slack.com/archives/C03TN5G758R/p1706866472714979?thread_ts=1706801092.664529&cid=C03TN5G758R
- [2] https://neonprod.grafana.net/goto/Flr52spSg?orgId=1

Omrigan:
Fixes #784 

I didn't find a way to set ulimit globally, as our init system doesn't call `/etc/initscript`


neonvm/runner/main.go
```diff
@@ -676,11 +676,14 @@ func run(logger *zap.Logger) error {
 	}
 
 	// create iso9660 disk with runtime options (command, args, envs, mounts)
-	sysctl := []string{}
+	sysctl := []string{
+		"kernel.core_pattern=/tmp/core.%e.%p.%h.%t",
```
sharnoff, line 680: Not sure if storing it in `/tmp` is a good idea... we've previously seen issues with kernel panics from tmpfs usage (see #450). Maybe the coredumps are small enough in practice that it'll be fine? WDYT?

Omrigan: Right, I changed it back to `/proc/sys/kernel/core_pattern -> core`. That's actually how it was requested, in an issue, so I hope it will be even easier for people to find coredumps.

neonvm/controllers/virtualmachine_controller.go
```diff
@@ -167,6 +167,7 @@ func (r *VirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	statusBefore := virtualmachine.Status.DeepCopy()
+
```
sharnoff, line 170: extra newline?

sharnoff (approved):

neonvm/runner/main.go
sharnoff: Fair enough, sounds good

neonvm/controllers/virtualmachine_controller.go
Omrigan: yep, thanks

Issue #763, opened 2024-01-25
Unit testing coverage for neonvm-controller


sharnoff:
Optional for neondatabase/company_projects#187.

## Problem description / Motivation

The neonvm-controller's reconcile functions are huge and complex, with the majority of testing coming from system-wide end-to-end tests.

This leaves us over-exposed to uncommon scenarios or edge cases, along with little testing against degradations.

## Feature idea(s) / DoD

More testing for the reconcile functions, mostly just to get it started. This is a continuing process

## Implementation ideas

We _may_ want to do something like what `pkg/agent/core/state.go` does, where the core of the logic is pure, and just returns the desired action to take.

While that would be a lot of work, _maybe_ it's worth it to "force" all of the state to exist outside the control flow.

## Related issues

- #716
- #733

sharnoff:
@Omrigan was this finished with #935? or is there more we should include here

Issue #885, opened 2024-03-29
Bug: `make manifests` fails with Go >1.21


sharnoff:
## Environment

Local development, with Go 1.22, for example.

```
$ go version
go version go1.22.1 linux/amd64
```

## Steps to reproduce

1. Clone the repository
2. Run `make manifests`
3. ~~Profit~~

## Expected result

`make manifests` should complete successfully

## Actual result

It fails with a null pointer deref.

<details>
<summary>Command output</summary>

```
$ make manifests
mkdir -p /path/to/repo/bin
go: downloading sigs.k8s.io/controller-tools v0.10.0
go: downloading github.com/spf13/cobra v1.4.0
go: downloading github.com/gobuffalo/flect v0.2.5
go: downloading github.com/fatih/color v1.12.0
go: downloading golang.org/x/tools v0.1.12
go: downloading github.com/mattn/go-isatty v0.0.12
go: downloading github.com/mattn/go-colorable v0.1.8
go: downloading golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4
/path/to/repo/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./neonvm/..." \
	output:crd:artifacts:config=neonvm/config/crd/bases \
	output:rbac:artifacts:config=neonvm/config/rbac \
	output:webhook:artifacts:config=neonvm/config/webhook
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa047cf]

goroutine 127 [running]:
go/types.(*Checker).handleBailout(0xc0001b3200, 0xc0013f7d40)
	/usr/lib/go/src/go/types/check.go:367 +0x88
panic({0xbbb960?, 0x129ace0?})
	/usr/lib/go/src/runtime/panic.go:770 +0x132
go/types.(*StdSizes).Sizeof(0x0, {0xdb0c18, 0x12a34c0})
	/usr/lib/go/src/go/types/sizes.go:228 +0x30f
go/types.(*Config).sizeof(...)
	/usr/lib/go/src/go/types/sizes.go:333
go/types.representableConst.func1({0xdb0c18?, 0x12a34c0?})
	/usr/lib/go/src/go/types/const.go:76 +0x9e
[...]

PR #935, opened 2024-05-14
neonvm-controller: introduce unit tests

Omrigan:
First PR, introducing unit tests. The old functional tests are moved to a `functest` directory.

Part of #763


sharnoff:
Not a full review, but brief thoughts: At first glance, seems like there's a couple other changes that might be good to separate out into other PRs (to make them separate, and merged faster) — in particular, thinking of golangci-lint bump + renaming the files in `neonvm/controllers`.

Also, not sure if it's still the case, but IIRC there's some requirements around exact matching between golangci-lint versions & the go toolchain version. IIRC #885 was blocking upgrading the Go version, so maybe we can do that now?

pkg/util/ptr.go
```diff
@@ -0,0 +1,5 @@
+package util
+
+func Ptr[T any](x T) *T {
+	return &x
+}
```
sharnoff, line 5: Let's use `github.com/samber/lo` instead? Previous discussion here: https://github.com/neondatabase/autoscaling/pull/895#discussion_r1576977602

Makefile
```diff
@@ -107,7 +108,8 @@ test: fmt vet envtest ## Run tests.
 	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)"; \
 	find $(KUBEBUILDER_ASSETS) -type d -exec chmod 0755 {} \; ; \
 	CGO_ENABLED=0 \
-		go test ./... -coverprofile cover.out
+		go test $(TESTARGS) -coverprofile cover.out
+	go tool cover -html=cover.out -o cover.html
```
sharnoff, line 112: Do we have a way of displaying this from CI?

sharnoff (commented):
Some comments, didn't go super deep.

pkg/util/ptr.go
sharnoff: New PR here: #946

Makefile
Omrigan: We most certainly do, but I decided to leave it over to the follow-up.

pkg/util/ptr.go
Omrigan: Fixed in 71558992d633a760c8f5e04535ce5869118df2b0

neonvm/controllers/vm_controller_test.go
```diff
@@ -1,136 +1,342 @@
-/*
```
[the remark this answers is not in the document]
Omrigan: changed the name in 2a0db8c90342ad394355e5b4b4cf28eb73dcea0d

neonvm/controllers/vm_controller_test.go
```diff
@@ -1,136 +1,342 @@
...
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
+	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
-var _ = Describe("VirtualMachine controller", func() {
-	Context("VirtualMachine controller test", func() {
+type clientMock struct {
+	objects map[types.NamespacedName][]byte
+	t       *testing.T
+}
```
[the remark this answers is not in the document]
sharnoff: > Do you think there are arguments against manual reimplementation?

I do, yeah. First, the K8s API is quite complicated. For example, neonvm-controller uses JSON patch to update pod metadata, but I see that's not implemented here. You _could_ reimplement all that, but why bother?

Second: I think it's better to use the mock generated by the k8s related tools rather than writing our own, otherwise we risk mocking based on our own (potentially flawed) understanding, increasing the risk that our tests are blind to behaviors we didn't expect from k8s.

To your point about it making it easier to simulate failures... AFAICT it should be possible by adding a `Reactor` to the [`ReactionChain`](https://github.com/kubernetes/client-go/blob/94cabc85d99585c7437d1b80f9c2571e5e6cf879/testing/fake.go#L38) in `Fake` to allow externally injected failures. (could use e.g. `PrependReactor` but then it's restricted to a single resource + action)

sharnoff (commented):
Replied to a comment, resolved others. + asked via DM about why move files to the `functest` directory.

neonvm/controllers/vm_controller_test.go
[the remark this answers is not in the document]
Omrigan: Ok, good points. Let's merge it like this, and I'll fix it in the follow-up?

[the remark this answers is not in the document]
Omrigan: Actually fixed it right here, please, take a look.

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+					Min: lo.ToPtr(vmv1.MilliCPU(1000)),
+					Max: lo.ToPtr(vmv1.MilliCPU(2000)),
+					Use: lo.ToPtr(vmv1.MilliCPU(1500)),
+				},
+				MemorySlots: vmv1.MemorySlots{
+					Min: lo.ToPtr(int32(512)),
+					Max: lo.ToPtr(int32(2048)),
+					Use: lo.ToPtr(int32(1024)),
+				},
+				MemorySlotSize: *resource.NewQuantity(1, resource.DecimalSI),
```
sharnoff, line 70: let's make these values more reasonable?

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+			RestartCount:  0,
+			PodName:       "",
+			PodIP:         "",
+			ExtraNetIP:    "",
+			ExtraNetMask:  "",
+			Node:          "",
+			CPUs:          lo.ToPtr(vmv1.MilliCPU(100)),
+			MemorySize:    resource.NewQuantity(123, resource.DecimalSI),
+			SSHSecretName: "",
+		},
```
sharnoff, line 85: IIUC the VM starts with a completely empty status object and it's up to the controller to fill it out after reconciling. Should we have all fields empty here? (instead of including cpus & memory size?)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+	r            *VMReconciler
+	client       client.Client
+	origVM       *vmv1.VirtualMachine
+	mockRecorder *mockRecorder
+}
+
+var reconcilerMetrics = MakeReconcilerMetrics()
+
+func newTestParams(t *testing.T) *testParams {
+	os.Setenv("VM_RUNNER_IMAGE", "vm-runner-img")
```
sharnoff, line 101: This isn't thread-safe, right? can we get away with not setting this for now? (if no: let's extract it into `ReconcilerConfig`?)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+			FailurePendingPeriod:    time.Minute,
+			FailingRefreshInterval:  time.Minute,
+		},
+		Metrics: reconcilerMetrics,
+	}
+
+	return params
+}
+
+func (p *testParams) initVM(vm *vmv1.VirtualMachine) *vmv1.VirtualMachine {
```
sharnoff, line 139: what is this method for? (can you add a comment?)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+	assert.NotEmpty(t, vm.Status.PodName)
+	// Spec is unchanged
+	assert.Equal(t, vm.Spec, origVM.Spec)
+
+	// Round 4
+	res, err = params.r.Reconcile(params.ctx, req)
+	assert.NoError(t, err)
+	assert.Equal(t, false, res.Requeue)
+
+	// Pod is pending, so nothing changes
```
sharnoff, line 200: Strictly speaking, nothing is moving the pod along, so it'll still have a phase of `""`, right? (or, do you mean "pending" in the generic sense here?)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+	assert.Equal(t, vm, params.getVM())
+}
+
+func PrettyPrint(t *testing.T, obj any) {
```
sharnoff, line 204: Should this function be private?

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+	require.NoError(t, err)
+
+	assert.Len(t, pod.Spec.Containers, 1)
+	assert.Equal(t, "neonvm-runner", pod.Spec.Containers[0].Name)
+	assert.Equal(t, "vm-runner-img", pod.Spec.Containers[0].Image)
+	assert.Len(t, pod.Spec.InitContainers, 2)
+	assert.Equal(t, "init", pod.Spec.InitContainers[0].Name)
+	assert.Equal(t, "init-kernel", pod.Spec.InitContainers[1].Name)
+
+	PrettyPrint(t, pod)
```
sharnoff, line 245: This is just for debugging, right? is it intentionally left in? (I could go either way, just curious what you think)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+	return params
+}
+
+func (p *testParams) initVM(vm *vmv1.VirtualMachine) *vmv1.VirtualMachine {
+	err := p.client.Create(p.ctx, vm)
+	require.NoError(p.t, err)
+	p.origVM = vm
+
+	// Do serialize/deserialize, to normalize resource.Quantity
+	return p.getVM()
```
sharnoff, line 145: out of curiosity, does this actually work? (asking because it has somewhat surprising (IMO) implications about the mock client)

neonvm/controllers/functests/vm_controller_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package controllers
+package functests
```
sharnoff, line 17: from DMs: if the reason to move this into a separate package is just so that it can be individually tested, that can also be accomplished with `go test <file>` -- so I'd prefer leaving this as-is unless there's other reasons as well

sharnoff (commented):
looks good, mostly just nits

neonvm/controllers/vm_controller_test.go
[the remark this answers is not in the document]
sharnoff: nice!

neonvm/controllers/vm_controller_unit_test.go
sharnoff: Huh ok, fair enough. Yeah, +1 to refactoring in a follow-up

neonvm/controllers/vm_controller_unit_test.go
sharnoff: Perfect, nice

neonvm/controllers/vm_controller_unit_test.go
sharnoff: Ack, fair enough

neonvm/controllers/vm_controller_unit_test.go
sharnoff: Huh! TIL :)

Yep, makes sense.

neonvm/controllers/functests/vm_controller_test.go
sharnoff: I still don't quite get what you mean, but if you get value from having them separate, then fair enough, let's go with it :)

neonvm/controllers/vm_controller_unit_test.go
```diff
@@ -0,0 +1,252 @@
+package controllers
```
sharnoff, line 1: nit: let's make this `controllers_test` ?

AFAICT the following diff would be required to do that (mostly just doing a couple renames)

<details>
<summary>Diff</summary>

```diff
diff --git a/neonvm/controllers/vm_controller.go b/neonvm/controllers/vm_controller.go
index e9ebfcf2..5f6b2254 100644
--- a/neonvm/controllers/vm_controller.go
+++ b/neonvm/controllers/vm_controller.go
@@ -60,15 +60,15 @@ import (
 )
 
 const (
-	virtualmachineFinalizer = "vm.neon.tech/finalizer"
+	VirtualmachineFinalizer = "vm.neon.tech/finalizer"
 )
 
 // Definitions to manage status conditions
 const (
-	// typeAvailableVirtualMachine represents the status of the Deployment reconciliation
-	typeAvailableVirtualMachine = "Available"
-	// typeDegradedVirtualMachine represents the status used when the custom resource is deleted and the finalizer operations are must to occur.
-	typeDegradedVirtualMachine = "Degraded"
+	// TypeAvailableVirtualMachine represents the status of the Deployment reconciliation
+	TypeAvailableVirtualMachine = "Available"
+	// TypeDegradedVirtualMachine represents the status used when the custom resource is deleted and the finalizer operations are must to occur.
+	TypeDegradedVirtualMachine = "Degraded"
 )
 
 const (
@@ -133,9 +133,9 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		// The object is not being deleted, so if it does not have our finalizer,
 		// then lets add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
-		if !controllerutil.ContainsFinalizer(&vm, virtualmachineFinalizer) {
+		if !controllerutil.ContainsFinalizer(&vm, VirtualmachineFinalizer) {
 			log.Info("Adding Finalizer for VirtualMachine")
-			if ok := controllerutil.AddFinalizer(&vm, virtualmachineFinalizer); !ok {
+			if ok := controllerutil.AddFinalizer(&vm, VirtualmachineFinalizer); !ok {
 				log.Info("Failed to add finalizer from VirtualMachine")
 				return ctrl.Result{Requeue: true}, nil
 			}
@@ -147,14 +147,14 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		}
 	} else {
 		// The object is being deleted
-		if controllerutil.ContainsFinalizer(&vm, virtualmachineFinalizer) {
+		if controllerutil.ContainsFinalizer(&vm, VirtualmachineFinalizer) {
 			// our finalizer is present, so lets handle any external dependency
 			log.Info("Performing Finalizer Operations for VirtualMachine before delete it")
 			r.doFinalizerOperationsForVirtualMachine(ctx, &vm)
 
 			// remove our finalizer from the list and update it.
 			log.Info("Removing Finalizer for VirtualMachine after successfully perform the operations")
-			if ok := controllerutil.RemoveFinalizer(&vm, virtualmachineFinalizer); !ok {
+			if ok := controllerutil.RemoveFinalizer(&vm, VirtualmachineFinalizer); !ok {
 				log.Info("Failed to remove finalizer from VirtualMachine")
 				return ctrl.Result{Requeue: true}, nil
 			}
@@ -312,7 +312,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 	// Let's check and just set the condition status as Unknown when no status are available
 	if vm.Status.Conditions == nil || len(vm.Status.Conditions) == 0 {
 		// set Unknown condition status for AvailableVirtualMachine
-		meta.SetStatusCondition(&vm.Status.Conditions, metav1.Condition{Type: typeAvailableVirtualMachine, Status: metav1.ConditionUnknown, Reason: "Reconciling", Message: "Starting reconciliation"})
+		meta.SetStatusCondition(&vm.Status.Conditions, metav1.Condition{Type: TypeAvailableVirtualMachine, Status: metav1.ConditionUnknown, Reason: "Reconciling", Message: "Starting reconciliation"})
 	}
 
 	// NB: .Spec.EnableSSH guaranteed non-nil because the k8s API server sets the default for us.
@@ -434,7 +434,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 			vm.Status.PodIP = vmRunner.Status.PodIP
 			vm.Status.Phase = vmv1.VmRunning
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) created successfully", vm.Status.PodName, vm.Name)})
@@ -452,21 +452,21 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		case runnerSucceeded:
 			vm.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionFalse,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) succeeded", vm.Status.PodName, vm.Name)})
 		case runnerFailed:
 			vm.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeDegradedVirtualMachine,
+				metav1.Condition{Type: TypeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) failed", vm.Status.PodName, vm.Name)})
 		case runnerUnknown:
 			vm.Status.Phase = vmv1.VmPending
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionUnknown,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) in Unknown phase", vm.Status.PodName, vm.Name)})
@@ -484,7 +484,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 					vm.Status.PodName))
 			vm.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeDegradedVirtualMachine,
+				metav1.Condition{Type: TypeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) not found", vm.Status.PodName, vm.Name)})
@@ -572,21 +572,21 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		case runnerSucceeded:
 			vm.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionFalse,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) succeeded", vm.Status.PodName, vm.Name)})
 		case runnerFailed:
 			vm.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeDegradedVirtualMachine,
+				metav1.Condition{Type: TypeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) failed", vm.Status.PodName, vm.Name)})
 		case runnerUnknown:
 			vm.Status.Phase = vmv1.VmPending
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionUnknown,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) in Unknown phase", vm.Status.PodName, vm.Name)})
@@ -605,7 +605,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 					vm.Status.PodName))
 			vm.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeDegradedVirtualMachine,
+				metav1.Condition{Type: TypeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) not found", vm.Status.PodName, vm.Name)})
@@ -625,7 +625,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		case runnerSucceeded:
 			vm.Status.Phase = vmv1.VmSucceeded
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionFalse,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) succeeded", vm.Status.PodName, vm.Name)})
@@ -633,7 +633,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		case runnerFailed:
 			vm.Status.Phase = vmv1.VmFailed
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeDegradedVirtualMachine,
+				metav1.Condition{Type: TypeDegradedVirtualMachine,
 					Status:  metav1.ConditionTrue,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) failed", vm.Status.PodName, vm.Name)})
@@ -641,7 +641,7 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		case runnerUnknown:
 			vm.Status.Phase = vmv1.VmPending
 			meta.SetStatusCondition(&vm.Status.Conditions,
-				metav1.Condition{Type: typeAvailableVirtualMachine,
+				metav1.Condition{Type: TypeAvailableVirtualMachine,
 					Status:  metav1.ConditionUnknown,
 					Reason:  "Reconciling",
 					Message: fmt.Sprintf("Pod (%s) for VirtualMachine (%s) in Unknown phase", vm.Status.PodName, vm.Name)})
diff --git a/neonvm/controllers/vm_controller_unit_test.go b/neonvm/controllers/vm_controller_unit_test.go
index fc7f4429..5db7c824 100644
--- a/neonvm/controllers/vm_controller_unit_test.go
+++ b/neonvm/controllers/vm_controller_unit_test.go
@@ -1,4 +1,4 @@
-package controllers
+package controllers_test
 
 import (
 	"context"
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
+	"github.com/neondatabase/autoscaling/neonvm/controllers"
 )
 
 type mockRecorder struct {
@@ -78,13 +79,13 @@ func defaultVm() *vmv1.VirtualMachine {
 type testParams struct {
 	t            *testing.T
 	ctx          context.Context
-	r            *VMReconciler
+	r            *controllers.VMReconciler
 	client       client.Client
 	origVM       *vmv1.VirtualMachine
 	mockRecorder *mockRecorder
 }
 
-var reconcilerMetrics = MakeReconcilerMetrics()
+var reconcilerMetrics = controllers.MakeReconcilerMetrics()
 
 func newTestParams(t *testing.T) *testParams {
 	os.Setenv("VM_RUNNER_IMAGE", "vm-runner-img")
@@ -107,11 +108,11 @@ func newTestParams(t *testing.T) *testParams {
 		origVM:       nil,
 	}
 
-	params.r = &VMReconciler{
+	params.r = &controllers.VMReconciler{
 		Client:   params.client,
 		Recorder: params.mockRecorder,
 		Scheme:   scheme,
-		Config: &ReconcilerConfig{
+		Config: &controllers.ReconcilerConfig{
 			IsK3s:                   false,
 			UseContainerMgr:         false,
 			MaxConcurrentReconciles: 10,
@@ -159,7 +160,7 @@ func TestReconcile(t *testing.T) {
 	assert.Equal(t, reconcile.Result{
 		Requeue: true,
 	}, res)
-	assert.Contains(t, params.getVM().Finalizers, virtualmachineFinalizer)
+	assert.Contains(t, params.getVM().Finalizers, controllers.VirtualmachineFinalizer)
 
 	// Round 2
 	res, err = params.r.Reconcile(params.ctx, req)
@@ -199,7 +200,7 @@ func prettyPrint(t *testing.T, obj any) {
 func TestRunningPod(t *testing.T) {
 	params := newTestParams(t)
 	origVM := defaultVm()
-	origVM.Finalizers = append(origVM.Finalizers, virtualmachineFinalizer)
+	origVM.Finalizers = append(origVM.Finalizers, controllers.VirtualmachineFinalizer)
 	origVM.Status.Phase = vmv1.VmPending
 
 	origVM = params.initVM(origVM)
@@ -248,5 +249,5 @@ func TestRunningPod(t *testing.T) {
 	// VM is now running
 	assert.Equal(t, vmv1.VmRunning, vm.Status.Phase)
 	assert.Len(t, vm.Status.Conditions, 1)
-	assert.Equal(t, vm.Status.Conditions[0].Type, typeAvailableVirtualMachine)
+	assert.Equal(t, vm.Status.Conditions[0].Type, controllers.TypeAvailableVirtualMachine)
 }
```

<details>

Alternately, I'm ok to leave that as a follow-up if you like.

sharnoff (approved):
lgtm, nice! one nit regarding `package controllers` vs `package controllers_test`

neonvm/controllers/functests/vm_controller_test.go
Omrigan: I meant there is `vm_controller_test.go` and `vm_qmp_test.go`, so by `go test ./neonvm/controllers` we can run them both.

Also it is "aesthetically" looks better IMO that lightweight unit tests are in the same folder, but heavier functional tests are in a different folder.

neonvm/controllers/vm_controller_unit_test.go
Omrigan: I want to leave the opportunity to test private functions also, these are unit tests.

PR #936, opened 2024-05-20
plugin: Transaction-based speculative reserve and logic unification

sharnoff:
Trying out this approach as a potential way to make implementing overcommit (#517) easier, to avoid dealing with rounding errors when responding to the autoscaler-agents.

I think this "speculative reserve" may be a reasonable change in isolation, and reducing the number of separate implementations of the same logic should also help with overcommit (because of how it affects every single resource calculation).

Not 100% sure we want to go this route, but it's my best guess so far.

---

Tested this locally and it seemed to be ok; already fixed a couple bugs based on suspicious logs. There might be more.

### DIFF
--- a/pkg/plugin/trans.go
+++ b/pkg/plugin/trans.go
@@ -64,59 +64,6 @@ func (s verdictSet) MarshalLogObject(enc zapcore.ObjectEncoder) error {
 	return nil
 }
 
-// handleLastPermit updates reserved values of r.pod and r.node with the most
-// recent info about the last permit received by the agent. This allows a new
-// scheduler to learn about previous scheduler's state.
-//
-// Why the new scheduler needs this info? It's always possible for the scheduler
-// to get killed and immediately restart, without the agent believing there was
-// any disconnect, which could lead to unintentional over-committing of resources
-// from the Buffer values if too many agents request upscaling on the first
-// request to the scheduler.
-func (r resourceTransitioner[T]) handleLastPermit(lastPermit T) (verdict string) {
-	oldState := r.snapshotState()
-
-	if lastPermit <= r.pod.Reserved {
-		r.node.Reserved -= r.pod.Reserved - lastPermit
-		r.pod.Reserved = lastPermit
-
-		var podBuffer string
-		var oldNodeBuffer string
-		var newNodeBuffer string
-		if r.pod.Buffer != 0 {
-			podBuffer = fmt.Sprintf(" [buffer %d]", r.pod.Buffer)
-			oldNodeBuffer = fmt.Sprintf(" [buffer %d]", oldState.node.Buffer)
-
-			r.node.Buffer -= r.pod.Buffer
-			r.pod.Buffer = 0
-
-			newNodeBuffer = fmt.Sprintf(" [buffer %d]", r.node.Buffer)
-		}
-
-		totalReservable := r.node.Total
-		verdict = fmt.Sprintf(
-			"pod reserved %d%s -> %d, "+
-				"node reserved %d%s -> %d%s (of %d)",
-			oldState.pod.Reserved, podBuffer, r.pod.Reserved,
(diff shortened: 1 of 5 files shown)


Omrigan (commented):
Partial review

pkg/plugin/trans.go
```diff
@@ -64,81 +64,65 @@ func (s verdictSet) MarshalLogObject(enc zapcore.ObjectEncoder) error {
...
 // handleReserve adds the resources from the pod to the node, reporting if the node was over-budget
 //
 // This is used in combination with Xact to speculatively *try* reserving a pod, and then revert if
 // it would result in being over-budget.
 func (r resourceTransitioner[T]) handleReserve() (overbudget bool, verdict string) {
-	oldState := r.snapshotState()
-
-	r.node.Reserved += r.pod.Reserved
-	r.node.Buffer += r.pod.Buffer
+	callback := func(oldState, newState resourceState[T]) string {
```
Omrigan, line 72: As discussed, perhaps all verdict formatting should be gathered at a single place, like a structure `verdictGenerator`, which will be used by `handleRequestedGeneric`.

pkg/plugin/trans.go
```diff
@@ -150,26 +134,130 @@ func (r resourceTransitioner[T]) handleReserve() (overbudget bool, verdict strin
 // Any permitted increases are required to be a multiple of factor.
 //
 // A pretty-formatted summary of the outcome is returned as the verdict, for logging.
```
Omrigan, line 136: It would be easier to understand, if those comments would more clearly reflect the difference between `handleRequested` and `handleReserve`.

pkg/plugin/trans.go
```diff
@@ -117,6 +117,33 @@ func (r resourceTransitioner[T]) handleLastPermit(lastPermit T) (verdict string)
...
+			oldState.node.Reserved, oldState.node.Buffer, r.pod.Reserved, r.pod.Buffer, r.node.Reserved, r.node.Buffer, oldState.node.Total,
+		)
+	} else {
+		verdict = fmt.Sprintf(
+			"node reserved %v + %v -> %v of total %v",
+			oldState.node.Reserved, r.pod.Reserved, r.node.Reserved, oldState.node.Total,
+		)
+	}
+
+	overbudget = r.node.Reserved > r.node.Total
```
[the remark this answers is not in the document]
sharnoff: `Buffer` is included in `Reserved`, so it should be ok as is

pkg/util/xact/xact.go
```diff
@@ -0,0 +1,31 @@
...
+// application.
+type Xact[T any] struct {
+	tmp  T
+	base *T
+}
+
+// New returns a new transaction object (called Xact) operating on the given pointer
+func New[T any](ptr *T) *Xact[T] {
+	return &Xact[T]{
+		tmp:  *ptr,
```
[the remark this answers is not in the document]
sharnoff: That's fair. Added a note about this in 0972c4ce6e1af352f4d6423cb99a97281dde3c65 -- the types I intended to use for this don't have pointers (and at least IMO, that's a reasonable restriction for now).

pkg/plugin/trans.go
sharnoff: Fair enough - added some comments in bbfca613b2a4fa8b8d650625b4049296475c89f6, although not sure if there's more I should elaborate on.

pkg/plugin/trans.go
sharnoff: I like that idea -- I think it _may_ end up somewhat complex; wdyt about leaving it as a follow-up?

sharnoff:
Updates:

* Rebased onto main
* Add note about limitations w.r.t. pointers: 0972c4ce6e1af352f4d6423cb99a97281dde3c65
* Added comments about the difference between `handleRequested` vs `handleReserve`: bbfca613b2a4fa8b8d650625b4049296475c89f6

Omrigan (approved):

Issue #932, opened 2024-05-09
neonvm: separate maps / counts for failure vs conflict


Omrigan:
This might help to debug the issues when we have a lot of VM failing to reconcile. Although, it is unclear if repeated conflicts for the same VM is likely failure scenario. 

_Originally posted by @sharnoff in https://github.com/neondatabase/autoscaling/pull/920#discussion_r1595561357_

sharnoff:
To add onto this, I think in particular, this would help with making our alerting more sensitive — having 10 minutes of >1 VM failing to reconcile may be expected as there's always _something_ affected by conflicts; but having 10 minutes of >1 VM _truly failing_ may not be expected.

Alternatively -- something I'd discussed as part of #757 is that we may be better off having metrics like "number of VMs failing reconcile for N seconds" or something — that's probably much easier to have higher-quality alerting for, rather than our gauge of binary "is it stuck" approach we currently have.

PR #937, opened 2024-05-20
neonvm: separate failing/conflict for the number of failing objects

Omrigan:
Before this, failing and conflicts were counted together, which resulted in a lot of false positive alert firings because of the ever-present background of conflicts.

Part of the #932.

### DIFF
--- a/neonvm/controllers/metrics.go
+++ b/neonvm/controllers/metrics.go
@@ -40,7 +40,7 @@ func MakeReconcilerMetrics() ReconcilerMetrics {
 				Name: "reconcile_failing_objects",
 				Help: "Number of objects that are failing to reconcile for each specific controller",
 			},
-			[]string{"controller"},
+			[]string{"controller", OutcomeLabel},
 		)),
 		vmCreationToRunnerCreationTime: util.RegisterMetric(metrics.Registry, prometheus.NewHistogram(
 			prometheus.HistogramOpts{
@@ -100,8 +100,9 @@ type wrappedReconciler struct {
 	Reconciler     reconcile.Reconciler
 	Metrics        ReconcilerMetrics
 
-	lock    sync.Mutex
-	failing map[client.ObjectKey]struct{}
+	lock        sync.Mutex
+	failing     map[client.ObjectKey]struct{}
+	conflicting map[client.ObjectKey]struct{}
 }
 
 // ReconcilerWithMetrics is a Reconciler produced by WithMetrics that can return a snapshot of the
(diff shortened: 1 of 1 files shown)


neonvm/controllers/metrics.go
```diff
@@ -156,36 +162,48 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	defer d.lock.Unlock()
 	outcome := SuccessOutcome
 	if err != nil {
-		d.failing[req.NamespacedName] = struct{}{}
 		if errors.IsConflict(err) {
 			outcome = ConflictOutcome
+			d.conflicting[req.NamespacedName] = struct{}{}
```
sharnoff, line 167: need to additionally `delete(...)` failing/conflicting when there's an error in the other one, right?

sharnoff (commented):
LGTM, one potential bug

neonvm/controllers/metrics.go
Omrigan: I don't think so, if it keeps failing, and a conflict occurs, we want it to keep being registered as failing.

sharnoff: Hm, I see. Fair enough. Does it make sense to go the other way as well (i.e. conflict -> error continues registering as conflict)

Omrigan: Ok, let's imagine the VM keeps conflicting, and there are no other failures -> it will register as conflicting. 
But, if the failure were to arise (e.g. QMP errors), then it will stop being conflicting, and starts to be failing only.
This means that an unrelated error impacts the semantics of "conflict", which is why I think it should be kept as-is.

Also, it is simpler to reason, and we might at some point decide to handle have more error types, and it is easier to have the semantics that different modes of failure are independent, and can only be finished when a successful reconciliation occurs.

sharnoff: My mental model for conflicts vs errors is that conflicts are practically guaranteed to be transient (although we should still track in practice because perhaps there's a bug), whereas errors _may_ be persistent.

So if a VM has several conflicts, but then starts getting errors, I would assume it's more likely that the conflicts were resolved than the errors being spurious (i.e. IMO it would make more sense to continue tracking as a conflict when there's errors)

sharnoff (approved):

neonvm/controllers/metrics.go
```diff
@@ -156,36 +162,57 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 			outcome = FailureOutcome
+			d.failing[req.NamespacedName] = struct{}{}
+
+			// If the VM is now getting non-conflict errors, it probably
+			// means transient conflicts has been resolved.
+			//
+			// Notably, the other way around is not true:
+			// if a VM is getting conflict errors, it doesn't mean
+			// non-conflict errors are resolved, as they are more
+			// likely to be persistent.
```
sharnoff, line 178: perfect explanation - thanks for writing it up :)

PR #941, opened 2024-05-21
Update golangci-lint to v1.58.1

Omrigan:
And fix related warnings.


sharnoff (approved):
if it works, it works - thanks :)

PR #943, opened 2024-05-24
neonvm-controller: expand the tests with QMP mocked

### LINKED ISSUE #716: move neonvm controller qmp logic into separate package

ref https://github.com/neondatabase/autoscaling/pull/704#issuecomment-1870632614

> I think we should make a `qmp` package and put all the qmp related stuff in there (separation of concerns). Prefixing related functions with a name (like `Qmp`) is a practice coming from languages such as C that didn't have package like Go. With the `qmp` package we would have a nicer API, for example, `qmp.DelMemoryBackend`.

Basically, we have a bunch of QMP-related types and functions in a single file in the controllers package, but it would make more sense to have those in a separate package, not just a separate file.

Omrigan:
Bases on top of the https://github.com/neondatabase/autoscaling/pull/935 

Part of #763, resolves #716


sharnoff:
I see this PR moves the qmp files into a separate package. That would resolve #716, right? (maybe that can be a separate PR? up to you)

Omrigan:
Superseded by #1032

Issue #939, opened 2024-05-20
Bug: compute_ctl can't open neon.tech.log.0 after restart


Omrigan:
The following error message is observed: `cannot create /dev/virtio-ports/neon.tech.log.0: Device or resource busy`. This has happened together with memory pressure, and OOM-kill of compute_ctl.

The leading hypothesis on this is: while compute_ctl got killed, postgres kept working, thus a descriptor for  `neon.tech.log.0` remained open (postgres still writes to it). Once compute_ctl was restarted, it tried to open `neon.tech.log.0`, but it is impossible to open a serial interface more than once.

## Environment

Production

## Steps to reproduce

Kill compute_ctl, while leaving postgres working. 

## Possible solutions

1. Have a layer between logs source and serial device, e.g. `socat`, which will multiplex output into the serial device.
2. #578 could be helpful.

## Other logs, links

Thread:
https://neondb.slack.com/archives/C03TN5G758R/p1716165317982799

PR #944, opened 2024-05-24
pg16-disk-test: add mkfifo

Omrigan:
The virtio-serial interface can be opened only once.
Consider the following scenario:

1. Process A starts writing to the serial device.
2. Process A spawns a fork, Process B. It inherits the open file descriptor.
3. Process A dies; Process B survives and preserves the file descriptor.
4. Process A is restarted but cannot open the serial device again,
    causing it to crash-loop.

To fix it, we are creating FIFO special file, which supports multiple
writers, and spawning cat to redirect it to the virtio-serial.

Part of the #939.


sharnoff (approved):
let's update branch to get the codespell fix, otherwise lgtm, as long as you've tested locally (as a follow-up, it'd be nice to check in CI that log forwarding works as expected).

Also, keep in mind that setting this in `neondatabase/neon` may not be so simple, because the control plane specifies the redirection.

PR #949, opened 2024-05-29
neonvm-controller: replace failing reconciliation with per-VM failure interval 

### LINKED ISSUE #932: neonvm: separate maps / counts for failure vs conflict

olegbbtr:
The old method had frequent false positive, because there might be a lot of intermittent failures, but overall the system does progress, and every particular VM is getting reconciled. 

Fixes #932

### DIFF
--- a/neonvm/config/controller/deployment.yaml
+++ b/neonvm/config/controller/deployment.yaml
@@ -60,7 +60,7 @@ spec:
         # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
         - "--failure-pending-period=1m"
-        - "--failing-refresh-interval=1m"
+        - "--failing-refresh-interval=15s"
         env:
         - name: NAD_IPAM_NAME
           value: $(NAD_IPAM_NAME)


neonvm/controllers/alerttracker/tracker.go
```diff
@@ -0,0 +1,87 @@
+package alerttracker
+
+import "time"
+
+var Now = time.Now
+
+// Tracker is a struct that keeps track of failure/success events for a given key.
+// Failure state over a given interval will trigger a firing alert.
```
sharnoff, line 8: Could you expand on the expected semantics here?

neonvm/controllers/metrics.go
```diff
@@ -176,28 +182,27 @@ func (d *wrappedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	} else {
-		delete(d.failing, req.NamespacedName)
-		delete(d.conflicting, req.NamespacedName)
+		d.failing.RecordSuccess(req.NamespacedName)
 		log.Info("Successful reconciliation", "duration", duration.String())
 	}
 	d.Metrics.ObserveReconcileDuration(outcome, duration)
 	d.Metrics.failing.WithLabelValues(d.ControllerName,
-		string(FailureOutcome)).Set(float64(len(d.failing)))
+		string(FailureOutcome)).Set(float64(d.failing.FiringCount()))
```
sharnoff, line 196: Technically, things can become firing outside of reconcile cycles, right?

It probably doesn't matter in practice, but should we instead just have a periodic job to update the metrics (rather than updating it every time?)

neonvm/controllers/alerttracker/tracker_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func newNowMock() *nowMock {
+	ts, _ := time.Parse("2006-01-02", "2024-01-01")
+	return &nowMock{ts: ts}
+}
+
+func TestTracker(t *testing.T) {
+	now := newNowMock()
+	alerttracker.Now = now.Now
```
sharnoff, line 31: This seems sketchy... (does it produce data races if we run tests in parallel?)

I would recommend something more like what we went with in `pkg/util/recentcnt[_test].go`

neonvm/controllers/config.go
```diff
@@ -22,6 +24,10 @@ type ReconcilerConfig struct {
 	// This field is passed to neonvm-runner as the `-qemu-disk-cache-settings` arg, and is directly
 	// used in setting up the VM disks via QEMU's `-drive` flag.
 	QEMUDiskCacheSettings string
+
+	// ReconcileFailureInterval is the interval after which a VM will be marked as failing if it has
+	// not reconciled successfully.
```
sharnoff, line 29: Strictly speaking, if the last reconcile was a success and there just haven't been any more it'll still be considered successful, right?
(should we somehow mark the start of reconciling so that if a reconcile operation is sufficiently long, that _also_ gets marked as stuck?)

neonvm/controllers/alerttracker/tracker.go
```diff
@@ -0,0 +1,87 @@
...
+func NewTracker[T comparable](interval time.Duration) *Tracker[T] {
+	return &Tracker[T]{
+		interval:     interval,
+		pendingSince: make(map[T]time.Time),
+		firing:       make(map[T]struct{}),
+		fireAt:       []fireAt[T]{},
+	}
+}
+
+func (t *Tracker[T]) forward(now time.Time) {
```
sharnoff, line 31: The logic here seems tricky... it'd be good to:

1. Have a brief high-level explainer about what's going on
2. Have some tests for multiple keys

neonvm/controllers/alerttracker/tracker.go
```diff
@@ -0,0 +1,87 @@
+package alerttracker
```
sharnoff, line 1: not sure how I feel about this package name. Strictly speaking, we're not tracking _alerts_... I can't immediately think of a better name though...

neonvm/config/controller/deployment.yaml
```diff
@@ -59,6 +59,7 @@ spec:
         # * cache.direct=on    - use O_DIRECT (don't abuse host's page cache!)
         # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
+        - "--reconcile-failure-interval=10m"
```
sharnoff, line 62: 10 minutes is a very long time if we expect to reconcile every second. Maybe let's set this to 30s or 1m to start with?

neonvm/controllers/config.go
```diff
@@ -22,6 +24,10 @@ type ReconcilerConfig struct {
...
+	ReconcileFailureInterval time.Duration
```
sharnoff, line 30: `ReconcileFailureInterval` (at least IMO) implies that something happens each interval. Maybe rename to something like `ReconcileFailureWindow` or `ReconcileFailingPendingPeriod` (taken from grafana :smile:) or something?

(maybe even just `FailurePendingPeriod`? these are the _reconciler_ config options after all; I guess "reconcile" may be assumed already...)

WDYT?

sharnoff (commented):
Looks quite good. Most of my comments are either nits or requests for more words :)

neonvm/controllers/alerttracker/tracker.go
Omrigan: I'd love to, but I think we should first settle on better "legend" for what it does, so the fact that those alerts are not grafana alerts doesn't need to be explained in the comments.

neonvm/controllers/metrics.go
Omrigan: That's right. I guess I could spawn another goroutine, which would fetch the counter and put it into `GaugeVec`. This would mean to move [that](https://github.com/neondatabase/autoscaling/pull/757/files#diff-2d9cf99104a80ac4cc6080fbe379163550a313e7e3734c631fb896f8f4630ed4R55) mutex into the `Tracker`, and make `Tracker` thread-safe, but this means mutex is on a hotpath. 

I don't expect any significant contention around mutex wrapping a map, not sure why Shayan was worried. 

Another approach would be to have a [GaugeFunc](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#GaugeFunc) instead of a goroutine, but it is a little inconvenient, as there is no `GaugeVecFunc`. 

Thoughts?

neonvm/controllers/config.go
sharnoff: Oh, TIL, nice!

neonvm/controllers/alerttracker/tracker_test.go
sharnoff: That can work too; I would still rather avoid dependency injection if we can, but wouldn't hold up approval over it.

neonvm/controllers/alerttracker/tracker.go
sharnoff: Ah well. Can bikeshed the name at a later date haha :)

neonvm/controllers/metrics.go
sharnoff: Hm... there's a few options here, huh.

Acquiring many mutexes per reconcile instead of a single one might not be great, although we'd have to see in practice whether it affects reconcile latency (might be good to test out this PR on staging?). Off-hand, I suspect it's _probably_ fine.

I've hit the "no GaugeVecFunc" issue before... its unfortunate. Could write our own GaugeVecFunc equivalent, but that's probably a lot of work.

One related thought:

For the "stuck VMs" dashboard, it's helpful that the autoscaler-agent periodically logs that the VM is marked "stuck" (that way, we can see which VMs were having issues after-the-fact -- without requiring some debug endpoint to fetch the current list). What if we did something similar here? Idea would be:

> Periodically:
> 1. Fetch all firing VMs
> 2. Update the metrics
> 3. Log which ones are firing

Could have a new config parameter `RefreshFailingInterval` ;)

neonvm/config/controller/deployment.yaml
```diff
@@ -59,6 +59,8 @@ spec:
         # * cache.direct=on    - use O_DIRECT (don't abuse host's page cache!)
         # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
+        - "--failure-pending-period=1m"
+        - "--failing-refresh-interval=1m"
```
sharnoff, line 63: Could we have this slightly smaller? e.g. 15 seconds? asking only so that the logs can have higher granularity than metrics (and so that the metrics are delayed by at most 15s, rather than at most 1 minute)

sharnoff (approved):
Looks good! Left a few comments, all small things

sharnoff:
Pinged @Omrigan, agreed I'll finish up & merge this in order to get it into the release.

neonvm/controllers/metrics.go
```diff
@@ -135,39 +137,82 @@ type ReconcileSnapshot struct {
...
+) {
+	degraded := tracker.Degraded()
+	d.Metrics.failing.WithLabelValues(d.ControllerName, string(outcome)).
+		Set(float64(len(degraded)))
+
+	log.Info("Currently failing objects",
+		"outcome", outcome,
+		"count", len(degraded),
+		"objects", degraded,
+	)
```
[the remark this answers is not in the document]
sharnoff: Implemented in f37f754f3ca4eac5adc0382ec308893306dee757

neonvm/config/controller/deployment.yaml
sharnoff: Implemented in 5454f0f94f9c9b56e2ab0c971ed8701e46c9ddd4

PR #950, opened 2024-06-04
neonvm: Switch to kernel 6.1.92

sharnoff:
As of 2024-06-03, www.kernel.org says this is the most recent version of the 6.1.x LTS branch.

Basically just followed the instructions in neonvm/README.md, copying in linux-config-6.1.63 as .config, and .config back to /host once done with menuconfig.

Exploring this as a potential solution to neondatabase/cloud#10622. More info there.


sharnoff:
This seems to dramatically improve the situation with neondatabase/cloud#10622, although it doesn't completely fix it. 6.6.32 (#952) does _not_ do improve the situation (for some reason...) so we should leave that for now.

Omrigan (approved):

PR #952, opened 2024-06-04
neonvm: Switch to kernel 6.6.32

sharnoff:
As of 2024-06-03, www.kernel.org says this is the most recent version of the 6.6.x LTS branch.

Did the same thing as in #950. Like that PR, this is also in order to explore if it's a potential solution to neondatabase/cloud#10622.


sharnoff:
This did not improve the situation with neondatabase/cloud#10622, even though #950 (kernel 6.1.92) did. Closing this PR -- it's easy enough to resurrect at a future date, if we want.

PR #953, opened 2024-06-04
neonvm/kernel-builder: Improve caching across kernel versions

sharnoff:
Changing `KERNEL_VERSION` means you have re-run `apt-get install`, which can take a while. Hit this while working on https://www.notion.so/neondatabase/59c9e9b2619e4ccd8fbb99e9b0cb1a89.

~~TODO - double-check that it actually doesn't cache without this, and this works as intended.~~


neonvm/hack/kernel/Dockerfile.kernel-builder
```diff
@@ -1,11 +1,14 @@
...
+FROM ubuntu:23.10 AS check-arg
 ARG KERNEL_VERSION
+WORKDIR /build
 
 RUN set -e \
-    && echo "Build linux  kernel ${KERNEL_VERSION}" \
-    && test -n "${KERNEL_VERSION}"
+    && echo "Build linux kernel ${KERNEL_VERSION}" \
+    && test -n "${KERNEL_VERSION}" \
+    && echo "force this as a requirement for build-deps" > /build/arg-check-succeeded
```
Omrigan, line 9: Why not just put this part after the apt-get?

sharnoff: We could. I implemented it this way because I liked that previously, if there was an issue with `KERNEL_VERSION`, it would show up right away -- instead of needing to wait for `apt-get install` to run first (which can take a while).

Omrigan (approved):

PR #928, opened 2024-05-07
Upgrade k8s dependencies to 1.26.15 

SergeyMelnikov:
The plan is after all clusters are upgraded to rebase this on top of main and merge.

- currently based on main
- uses 1.26 k3d, kind, kuttl versions for tests
- cluster-autoscaler patch just works with the new version

cc: https://github.com/neondatabase/cloud/issues/7922

PR #956, opened 2024-06-06
Fix k3s cluster detection

SergeyMelnikov:
When upgrading k3d cluster to 1.26 in https://github.com/neondatabase/autoscaling/pull/928 k3d cluster nodes are getting renamed into `K3s v1.26.15+k3s1` instead of `K3s dev`.
We rely on this to find containerd socket in e2e tests, so they breaks.
Let's detect if we are running inside k3d cluster using `K3s` prefix.


neonvm/main.go
```diff
@@ -245,7 +246,7 @@ func checkIfRunningInK3sCluster(cfg *rest.Config) (bool, error) {
 	}
 
 	for _, node := range nodes.Items {
-		if node.Status.NodeInfo.OSImage == "K3s dev" {
+		if strings.Contains(node.Status.NodeInfo.OSImage, "K3s") {
```
sharnoff, line 249: WDYT about checking if it starts with k3s, rather than contains?

sharnoff (approved):
Nice- thanks! approval, pending one comment

PR #960, opened 2024-06-13
Use debug logging for normal operations in network setup

MMeent:
This reduces log rate by a good margin, while still keeping enough info in logs for debugging if and when errors occur.


stradig:
Hi Matthias, we were recently going through the list of old PRs in draft status. Is this one still of any value or could you close it?

PR #961, opened 2024-06-13
Flatten & unify component structure

sharnoff:
This basically encompasses two logical changes:

1. More fully integrating neonvm into the rest of the repository; and
2. Using a consistent directory structure for all components

The new structure is:

- All components get their own top-level directory (e.g. `autoscaler-agent`, `neonvm-controller`, `vm-builder`)
- Inside the directory, there is:
  - `cmd/` - directory containing main.go for the component
  - `Dockerfile` - for building the component
  - `*.yaml` - various YAML files for deploying, if applicable

Additionally, all code outside the `cmd` package are moved into the `pkg/` subdirectory. In particular, there is now `pkg/neonvm/`.

Also of note: `neonvm/hack/kernel` has been relocated to the new top-level directory `neonvm-kernel/`.

Under this structure, the user-visible changes are:

1. Introduction of new YAML: neonvm-controller.yaml
2. Introduction of new YAML: neonvm-vxlan-controller.yaml
3. neonvm.yaml now is *mostly* just the CRD.

The primary reasons for this change are:

1. Because it's been long enough since we [moved the neonvm repository here](https://github.com/neondatabase/autoscaling/pull/91) that we might as well make it more well-integrated.
2. Because as we look to potentially add components in other languages, we need a directory structure that supports it.

---

**Notes for review**: Splitting the YAMLs will be somewhat tricky to manage safely. I think we should have a clear idea of how we'll do that before merging.

If we go with this approach, we should aim to cut a release soon after merging.


pkg/neonvm/controllers/vm_controller.go
Omrigan, line 1: This move of `neonvm` files into common `pkg/` (vs keeping it in `neonvm/pkg`) hints towards erasing the boundary between `neonvm` and `autoscaling`. Personally, I am for that change, as I believe it might eventually help us to simplify the architecture and reduce the number of components. 

But, we should agree on this explicitly.

Omrigan (commented):
Regardless of the comment, I support this change, let's get it done and merged!

sharnoff:
Will be affected by #988 (and also #989, if we go that route). I plan to rebase after those are merged.

sharnoff:
One issue in CI -- for some reason we're incorrectly using `runner:dev` as the image in CI, although everything appears to work ok locally.

Haven't fully investigated what's up there, but otherwise this PR should be good.

ARCHITECTURE.md
```diff
@@ -81,48 +84,49 @@ discussed more in the [high-level consequences] section below.
...
+At a high level, each component gets its own directory and resulting YAML for its deployment, where
+applicable.
+
+These are:
+
+* `autoscale-scheduler` — the scheduler (with our plugin)
+* `autoscaler-agent`
+* `cluster-autoscaler` — patch for building a NeonVM-compatible [cluster-autoscaler]
+* `cluster-autoscaler` — our patch to cluster-autoscaler so it understands the resource usage from
+    NeonVM VMs
```
Omrigan, line 96:
```suggestion
```

Duplicate.

ARCHITECTURE.md
```diff
@@ -36,20 +36,23 @@ This isn't the only architecture document. You may also want to look at:
 
 ## High-level overview
 
-At a high level, this repository provides two components:
+At a high level, this repository provides three components:
```
Omrigan, line 39:
```suggestion
At a high level, this repository provides 5 components with non-trivial amount of code:
```
1. neonvm-runner
2. neonvm-controller
3. autoscale-scheduler (plugin)
4. autoscaler-agent (side node: those autoscale vs autoscaler vs autoscaling produce mildly annoying typos)
5. vm-builder

I think it makes navigating the repo easier.

README-NeonVM.md
Omrigan, line 1: why this move?

neonvm-vxlan-controller/Dockerfile
Omrigan, line 1: I am tempted to suggest `neonvm-vxlan-controller` -> `neonvm/vxlan-controller`

README-NeonVM.md
sharnoff: NeonVM is broader than just the CRD (e.g., neonvm-controller now has its own top-level directory). So I figured that the README for it should be at the highest common directory.

neonvm-vxlan-controller/Dockerfile
sharnoff: Could you elaborate? (or, why just `neonvm-vxlan-controller` and not also `neonvm-controller` ?)

pkg/neonvm/controllers/vm_controller.go
petuhovskiy: I'm +1 for "erasing the boundary between neonvm and autoscaling", so probably it's a consensus already :)

petuhovskiy (approved):
I'm +1 for everything, but not sure how to ensure that all docs/yamls/scripts/tests won't have a broker path in them. Let's merge this sooner rather than later.

ARCHITECTURE.md
sharnoff: Good catch, thanks.

ARCHITECTURE.md
sharnoff: I've updated it now. LMK what you think!

neonvm-vxlan-controller/Dockerfile
Omrigan: Because it has an insignificant amount of code and could tucked away from top-level.

README-NeonVM.md
Omrigan: I guess the reasoning the same as for vxlan controller -> so that there is less stuff in top level dir.

Omrigan (approved):

ARCHITECTURE.md
Omrigan: Good!

PR #963, opened 2024-06-16
neonvm: hacking on adding virtio-fs support

sharnoff:
virtio-fs v1.11.0 supports live migration (albeit, with some major caveats).

There's maybe some cool opportunities there - figured it's worth playing around with.

**Current status**: using current vm-deploy fails with QEMU saying

    Device vhost-user-fs-device is not migratable, but --only-migratable was specified

We're using an old version of QEMU, but I'm not actually sure if support for migratable vhost-user fs devices has been released yet (if it's even been merged at all!).

More digging required...

See also https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/206
See also https://www.notion.so/neondatabase/25e6aa0142e94e8fab15fbb2ba237757

---

Also, included some fixes to issues I noticed; those will be split off into separate PRs.


sharnoff:
Closing this, as we didn't end up doing anything here.

PR #705, opened 2023-12-21
WIP [NeonVM] virtio-mem for VM memory management

cicdteam:
### this PR:
- add ability to use virtio-mem device for VM memory management

### how to specify memory in VM psec

old way
```yaml
spec:
  guest:
    memorySlotSize: 1Gi
    memorySlots:
      min: 1
      max: 4
      use: 2
```

new (virtio-mem) way
```yaml
spec:
  guest:
    memory:
      min: 1Gi
      max: 4Gi
      use: 2Gi
```
 > [!NOTE]
> `.spec.memory.use` can be any value from `.min` to `.max` and must be a multiple of **8MiB** (hardcoded `block size` in virtio-mem settings), e.g. `1536Mi`, `1776Mi`, `3Gi` and so on  

> [!IMPORTANT]
> The maximum size of `.spec.memory.max` can be **`256Gi`** (32k blocks of 8MiB each)

### remaining tasks
- [ ] implement autoscaling for `virtio-mem` managed memory
- [x] implement VM migration with `virtio-mem` device
- [x] check boundary for `.spec.memory.use` in VM controller webhook

PR #970, opened 2024-06-18
neonvm: Add support for virtio-mem

### LINKED ISSUE #117: Impossible to scale memory more than x60
Original title: "vhost backends exhausted"

## Steps to reproduce
related to large VMs

## Logs, links
-

### LINKED ISSUE #967: Bug: neonvm-controller spams new pod names when VM fails and RestartPolicy = Never
## Environment

Local development (not staging/prod because we have RestartPolicy = Always there)

## Steps to reproduce

Using `vm-deploy.yaml`, set `spec.restartPolicy: Never` and then `kill -6 1` inside the runner pod (`kill -9` doesn't work).

```diff
diff --git a/vm-deploy.yaml b/vm-deploy.yaml
index 09588f6..7d84d67 100644
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -13,6 +13,7 @@ metadata:
 spec:
   schedulerName: autoscale-scheduler
   enableSSH: true
+  restartPolicy: Never
   guest:
     cpus: { min: 0.25, use: 0.25, max: 1.25 }
     memorySlotSize: 1Gi
```

```sh
kubectl exec postgres16-disk-test-<SUFFIX> -- kill -6 1
```

## Expected result

The pod name should either:

1. Remain what it was when the VM failed; or
2. Remain set to `""` until the VM is deleted

There are good arguments for both cases here, I think.

## Actual result

If you `kubectl get -w neonvm` while using the reproduction steps above, you'll see a rapid stream of changes that looks something like:

```sh-session
$ kubectl get -w neonvm
NAME                   CPUS   MEMORY   POD                          EXTRAIP   STATUS    RESTARTS   AGE
postgres16-disk-test   250m   1Gi      postgres16-disk-test-j5sb8             Running              14s
postgres16-disk-test   250m   1Gi      postgres16-disk-test-j5sb8             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-mpl6w             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-pdh8f             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-zzzm6             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-9qfcm             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-nq4vf             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-dwml5             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-r78fc             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-965jp             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-fqmxd             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-zgnrv             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-bcpw7             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-ck9kz             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-m9n97             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-h46sz             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-x49h7             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-dqhss             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-mc6xm             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-zh682             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-r87cv             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-ldmhd             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-mqfrb             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-gs8pc             Failed               38s
postgres16-disk-test                                                          Failed               38s
postgres16-disk-test                   postgres16-disk-test-8twbx             Failed               38s
```

## Other logs, links

- ...

sharnoff:
Adapted from #705.

This implementation is focused on making virtio-mem a drop-in replacement for our existing DIMM hotplug -based memory scaling.

As such, the externally visible differences are:

1. VirtualMachines have a new optional field: `.spec.guest.memoryProvider`
2. VirtualMachines have a new optional field: `.status.memoryProvider`
3. VirtualMachine `.status.memorySize` changes in smaller increments if using virtio-mem.

This change introduces the concept of a "memory provider", which is either "DIMMSlots" or "VirtioMem".

If a VM is created without setting the memory provider in the spec, the controller will use the default one, which is specified by its `--default-memory-provider` flag.
This will only persist for the lifetime of the runner pod. If the VM is restarted, then it may change to a new default.

Additionally, because our usage of virtio-mem has 8MiB block sizes, we actually *don't* default to virtio-mem if the VM was created with some odd memory slot size that isn't divisible by 8MiB.

The memory provider of the current runner pod is stored in the new status field. For old VMs created before this change, the status field is set to DIMMSlots iff the pod name is set without the status field.

Also, we allow mutating `.spec.guest.memoryProvider` IFF it is being set equal to the current `.status.memoryProvider`, in case we we want old VMs to keep their dimm slots after switching to virtio-mem by default.

---

Changes to split out from this PR:

1. #971
2. #973
3. #974

On top of those, there's a couple places where code was indented that should probably be split into its own commit so it can be properly added to .git-blame-ignore-revs :)

---

Remaining testing for this PR before merging:

- [x] Test that backwards compatibility works as expected
   - [x] Old VMs are updated while running as described above
   - [x] Old VMs switch to the new default on restart
- [x] Test rollback safety: that it's ok to go between the last release and this PR (with `--default-memory-provider=DIMMSlots`).
   - Note: If `--default-memory-provider=VirtioMem` when this is released, we cannot make it rollback-safe.
- [ ] ~~Test that updating `.spec.guest.memoryProvider` works as expected~~

---

Future/follow-up work:

1. #981
2. Switch to virtio-mem by default
3. Remove support for DIMM hotplug
4. Move from representing memory as "memory slots" towards the implementation
5. Fix the raciness issue with migration target pod creation.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/samber/lo"
 
@@ -172,6 +173,8 @@ type Guest struct {
 	// +optional
 	MemorySlots MemorySlots `json:"memorySlots"`
 	// +optional
+	MemoryProvider *MemoryProvider `json:"memoryProvider,omitempty"`
+	// +optional
 	RootDisk RootDisk `json:"rootDisk"`
 	// Docker image Entrypoint array replacement.
 	// +optional
@@ -192,6 +195,24 @@ type Guest struct {
 	// Cannot be updated.
 	// +optional
 	Settings *GuestSettings `json:"settings,omitempty"`
+}
+
+const virtioMemBlockSizeBytes = 8 * 1024 * 1024 // 8 MiB
+
+// ValidateForMemoryProvider returns an error iff the guest memory settings are invalid for the
+// MemoryProvider.
+//
+// This is used in two places. First, to validate VirtualMachine object creation. Second, to handle
+// the defaulting behavior for VirtualMachines that would be switching from DIMMSlots to VirtioMem
+// on restart. We place more restrictions on VirtioMem because we use 8MiB block sizes, so changing
+// to a new default can only happen if the memory slot size is a multiple of 8MiB.
+func (g Guest) ValidateForMemoryProvider(p MemoryProvider) error {
+	if p == MemoryProviderVirtioMem {
+		if g.MemorySlotSize.Value()%virtioMemBlockSizeBytes != 0 {
+			return fmt.Errorf("memorySlotSize invalid for memoryProvider VirtioMem: must be a multiple of 8Mi")
+		}
+	}
+	return nil
 }
 
 type GuestSettings struct {
@@ -353,6 +374,29 @@ type MemorySlots struct {
 	Use int32 `json:"use"`
 }
 
+// +kubebuilder:validation:Enum=DIMMSlots;VirtioMem
+type MemoryProvider string
+
+const (
+	MemoryProviderDIMMSlots MemoryProvider = "DIMMSlots"
+	MemoryProviderVirtioMem MemoryProvider = "VirtioMem"
+)
+
+// FlagFunc is a parsing function to be used with flag.Func
+func (p *MemoryProvider) FlagFunc(value string) error {
+	possibleValues := []string{
+		string(MemoryProviderDIMMSlots),
+		string(MemoryProviderVirtioMem),
+	}
+
+	if !slices.Contains(possibleValues, value) {
+		return fmt.Errorf("Unknown MemoryProvider %q, must be one of %v", value, possibleValues)
+	}
+
+	*p = MemoryProvider(value)
+	return nil
+}
+
 type RootDisk struct {
 	Image string `json:"image"`
 	// +optional
@@ -485,6 +529,8 @@ type VirtualMachineStatus struct {
 	// +optional
 	MemorySize *resource.Quantity `json:"memorySize,omitempty"`
 	// +optional
+	MemoryProvider *MemoryProvider `json:"memoryProvider,omitempty"`
+	// +optional
 	SSHSecretName string `json:"sshSecretName,omitempty"`
 }
 
@@ -550,6 +596,7 @@ func (vm *VirtualMachine) Cleanup() {
 	vm.Status.Node = ""
 	vm.Status.CPUs = nil
 	vm.Status.MemorySize = nil
+	vm.Status.MemoryProvider = nil
 }
 
 func (vm *VirtualMachine) HasRestarted() bool {
--- a/neonvm/apis/neonvm/v1/virtualmachine_webhook.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_webhook.go
@@ -61,6 +61,13 @@ func (r *VirtualMachine) ValidateCreate() error {
 			r.Spec.Guest.CPUs.Max)
 	}
 
+	// validate .spec.guest.memorySlotSize w.r.t. .spec.guest.memoryProvider
+	if r.Spec.Guest.MemoryProvider != nil {
+		if err := r.Spec.Guest.ValidateForMemoryProvider(*r.Spec.Guest.MemoryProvider); err != nil {
+			return fmt.Errorf(".spec.guest: %w", err)
+		}
+	}
+
 	// validate .spec.guest.memorySlots.use and .spec.guest.memorySlots.max
 	if r.Spec.Guest.MemorySlots.Use < r.Spec.Guest.MemorySlots.Min {
 		return fmt.Errorf(".spec.guest.memorySlots.use (%d) should be greater than or equal to the .spec.guest.memorySlots.min (%d)",
@@ -124,6 +131,10 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) error {
 		{".spec.guest.cpus.max", func(v *VirtualMachine) any { return v.Spec.Guest.CPUs.Max }},
 		{".spec.guest.memorySlots.min", func(v *VirtualMachine) any { return v.Spec.Guest.MemorySlots.Min }},
 		{".spec.guest.memorySlots.max", func(v *VirtualMachine) any { return v.Spec.Guest.MemorySlots.Max }},
+		// nb: we don't check memoryProvider here, so that it's allowed to be mutable as a way of
+		// getting flexibility to solidify the memory provider or change it across restarts.
+		// ref https://github.com/neondatabase/autoscaling/pull/970#discussion_r1644225986
+		{".spec.guest.memoryProvider", func(v *VirtualMachine) any { return v.Spec.Guest.MemoryProvider }},
 		{".spec.guest.ports", func(v *VirtualMachine) any { return v.Spec.Guest.Ports }},
 		{".spec.guest.rootDisk", func(v *VirtualMachine) any { return v.Spec.Guest.RootDisk }},
 		{".spec.guest.command", func(v *VirtualMachine) any { return v.Spec.Guest.Command }},
--- a/neonvm/controllers/vm_controller.go
+++ b/neonvm/controllers/vm_controller.go
@@ -719,41 +719,10 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		r.updateVMStatusCPU(ctx, vm, vmRunner, pluggedCPU, cgroupUsage)
 
 		// do hotplug/unplug Memory
-		memSlotsMin := vm.Spec.Guest.MemorySlots.Min
-		targetSlotCount := int(vm.Spec.Guest.MemorySlots.Use - memSlotsMin)
-
-		realSlots, err := QmpSetMemorySlots(ctx, vm, targetSlotCount, r.Recorder)
-		if realSlots < 0 {
-			return err
-		}
-
-		if realSlots != int(targetSlotCount) {
-			log.Info("Couldn't achieve desired memory slot count, will modify .spec.guest.memorySlots.use instead", "details", err)
-			// firstly re-fetch VM
-			if err := r.Get(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace}, vm); err != nil {
-				log.Error(err, "Unable to re-fetch VirtualMachine")
-				return err
-			}
-			memorySlotsUseInSpec := vm.Spec.Guest.MemorySlots.Use
-			memoryPluggedSlots := memSlotsMin + int32(realSlots)
-			vm.Spec.Guest.MemorySlots.Use = memoryPluggedSlots
-			if err := r.tryUpdateVM(ctx, vm); err != nil {
-				log.Error(err, "Failed to update .spec.guest.memorySlots.use",
-					"old value", memorySlotsUseInSpec,
-					"new value", memoryPluggedSlots)
-				return err
-			}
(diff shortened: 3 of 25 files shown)


Omrigan (commented):
Just a couple thoughts

sharnoff:
Cleaned up the commits in this PR -- added single commits for #973 and #974, and split some indenting into a separate commit. May move more indenting, need to see.

Anyways, hopefully should make the final commit easier to review :crossed_fingers:

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -558,6 +604,7 @@ func (vm *VirtualMachine) Cleanup() {
 	vm.Status.Node = ""
 	vm.Status.CPUs = nil
 	vm.Status.MemorySize = nil
+	vm.Status.MemoryProvider = nil
```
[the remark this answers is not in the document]
sharnoff: There's a couple reasons I went for this approach. This was my thought process, lmk what you think:

1. We need _some_ way to know how to scale the memory on the runner pod. This can either be a field in the spec or status.
2. We don't want to _need_ the control plane to set it; that way, we get control over rollout. So, as you mentioned -- either spec or status, but we have to auto-set _something_.
3. IMO: We should have an escape hatch, once virtio-mem is deployed. I.e., for a particular VM, we should be able to say "hey, please use DIMM slots instead". Eventually this should be a project/endpoint/compute setting (until we decide it's stable enough), and then we can remove all the old code. This belongs in the spec.
4. IMO: Once the VM is up and running, we _shouldn't_ allow changes to MemoryProvider setting. Actually, it needs to be _before_ the VM is up and running, to prevent race conditions. This can either be done with a status field (because special permissions are required), or with webhook update validation (i.e. only allow updates if pending and the podName wasn't set). We've had [issues around rollout of new fields + webhook validation in the past](https://www.notion.so/neondatabase/d66861b2d8a8444887282d8558428791?pvs=4#3ddf05468a8a44378b4c9931ad3dca95), so I'm hesitant to go that route.
5. IMO: _it would be nice_ if forcing VMs to pick up a change in MemoryProvider could just be done by deleting the pods. Otherwise we have to do object "delete + recreate with modified spec" (which is kind of a pain). I'm thinking of e.g. "oh no, there's an issue with virtio-mem, we need to switch out all affected runner pods". This requires a separate field in the status, if we also want to respect the spec.
6. IMO: _it would be nice_ if we could avoid neonvm-controller updating the spec (because I view "spec" as kind of "here's what the user asked", so to me it feels a little odd to update the spec. but... we have defaults anyways, so maybe this is analogous to that?)

From going through it again, I think it would be _fine_ to just use the spec, but I do also like having the status be separate. With (5) I'm probably just being paranoid.

WDYT?

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -187,6 +195,19 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) error {
 		}
 	}
 
+	// Only allow changing .spec.guest.memoryProvider if it was previously nil, and the new version
+	// is equal to .status.memoryProvider.
+	if !reflect.DeepEqual(before.Spec.Guest.MemoryProvider, r.Spec.Guest.MemoryProvider) {
```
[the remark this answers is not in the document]
sharnoff: Yeah fair enough. Done in 072fe6a130bb1884f79319ec68d0452d92935c4b. Won't squash until the discussion below is resolved, because IIUC we'll have to make this mutable if we put MemoryProvider in the spec, to allow neonvm-controller to set it

sharnoff:
Did backwards-compatibility testing locally (started a VM, upgraded the version, checked status was as expected, assuming that our existing e2e tests can take it from there). Found & fixed a bug.
Also checked rollback safety; yeah, seems fine :)
One interesting thing, posted it here: https://neondb.slack.com/archives/C039YKBRZB4/p1718950172991649

neonvm/controllers/vm_controller.go
```diff
@@ -715,41 +729,71 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+				if targetVirtioMemSize < previousTarget {
+					reason = "ScaleDown"
+				}
+				r.Recorder.Eventf(vm, "Normal", reason, "Set virtio-mem size for %v total memory", goalTotalSize)
+			}
+
+			// Maybe we're already using the amount we want?
+			// Update the status to reflect the current size - and if it matches goalTotalSize, ram
+			// is done.
+			currentTotalSize, err := QmpGetMemorySize(QmpAddr(vm))
```
Omrigan, line 764: I think this is a redundant check, we could get the current memory size from where it is set. 

But no worries, I'll clean it up in #943

neonvm/controllers/vm_controller.go
```diff
@@ -715,41 +729,71 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+				reason := "ScaleUp"
+				if targetVirtioMemSize < previousTarget {
+					reason = "ScaleDown"
+				}
+				r.Recorder.Eventf(vm, "Normal", reason, "Set virtio-mem size for %v total memory", goalTotalSize)
+			}
+
+			// Maybe we're already using the amount we want?
+			// Update the status to reflect the current size - and if it matches goalTotalSize, ram
+			// is done.
```
Omrigan, line 763:
```suggestion
			// scaling is done.
```

neonvm/controllers/vm_controller.go
```diff
@@ -812,6 +861,23 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 	return nil
 }
 
+func pickMemoryProvider(config *ReconcilerConfig, vm *vmv1.VirtualMachine) vmv1.MemoryProvider {
+	if p := vm.Spec.Guest.MemoryProvider; p != nil {
+		return *p
+	}
+	if p := vm.Status.MemoryProvider; p != nil {
```
Omrigan, line 868: Won't this prevent us from changing from one provider to the other by deleting the pod, because the old value be preserved through Status?

neonvm/controllers/vm_qmp_queries.go
```diff
@@ -333,6 +333,65 @@ type QMPRunner interface {
...
+// If the new target size is equal to the previous one, this function does nothing but query the
+// target.
+func QmpSetVirtioMem(vm *vmv1.VirtualMachine, targetVirtioMemSize int64) (previous int64, _ error) {
+	// Note: The virtio-mem device only exists when max mem != min mem.
+	// So if min == max, we should just short-cut, skip the queries, and say it's all good.
+	// Refer to the instantiation in neonvm-runner for more.
+	if vm.Spec.Guest.MemorySlots.Min == vm.Spec.Guest.MemorySlots.Max {
+		// if target size is non-zero even though min == max, something went very wrong
+		if targetVirtioMemSize != 0 {
+			panic(fmt.Errorf(
```
Omrigan, line 347:
```suggestion
			panic(fmt.Sprintf(
```

Usually panic's arg is the string.

tests/e2e/autoscaling.virtio-mem/01-assert.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 90
+---
+apiVersion: vm.neon.tech/v1
+kind: VirtualMachine
+metadata:
+  name: example
+status:
```
Omrigan, line 9: can we assert on the provider here?

Omrigan (approved):
Looks good overall

neonvm/controllers/vm_controller.go
sharnoff: I _think_ in testing I actually saw this value change in between QMP queries in the same reconcile cycle. Not totally sure though...

neonvm/controllers/vm_controller.go
sharnoff: the tl;dr on that is that we reset `.status.memoryProvider` in the same place we reset `.status.podName` (there's some method on `VirtualMachineStatus` that does it for us).

I tested it locally and it works as expected :)

neonvm/controllers/vm_controller.go
```diff
@@ -359,16 +366,23 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		// VirtualMachine just created, change Phase to "Pending"
 		vm.Status.Phase = vmv1.VmPending
 	case vmv1.VmPending:
-		// Generate runner pod name
+		// Generate runner pod name and set desired memory provider. After this, we know that
+		// MemoryProvider will be set, because earlier (above the 'switch vm.Status.Phase'), we set
+		// the MemoryProvider iff PodName was set but it wasn't.
```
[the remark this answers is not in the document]
sharnoff: Nice - I like that a lot better :)

tests/e2e/autoscaling.virtio-mem/01-assert.yaml
sharnoff: Done

neonvm/controllers/vm_controller.go
```diff
@@ -722,38 +722,38 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
-		if realSlots < 0 {
-			return err
-		}
-
-		if realSlots != int(targetSlotCount) {
-			log.Info("Couldn't achieve desired memory slot count, will modify .spec.guest.memorySlots.use instead", "details", err)
-			// firstly re-fetch VM
-			if err := r.Get(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace}, vm); err != nil {
-				log.Error(err, "Unable to re-fetch VirtualMachine")
+			realSlots, err := QmpSetMemorySlots(ctx, vm, targetSlotCount, r.Recorder)
```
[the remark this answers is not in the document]
sharnoff: Extracted in d2b8d897c917835790acb35b157e2833a6109d0d, replaced the indent commit with 1f05547e8c0f88e361229556fcd5b695a69b40d2 (splitting the first commit).

PR #971, opened 2024-06-18
neonvm: Make spec min/max/use resources required

sharnoff:
This already was de-facto the case, which made much of the logic around handling "but what if .max is nil?" somewhat tricky to reason about.

Extracted from #970 and fleshed out here.

All our staging & production clusters have VMs meeting this criteria, with the notable exception of a single, already-broken VM in dev-us-east-2-beta (it's a long story).

More here: https://neondb.slack.com/archives/C03TN5G758R/p1718742308251569

---

Checked each cluster with:

```sh
kubectl get neonvm -o json \
  |  jq -r '.items[] | select(.spec.guest | [.cpus.max,.cpus.min,.memorySlots.max,.memorySlots.min] | map(select(. == null)) | (length != 0)) | .metadata.name'
```


neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan, line 1: Pls change "min/max" to "min/max/use" in the PR title.

Omrigan (approved):
Nice! It's always pleasant to see when the type checker does the work instead of us.

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Ah, good catch, thanks :)

PR #972, opened 2024-06-18
Update Rust: 1.78 -> 1.79

sharnoff:
Good to keep in sync with `neondatabase/neon` because we pull in vm-monitor.
Corresponding PR there was neondatabase/neon#8048.


Omrigan (approved):
Perhaps a more long-term solution could be to have a `neondatabase/rust` base image, to keep it always in sync. Same for go.

PR #974, opened 2024-06-19
neonvm-controller: Emit CpuInfo/MemoryInfo every reconcile

sharnoff:
Extracted from #970.

Currently, when leaving the Scaling phase, we emit events like

```
MemoryInfo  VirtualMachine vm-foo uses 4GiB memory
CpuInfo     VirtualMachine vm-foo uses 1 cpu cores
```

To provide more up-to-date information in the event log, we can emit these events every time `.status.{cpus,memorySize}` changes.

This is relevant for #970 because with virtio-mem, the current state of the memory is performed asychronously and so otherwise wouldn't be available in the logs.


Omrigan (approved):
> we can emit these events every time .status.{cpus,memorySize} changes.

This was the case before this PR. The thing that changed is when we update the status.

PR #978, opened 2024-06-19
k3d: Pass through auth for docker.io in CI

### LINKED ISSUE #975: Docker rate limiting causes e2e-tests to fail
## Problem

We sometimes get rate limited by docker in the e2e tests. When this happens, image pulls fail - and therefore the entire e2e test job fails as a result.

As a recent example, I saw a couple cases where deploying the components failed with:

```
Waiting for daemon set "neonvm-device-plugin" rollout to finish: 0 of 3 updated pods are available...
Error: The action 'deploy components' has timed out after 3 minutes.
```

and when looking at the events, we see:

```
LAST SEEN   TYPE      REASON           OBJECT                           MESSAGE
2m50s       Normal    Scheduled        pod/neonvm-device-plugin-blskl   Successfully assigned neonvm-system/neonvm-device-plugin-blskl to k3d-neonvm-agent-0
2m49s       Normal    AddedInterface   pod/neonvm-device-plugin-blskl   Add eth0 [10.0.0.154/32] from cilium
75s         Normal    Pulling          pod/neonvm-device-plugin-blskl   Pulling image "squat/generic-device-plugin"
72s         Warning   Failed           pod/neonvm-device-plugin-blskl   Failed to pull image "squat/generic-device-plugin": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/squat/generic-device-plugin:latest": failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/squat/generic-device-plugin/manifests/sha256:ba6f0b4cf6c858d6ad29ba4d32e4da11638abbc7d96436bf04f582a97b2b8821: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
72s         Warning   Failed           pod/neonvm-device-plugin-blskl   Error: ErrImagePull
57s         Warning   Failed           pod/neonvm-device-plugin-blskl   Error: ImagePullBackOff
45s         Normal    BackOff          pod/neonvm-device-plugin-blskl   Back-off pulling image "squat/generic-device-plugin"
```

ref

- https://github.com/neondatabase/autoscaling/actions/runs/9574243656/job/26397291740
- https://github.com/neondatabase/autoscaling/actions/runs/9574245249/job/26397290700

## Potential solutions

Maybe we can specify credentials for dockerhub with this registries configuration file? https://k3d.io/v5.6.0/usage/registries/

We _might_ also want to look into implementing this for kind, but that's lower priority because we aren't regularly using it in CI.

sharnoff:
Fixes #975.

~~TODO: figure out why it isn't working~~


sharnoff:
Turns out it's not quite fixed, unfortunately: https://github.com/neondatabase/autoscaling/actions/runs/9576387644/job/26402955386#step:15:2360

sharnoff:
Turns out it just needed to be `registry-1.docker.io` instead of `docker.io`, see more here: https://devops.stackexchange.com/a/16154

.github/workflows/e2e-test.yaml
```diff
@@ -142,7 +142,28 @@ jobs:
           username: ${{ secrets.NEON_DOCKERHUB_USERNAME }}
           password: ${{ secrets.NEON_DOCKERHUB_PASSWORD }}
 
+      # https://docs.k3s.io/installation/private-registry#registries-configuration-file
+      # https://github.com/neondatabase/autoscaling/issues/975
+      - name: set k3d registries.yaml
+        # TODO: Implement an equivalent for kind?
```
Omrigan, line 148: Maybe just drop kind support?

Omrigan (approved):

.github/workflows/e2e-test.yaml
sharnoff: I still use kind locally because it says "have a nice day" :sweat_smile:

Given that k3d _is_ a little bit different from a more normal k8s distribution, I'd rather we keep kind around for now -- if only so we make sure that we don't implement something that works only in k3d (but not EKS).

WDYT?

Omrigan: Sure thing!

PR #959, opened 2024-06-11
temporary PR to help investigate ZONE_NORMAL

sharnoff:
It's more convenient to have CI build & push the images than do it myself

ref: https://www.notion.so/neondatabase/59c9e9b2619e4ccd8fbb99e9b0cb1a89?pvs=4#dcf1ea2e65464f618db917b77e3dcf42
ref: https://www.notion.so/neondatabase/59c9e9b2619e4ccd8fbb99e9b0cb1a89?pvs=4#8df8a3aa3cf64ecb8aa68266913b7182
ref: https://www.notion.so/neondatabase/59c9e9b2619e4ccd8fbb99e9b0cb1a89?pvs=4#3fec5babb1474eeebffe72619728ea68

PR #981, opened 2024-06-20
neonvm: Use memory_hotplug.online_policy=auto-movable for virtio-mem

### LINKED ISSUE #117: Impossible to scale memory more than x60

sharnoff:
In short: Currently we hotplug all memory as ZONE_MOVABLE. This has known issues because there are theoretical and practical limits on the ratio between ZONE_MOVABLE and ZONE_NORMAL. With `memory_hotplug.online_policy=auto-movable`, we can use the config value `memory_hotplug.auto_movable_ratio` to set a ratio of MOVABLE:NORMAL to uphold during hotplug and the kernel will Do The Right Thing™.

This change adds a new flag to `neonvm-controller` and `neonvm-runner` to set the value of this ratio: `-memhp-auto-movable-ratio`.

However, please note: The new behavior is only available for VMs using virito-mem because `memory_hotplug.auto_movable_ratio` does not count hotplugged ZONE_NORMAL DIMM slots towards the ratio, so IMO it's not worth changing the existing behavior. [1] [2] [3]

This is the productionized version of the experimentation from #959. For more on that, see [4].

[1]: https://docs.kernel.org/admin-guide/mm/memory-hotplug.html#module-parameters
[2]: https://lwn.net/Articles/865618/
[3]: https://neondb.slack.com/archives/C03TN5G758R/p1718225978299459?thread_ts=1718158200.324759
[4]: https://www.notion.so/neondatabase/59c9e9b2619e4ccd8fbb99e9b0cb1a89

---

**Notes for review:** This is split into two commits; the first is a small refactoring in neonvm-runner to make the diff more clear. I _think_ I'll end up doing rebase-and-merge.

~~**Also, this PR builds on #970, and must not be merged before it.**~~

This PR fixes #117 and also fixes neondatabase/cloud#10622.

### DIFF
--- a/neonvm/config/controller/deployment.yaml
+++ b/neonvm/config/controller/deployment.yaml
@@ -60,6 +60,7 @@ spec:
         # * cache.no-flush=on  - ignores disk flush operations (not needed; our disks are ephemeral)
         - "--qemu-disk-cache-settings=cache.writeback=on,cache.direct=on,cache.no-flush=on"
         - "--default-memory-provider=DIMMSlots"
+        - "--memhp-auto-movable-ratio=801" # for virtio-mem, set memory_hotplug.auto_movable_ratio=801
         - "--failure-pending-period=1m"
         - "--failing-refresh-interval=15s"
         env:
--- a/neonvm/controllers/config.go
+++ b/neonvm/controllers/config.go
@@ -33,6 +33,14 @@ type ReconcilerConfig struct {
 	// new VMs (or, when old ones restart) if nothing is explicitly set.
 	DefaultMemoryProvider vmv1.MemoryProvider
 
+	// MemhpAutoMovableRatio specifies the value that new neonvm-runners will set as the
+	// kernel's 'memory_hotplug.auto_movable_ratio', iff the memory provider is virtio-mem.
+	//
+	// This value is passed directly to neonvm-runner as the '-memhp-auto-movable-ratio' flag.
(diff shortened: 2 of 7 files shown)


neonvm/controllers/vm_controller_test.go
```diff
@@ -111,6 +111,7 @@ var _ = Describe("VirtualMachine controller", func() {
 					MaxConcurrentReconciles: 1,
 					QEMUDiskCacheSettings:   "cache=none",
 					DefaultMemoryProvider:   vmv1.MemoryProviderDIMMSlots,
+					MemhpAutoMovableRatio:   "301",
```
Omrigan, line 114: Should have 801 everywhere?

Omrigan (approved):

neonvm/controllers/vm_controller_test.go
sharnoff: My thinking here was that 301 is a more "safe" default, but we know we can get away with 801 in practice (kind of like the cache=none situation). (so, this 301 here matches the default of 301 for the neonvm-controller flag, which we then always override)

Thoughts?

Omrigan: It's just that this PR request has both 301 and 801 in different places. Just wondering is that is intentional, ok with me either way.

PR #982, opened 2024-06-20
Upload autoscaling images to ECR

SergeyMelnikov:
Test in https://github.com/neondatabase/autoscaling/actions/runs/9595639251

cc: https://github.com/neondatabase/cloud/issues/10177

### DIFF
--- a/neonvm/controllers/vm_controller.go
+++ b/neonvm/controllers/vm_controller.go
@@ -323,16 +323,6 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		vm.Status.SSHSecretName = fmt.Sprintf("ssh-neonvm-%s", vm.Name)
 	}
 
-	// Generate runner pod name
-	if len(vm.Status.PodName) == 0 {
-		vm.Status.PodName = names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", vm.Name))
-		// Update the .Status on API Server to avoid creating multiple pods for a single VM
-		// See https://github.com/neondatabase/autoscaling/issues/794 for the context
-		if err := r.Status().Update(ctx, vm); err != nil {
-			return fmt.Errorf("Failed to update VirtualMachine status: %w", err)
-		}
-	}
-
 	switch vm.Status.Phase {
 
 	case "":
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images.yaml
```diff
@@ -219,6 +223,42 @@ jobs:
...
+          username: ${{ secrets.DEV_GHA_RUNNER_LIMITED_AWS_ACCESS_KEY_ID }}
+          password: ${{ secrets.DEV_GHA_RUNNER_LIMITED_AWS_SECRET_ACCESS_KEY }}
+
+      - name: Login to prod ECR
+        if: ${{ format('{0}', inputs.upload-to-ecr) == 'true' }}
+        uses: docker/login-action@v3
+        with:
+          registry: 093970136003.dkr.ecr.eu-central-1.amazonaws.com
+          username: ${{ secrets.PROD_GHA_RUNNER_LIMITED_AWS_ACCESS_KEY_ID }}
+          password: ${{ secrets.PROD_GHA_RUNNER_LIMITED_AWS_SECRET_ACCESS_KEY }}
```
sharnoff, line 240: Let's do this at the top, after the initial docker login? (unless we think it's more reliable to delay login, because at least we'd push to dockerhub if login to ECR isn't working)

Related: _may_ also want to consider whether we should "build and push" each image or just build locally and _then_ push everything in one go (allowing us to push to other registries if dockerhub is failing...)

Incremental benefit of that change is probably small though; happy to leave it for a follow-up.

.github/workflows/build-images.yaml
```diff
@@ -219,6 +223,42 @@ jobs:
...
+            autoscale-scheduler \
+            autoscaler-agent \
+            cluster-autoscaler-neonvm \
+            ; do
+            echo Copy ${image}:${{ inputs.tag }} to dev ECR
+            docker buildx imagetools create -t 369495373322.dkr.ecr.eu-central-1.amazonaws.com/${image}:${{ inputs.tag }} \
+                                               neondatabase/${image}:${{ inputs.tag }}
+            echo Copy ${image}:${{ inputs.tag }} to prod ECR
+            docker buildx imagetools create -t 093970136003.dkr.ecr.eu-central-1.amazonaws.com/${image}:${{ inputs.tag }} \
+                                               neondatabase/${image}:${{ inputs.tag }}
```
sharnoff, line 259: Can we make the registries an environment variable? either for just this step, or for the entire job

Just thinking of how it can be cleaner :)

sharnoff (approved):
lgtm, nice! Approved, pending a couple comments

PR #983, opened 2024-06-20
Implement scaling latency metrics through revisions

### LINKED ISSUE #594: Epic: Scaling latency metrics
## Motivation

Two reasons:

1. We're currently flying blind w.r.t. how long scaling takes
2. Scaling latency should be part of any autoscaling SLOs

## DoD

We should have histogram metrics recording:
- end-to-end latency of scaling (down and up; cpu and memory)
- latency of all the components:
    - requests to scheduler plugin (including retries)
    - requests to vm-monitor (including retries)
    - delay between initial NeonVM patch request and when status was updated

## Implementation ideas

AFAICT the basic idea is that we store some extra info in `agent/core.State` and add some extra callbacks in `agent/core.Config` to increment some metrics when we determine that various parts of scaling (and the entire thing) have occurred.

More design work is required, because the edge cases are quite subtle.

## Tasks

```[tasklist]
# Blockers
- [ ] #453
- [ ] #462
- [ ] #592
```

```[tasklist]
# Implementation
- [x] Decide how we should measure latency (difficult due to eventual consistency)
- [ ] agent/core: Measure scaling latency
```

```[tasklist]
### Follow-ups
- [ ] https://github.com/neondatabase/autoscaling/issues/1012
```

## Other related tasks, Epics, and links

- [Proposed RFC](https://www.notion.so/neondatabase/RFC-scaling-latency-metrics-4010b9b0b94f4ac8801171dbab6a55da?pvs=4)

Omrigan:
Newly introduced `Revision` is (basically) an integer value, which can be associated with different parts of a system: Monitor, Plugin, NeonVM and scaling algorithm itself. There are two types of such association:
 - `TargetRevision` corresponds to the desired state of a particular part.
 - `CurrentRevision` corresponds to the state which was already achieved.

As the system makes progress `TargetRevision` is propagated to `CurrentRevision` field, simultaneously tracking how long it took for this propagation.

When the same revision value is passed through multiple parts, we can measure end-to-end latency of multi-component operations.

Fixes #594.


Omrigan:
@sharnoff can you take a look, if you have a moment? This is very WIP, I didn't update the tests or tested it. 

Looking at the logic around agent's state -> does this logic of clock propagation looks sound to you? Can I make it simpler and more compact somehow?

sharnoff (commented):
Had a look - I think I'm struggling to get the big picture of what the intended flow within the autoscaler-agent state machine is.

Broadly I _think_ it should be ok, but do you have a quick (1-2 paragraph) explanation of how the clocks are supposed to flow?

(IIUC, currently the "desired logical time" stored in the plugin/monitor/neonvm state is basically the logical time of the most recent scaling that the component has completed successfully? If so, AFAICT there are still some subtle edge cases, but they shouldn't require major changes to accommodate.)

Would be good to discuss on Monday :sweat_smile:

---

All that aside, one thing I noticed: In a lot of places, there's variables like `desiredClock` or `desiredLogicalTime` -- IMO, this reads with desired as an adjective affecting the clock/logical time, which I guess is not what's intended. I wonder if it'd be better to refer to these more like timestamps, e.g. "tsOfDesired" or "desiredAtTime" etc. (or event just "desiredAt"?)

pkg/agent/core/state.go
```diff
@@ -791,12 +838,41 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
 			return nil
 		}
 	}
+	s.updateDesiredLogicalTime(now, result, s.VM.Using(), requestedUpscalingAffectedResult)
+	s.updateCurrentLogicalTime(s.VM.CurrentLogicalTime)
```
sharnoff, line 842: why `updateCurrentLogicalTime` here?

pkg/agent/core/state.go
```diff
@@ -429,7 +457,21 @@ func (s *state) calculateNeonVMAction(
...
+
+	if desiredResources.HasFieldLessThan(s.VM.Using()) {
+		// We are downscaling, so we needed a permit from the¡ monitor
+		desiredTime = desiredTime.Earliest(s.Monitor.CurrentLogicalTime)
+	}
+
+	if desiredResources.HasFieldGreaterThan(s.VM.Using()) {
+		// We are upscaling, so we needed a permit from the plugin
+		desiredTime = desiredTime.Earliest(s.Plugin.CurrentLogicalTime)
+	}
```
sharnoff, line 473: I think this is not sound under the following sequence of events:

1. At some prior time, we scaled down with logical time `A`.
1. We decide to scale up, recording logical time `B`
1. Record logical time `B`
1. Scheduler plugin approves scale-up with logical time `B`
1. We update neonvm to logical time `B` for the scale-up
1. We change our minds, deciding actually to scale back down. New logical time `C`
1. We update neonvm, but because it's scaling down, we use `min(A, C)` because we never got around to updating the vm-monitor

So, we end up with an arbitrarily old desired time if we change our minds (?)

Not sure if I missed something, or if this is actually ok — lmk!

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -191,6 +192,10 @@ type Guest struct {
 	// +optional
 	Ports []Port `json:"ports,omitempty"`
 
+	// Logical clock value corresponding to the desired resources of the VM.
+	// +optional
+	DesiredLogicalTime *LogicalTime `json:"desiredLogicalTime,omitempty"`
+
```
sharnoff, line 198: IMO this should be a field on `VirtualMachineSpec` not `Guest` — `Guest` is more about the virtual machine itself; `spec` tends to be higher-level stuff relating to the VM object.

WDYT?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -532,6 +570,8 @@ type VirtualMachineStatus struct {
 	MemoryProvider *MemoryProvider `json:"memoryProvider,omitempty"`
 	// +optional
 	SSHSecretName string `json:"sshSecretName,omitempty"`
+	// +optional
+	CurrentLogicalTime *LogicalTime `json:"currentLogicalTime,omitempty"`
```
sharnoff, line 574: What does "current" mean here? (is there a better name that's more descriptive of the semantics we're promising? regardless, please add a comment explaining this)

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -191,6 +192,10 @@ type Guest struct {
 	// +optional
 	Ports []Port `json:"ports,omitempty"`
 
+	// Logical clock value corresponding to the desired resources of the VM.
+	// +optional
+	DesiredLogicalTime *LogicalTime `json:"desiredLogicalTime,omitempty"`
```
sharnoff, line 197: Still not a fan of the "desired logical time" names... WDYT about `DesiredAt` or equivalent? (see my prior comment on the PR)

(also not sure that "desired" is the right word for this either... I often find that writing an explanation of the semantics helps make it clear what the right words to describe things are -- maybe some documentation would help with both here?)

pkg/agent/core/logiclock/logiclock.go
```diff
@@ -0,0 +1,95 @@
+package logiclock
```
sharnoff, line 1: Please add comments here!

(as I'm reading this, I have no context for how the pieces fit together & have to build that by re-reading multiple times rather than skimming some comments :sweat_smile:)

.golangci.yml
```diff
@@ -79,6 +79,7 @@ linters-settings:
       - '^github\.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1\.VirtualMachine(Migration)?(Spec)?$'
       - '^github\.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1\.IPPool$'
       - '^github\.com/neondatabase/autoscaling/pkg/agent/core\.ActionSet$'
+      - '^github\.com/neondatabase/autoscaling/pkg/agent/core\.Action.*$'
```
sharnoff, line 82: Why?

(if it's to reduce the diff on the tests, I would rather be explicit there)

pkg/agent/core/state_test.go
```diff
@@ -202,19 +210,24 @@ func getDesiredResources(state *core.State, now time.Time) api.Resources {
...
-	resources api.Resources,
-) {
+func logicalTime(clock *helpers.FakeClock, value int64) *vmv1.LogicalTime {
+	return &vmv1.LogicalTime{
+		Value:     value,
+		UpdatedAt: v1.NewTime(clock.Now()),
+	}
+}
+
+func doInitialPluginRequest(a helpers.Assert, state *core.State, clock *helpers.FakeClock, requestTime time.Duration, metrics *api.Metrics, resources api.Resources, enableLogicalClock bool) {
```
sharnoff, line 220: nit: very long line

(WDYT about adding a lint for this?)

pkg/agent/executor/exec_monitor.go
```diff
@@ -107,6 +107,7 @@ func (c *ExecutorCoreWithClients) DoMonitorDownscales(ctx context.Context, logge
 				logger.Info("vm-monitor approved downscale", logFields...)
 				if unchanged {
 					state.Monitor().DownscaleRequestAllowed(endTime)
+					state.Monitor().UpdateLogicalTime(action.DesiredLogicalTime.Rewind(endTime))
```
sharnoff, line 110: I noticed we have a these new `UpdateLogicalTime` methods. Can they be merged with the response handling methods instead? (e.g., just passing `endTime` and `action.DesiredLogicalTime` to `.Monitor().DownscaleRequestAllowed()` ?)

pkg/agent/runner.go
```diff
@@ -194,6 +196,19 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
+		if flag.Has(logiclock.Upscale) {
+			labels[0] = "true"
+		}
+		if flag.Has(logiclock.Downscale) {
+			labels[1] = "true"
+		}
+		if flag.Has(logiclock.Immediate) {
+			labels[2] = "true"
+		}
+		r.global.metrics.scalingLatency.WithLabelValues(labels...).Observe(duration.Seconds())
```
sharnoff, line 210: This seems funky... Is there a reason we can't just have a fixed enum? (if we need to accommodate scaling down & up at the same time, maybe "upscale"/"downscale"/"mixed" ?)

If no, let's have a comment here explaining what's going on?

sharnoff (commented):
Some thoughts. Not the most thorough review -- rough expectations: next round will include more nits, then one more as final thoughts.

pkg/agent/core/state.go
Omrigan: So, in this case:
 - After step 1 we have current=desired=A, so we've measured the latency for normal scale down.
 - After step 5 we have status.current=A and spec.desired=B.
    - If neonvm-controller will observe this, it will propagate status.current=B, this will loop back to the agent, and we observe latency for normal scale down.
    - If neonvm-controller doesn't observe this, nothing changes.
 - After step 7
    - If current time was propagated to B, then we have a weird situation where the desired time is less than current. This should be fixed (see note below).
    - If current time wasn't propagated to B between steps 5 and 6, then desired=A and current=A, and also we don't need to do any work (wanted to upscale, but decided not to).
 
Regardless, no new measurement will be added, since we already measured the latency to get current=A, and we can't measure the same latency again. This is equivalent to saying "steps 2-7 are actually part of the scale down, which happened in step 1, because they are started by that monitor request".

But what do we even want to measure in this case? I think measuring only clean scale down during step 1 (and, potenitally, scale-up 2-4, if it completed) is fine, otherwise it becomes harder to reason about what are we even measuring.

Note how to fix reversed time flow:
 - The interpretation is "we want neonvm to reach the state which was intended before neonvm got to it's current state".
 - I think, just check `desired > current` before doing `current = desired` should be fine.

I don't know if what I wrote makes sense to you, feel free to ask more.

pkg/agent/core/state.go
Omrigan: Moved to `UpdatedVM`.

pkg/agent/executor/exec_monitor.go
Omrigan: I tried that, and the diff was too much, especially in tests.

pkg/agent/runner.go
Omrigan: I wanted to have more flexibility to carry more information (e. g. Immediate flag there).

neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan: Just thought: maybe I should rename `LogicalTime` to `ResourcesVersion`, so it better conveys the intended semantics?

pkg/agent/runner.go
Omrigan: I've cleaned it up in 3118e51807e3af62b08bd47b24c02657cd66316c.

pkg/agent/executor/exec_monitor.go
Omrigan: I tried it in 896d0df8, a little ugly, but manageable. WDYT?

pkg/agent/core/state_test.go
Omrigan: Lint 👍 , added to #921

neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan: "Current" means the associated changes have propagated through neonvm.

pkg/agent/core/state.go
sharnoff: > I think measuring only clean scale down during step 1 (and, potenitally, scale-up 2-4, if it completed) is fine, otherwise it becomes harder to reason about what are we even measuring.

Yep, +1

> I think, just check `desired > current` before doing `current = desired` should be fine.

Not sure which you're referring to, but if you think it fixes the issue, good by me :)

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Copying from a DM:

`ResourcesVersion` is very close to `ResourceVersion`, which is a k8s metadata thing. But otherwise, I think that's a good direction. Maybe something like `UpdatedTo`? or `ResourcesRequestedAt`? (trying to avoid "version" as well because it can be confused for schema version of the CRD)

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: > No, that would be `CurrentLogicalTime`

My thinking was that "desired at" would be short for "what time was the current state desired at?"

> This is analogous to the scaling target itself: we have "current memory size" and "desired memory size", and the system works toward current becoming equal to the desired.

I don't think the analogy holds up — the "current" time is not actually equal to the value of the clock; it's equal to the time at which the current resources were requested.

Unless I'm missing something?

> > variables like desiredClock or desiredLogicalTime -- IMO, this reads with desired as an adjective affecting the clock/logical time, which I guess is not what's intended.
>
> On the contrary, this is intended.

I think I didn't express this well before ­— I'm reading `CurrentLogicalTime` ≈ "logical time right now" (i.e. literally the clock right now) and `DesiredLogicalTime` ≈ "logical time that we desire" (i.e., we want the time to be X).

Part of the reason I keep suggesting `FooAt`-like names is because (a) they don't have this issue, and (b) there's already precedent in k8s for that naming convention (see e.g. `metadata.{createdAt,deletedAt}`)

pkg/agent/core/state_test.go
sharnoff: (line is still very long)

pkg/agent/executor/exec_monitor.go
sharnoff: Fair enough — testing locally, the following also works:

```diff
diff --git a/pkg/agent/core/state_test.go b/pkg/agent/core/state_test.go
index f019e4c5..26f26354 100644
--- a/pkg/agent/core/state_test.go
+++ b/pkg/agent/core/state_test.go
@@ -19,7 +19,7 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
-var NilLogicalTime = helpers.SafeVal[vmv1.LogicalTime](nil)
+var NilLogicalTime = lo.Empty[*vmv1.LogicalTime]()
 
 func Test_DesiredResourcesFromMetricsOrRequestedUpscaling(t *testing.T) {
 	slotSize := api.Bytes(1 << 30 /* 1 Gi */)
```

(I remember putting a lot of effort into making sure that the nil-handling in testhelpers worked properly :sweat_smile:)

pkg/agent/runner.go
sharnoff: I still don't quite understand why we can't just use a fixed enum... could you elaborate?

(AFAICT "immediate" is just used to indicate whether a scale-up was requested by vm-monitor... I would remove it for now, potentially adding it back later, if we have a need for it. Maybe in the future it'd make sense to have a "requester" field instead, where it could be "bounds"/"agent"/"monitor"...?)

sharnoff: I also don't quite get why the clock needs to care about scaling flags -- feels like it's breaking separation of concerns. But presumably I'm missing something?

pkg/agent/core/logiclock/logiclock.go
```diff
@@ -0,0 +1,122 @@
...
```
sharnoff, line 1: is it intentional that this package is named "logiclock" and not "logicclock"?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -215,6 +221,46 @@ func (g Guest) ValidateForMemoryProvider(p MemoryProvider) error {
...
+
+func (t *LogicalTime) Rewind(now time.Time) *LogicalTime {
+	if t == nil {
+		return nil
+	}
+	return &LogicalTime{
+		Value:     t.Value,
+		UpdatedAt: metav1.NewTime(now),
+	}
+}
```
sharnoff, line 245: What does `Rewind` do? Without context, the name of the function does not give me any clues. (maybe it should be called e.g. `OverrideRealTime` or something?)

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -215,6 +221,46 @@ func (g Guest) ValidateForMemoryProvider(p MemoryProvider) error {
 	return nil
 }
 
+// LogicalTime allows to track progress of changes to a VM.
+type LogicalTime struct {
+	Value     int64       `json:"value"`
+	UpdatedAt metav1.Time `json:"updatedAt"`
```
sharnoff, line 227: why `UpdatedAt` ? (asking because it seems odd that a logical timestamp cares about semantics like updates -- my naive assumption is that `LogicalTime` would have two components: "logical" (int) and "real" (metav1.Time), and it's up to the "application logic" (in this case, everyone else) to apply meaning on top of that)

pkg/agent/core/testhelpers/construct.go
```diff
@@ -158,3 +178,15 @@ func WithCurrentCU(cu uint16) VmInfoOpt {
 		vm.SetUsing(c.ComputeUnit.Mul(cu))
 	})
 }
+
+func WithLogicalTime(t *vmapi.LogicalTime) VmInfoOpt {
```
sharnoff, line 182: Should this be `WithInitialLogicalTime` or something? asking because `WithLogicalTime` IMO makes it seem like the VM only refers to a particular instant in time

pkg/agent/core/testhelpers/construct.go
```diff
@@ -36,12 +37,24 @@ type VmInfoOpt interface {
 	modifyVmInfoWithConfig(InitialVmInfoConfig, *api.VmInfo)
 }
 
+type ClockSourceOpt interface {
+	InitialStateOpt
+
+	clock() core.LogicClock
+}
```
sharnoff, line 44: I think you don't need to add this if the clock is included in `core.Config` -- wdyt?

(no opinions, just noticed it seemed like a lot of effort here)

pkg/agent/prommetrics.go
```diff
@@ -217,6 +220,13 @@ func makeGlobalMetrics() (GlobalMetrics, *prometheus.Registry) {
 				Help: "Number of times (*core.State).NextActions() has been called",
 			},
 		)),
+
+		scalingLatency: *util.RegisterMetric(reg, prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: "autoscaling_agent_scaling_latency_seconds",
+				Help: "End-to-end scaling latency",
+			}, logiclock.AllFlagNames,
```
sharnoff, line 228:
```suggestion
			},
			logiclock.AllFlagNames,
```


(another reason to have a formatter... :melting_face:)

pkg/agent/prommetrics.go
```diff
@@ -217,6 +220,13 @@ func makeGlobalMetrics() (GlobalMetrics, *prometheus.Registry) {
 				Help: "Number of times (*core.State).NextActions() has been called",
 			},
 		)),
+
+		scalingLatency: *util.RegisterMetric(reg, prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: "autoscaling_agent_scaling_latency_seconds",
+				Help: "End-to-end scaling latency",
```
sharnoff, line 227: We need to also set the buckets here, right?

sharnoff (commented):
more thorough this time

pkg/agent/prommetrics.go
Omrigan: Yeah, the defaults have low quite resolution.

pkg/agent/core/testhelpers/assert.go
```diff
@@ -79,6 +79,11 @@ func (a Assert) Call(f any, args ...any) PreparedFunctionCall {
 
 	var argValues []reflect.Value
 	for _, a := range args {
+		if _, ok := a.(reflect.Value); ok {
+			// This is a SafeVal value, so we can just use it directly
+			argValues = append(argValues, a.(reflect.Value))
+			continue
+		}
```
sharnoff, line 86: This is no longer needed, right?

pkg/agent/core/revsource/revsource.go
```diff
@@ -0,0 +1,131 @@
...
+
+const (
+	Upscale vmv1.Flag = 1 << iota
+	Downscale
+	Immediate
+)
+
+// AllFlags and AllFlagNames must have the same order, so the metrics work correctly.
+var AllFlags = []vmv1.Flag{Upscale, Downscale, Immediate}
+var AllFlagNames = []string{"upscale", "downscale", "immediate"}
```
sharnoff, line 20: Same comments here as before

pkg/agent/core/revsource/revsource.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (c *NilRevisionSource) Next(_ time.Time, _ vmv1.Flag) vmv1.Revision {
+	return vmv1.Revision{
+		Value: 0,
+		Flags: 0,
+	}
+}
+func (c *NilRevisionSource) Observe(_ time.Time, _ vmv1.Revision) error { return nil }
+
+type MetricCB func(dur time.Duration, flags vmv1.Flag)
```
sharnoff, line 103: let's be explicit in the name here? i.e. `MetricsCallback` ?

edit: Actually, is this required to be related to metrics at all, other than how we use it? if no, let's call it something like `ObserveCallback`, because that seems more in line with its semantics ? (and, we use it in tests, but not for metrics :sweat_smile:)

neonvm/controllers/vm_controller.go
```diff
@@ -800,6 +800,11 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		// do nothing
 	}
 
+	if vm.Status.Phase == vmv1.VmRunning && vm.Spec.TargetRevision != nil {
+		rev := vm.Spec.TargetRevision.WithTime(time.Now())
+		vm.Status.CurrentRevision = &rev
+	}
```
sharnoff, line 806: Just double-checking I understand correctly: This means that spec -> status propagation is only performed once the VM reaches the Running phase, right?

I think _in practice_ this should be fine, but it's worth considering what cases this doesn't catch...

sharnoff, line 806: I'm also a little nervous here about adding new stuff at the end of `doReconcile`. It's easy to write an early return above, assuming that there's nothing relevant at the bottom of the function.

Have you checked we have no such early returns?

If so: I think it's ok for now, but should be reconsidered as part of broader neonvm-controller refactoring.

pkg/agent/core/state.go
```diff
@@ -190,23 +216,32 @@ type neonvmState struct {
...
 }
 
 func (ns *neonvmState) ongoingRequest() bool {
 	return ns.OngoingRequested != nil
 }
 
+type RevisionSource interface {
+	Next(ts time.Time, flags vmv1.Flag) vmv1.Revision
+	Observe(moment time.Time, rev vmv1.Revision) error
+}
```
sharnoff, line 231: nit: let's move this definition up next to `Config` and `LogConfig`

pkg/agent/core/state.go
```diff
@@ -72,6 +80,12 @@ type Config struct {
 	// Log provides an outlet for (*State).NextActions() to give informative messages or warnings
 	// about conditions that are impeding its ability to execute.
 	Log LogConfig `json:"-"`
+
+	// RevisionSource is the source of revisions to track the progress during scaling.
+	RevisionSource RevisionSource `json:"-"`
+
+	// PromMetricsCallbacks are the callbacks to update the Prometheus metrics.
+	PromMetricsCallbacks PromMetricsCallbacks `json:"-"`
```
sharnoff, line 88: nit: s/PromMetrics/Metrics/ — we don't care whether it's prometheus, right?

pkg/agent/core/state.go
```diff
@@ -430,6 +478,17 @@ func (s *state) calculateNeonVMAction(
...
+	targetRevision := s.TargetRevision
+	if desiredResources.HasFieldLessThan(s.VM.Using()) {
+		// We are downscaling, so we needed a permit from the monitor
+		targetRevision = targetRevision.Min(s.Monitor.CurrentRevision)
+	}
+
+	if desiredResources.HasFieldGreaterThan(s.VM.Using()) {
+		// We are upscaling, so we needed a permit from the plugin
+		targetRevision = targetRevision.Min(s.Plugin.CurrentRevision)
+	}
```
sharnoff, line 490: Revisiting this — the implicit assumption is that _if_ we're downscaling/upscaling, we already have the necessary permission, right?

If so: I think this `targetRevision` logic must go after `desiredResources = s.clampResources(...)` below, but please double-check me on that (maybe, write a test case that fails without that change?)

pkg/agent/core/state_test.go
```diff
@@ -405,8 +507,15 @@ func TestPeriodicPluginRequest(t *testing.T) {
...
+	cfg := DefaultInitialStateConfig
+	latencyObserver := &latencyObserver{t: t, observations: nil}
+	defer latencyObserver.assertEmpty()
+	// This time, we will test plugin latency
+	cfg.Core.PromMetricsCallbacks.PluginLatency = latencyObserver.observe
+
+	cfg.Core.RevisionSource = revsource.NewRevisionSource(nil)
 	state := helpers.CreateInitialState(
-		DefaultInitialStateConfig,
+		cfg,
```
sharnoff, line 518: May be cleaner to use `testhelpers.WithConfigSetting`, wdyt?

pkg/agent/core/state_test.go
```diff
@@ -376,20 +460,38 @@ func TestBasicScaleUpAndDownFlow(t *testing.T) {
 	})
 	a.Do(state.NeonVM().RequestSuccessful, clock.Now())
 
+	// Request to NeonVM is successful, but let's wait one more tick for
+	// NeonVM to pick up the changes and apply those.
+	clockTick().AssertEquals(duration("0.9s"))
+	vmInfo = helpers.CreateVmInfo(
+		DefaultInitialStateConfig.VM,
+		helpers.WithCurrentCU(1),
+		helpers.WithMinMaxCU(1, 1),
```
sharnoff, line 469: This would change max CU from 4 -> 1, is that intended?

pkg/agent/core/state_test.go
```diff
@@ -234,6 +253,37 @@ func duration(s string) time.Duration {
 	return d
 }
 
+func zeroRev(t time.Time) vmv1.RevisionWithTime {
+	return vmv1.ZeroRevision.WithTime(t)
+}
```
sharnoff, line 258: FYI there's some places where you use this function, and some where you use `vmv1.ZeroRevision.WithTime` explicitly. I have no preference, although `zeroRev` has less syntax :)

sharnoff: That said -- I see there's a lot of places where you're passing `clock.Now()`. Maybe worthwhile to define a local anonymous function that does that automatically?

e.g.
```go
clock := ...
expectedRevision := ...

zeroRev := func() vmv1.RevisionWithTime {
    return expectedRevision.WithTime(clock.Now())
}
```

pkg/agent/prommetrics.go
```diff
@@ -48,6 +66,11 @@ const (
 	runnerMetricStatePanicked runnerMetricState = "panicked"
 )
 
+// Copied bucket values from controller runtime latency metric. We can
+// adjust them in the future if needed.
+var buckets = []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
+	1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60}
```
sharnoff, line 72: That's a ton of buckets. Can we get some data from staging first about how often this typically takes?

pkg/agent/executor/core.go
```diff
@@ -53,7 +53,11 @@ type ClientSet struct {
 	Monitor MonitorInterface
 }
 
-func NewExecutorCore(stateLogger *zap.Logger, vm api.VmInfo, config Config) *ExecutorCore {
+func NewExecutorCore(
+	stateLogger *zap.Logger,
+	vm api.VmInfo,
+	config Config,
+) *ExecutorCore {
```
sharnoff, line 60: Unnecessary diff?

pkg/agent/core/state.go
```diff
@@ -289,7 +331,12 @@ func (s *state) nextActions(now time.Time) ActionSet {
 	// Requests to vm-monitor (downscaling)
 	plannedUpscale := actions.MonitorUpscale != nil
 	var monitorDownscaleRequiredWait *time.Duration
-	actions.MonitorDownscale, monitorDownscaleRequiredWait = s.calculateMonitorDownscaleAction(now, desiredResources, plannedUpscale)
+	actions.MonitorDownscale, monitorDownscaleRequiredWait =
+		s.calculateMonitorDownscaleAction(
+			now,
+			desiredResources,
+			plannedUpscale,
+		)
```
sharnoff, line 339: Unnecessary diff?

pkg/agent/core/revsource/revsource.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// Forget the measurement, and all the measurements before it.
+	c.offset = rev.Value + 1
+	c.measurements = c.measurements[idx+1:]
+
+	return nil
+}
+
+type NilRevisionSource struct{}
```
sharnoff, line 93: Is this used in practice? (if only for tests, maybe we just define it in the test file, instead of here?)

pkg/agent/core/state.go
```diff
@@ -995,6 +1118,11 @@ func (h PluginHandle) RequestSuccessful(now time.Time, resp api.PluginResponse)
 	// the process of moving the source of truth for ComputeUnit from the scheduler plugin to the
 	// autoscaler-agent.
 	h.s.Plugin.Permit = &resp.Permit
+	revsource.Propagate(now,
+		targetRevision,
+		&h.s.Plugin.CurrentRevision,
+		h.s.Config.PromMetricsCallbacks.PluginLatency,
+	)
```
sharnoff, line 1125: Reading through the related code — will the plugin latency get recorded as:

1. end-to-end time from decision to plugin request completed; or
2. latency between end of previous request to end of plugin request

?

OTOH I thought from reading that it was (1), and I would've expected (2). But I'm not sure if I missed something

pkg/agent/core/testhelpers/construct.go
```diff
@@ -38,6 +38,7 @@ type VmInfoOpt interface {
 
 func CreateInitialState(config InitialStateConfig, opts ...InitialStateOpt) *core.State {
 	vmOpts := []VmInfoOpt{}
+
```
sharnoff, line 41: Unnecessary diff?

sharnoff (commented):
Big updates! Kudos on the name, "revisions" seems like a much more natural construct imo.

There's definitely stuff I missed, tried to comment on everything though to reduce RTT :)

At a glance, I think a handful of my comments from my previous review are still valid & unaddressed as well.

neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan: Resolved, as there isn't any logical time left.

neonvm/apis/neonvm/v1/virtualmachine_types.go
Omrigan: Resolved, as there isn't any logical time left.

pkg/agent/core/revsource/revsource.go
Omrigan: Responding:

> I also don't quite get why the clock needs to care about scaling flags -- feels like it's breaking separation of concerns. But presumably I'm missing something?

> edit: OTOH, an alternate approach would be making LogicClock generic over the flag type, and then providing the flags when it's instantiated in agent/core or wherever.

Yeah, currently, the flag type is defined together with other NeonVM types (because it is used in the CRD), but the actual flags and their meaning defined in revsource. This gives a little bit of separation of concern, and kinda works like generics. Bringing full-fledged generics here might be a bad idea, because:
1. CRD generation from generic types will probably break
2. I try to avoid generics in the business logic code

neonvm/controllers/vm_controller.go
Omrigan: > Just double-checking I understand correctly: This means that spec -> status propagation is only performed once the VM reaches the Running phase, right?

Yes

> I think in practice this should be fine, but it's worth considering what cases this doesn't catch...

Well, the main case where we want to propagate the revision is when the VM transitions Scaling -> Running. Or when it didn't need scaling, and keeps being Running. I don't know of any other cases where it is need, actually.

pkg/agent/executor/core.go
sharnoff: All good :)

pkg/agent/core/state.go
```diff
@@ -791,12 +841,57 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
 			return nil
 		}
 	}
+	s.updateDesiredLogicalTime(now, result, s.VM.Using(), requestedUpscalingAffectedResult)
+
+	s.LastDesiredResources = &result
```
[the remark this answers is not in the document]
sharnoff: > This is also doubtful, as the dataflow becomes a bit harder to follow.

That's exactly my thoughts & why I think there isn't anything better for now :sweat_smile:

pkg/agent/core/revsource/revsource.go
Omrigan: I'll look into how other metrics in the agent expose this info and will see if I can do something similar.

Omrigan: Previous discussion: https://github.com/neondatabase/autoscaling/pull/983#discussion_r1669835735

pkg/agent/core/state.go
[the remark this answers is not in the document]
Omrigan: I'll leave a comment explaining this issues and leave as-is for now.

pkg/agent/core/state.go
sharnoff: I'll think about this some more & report back

pkg/agent/core/state_test.go
Omrigan: Check again there is no expectedRevision.Revision

pkg/agent/core/state.go
Omrigan: We should ideally have 2 metrics: one with retries, and one without. It is sufficient to have only one in this PR.

I'll think how to best implement it.

Omrigan: Plugin/NeonVM/Monitor metrics we currently have do not count retries, i. e. measure each individual request separately. 

In order to count the retries also, we must not do `TargetRevision: s.TargetRevision.WithTime(now)` every time a request is planned. Rather, we update the time once per concrete revision value, and not update it if more request happen with the same revision.

pkg/agent/core/revsource/revsource.go
Omrigan: Reused directionLabel in be700fff

pkg/agent/core/state.go
```diff
@@ -791,12 +852,54 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
...
+func (s *state) updateNeonVMCurrentRevision(currentRevision vmv1.RevisionWithTime) {
+	revsource.Propagate(currentRevision.UpdatedAt.Time,
+		s.NeonVM.TargetRevision,
+		&s.NeonVM.CurrentRevision,
+		s.Config.ObservabilityCallbacks.NeonVMLatency,
+	)
+	err := s.Config.RevisionSource.Observe(currentRevision.UpdatedAt.Time, currentRevision.Revision)
+	if err != nil {
+		s.warnf("Failed to observe clock source: %v", err)
+	}
```
sharnoff, line 900: If we get an error here, do you think it's likely that the error is due to some persistent bad state?

If so, can we change this to panic? that will reset all of the internal state that the agent tracks for this VM

pkg/agent/core/state.go
```diff
@@ -995,6 +1106,11 @@ func (h PluginHandle) RequestSuccessful(now time.Time, resp api.PluginResponse)
 	// the process of moving the source of truth for ComputeUnit from the scheduler plugin to the
 	// autoscaler-agent.
 	h.s.Plugin.Permit = &resp.Permit
+	revsource.Propagate(now,
+		targetRevision,
+		&h.s.Plugin.CurrentRevision,
+		h.s.Config.ObservabilityCallbacks.PluginLatency,
+	)
```
sharnoff, line 1113: What happens if the request is successful but does not approve the requested resources?

pkg/agent/core/state.go
```diff
@@ -1061,19 +1178,29 @@ func (h MonitorHandle) StartingDownscaleRequest(now time.Time, resources api.Res
...
 		At:        now,
 		Current:   *h.s.Monitor.Approved,
 		Requested: h.s.Monitor.OngoingRequest.Requested,
 	}
 	h.s.Monitor.OngoingRequest = nil
+	revsource.Propagate(now,
+		targetRevision,
+		&h.s.Monitor.CurrentRevision,
+		h.s.Config.ObservabilityCallbacks.MonitorLatency,
+	)
```
sharnoff, line 1203: Similar question here as with the scheduler - what happens when downscale is denied?

pkg/agent/runner.go
```diff
@@ -194,6 +196,9 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 	pluginRequestJitter := util.NewTimeRange(time.Millisecond, 0, 100).Random()
 
 	coreExecLogger := execLogger.Named("core")
+	revisionSource := revsource.NewRevisionSource(
+		WrapHistogramVec(&r.global.metrics.scalingLatency),
+	)
```
sharnoff, line 201: Can we pass in the VM spec's current revision here so that we pick up where the previous agent left off? (otherwise IIUC we'll get a bunch of warnings on restart, right?)

sharnoff (commented):
basically final review, a few questions left

pkg/agent/core/state.go
Omrigan: The propagation doesn't happen -> we don't measure the latency.

pkg/agent/core/state.go
Omrigan: The propagation doesn't happen -> we don't measure the latency.

pkg/agent/core/state.go
Omrigan: > If we get an error here, do you think it's likely that the error is due to some persistent bad state?

Not necessarily. For example, the agent might've restarted, and we are getting the revisions generated by a previous agent.

Even if I initialize the revision source with last known revision, it can still race.

Also, revision in status is "external input", meaning something external could've change that, and I think we shouldn't panic on external input.

pkg/agent/core/state.go
```diff
@@ -215,15 +253,22 @@ func NewState(vm api.VmInfo, config Config) *State {
...
+				CurrentRevision:  vmv1.ZeroRevision,
 			},
-			Metrics: nil,
+			Metrics:              nil,
+			LastDesiredResources: nil,
+			TargetRevision:       vmv1.ZeroRevision,
 		},
 	}
+
+	return state
```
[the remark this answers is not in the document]
Omrigan: done

pkg/agent/runner.go
Omrigan: That's a great idea! Implemented it in 0df27ac9974d4e06dd32eb186053e1e062c3d00f, take a look, pls.

pkg/agent/core/state.go
sharnoff: > Not necessarily. For example, the agent might've restarted, and we are getting the revisions generated by a previous agent.
>
> Even if I initialize the revision source with last known revision, it can still race.

Sure, yeah -- I'm just thinking that if we get in a bad state that we can't recover from, it would be better to reset the state. But if you think that's unlikely relative to the amount of false-positives we'd get around agent restarts, let's leave as is.

> I think we shouldn't panic on external input

For the sake of general discussion: In most other contexts I would agree with you; here however, panicking triggers a clean restart of _just_ the tasks associated with a particular VM. IMO it's a clean way to recover from a degraded state.

pkg/agent/runner.go
```diff
@@ -194,7 +196,14 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 	pluginRequestJitter := util.NewTimeRange(time.Millisecond, 0, 100).Random()
 
 	coreExecLogger := execLogger.Named("core")
-	executorCore := executor.NewExecutorCore(coreExecLogger, getVmInfo(), executor.Config{
+
+	vmInfo := getVmInfo()
+	var initialRevision int64
+	if vmInfo.CurrentRevision != nil {
+		initialRevision = vmInfo.CurrentRevision.Value
+	}
```
sharnoff, line 204: nit: we can use `lo.FromPtrOr` here

pkg/agent/runner.go
sharnoff: nice! One nit, left a separate comment about it

pkg/agent/core/state.go
sharnoff: > The propagation doesn't happen

Could you elaborate on that? AFAICT there is no special detection of "scheduler did not approve what we asked for", so we'll still call `revsource.Propagate`.

(do you mean that we won't make follow-up requests, so end-to-end latency isn't recorded until the request is approved? in that case: we'd still record plugin latency incorrectly though, right? and- that would mean that end-to-end latency records the _first_ change rather than the _full_ change — e.g. if we want to scale from 1 → 4 CU, but can only get approval for 3 CU first, latency of 1 → 3 will get recorded, but 1 → 4 won't, IIUC?)

pkg/agent/core/state.go
sharnoff: Here we still measure the latency for the vm-monitor even though it was denied, right? Is that intentional? (if so: what are the expected semantics for component latency?)

pkg/agent/core/state.go
Omrigan: Anyway, the recovery here should happen once autoscaler-agent writes fresh `targetRevision`, and it is propagated.

pkg/agent/core/state.go
Omrigan: > Could you elaborate on that? AFAICT there is no special detection of "scheduler did not approve what we asked for", so we'll still call revsource.Propagate.

Sorry, I misread the question. Actually yes, the propagation happens as usual. 

> that would mean that end-to-end latency records the first change rather than the full change

Very good question. I wrote a test to verify this 8dcd87ac689e641384fe2cd5af5aece9fda3e3ad.

And then I changed the implementation so that sequence is considered as two upscalings: 
 - 1 -> 4 desired / 3 real
 - 3 -> 4

In 49bd61af4dcb1678bbaafaaca1bbea643868a3b9

pkg/agent/core/state.go
sharnoff: > Well, the distribution of latency for successful requests

My current understanding is either:

* Component latency should only look at individual request latency
* Component latency should only be related end-to-end scaling

If it's the first one, then presumably we shouldn't be using `revsource` for this (we'd just want a simple histogram metric looking at the time difference since we started the request, right?).

If it's the second one, then we should treat denial as failure, because that doesn't get us closer to scaling.

Does that make sense?

pkg/agent/core/state.go
```diff
@@ -791,12 +850,59 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
...
+	revsource.Propagate(currentRevision.UpdatedAt.Time,
+		s.NeonVM.TargetRevision,
+		&s.NeonVM.CurrentRevision,
+		s.Config.ObservabilityCallbacks.NeonVMLatency,
+	)
+	err := s.Config.RevisionSource.Observe(currentRevision.UpdatedAt.Time, currentRevision.Revision)
+
+	// We also zero out LastDesiredResources, because we are now starting from
+	// a new current resources.
+	s.LastDesiredResources = nil
```
sharnoff, line 899: Should we only do this when `LastDesiredResources == current neonvm resources` ?

pkg/agent/core/state.go
```diff
@@ -791,12 +850,59 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
...
+
+	if err != nil {
+		s.warnf("Failed to observe clock source: %v", err)
+	}
```
sharnoff, line 903: why `if err != nil` not immediately after where it's generated?

pkg/agent/core/state.go
sharnoff: Nice!

pkg/agent/core/state.go
Omrigan: It is "Component latency should only look at individual request latency".

The implementation will remain as-is for now, and later can be simplified.

I should check the metric name, that it is clearly expresses semantics.

pkg/agent/runner.go
sharnoff: After re-reading we actually can't use this here

pkg/agent/core/state.go
sharnoff: discussed -- the current behavior is actually required.

pkg/agent/core/state.go
Omrigan: Done.

PR #984, opened 2024-06-21
vm-builder: don't use ANSI terminal formatting in vector logs

antonyc:
It is not possible for all of the log collectors to correctly strip them out.

```
vector --help | grep '\-\-color' -A 5
      --color <COLOR>
          Control when ANSI terminal formatting is used.

          By default `vector` will try and detect if `stdout` is a terminal, if it is ANSI will be enabled. Otherwise it will be disabled. By providing this flag with the `--color always` option will always enable
          ANSI terminal formatting. `--color never` will disable all ANSI terminal formatting. `--color auto` will attempt to detect it automatically.

          [env: VECTOR_COLOR=]
          [default: auto]
          [possible values: auto, always, never]
```

### DIFF
--- a/neonvm/tools/vm-builder/files/inittab
+++ b/neonvm/tools/vm-builder/files/inittab
@@ -4,7 +4,7 @@
 ::respawn:/neonvm/bin/udevd
 ::wait:/neonvm/bin/udev-init.sh
 ::respawn:/neonvm/bin/acpid -f -c /neonvm/acpi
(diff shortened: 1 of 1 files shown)


Omrigan (approved):

sharnoff (approved):
lgtm - thanks! Odd that it makes the incorrect inference though...

Before merging, could you prefix the PR title with "vm-builder: " ?

PR #986, opened 2024-06-21
neonvm: Fix tests failing to compile

sharnoff:
Seems like #935 had issues after merging main into it due to #971, but test failure didn't prevent auto-merge.


Omrigan (approved):
> test failure didn't prevent auto-merge

Let's see if we can change that.

PR #987, opened 2024-06-22
plugin: Fix Score reserving when over-budget

sharnoff:
For calls to `(*AutoscaleEnforcer).speculativeReserve()`, the callback given to it returns a boolean which should indicate whether to actually reserve the Pod.

When handling Score, we were returning true (which we should never do in Score) exactly when the placement would have been over-budget. This meant that pods would be recorded on an over-full node early in the scheduling flow, disagreeing with the core scheduler, which causes even more calls to Score that we'd think are over-budget, exacerbating the problem.

ref https://neondb.slack.com/archives/C0798QG4D0T/p1719016313787749?thread_ts=1719014348.041469

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -655,7 +655,7 @@ func (e *AutoscaleEnforcer) Score(
 	}
 
 	// Special case: return minimum score if we don't have room
-	overbudget, verdict := e.speculativeReserve(node, vmInfo, pod, false, func(_ verdictSet, overBudget bool) bool {
+	overbudget, verdict := e.speculativeReserve(node, vmInfo, pod, false, func(_ verdictSet, _ bool) bool {
 		return false // never actually accept the pod; we're just doing this to ask if it's over-budget.
 	})
 	if overbudget {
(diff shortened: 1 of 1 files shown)


pkg/plugin/plugin.go
```diff
@@ -656,7 +656,7 @@ func (e *AutoscaleEnforcer) Score(
 
 	// Special case: return minimum score if we don't have room
 	overbudget, verdict := e.speculativeReserve(node, vmInfo, pod, false, func(_ verdictSet, overBudget bool) bool {
-		return overBudget
+		return false // never actually accept the pod; we're just doing this to ask if it's over-budget.
```
Omrigan, line 659: How we could've prevented it in theory? I see two ways which could've helped:
1. Passing the arguments (and especially callbacks) through the struct, thus giving it a name on the call site, increasing the chance one of us would spot the bug.
2. Having a custom return type (wrapper over bool).

Omrigan (approved):

pkg/plugin/plugin.go
sharnoff: Yeah, I was thinking that as well... Good to know going forward w.r.t. avoiding plain bools in return types for things like this where the meaning isn't obvious from context.

PR #988, opened 2024-06-23
neonvm: Move runner image env var into YAML

sharnoff:
For neondatabase/cloud#10177, we want to be able to override the neonvm-runner image that the controller uses *after* it's built and the YAML is generated.

To do this, we need to actually define the runner image in the YAML somwhere.

This PR:

1. Moves VM_RUNNER_IMAGE env var from the neonvm-controller dockerfile into its Pod spec.
2. Adds neonvm-runner-image-loader to neonvm.yaml (required for 1)

---

Tested compared to previous main the difference in the YAMLs when using v0.32.0 as the tag, and the diff to neonvm.yaml (aside from adding the image loader) was just the env var in the expected place.


sharnoff:
From discussing: Should add a comment explaining the source of truth for the neonvm-runner image and where it's used -- comment should be both in controller spec and runner loader spec.

Omrigan (approved):
Approved, pending the addition mentioned comments.

PR #989, opened 2024-06-23
Use go-chef for image builds

sharnoff:
It *dramatically* reduces the amount of time it takes to re-run image building if the set of packages used hasn't changed. It's hard to get an objective measurement (because we do get *some* caching from being careful what we add), but:

On my machine, for a rebuild after a change in 'pkg/api', it reduces `make docker-build` time from 249s -> 22s.

And, it also slightly improves the time it takes to run a clean build. After `docker system prune -af`, it reduces `make docker-build` on my machine from 345s -> 126s.

### DIFF
--- a/build/autoscale-scheduler/Dockerfile
+++ b/build/autoscale-scheduler/Dockerfile
@@ -24,6 +24,6 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     -ldflags "-X 'github.com/neondatabase/autoscaling/pkg/util.BuildGitInfo=$GIT_INFO'" \
     cmd/autoscale-scheduler/main.go
 
-FROM alpine
+FROM alpine:3.19
 COPY --from=builder /workspace/main /usr/bin/kube-scheduler
 ENTRYPOINT ["/usr/bin/kube-scheduler"]
--- a/build/autoscale-scheduler/Dockerfile
+++ b/build/autoscale-scheduler/Dockerfile
@@ -1,26 +1,12 @@
-FROM golang:1.23-alpine AS builder
-WORKDIR /workspace
-
-RUN apk add gcc musl-dev # gcc (and therefore musl-dev) is required for cgo extensions
-
-COPY go.mod go.mod
-COPY go.sum go.sum
-RUN go mod download
-
-COPY neonvm/apis             neonvm/apis
-COPY neonvm/client           neonvm/client
-COPY pkg/api                 pkg/api
-COPY pkg/plugin              pkg/plugin
-COPY pkg/util                pkg/util
-COPY cmd/autoscale-scheduler cmd/autoscale-scheduler
+ARG GO_BASE_IMG=autoscaling-go-base:dev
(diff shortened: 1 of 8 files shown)


sharnoff:
Another fun datapoint: Seems like it _may_ drop clean image builds in CI from ~[2m50s](https://github.com/neondatabase/autoscaling/actions/runs/9636113287/job/26573733361?pr=987) → ~[1m45s](https://github.com/neondatabase/autoscaling/actions/runs/9636513834/job/26574578415?pr=989)

build/autoscale-scheduler/Dockerfile
```diff
@@ -1,11 +1,19 @@
-FROM golang:1.21-alpine AS builder
+FROM golang:1.21-alpine as chef
```
bayandin, line 1: "as" should be "AS"

From [CI logs](https://github.com/neondatabase/autoscaling/actions/runs/9636513834/job/26574578415?pr=989):
```
#1 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
```

sharnoff: Ah yup, good catch - thanks!

neonvm/Dockerfile
```diff
@@ -27,7 +34,7 @@ COPY pkg/util pkg/util
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -tags=${BUILDTAGS} -o manager neonvm/main.go
```
Omrigan, line 30: This change loses GOOS and GOARCH, is that intentional?

Omrigan (commented):
The main thing I am concerned here is the status of `go-chef` going forward. Who is going to be the owner of it? Maybe @neondatabase/developer-productivity? Do we want it to use for other go-based images?

That's unfortunate that there is no standard way of solving this problem in the go community.

neonvm/Dockerfile
sharnoff: It's somewhat intentional (all the environment variables, etc. need to match in order for `go-chef --cook` to work) but I'm not actually sure what these should be.

Perhaps @cicdteam has more context on what breaks without these? (and, why it wasn't needed for the binaries outside of neonvm?)

cicdteam: These variables are came from kubebuilder templates. It can be used for cross compiling (not a multi-platform docker image, just compiling). I don't think we need them at all.

sharnoff: Ah ok! that makes a lot of sense - thanks!

bayandin (approved):
@sharnoff said it works correctly

PR #990, opened 2024-06-23
neonvm-controller: Add '--skip-update-validation-for' 

sharnoff:
Copying the second commit's message:

> This flag is a new escape hatch to allow arbitrary updates to an object, bypassing the webhook update validation (but still within the CRD).
>
> Sometimes when trialing things on staging, we end up with objects in a bad state that can't be deleted because update validation fails when removing the finalizer (e.g., due to destructive CRD changes).

This PR comes in two parts:

1. Refactoring — moving webhook setup in to the controller, so that it's easier to control from there
2. Feature — adding the `--skip-update-validation-for` flag to neonvm-controller, and using our newfound control over webhooks to skip the update webhook for objects named with the flag.

_Probably_ worth merging via rebase-and-merge.

### DIFF
--- a/neonvm/controllers/webhook.go
+++ b/neonvm/controllers/webhook.go
@@ -6,16 +6,65 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
+	"github.com/neondatabase/autoscaling/pkg/util/stack"
 )
+
+func validateUpdate(
+	ctx context.Context,
+	cfg *ReconcilerConfig,
+	recorder record.EventRecorder,
+	oldObj runtime.Object,
+	newObj interface {
+		webhook.Validator
+		metav1.Object
+	},
+) (admission.Warnings, error) {
+	log := log.FromContext(ctx)
+
+	namespacedName := client.ObjectKeyFromObject(newObj)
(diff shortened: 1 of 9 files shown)


neonvm/controllers/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+	oldObj runtime.Object,
+	newObj interface {
+		webhook.Validator
+		metav1.Object
+	},
+) error {
+	namespacedName := types.NamespacedName{
+		Namespace: newObj.GetNamespace(),
+		Name:      newObj.GetName(),
+	}
```
sharnoff, line 33: Note to self: Should probably use `client.GetObjectKeyFromObject` instead of filling out the `NamespacedName` manually

Omrigan:
Does `kubectl apply --validate=false` do something different? From the man:
```
    --validate='strict':
	Must be one of: strict (or true), warn, ignore (or false). 	"true" or "strict" will use a schema to validate the input and fail the request if invalid. It will perform server side validation if ServerSideFieldValidation is enabled on the api-server, but will fall back to less reliable client-side validation if not. 		"warn" will warn about unknown or duplicate fields without blocking the request if server-side field validation is enabled on the API server, and behave as "ignore" otherwise. 		"false" or "ignore" will not perform any schema validation, silently dropping any unknown or duplicate fields
```

I think it would be much more convenient to have a bypass method always-on.

sharnoff:
I didn't know about `kubectl apply --validate=false`, but the main benefit I was thinking of here was allowing neonvm-controller to make changes without validation. Specifically because:

> Sometimes when trialing things on staging, we end up with objects in a bad state that can't be deleted because update validation fails when removing the finalizer (e.g., due to destructive CRD changes).

Could you elaborate on what you mean by

> I think it would be much more convenient to have a bypass method always-on.

?

neonvm/controllers/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+		recorder.Event(
+			newObj,
+			"Warning",
+			"SkippedValidation",
+			"Skipping webhook validation because of controller's '--skip-update-validation-for' flag",
+		)
+		return nil
+	}
+
+	return newObj.ValidateUpdate(oldObj)
```
Omrigan, line 44: How about writing the would-be validation verdict into the recorder/logger?

Omrigan (approved):

neonvm/controllers/webhook.go
sharnoff: Good idea, yeah. I'll have it record a warning event only when validation _would have_ failed?

PR #994, opened 2024-06-25
agent/billing: support Azure Blob Storage to store consumption events

antonyc:
This enables `billing` package to persist consumption events in Azure Blob storage. Tests are done with Azurite [^1] docker container through `gnomock` library.

[^1]: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite

https://github.com/neondatabase/cloud/issues/14786


pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,153 @@
...
+type AzureAuthType string
+
+const (
+	// AzureAuthTypeTests is used in tests
+	// It uses well-known storage account and key.
+	// See https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite
+	AzureAuthTypeTests AzureAuthType = "tests"
+	// AzureAuthTypeDefault is for pods running in Azure Kubernetes.
+	// Make sure you have provisioned Role and you have Managed Identity.
+	AzureAuthTypeDefault AzureAuthType = "default"
```
antonyc, line 26: this auth type has not been tested yet in Azure cloud, going to do that this week before merging.

antonyc:
Tested in azure, and it worked: https://github.com/neondatabase/cloud/issues/14786#issuecomment-2191013925

pkg/billing/azureblobstoragebilling.go
antonyc: https://github.com/neondatabase/cloud/issues/14786#issuecomment-2191013925 it worked

pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (c AzureClient) generateKey() string {
+	if c.cfg.generateKey != nil {
+		return c.cfg.generateKey()
+	}
+	// Example: prefixInContainer/year=2021/month=01/day=26/hh:mm:ssZ_{uuid}.ndjson.gz
+	now := time.Now()
+	id := shortuuid.New()
+
+	filename := fmt.Sprintf("year=%d/month=%02d/day=%02d/%s_%s.ndjson.gz",
```
Omrigan, line 90: If this is the same format as S3Client, perhaps, extract&reuse?

pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,158 @@
...
+		enc.AddString("container", c.cfg.Container)
+		enc.AddString("prefixInContainer", c.cfg.PrefixInContainer)
+		enc.AddString("endpoint", c.cfg.Endpoint)
+		return nil
+	}))
+}
+
+func (c AzureClient) generateKey() string {
+	if c.cfg.generateKey != nil {
+		return c.cfg.generateKey()
```
Omrigan, line 84: Instead of this, I'd add an argument `now time.Time`, and verified in tests that the key is what it is supposed to look like for the hardcoded timestamp.

pkg/billing/azureblobstoragebilling_test.go
Omrigan, line 1: FYI: that's how tests for S3 look like: https://github.com/neondatabase/autoscaling/tree/main/tests/e2e/s3

Do you think the similar approach (deploy azure-like storage to k8s and test e2e) would fit here?

However, the functional tests are preferred over e2e where possible.

pkg/billing/azureblobstoragebilling.go
Omrigan, line 1: Out of curiosity, did you consider the approach of turning Azure API into S3 (with e.g. [S3Proxy](https://github.com/gaul/s3proxy)), so that the alternative implementation is not needed?

pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,158 @@
...
+	//
+	// Unexported attributes follow this comment.
+	//
+
+	// Use generateKey for tests.
+	// Otherwise, keep empty.
+	generateKey func() string
+	// trackProgress is useful in tests, it's invoked when SDK sends blobs to Azure.
+	// Otherwise keep empty.
+	trackProgress func(bytesTransferred int64)
```
Omrigan, line 53: Is it used anywhere?

Omrigan (commented):
Some high-level questions.

pkg/billing/azureblobstoragebilling.go
antonyc: thanks, it's not used!

pkg/billing/azureblobstoragebilling.go
antonyc: I believe these third party proxies can bring new problems:
* how would you set it up
* how would you configure it
* is it 100% compatible
* what happens when it's down

etc. And I don't see much benefits. Frankly speaking this piece of "ok, there is Vector, but we still want to persist some events inside the applications" is a bit that I don't 100% understand: we're trying to make sure we still have events persisted even if vector is down, and this is hardly solvable with introducing another proxy.
I'd just try to make Vector more stable and remove all S3/Azure related code, but that is out of my scope.

pkg/billing/azureblobstoragebilling_test.go
antonyc: I saw the S3 tests. IIUC the tests in this PR are exactly functional, right? So is there need for e2e?

pkg/billing/azureblobstoragebilling.go
antonyc: I wanted to test that I can read what I wrote, not the format of the `AzureClient.generateKey()`. If you wish to test the format we can have another test on that.

pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,154 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+type AzureAuthType string
+
+const (
+	// AzureAuthTypeTests is used in tests
+	// It uses well-known storage account and key.
+	// See https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite
+	AzureAuthTypeTests AzureAuthType = "tests"
```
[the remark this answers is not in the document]
antonyc: I don't see support for ShareKeyCredentials which Azurite container uses. Tried to connect, without success.

.golangci.yml
```diff
@@ -4,6 +4,8 @@ run:
   issues-exit-code: 1
 
 issues:
+  exclude-files:
+    - '.*_test.go$'
```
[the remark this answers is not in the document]
antonyc: fixed, thanks.

pkg/billing/azureblobstoragebilling.go
Omrigan: I am just concerned about adding the field which is only used in tests.

Perhaps, you could extract the actual key generation into a standalone `func generateKey`, which would be the default value for the `cfg.generateKey`, also achieving the de-duplication with S3 code.

pkg/billing/azureblobstoragebilling.go
Omrigan: > third party proxies can bring new problems

Yeah, that's true. TBH, I'd just expect Azure to provide the S3 API. I wonder what the projects usually do: use proxy or have an alternative implementation. Maybe there are libraries on the client side, which can do this in a unified way 🤔 

>  Frankly speaking this piece of "ok, there is Vector, but we still want to persist some events inside the applications" is a bit that I don't 100% understand

Yeah, this [was discussed](https://neondb.slack.com/archives/C061CPK7UQL/p1711042495383509) before. Maybe this point, where we have to support additional storage, is a good moment to revisit that decision.

pkg/billing/azureblobstoragebilling_test.go
Omrigan: > IIUC the tests in this PR are exactly functional, right?

Yep, right, I am just hesitant to have 2 different types of tests for different storage providers. Maybe, that's fine.

pkg/billing/azureblobstoragebilling.go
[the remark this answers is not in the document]
Omrigan: Hmm, do they have incompatibility between `azidentity` and `azblob` 🤔 

Still, the test-related logic should reside in `_test.go`, and be dependency injected into the core logic.

sharnoff:
Thanks for tackling this! Haven't looked at the code yet, but a request about the PR title:

Could you rename the PR to the format `component/subsystem: title` (e.g. here, "agent/billing: use Azure ...")
(Here though, I'd also note that the PR doesn't switch the autoscaler-agent to _using_ Azure, it just adds support - IIUC?)

Probably we should have a workflow to enforce this -- I just haven't gotten around to opening an issue for it

antonyc:
> Could you rename the PR to the format component/subsystem: title (e.g. here, "agent/billing: use Azure ...")

I like https://www.conventionalcommits.org/en/v1.0.0/#summary, have you looked at it? There is a lot of tooling around it.

pkg/billing/azureblobstoragebilling.go
antonyc: > Maybe there are libraries on the client side, which can do this in a unified way 🤔

Sure, we once considered using https://github.com/google/go-cloud/tree/master/blob

pkg/billing/azureblobstoragebilling.go
antonyc: > I am just concerned about adding the field which is only used in tests.

It's not exposed to users of the package.

> Perhaps, you could extract the actual key generation into a standalone func generateKey, which would be the default value for the cfg.generateKey, also achieving the de-duplication with S3 code.

Yeah, thought about this too, and did that.

pkg/billing/azureblobstoragebilling.go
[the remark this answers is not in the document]
antonyc: Making  `AzureAuthTypeTests` -> `azureAuthTypeTests` is probably enough to not expose any test details to users, right?

pkg/billing/azureblobstoragebilling.go
antonyc: > Yeah, this [was discussed](https://neondb.slack.com/archives/C061CPK7UQL/p1711042495383509) before.

Participated voting for Vector as single sink to accept consumption events, dropping S3, Azure from the event producers.

antonyc:
>  (Here though, I'd also note that the PR doesn't switch the autoscaler-agent to using Azure, it just adds support - IIUC?)

It [does](https://github.com/neondatabase/autoscaling/pull/994/files#diff-0585b04c95eec9d44b4a16d1d3e73f15ef6e2033078f3451265ba1b5828770c4R111), did I miss something?

sharnoff:
> It [does](https://github.com/neondatabase/autoscaling/pull/994/files#diff-0585b04c95eec9d44b4a16d1d3e73f15ef6e2033078f3451265ba1b5828770c4R111), did I miss something?

Yeah, we'll need to enable it from the deploy side -- see e.g. [here](https://github.com/neondatabase/autoscaling-deploy/blob/1833bb9cad4c972af866489ebd5183e74f6aa883/patches.json#L137-L152). (in short: the YAMLs produced by the release don't have billing enabled, because there's nowhere to configure them; so enabling is done further downstream)

antonyc:
There is a PR in review (it's on you), and then the next followup PR that has not been published yet enables it for Azure.

pkg/billing/azureblobstoragebilling.go
Omrigan: > Yeah, thought about this too, and did that.

I imagined one more step - setting `keyTemplate` as the default value for `client.cfg.generateKey` (it will have to accept an argument), and overwriting it in tests, and always calling `cfg.generateKey`. This way there will be no test-only code in the main file.

pkg/billing/azureblobstoragebilling.go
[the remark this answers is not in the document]
Omrigan: But you still have AzureAuthType  as part of the json config.

I think, we can do similar to `generateKey` -> invent a callback `AzblobClient func() azblob.Client`, set it to the real-life implementation by default, and overwrite in the tests for the shared key implementation.

This, same is in `generateKey` case, achieves the state where the all testing-related implementations are contained in the `_test.go` files.

pkg/billing/azureblobstoragebilling.go
```diff
@@ -0,0 +1,143 @@
...
+		cfg: cfg,
+		c:   client,
+	}, nil
+}
+
+func handleAzureError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return AzureError{err}
```
Omrigan, line 142: The reason those error wrappers exist is to reduce the cardinality for the errors graph.

https://github.com/neondatabase/autoscaling/blob/8195df03ebe55f1dc185965a7f0649b9edd6e0bb/pkg/agent/billing/send.go#L154

I think the same should be done for azure.

pkg/billing/client.go
```diff
@@ -153,25 +157,33 @@ func (c S3Client) LogFields() zap.Field {
 	}))
 }
 
-func (c S3Client) send(ctx context.Context, payload []byte, _ TraceID) error {
-	// Source of truth for the storage format:
-	// https://github.com/neondatabase/cloud/issues/11199#issuecomment-1992549672
-
-	key := c.generateKey()
+func bytesForBilling(i []byte) ([]byte, error) {
```
Omrigan, line 160:
```suggestion
func compress(payload []byte) ([]byte, error) {
```

pkg/billing/client.go
```diff
@@ -153,25 +157,33 @@ func (c S3Client) LogFields() zap.Field {
...
 	buf := bytes.Buffer{}
 
 	gzW := gzip.NewWriter(&buf)
-	_, err := gzW.Write(payload)
+	_, err := gzW.Write(i)
```
Omrigan, line 164:
```suggestion
	_, err := gzW.Write(payload)
```

pkg/billing/azureblobstoragebilling_test.go
```diff
@@ -0,0 +1,181 @@
...
+			when: func(t *testing.T, i *input) {
+				i.payload = []byte("hello, billing data is here")
+				i.cfg.PrefixInContainer = "test-prefix"
+				i.cfg.Container = "test-container"
+				i.cfg.generateKey = func() string {
+					return "test-blob-name"
+				}
+				c, err := NewAzureBlobStorageClient(i.cfg)
+				require.NoError(t, err)
+				i.c = c
```
Omrigan, line 87: This common part can be set in input by default.

Omrigan (commented):
Looks close to done, few comments remaining!

Omrigan:
> I'd also note that the PR doesn't switch the autoscaler-agent to _using_ Azure, it just adds support - IIUC?

@antonyc  +1 for `s/use/support/` in PR subj.

pkg/billing/azureblobstoragebilling.go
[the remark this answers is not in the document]
antonyc: sure.

pkg/billing/azureblobstoragebilling.go
antonyc: sure.

antonyc:
@Omrigan please have a look.

pkg/billing/azureblobstoragebilling_test.go
Omrigan: I think that's probably fine. Is that what control plane uses? 

@sharnoff let's think if we need more of that kind of tests.

pkg/billing/azureblobstoragebilling.go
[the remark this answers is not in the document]
Omrigan: Looking perfect!

Omrigan (approved):
Great!

pkg/billing/azureblobstoragebilling_test.go
```diff
@@ -0,0 +1,189 @@
+package billing
```
sharnoff, line 1: Should this be `billing_test`?

pkg/billing/azureblobstoragebilling_test.go
antonyc: > I think that's probably fine. Is that what control plane uses?

this and test-containers package.

pkg/billing/azureblobstoragebilling_test.go
antonyc: that would hide the unexported attributes like getClient from the test package. I tend to not use different packages for testing. IIUC golang does the optimisation of not including testing dependencies unless they are explicitly used in non-test code.

sharnoff: > that would hide the unexported attributes like config.getClient from the billing package

Hm, I think these can be worked around with not too much effort, but it can be a follow-up PR.

> IIUC golang does the optimisation of not including testing dependencies unless they are explicitly used in non-test code.

I thought that was only the case for `_test` packages? if that's the case though, that's quite fancy!

sharnoff (approved):

pkg/billing/azureblobstoragebilling_test.go
sharnoff: Just read what you linked here: https://neondb.slack.com/archives/C04DTS134H0/p1720027537887639

Neat!

antonyc: > the customer reported that renaming all the (non-acceptance) test files in go-mysql-server to end in _test.go removed the testify symbols from his binary

yeah, seems like that.

PR #1000, opened 2024-07-04
Reduce log level of neonvm controller's successful reconcilation

MMeent:
"Successful reconcilation" is 99+% of the controller's logs, which in turn is ~11% of total log volume (in a 30-minute sample window taken just now), so reducing this will help save $$$.

https://neondb.slack.com/archives/C039YKBRZB4/p1720111839278329?thread_ts=1720109708.647239&cid=C039YKBRZB4


Omrigan:
These we deliberately added recently, in 3b08db3e2595922457b353f29888af25c7bbc163, so that we have 1 log line minimum per reconciliation. 

> reducing this will help save $$$

Do you happen to have an estimation, how much? Anyway, let's keep discussing in the attached slack thread.

sharnoff:
An alternate change was implemented in #1016. See justification for that approach there.

PR #1003, opened 2024-07-07
agent/core: Implement LFC-aware scaling

sharnoff:
Part of #872.
This builds on the metrics that will be exposed by neondatabase/neon#8298.

For now, we only look at the working set size metrics over various time windows.

The algorithm is somewhat straightforward to implement (see wss.go), but unfortunately seems to be difficult to understand *why* it's expected to work.

See also: https://www.notion.so/neondatabase/cca38138fadd45eaa753d81b859490c6

---

**Notes for review:**

~~Currently just a draft, pending validation that this does behave as expected.~~
Before merging, I want to:

1. ~~Add some tests for LFC-based scaling to `state_test.go`'s `Test_DesiredResourcesFromMetricsOrRequestedUpscaling`.~~
2. ~~Add some more tests in `wss_test.go`~~
3. ~~Add some more comments explaining some of the weirder things (the new algorithm, why `desiredResourcesFromMetricsOrRequestedUpscaling` now needs `hasMetrics`, etc.)~~

As a follow-up, I'd like to resurrect #737 to make this cleaner. I have a WIP branch for that locally; figured it'd be easier to get this PR in first, and then tidy up.

~~Also, for now this PR builds on #929 for consistency with what's deployed on staging.~~


deploy/agent/config_map.yaml
```diff
@@ -45,7 +48,7 @@ data:
           "secondsBetweenRequests": 5
         },
         "lfc": {
-          "port": 9399,
+          "port": 9499,
```
Omrigan, line 51: why?

pkg/agent/core/metrics.go
```diff
@@ -31,7 +35,8 @@ type LFCMetrics struct {
 	CacheMissesTotal float64
 	CacheWritesTotal float64
 
-	ApproximateWorkingSetSizeTotal float64 // approximate_working_set_size
+	// lfc_approximate_working_set_size_seconds
```
Omrigan, line 38: These are indexed by number of minutes?

pkg/agent/core/wss.go
```diff
@@ -0,0 +1,87 @@
...
+) float64 {
+	if cfg.WindowSize < 2 {
+		panic(fmt.Errorf("cfg.WindowSize must be >= 2 (got %v)", cfg.WindowSize))
+	}
+
+	w := cfg.WindowSize - 1
+
+	for t := cfg.InitialOffset; t < len(series)-w; t += 1 {
+
+		d0 := max(0.0, series[t]-series[t-w])
```
Omrigan, line 48: max with 0 is just a precaution, right? From how HLL is designed, this shouldn't be possible, right?

sharnoff:
Rebased onto main & squashed all commits -- merging left a lot of commits in the branch history.

Omrigan (commented):
Review in-progress.

pkg/agent/core/wss.go
sharnoff: Technically, the values in the metric aren't guaranteed to be computed at the same time, so it's in theory possible to get `series[t] < series[t-w]`

deploy/agent/config_map.yaml
sharnoff: 9399 is the "public" (stored in ext-metrics-*) metrics. 9499 is the metrics exporter just for autoscaling. With neondatabase/neon#8298, the latter actually is higher resolution, because we don't need to worry about storage.

pkg/agent/core/metrics.go
sharnoff: Yeah, currently the exposed metric name is `lfc_approximate_working_set_size_seconds`.

Although that metric _also_ is labeled with `duration_minutes`. Perhaps something there should change -- I'd be interested in your thoughts re: neondatabase/neon#8298

Omrigan: > Perhaps something there should change

Like what?

One thing metric `..._seconds` being labeled with `duration_minutes` is no ideal. See also https://github.com/neondatabase/neon/pull/8298#discussion_r1680794911.

pkg/agent/core/state.go
```diff
@@ -667,8 +671,11 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
 	// 2. Cap the goal CU by min/max, etc
 	// 3. that's it!
 
+	hasMetrics := s.Metrics != nil && (!*s.scalingConfig().EnableLFCMetrics || s.LFCMetrics != nil)
```
Omrigan, line 674: Maybe instead check separately for system metrics and lfc metrics, and if one exists, but the other doesn't, we just use the one which exists to compute maximum?

This can help in an event that e.g. LFC metric exposing/fetching breaks, and we would still be able to estimate goalCU based off CPU and memory.

pkg/agent/core/state.go
```diff
@@ -686,6 +693,44 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
...
+			if len(wssValues) < offsetIndex+windowSize {
+				s.warn("not enough working set size values to make scaling determination")
+			} else {
+				estimateWss := EstimateTrueWorkingSetSize(wssValues, WssEstimatorConfig{
+					MaxAllowedIncreaseFactor: 3.0, // hard-code this for now.
+					InitialOffset:            offsetIndex,
+					WindowSize:               windowSize,
+				})
+				projectLen := 0.5 // hard-code this for now.
+				predictedHighestNextMinute := ProjectNextHighest(wssValues, projectLen, estimateWss)
```
Omrigan, line 715: WDYT about filtering `wssValue` here for `< estimateWss`, so that you don't need to pass estimateWss, simplifying the semantics of ProjectNextHighest?

pkg/agent/core/state.go
```diff
@@ -922,6 +974,8 @@ func (s *State) Debug(enabled bool) {
 }
 
 func (s *State) UpdatedVM(vm api.VmInfo) {
+	hadLFCMetrics := *s.internal.scalingConfig().EnableLFCMetrics
```
Omrigan, line 977: Why not just do 
```
if !*s.internal.scalingConfig().EnableLFCMetrics {
     s.internal.LFCMetrics = nil
}
```
i.e. zeroing metrics when we detect "LFC is disabled", not "it was disabled, but became enabled".

pkg/agent/core/metrics.go
```diff
@@ -133,3 +141,72 @@ func (m *LFCMetrics) fromPrometheus(mfs map[string]*promtypes.MetricFamily) erro
...
+			duration: time.Minute * time.Duration(durationMinutes),
+			value:    m.GetGauge().GetValue(),
+		})
+	}
+
+	slices.SortFunc(pairs, func(x, y pair) int {
+		return cmp.Compare(x.duration, y.duration)
+	})
+
+	// Check that the values make are as expected: they should all be 1 minute apart, starting
```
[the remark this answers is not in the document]
sharnoff: Hm... I figured it'd be cleaner to "just" assume it's always a 1-minute gap for now, and then we can improve this after merging. But yeah, +1 that we should eventually land on that sort of implementation.

WDYT?

pkg/agent/core/state.go
sharnoff: Yeah, this is a tricky situation... the thing I was trying to prevent was flushing caches on autoscaler-agent restart because goalCU was low before fetching LFC metrics.

I thought about potentially making making it a "min" goalCU if one of the metrics is missing (i.e., allow scaling _up_ to that amount, but don't scale down if we're already above it). But that's somewhat complex to implement in the current function.

Curious for your thoughts on this.

pkg/agent/core/metrics.go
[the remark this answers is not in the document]
sharnoff: Plan is to merge as-is and tackle this in a follow-up.

deploy/agent/config_map.yaml
```diff
@@ -12,7 +12,10 @@ data:
         "defaultConfig": {
           "loadAverageFractionTarget": 0.9,
           "memoryUsageFractionTarget": 0.75,
-          "enableLFCMetrics": false
+          "enableLFCMetrics": false,
+          "lfcSizePerCU": 0.75,
```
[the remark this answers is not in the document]
sharnoff: Plan: go with the rename; currently this ratio is hard-coded in vm-monitor anyways — when we change to setting it via the cplane, we'll override here as well.

Once every VM has this, we can move towards requiring this field per each VM (or, warning if it isn't set, or something).

pkg/agent/core/metrics.go
sharnoff: Fixed in 496a333a5aa4452f7cbb1230ca7f3ca3f65ff9f0

deploy/agent/config_map.yaml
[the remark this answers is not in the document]
sharnoff: Added this to #1011

pkg/agent/core/state.go
sharnoff: Fixed in 5a65920df49af712b6e6a4271e87293a109ab8ad.

pkg/agent/core/state.go
sharnoff: Implemented this in 43a75dcd34103e3096c9bf722fe310cc3f10ea03.

Follow-up is in #1011.

pkg/agent/core/wss.go
```diff
@@ -0,0 +1,92 @@
...
+	// border between the two windows.
+	MaxAllowedIncreaseFactor float64
+	// InitialOffset is the index of the minimum working set size we must consider.
+	//
+	// In practice, this is taken from the scaling config's LFCMinWaitBeforeDownscaleMinutes, with
+	// the expectation that datapoints are all one minute apart, starting at 1m. So a value of 15m
+	// translates to an InitialOffset of 14 (-1 because indexes start at zero, but the first
+	// datapoint is 1m).
+	InitialOffset int
+	// WindowSize sets the number of datapoints averaged out in a slope calculation.
```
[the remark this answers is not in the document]
sharnoff: Yeah I guess technically they aren't averaged out, but if the values are monotonically non-decreasing, it's an equivalent calculation.

I'll add a comment clarifying that but keep the original - because I think "averaged out" is a much simpler thing to explain :sweat_smile:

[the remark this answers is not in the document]
sharnoff: Added in 2f81286c1fa3eb7541eb738fbc3a754440245f91

pkg/agent/core/state.go
sharnoff: Fair enough, yeah - implemented in a96af7b0c8985bf8381becb2c398d90fe9f77507

deploy/agent/config_map.yaml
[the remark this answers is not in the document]
Omrigan: What about the rename?

pkg/agent/core/wss.go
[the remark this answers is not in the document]
Omrigan: I don't understand why they are equivalent...

Code currently is `series[t]-series[t-w]`

"Averaged out" in my understanding would result in:
```
sum := 0
for( i := t-w; i < t; i++) {
  sum += series[i]
}
return sum/w
```

Are these equivalent?

pkg/agent/core/wss.go
```diff
@@ -0,0 +1,93 @@
...
+	series []float64,
+	cfg WssEstimatorConfig,
+) float64 {
+	if cfg.WindowSize < 2 {
+		panic(fmt.Errorf("cfg.WindowSize must be >= 2 (got %v)", cfg.WindowSize))
+	}
+
+	w := cfg.WindowSize - 1
+
+	for t := cfg.InitialOffset; t < len(series)-w; t += 1 {
```
Omrigan, line 50: If cfg.WindowSize > cfg.InitialOffset, is it possible to get out-of-bounds here?

deploy/agent/config_map.yaml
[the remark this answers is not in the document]
sharnoff: Oh good catch, thanks. Fixed in e18cf9ddd7b238ce8a9c63a673686dda0919ba2a

pkg/agent/core/wss.go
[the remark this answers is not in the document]
sharnoff: Ah I see what you mean. I was thinking from the perspective that the "slope" is averaged out (rather than the points), but I think your interpretation is probably more likely.

Reworded in c3b83a2aef62b275fda362d7051703f9e5640c54 & 7daeb3ec97d7dc6849c3c7614c7906c91883f2e8

> WindowSize sets the offset for datapoints used in the calculation of the slope before & after a point. For window size W, we calculate the slope at point P as value[P]-value[P-(W-1)].

Open to suggestions for further rewording.

pkg/agent/core/wss.go
sharnoff: _Technically_ we wouldn't because if `t >= len(series)` then we fail the condition in the for loop, but we could _technically_ get out of bounds if `len(series) == 0`.

However, in `core/state.go` we already check to make sure we have enough datapoints:

https://github.com/neondatabase/autoscaling/blob/7daeb3ec97d7dc6849c3c7614c7906c91883f2e8/pkg/agent/core/state.go#L714-L716

pkg/agent/core/wss.go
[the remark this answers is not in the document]
Omrigan: I think that's fine.

However, this `-1` is annoying. WDYT about this:

> SlopeDistance sets the distance between points used for slope calculation. At point i we would compare two slopes: (arr[i] - arr[i - SlopeDistance]) and (arr[i + SlopeDistance] - arr[i]). SlopeDistance must be >=1.

pkg/agent/core/wss.go
Omrigan: If InitialOffset=1, WindowSize=10, len(series)=100

then in the first iteration we will take `series[1] - series[1-9]` which is not good.

sharnoff: Oh, good point, I guess I misread your comment :sweat_smile:

Added a check in 253c8f16239149426023a83be0c777b3c1746830

pkg/agent/core/wss.go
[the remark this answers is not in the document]
sharnoff: discussed -- I'll explain the -1 in a comment next to it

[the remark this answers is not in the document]
sharnoff: Added a comment in d1fb91592ef8dd88863038963e3d4b4499655f4c

Omrigan (approved):

PR #1005, opened 2024-07-10
Upgrade k8s deps to 1.28

SergeyMelnikov:
cc: https://github.com/neondatabase/cloud/issues/7922


neonvm/apis/neonvm/v1/webhook_suite_test.go
```diff
@@ -90,12 +92,16 @@ var _ = BeforeSuite(func() {
...
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
-		Scheme:             scheme,
-		Host:               webhookInstallOptions.LocalServingHost,
-		Port:               webhookInstallOptions.LocalServingPort,
-		CertDir:            webhookInstallOptions.LocalServingCertDir,
-		LeaderElection:     false,
-		MetricsBindAddress: "0",
+		Scheme:         scheme,
+		LeaderElection: false,
+		WebhookServer: webhook.NewServer(webhook.Options{ //nolint:exhaustruct // Other fields set to default values by webhook Constructor
```
sharnoff, line 97: If you like, another option is adding the `Options` types to the set of types we exclude in `.golangci.yml`

sharnoff (commented):
Looks good!

I see that k8s 1.28.12 has since been released. Should we use that for deps?

sharnoff (approved):

Issue #889, opened 2024-04-02
neonvm-controller does not update VM status if reconcile fails


sharnoff:
## Problem description / Motivation

Maybe this is the desired behavior, however: Currently, if reconciling fails, the neonvm-controller does not update the VM status, even if it has changed.

This means that it's possible for more up-to-date status information to be arbitrarily delayed due to reconcile failures (either due to transient issues or bugs).

I think this is probably not desirable as it impacts liveness, and as we move towards greater dependence on what's reported by the controller, we may want to tighten this up.

## Feature idea(s) / DoD

Don't skip status update if there was some earlier failure.

PR #1009, opened 2024-07-14
neonvm-controller: patch status if reconcile fails (#889)

agadelshin:
Added status update on fail. There is a way to just run it in a defer, but it's only 2 places so I guess it's more straightforward solution.

Also changed Update to Patch, added request of virtualmachine before to avoid conflicts.

And a last one - removed requeue after succesful reconcile, it's not required to reconcile if nothing happens, reconcile will be queued if there will be a change in a VirtualMachine or Pod with VirtualMachine owner. If there is another reason to keep it in a constant loop - please tell, will update it. Otherwise it'll save some CPU cycles and make logs more readable.

Tested it manually on a virtualmachine inside GCP, if there is a way to run autoscaling on Silicon Apple, I'd be happy to know (x86-64 emulation in QEMU is too slow, took hours to build everything).

### DIFF
--- a/neonvm/controllers/vm_controller.go
+++ b/neonvm/controllers/vm_controller.go
@@ -171,19 +171,40 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 	if err := r.doReconcile(ctx, &vm); err != nil {
 		r.Recorder.Eventf(&vm, corev1.EventTypeWarning, "Failed",
 			"Failed to reconcile (%s): %s", vm.Name, err)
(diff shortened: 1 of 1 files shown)


agadelshin:
More I think about it - if requeue is still required for some reason - it's better to increase duration between runs to some reasonable period like 10-30 seconds.

agadelshin:
Removed requeAfter changes since it's a part of #1016

Omrigan:
Closed in favor of #1102 and #1103.

@agadelshin Thank you for the proactive proposal!

Issue #900, opened 2024-04-11
Alpine base images should be updated


sharnoff:
## Problem description / Motivation

We're currently using alpine 3.16. We should update to the latest (3.19).

## Feature idea(s) / DoD

All alpine images used are updated to 3.19, or whatever the latest is when this issues is resolved.

## Implementation ideas

See previous attempt here: #530.

This is difficult primarily because of the copying of shared libraries that we do in vm-builder. It _may_ be easier to first audit all the commands we use, then run `ldd` on them to see the shared libraries, and derive a new list from scratch, rather than by trial-and-error.

stradig:
I checked the repository for version 3.15 which is end-of-life since a year. The repository is still up and usable at https://dl-cdn.alpinelinux.org/alpine/v3.15/main/x86_64/. So at least we should still be able to build against it. But of course no more security updates, so we should upgrade in any case.

stradig:
This is not urgent because we can still build. We don't have the capacity to fix this now, will look at this in June.

stradig:
1. Still want to upgrade to newer version, but not before GA, currently also blocked in CI. Putting it into backlog, this is not urgent.

bayandin:
FYI: currently, we're on `alpine:3.19`, the EOL is 01 Nov 2025 (in 8 months)

from https://endoflife.date/alpine

PR #1010, opened 2024-07-16
Update alpine base images: 3.16 → 3.17

sharnoff:
Part of #900; extracted from #1008.

This PR:

1. Replaces the shared library copying in vm-builder with a script to handle that for us, using the output from `ldd`
2. Bumps the alpine versions everywhere to 3.17

The two parts are in two separate commits. It probably makes sense to merge them via rebase-and-merge.

Notably, the bump to 3.17 includes bumping QEMU from 7.0.0 to 7.1.0.


stradig (approved):
LGTM, thanks!

petuhovskiy (approved):

PR #1013, opened 2024-07-19
agent: Reduce logging

petuhovskiy:
Try to reduce some very common logs:
- Change interval for recurring scheduler requests (5s -> 15s)
- Log 1-2 log entries instead of 4 for each scheduler request
- Log successful healthchecks only once per 10 requests
- Do not log "Making metrics request to VM", same info as in "Updated metrics"

ref https://github.com/neondatabase/cloud/issues/15591
ref https://github.com/neondatabase/cloud/issues/15605

### DIFF
--- a/pkg/agent/dispatcher.go
+++ b/pkg/agent/dispatcher.go
@@ -4,7 +4,6 @@ package agent
 // connection through a simple RPC-style protocol.
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -323,12 +322,7 @@ func (disp *Dispatcher) send(ctx context.Context, logger *zap.Logger, id uint64,
 	// by base64 encoding it, so use RawMessage to avoid serializing to []byte
 	// (done by SerializeMonitorMessage), and then base64 encoding again
 	raw := json.RawMessage(data)
-	if bytes.HasPrefix(data, []byte("{\"content\":{},\"type\":\"HealthCheck\",\"id\":")) {
-		// don't log health checks, they're too frequent
-		logger.Debug("sending message to monitor", zap.ByteString("message", raw))
-	} else {
-		logger.Info("sending message to monitor", zap.ByteString("message", raw))
-	}
+	logger.Debug("sending message to monitor", zap.ByteString("message", raw))
 	return wsjson.Write(ctx, disp.conn, &raw)
 }
 
@@ -440,12 +434,7 @@ func (disp *Dispatcher) HandleMessage(
 		return fmt.Errorf("Error receiving message: %w", err)
 	}
 
-	if bytes.HasPrefix(message, []byte("{\"type\":\"HealthCheck\",\"id\":")) {
-		// don't log health checks, they're too frequent
-		logger.Debug("(pre-decoding): received a message", zap.ByteString("message", message))
-	} else {
-		logger.Info("(pre-decoding): received a message", zap.ByteString("message", message))
-	}
+	logger.Debug("(pre-decoding): received a message", zap.ByteString("message", message))
 
 	var unstructured map[string]interface{}
 	if err := json.Unmarshal(message, &unstructured); err != nil {
--- a/pkg/agent/dispatcher.go
+++ b/pkg/agent/dispatcher.go
@@ -4,6 +4,7 @@ package agent
 // connection through a simple RPC-style protocol.
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -165,6 +166,10 @@ func NewDispatcher(
 		var firstSequentialFailure *time.Time
 		continuedFailureAbortTimeout := time.Second * time.Duration(runner.global.config.Monitor.MaxHealthCheckSequentialFailuresSeconds)
 
+		// if we don't have any errors, we will log only every 10th successful health check
+		const maxSuccessiveSkips = 10
+		var successesSkipped int
+
 		for {
 			select {
 			case <-ctx.Done():
@@ -172,9 +177,18 @@ func NewDispatcher(
 			case <-ticker.C:
 			}
 
+			startTime := time.Now()
 			_, err := disp.Call(ctx, logger, timeout, "HealthCheck", api.HealthCheck{})
+			endTime := time.Now()
+
+			logFields := []zap.Field{
+				zap.Duration("duration", endTime.Sub(startTime)),
+				zap.Int("successesSkipped", successesSkipped),
+			}
+
 			if err != nil {
-				logger.Error("vm-monitor health check failed", zap.Error(err))
+				logger.Error("vm-monitor health check failed", append(logFields, zap.Error(err))...)
+				successesSkipped = 0
 
 				if firstSequentialFailure == nil {
 					now := time.Now()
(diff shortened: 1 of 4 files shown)


pkg/agent/core/state.go
```diff
@@ -199,9 +206,10 @@ func (ns *neonvmState) ongoingRequest() bool {
 func NewState(vm api.VmInfo, config Config) *State {
 	return &State{
 		internal: state{
-			Config: config,
-			Debug:  false,
-			VM:     vm,
+			Config:          config,
+			Debug:           false,
+			LoggedResources: api.Resources{VCPU: 0, Mem: 0},
```
sharnoff, line 211: This is a somewhat tricky change to do well, and will intersect with #983. Let's split into a separate PR?

(broader context is that `core.State` is complicated and we do want to periodically log the desired resources even if they haven't changed (either time-based or "metrics changed"), and sometimes we'll want to immediately log them even if they're equal to what the VM is using.)

pkg/agent/dispatcher.go
```diff
@@ -165,16 +166,29 @@ func NewDispatcher(
 		var firstSequentialFailure *time.Time
 		continuedFailureAbortTimeout := time.Second * time.Duration(runner.global.config.Monitor.MaxHealthCheckSequentialFailuresSeconds)
 
+		// if we don't have any errors, we will log only every 10th successful health check
+		const maxSuccessiveSkips = 10
+		var successesSkipped int
```
sharnoff, line 171: WDYT about not resetting this when we log, and instead just counting the number of sequential successes/failures and including the opposite in the logs?

(i.e., on first success, log the number of failures; and vice versa)

pkg/agent/dispatcher.go
```diff
@@ -302,7 +323,12 @@ func (disp *Dispatcher) send(ctx context.Context, logger *zap.Logger, id uint64,
 	// by base64 encoding it, so use RawMessage to avoid serializing to []byte
 	// (done by SerializeMonitorMessage), and then base64 encoding again
 	raw := json.RawMessage(data)
-	logger.Info("sending message to monitor", zap.ByteString("message", raw))
+	if bytes.HasPrefix(data, []byte("{\"content\":{},\"type\":\"HealthCheck\",\"id\":")) {
+		// don't log health checks, they're too frequent
+		logger.Debug("sending message to monitor", zap.ByteString("message", raw))
+	} else {
+		logger.Info("sending message to monitor", zap.ByteString("message", raw))
+	}
```
sharnoff, line 331: IMO it's ok to downgrade this to debug in all cases. Any messages we send that aren't health checks will already have log lines from `exec_monitor.go`

pkg/agent/dispatcher.go
```diff
@@ -413,7 +439,13 @@ func (disp *Dispatcher) HandleMessage(
...
 		return fmt.Errorf("Error receiving message: %w", err)
 	}
-	logger.Info("(pre-decoding): received a message", zap.ByteString("message", message))
+
+	if bytes.HasPrefix(message, []byte("{\"type\":\"HealthCheck\",\"id\":")) {
+		// don't log health checks, they're too frequent
+		logger.Debug("(pre-decoding): received a message", zap.ByteString("message", message))
+	} else {
+		logger.Info("(pre-decoding): received a message", zap.ByteString("message", message))
+	}
```
sharnoff, line 448: Similar thing here w.r.t. downgrading to "always debug"

pkg/agent/runner.go
```diff
@@ -778,7 +779,12 @@ func (r *Runner) DoSchedulerRequest(
 	}
 	request.Header.Set("content-type", "application/json")
 
-	logger.Info("Sending request to scheduler", zap.Any("request", reqData))
+	if reqData.LastPermit != nil && *reqData.LastPermit == reqData.Resources {
+		// If the last permit is the same as the current request, we can skip request logging.
+		logger.Debug("Sending request to scheduler", zap.Any("request", reqData))
+	} else {
+		logger.Info("Sending request to scheduler", zap.Any("request", reqData))
+	}
```
sharnoff, line 787: I'd rather keep this as debug-only and put the conditional in `exec_plugin.go`, so that it takes place while we hold the lock (just nicer to prioritize log lines that are guaranteed to be in the same order as the operations that took place :sweat_smile:)

pkg/agent/runner.go
```diff
@@ -807,7 +813,8 @@ func (r *Runner) DoSchedulerRequest(
 		return nil, fmt.Errorf("Bad JSON response: %w", err)
 	}
 
-	logger.Info("Received response from scheduler", zap.Any("response", respData))
+	// there will be "Plugin request successful" INFO log right after
```
sharnoff, line 816: I found this comment a little confusing ("where?" ; "in what cases?"). IMO it's ok to say nothing here, but adding more context about where the log line comes from would also be ok

sharnoff (commented):
nice! A few comments, but broadly pretty good

pkg/agent/core/state.go
petuhovskiy: Splitted into https://github.com/neondatabase/autoscaling/pull/1014

pkg/agent/runner.go
petuhovskiy: My idea was that `exec_plugin.go` doesn't know exact `reqBody` sent to the scheduler, so request log in `runner.go` has more context.

Also I'm not sure about the conditional, what is the best balance between logs cost / usefulness here in your opinion?

WDYT about such log levels and no conditions?
1. [info] Starting plugin request (exec_plugin.go)
2. [debug] Sending request to scheduler (runner.go)
3. [debug] Received response from scheduler (runner.go)
4. [info] Plugin request successful (exec_plugin.go)

pkg/agent/dispatcher.go
petuhovskiy: I didn't get the idea, do think just counting totals would work?

I.e. to maintain `totalSuccessCnt` and `totalFailCnt` and always add them to `logFields`?

Then print successful healthchecks only if `totalSuccessCnt % 10 == 0`?

sharnoff: Roughly, yeah- with the additions that

1. `totalSuccessCnt` is reset to 0 on failure
2. `totalFailCnt` is reset to 0 on success
3. Failure-then-success logs `totalFailCnt`
4. Success-then-failure logs `totalSuccessCnt`

petuhovskiy: Ok, made the change.

> Failure-then-success logs totalFailCnt
> Success-then-failure logs totalSuccessCnt

IMO it's not necessary to do this. I pushed a commit now which just counts oks/fails in a row.

pkg/agent/runner.go
sharnoff: > WDYT about such log levels and no conditions?

That makes sense, yeah - I think we _could_ also change (1) to debug in the cases where it doesn't change, but it's easy enough to change that in a follow-up

> My idea was that exec_plugin.go doesn't know exact `reqBody` sent to the scheduler, so request log in `runner.go` has more context.

Yeah, that's what I was thinking first as well, but eventually thought it'd be better to keep the guarantees around log lines being in the right order -- especially because IIRC `reqBody` can be exactly determined from the action given to `exec_plugin.go`

sharnoff (approved):
lgtm, nice!

PR #1015, opened 2024-07-19
plugin: Log only once on successful agent request

sharnoff:
Previously we'd log 4 times for each request.

This change moves all the fields into a single line that's emitted while we hold the state lock, so that we maintain the guarantee that log lines are produced in order of the changes that were made to the state.

---

Part of neondatabase/cloud#15591.


pkg/plugin/run.go
```diff
@@ -250,26 +243,34 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
...
 	resp := api.PluginResponse{
 		Permit:  permit,
 		Migrate: migrateDecision,
 	}
-	return &resp, 200, nil
+
+	logger.Info(
+		"Handled agent request",
+		zap.Object("verdict", verdict),
+		zap.Int("status", 200),
```
petuhovskiy, line 255:
```suggestion
		zap.Int("status", status),
```

just aesthetics :)

petuhovskiy (approved):

pkg/plugin/run.go
sharnoff: Oh, good catch - meant to do that. Thanks

pkg/plugin/run.go
```diff
@@ -250,26 +243,34 @@ func (e *AutoscaleEnforcer) handleAgentRequest(
...
-	logger *zap.Logger,
 	pod *podState,
 	node *nodeState,
 	cu api.Resources,
 	req api.Resources,
 	lastPermit *api.Resources,
 	startingMigration bool,
 	supportsFractionalCPU bool,
-) (api.Resources, int, error) {
+) (verdictSet, api.Resources, int, error) {
```
Omrigan, line 270: Never seen 4 return values 0_0

Doesn't seem like a problem though.

Omrigan (approved):

Issue #764, opened 2024-01-25
Basic unit testing coverage for scheduler plugin


sharnoff:
Optional for neondatabase/company_projects#187.

## Problem description / Motivation

Currently the only testing we have for the scheduler plugin comes from the end-to-end tests. The scheduler actually must handle a wide variety of edge cases, and without proper testing of this component, we're at risk for unexpected regressions.

## Feature idea(s) / DoD

Just _some_ amount of tests. It's a gradual process, we'll add more as we go.

## Implementation ideas

Similar to #763, we may need significant refactoring to make this feasible.

## Related issues

- #666
- #763

sharnoff:
I think this was mostly handled by #1163 with the new `state` package; closing as completed.

cc @Omrigan - in case you have thoughts, feel free to reopen

PR #1023, opened 2024-08-06
plugin: Add basic unit tests for resourceTransitioner

sharnoff:
Found myself wanting these as part of #517, and it's been on the todo list for a while from #764.

Some of the tests currently aren't working as expected. They are configured to pass, acting as if the undesired outcome is expected. The bugfix will be in a separate PR.

---

**Notes for review:** There's still other methods I'd like to get test coverage for, so I think this doesn't completely fulfill #764. But, IMO it's a good start, at least -- and should definitely be useful for #517 :)

### DIFF
--- a/pkg/plugin/trans_test.go
+++ b/pkg/plugin/trans_test.go
@@ -112,16 +112,16 @@ func Test_handleReserve(t *testing.T) {
 		{
 			name: "AlreadyOverbudgetWithBuffer",
 
-			podMin:             8,
+			podMin:             4,
 			podMax:             8,
 			nodeTotal:          12,
 			nodeReservedBefore: 16,
-			nodeBufferBefore:   0,
+			nodeBufferBefore:   8,
 
 			overbudget:        true,
-			verdict:           "node reserved 16 + 8 -> 24 of total 12",
+			verdict:           "node reserved 16 [buffer 8] + 8 [buffer 4] -> 24 [buffer 12] of total 12",
 			nodeReservedAfter: 24,
-			nodeBufferAfter:   0,
+			nodeBufferAfter:   12,
 		},
 	}
 
--- a/pkg/plugin/trans_test.go
+++ b/pkg/plugin/trans_test.go
@@ -1,0 +1,521 @@
+package plugin
+
+// tests for the core logic in resourceTransitioner and related methods.
+
+import (
+	"testing"
+
+	"github.com/samber/lo"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_handleReserve(t *testing.T) {
+	cases := []struct {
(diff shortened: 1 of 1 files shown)


pkg/plugin/trans_test.go
```diff
@@ -0,0 +1,521 @@
...
+			nodeReservedBefore: 16,
+			nodeBufferBefore:   0,
+
+			overbudget:        true,
+			verdict:           "node reserved 16 + 8 -> 24 of total 12",
+			nodeReservedAfter: 24,
+			nodeBufferAfter:   0,
+		},
+		{
+			name: "AlreadyOverbudgetWithBuffer",
```
petuhovskiy, line 113: This config is the same as `AlreadyOverbudget`, I think you forgot to add buffer here

petuhovskiy:
https://github.com/neondatabase/autoscaling/blob/2bf8d3e9e703e902f5963252d46da222528fa1ef/pkg/plugin/trans.go#L275

FYI this condition is always true, looks like a bug probably?

pkg/plugin/trans_test.go
```diff
@@ -0,0 +1,521 @@
...
+				buffer:   0,
+				pressure: 0,
+			},
+			nodeAfter: node{
+				reserved: 13,
+				buffer:   0,
+				pressure: 0,
+			},
+		},
+	}
```
petuhovskiy, line 485: Probably related to the FIXMEs above, found something in the `trans.go`:

First we set `forceApprovalMinimum` to:
https://github.com/neondatabase/autoscaling/blob/2bf8d3e9e703e902f5963252d46da222528fa1ef/pkg/plugin/trans.go#L191-L194

Then we use it as the minimum increase here:
https://github.com/neondatabase/autoscaling/blob/2bf8d3e9e703e902f5963252d46da222528fa1ef/pkg/plugin/trans.go#L315-L316

Currently it means that that plugin always allows scaling up to 2 times from the `lastPermit` value, probably a bug?

petuhovskiy (approved):

sharnoff:
> FYI this condition is always true, looks like a bug probably?

Ah good catch, that is a bug, yeah. It should be `oldState.pod.Reserved`.

pkg/plugin/trans_test.go
sharnoff: Exactly, yup (previously opened a fix PR in #1025).

pkg/plugin/trans_test.go
sharnoff: Yup, good catch - thanks. Fixed in 2e8dbc2eff31dcdb09c13f77ab8b0cfd7ff22e2f

PR #1025, opened 2024-08-06
plugin/trans: Fix over-allowing from forceApprovalMinimum

sharnoff:
Noticed as part of #1023.

The current (incorrect) behavior is that `forceApprovalMinimum` is relative to the current resources when it's actually intended (and documented as such) to be absolute. So in practice, we're approving autoscaler-agent requests that we shouldn't -- allowing resource increases above the node maximum when we don't need to.

`forceApprovalMinimum` was added as part of #936, and presumably we didn't catch this because of the lack of unit tests.

### DIFF
--- a/pkg/plugin/trans.go
+++ b/pkg/plugin/trans.go
@@ -313,8 +313,7 @@ func (r resourceTransitioner[T]) handleRequestedGeneric(
 		// the factor.
 		maxIncrease := (remainingReservable / opts.factor) * opts.factor
 		// ... but we must allow at least opts.forceApprovalMinimum
-		remainingBelowForcedApproval := util.SaturatingSub(opts.forceApprovalMinimum, r.pod.Reserved)
-		increaseFromForceApproval := (remainingBelowForcedApproval / opts.factor) * opts.factor
+		increaseFromForceApproval := util.SaturatingSub(opts.forceApprovalMinimum, r.pod.Reserved)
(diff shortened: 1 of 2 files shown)


pkg/plugin/trans.go
```diff
@@ -313,7 +313,9 @@ func (r resourceTransitioner[T]) handleRequestedGeneric(
 		// the factor.
 		maxIncrease := (remainingReservable / opts.factor) * opts.factor
 		// ... but we must allow at least opts.forceApprovalMinimum
-		maxIncrease = util.Max(maxIncrease, opts.forceApprovalMinimum)
+		remainingBelowForcedApproval := util.SaturatingSub(opts.forceApprovalMinimum, r.pod.Reserved)
+		increaseFromForceApproval := (remainingBelowForcedApproval / opts.factor) * opts.factor
```
petuhovskiy, line 317: Why do we need to align it with the `factor` here? What will happen if we don't?

petuhovskiy (approved):

pkg/plugin/trans.go
sharnoff: Fair question — after thinking about it some more, I think it's not needed. In theory we can hit that if a VM is created with bounds that aren't a multiple of the configured compute units, but there isn't _really_ a correct way to handle that.

Removed it in a204813627eb0b4d9da828c5931b1e6c0ea3c413

PR #1026, opened 2024-08-07
neonvm-controller: Fix immutable `.spec.guest.memoryProvider`

sharnoff:
In the original PR (#970), I'd intended to make this field mutable and added a comment about it, but forgot to actually remove it from the list of immutable fields.

*Probably* doesn't matter much, but IMO it's worth cleaning up.


stradig (approved):
LGTM

PR #1029, opened 2024-08-08
agent: add metric and log warning for denied scaling

### LINKED ISSUE #940: Metrics and clearer logs for denied scaling

We don't have clear signals for when the scheduler denies upscaling, or the vm-monitor denies downscaling.

This came up here: https://neondb.slack.com/archives/C03TN5G758R/p1716167732610919?thread_ts=1716166712.233079

## Feature idea(s) / DoD

The autoscaler-agent can:

1. Expose new instance-level metrics for amount of scaling requests that were not satisfied
2. Change the log level of scheduler responses that didn't grant the full request to "warn", rather than just "info"

## Implementation ideas

Should be simple enough to inject these alongside the existing logs & metrics, wherever they're defined (IIRC exec_bridge / executor ?)

mikhail-sakhnov:
Add new VM metrics called autoscaling_agent_scaling_full_denials_total and autoscaling_agent_scaling_partial_approvals_total.

Closes #940


pkg/agent/runner.go
```diff
@@ -830,7 +830,12 @@ func (r *Runner) DoSchedulerRequest(
 		return nil, fmt.Errorf("Bad JSON response: %w", err)
 	}
 
-	logger.Debug("Received response from scheduler", zap.Any("response", respData))
+	if respData.Permit.HasFieldLessThan(resources) {
```
Omrigan, line 833: There are similar metrics added by Shayan recently: https://github.com/neondatabase/autoscaling/pull/855/files#diff-80865f7ddcbe6bce5bd007849eb00e5431d6085bb2e133f12528733b7bd8e764R54

It is reported in slightly different way - not directly through prometheus export, but through periodical "isStuck" check. 

So, we can have four outcomes:
1. Request failed
2. Request was fully denied
3. Request was partially approved
4. Request can be fully approved

Shayan's "metric" is incremented when (1) or (2) happens, yours is incremented if (2) or (3) happens.

I wonder, perhaps we actually want two metrics, one for (2), one for (3). With this distinction, the metric name could also easier convey the semantics, e. g. `autoscaling_agent_scheduler_plugin_partial_approvals_total` and `autoscaling_agent_scheduler_plugin_full_denials_total`.

pkg/agent/runner.go
```diff
@@ -830,7 +830,12 @@ func (r *Runner) DoSchedulerRequest(
...
+		r.global.metrics.schedulerSchedulingDeniedResponses.Inc()
+		logger.Warn("Received response from scheduler, permitted resources are insufficient", zap.Any("requested", resources), zap.Any("response", respData))
```
Omrigan, line 835: WDYT about keeping the same log message `"Received response from scheduler", zap.Any("response", respData),  zap.Any("requested", resources)`, but changing the log level? 

This way it can be easily filtered in loki by any predicate: `response < requested` or  `response > 0` or `response == 0`. 

CC @sharnoff for thoughts

Omrigan:
About the commit message:

> issue940: add metric and log warning for denied scaling
The subject format is "component: what changes", so in this case should be smth like

`agent: add metric and emit warn when the scaling is not fully approved`

> Closes https://github.com/neondatabase/autoscaling/issues/940

This line usually goes to the end. Also, it will close the ticket, but I prefer to defer the closure until the new metrics are configured on the dashboards, so those are not forgotten. In such cases, I write "Part of #940".

> issue940: add metric and log warning for denied scaling

No need to repeat it

> The PR adds new VM metric called autoscaling_agent_scheduler_plugin_denied_responses_total
> The metric is incremented in case if returned response has lower resources than requested.

Should use imperative mood, e. g. "Add new VM metric..."

Omrigan (commented):
Good job! 

However, this only concerns the upscaling. For downscaling, the agent has to obtain permit from `vm-monitor`, not `plugin`. But the logic is more-or-less symmetrical.

mikhail-sakhnov:
I work on adding the metric for downscaling and fixing other notices. @Omrigan thanks for the feedback.

pkg/agent/runner.go
mikhail-sakhnov: I want to clarify, what do you mean by full and partial denials? 

Do I understand correctly, that the full deny is denied on the grounds of both cpu and memory requested, and partial if on the grounds of only one type of resource?

pkg/agent/runner.go
mikhail-sakhnov: In general, it sounds good, but does it mean we will have a warning-level message for each response, unconditionally, if it fails or succeeds?

Omrigan: Not necessarily, we can have the same log message, but printed with a different log level.

pkg/agent/execbridge.go
```diff
@@ -52,20 +67,26 @@ func (iface *execPluginInterface) Request(
...
+			return true, scalingResponseTypePartiallyApproved
 		}
-		return false // scheduler denied the request
+		return false, scalingResponseTypeDenied // scheduler denied the request
 	}()
 	iface.runner.status.update(iface.runner.global, func(ps podStatus) podStatus {
 		if !successful {
+			switch responseType {
+			case scalingResponseTypePartiallyApproved:
+				iface.runner.global.metrics.scalingPartialApprovalsTotal.WithLabelValues(upscaleDirectionLabel).Inc()
```
Omrigan, line 86: Partial approval is considered a success, so this is code is unreachable

pkg/agent/execbridge.go
```diff
@@ -52,20 +67,26 @@ func (iface *execPluginInterface) Request(
 		iface.runner.recordResourceChange(*lastPermit, resp.Permit, iface.runner.global.metrics.schedulerApprovedChange)
 	}
 
-	successful := func() bool {
+	successful, responseType := func() (bool, scalingResponseType) {
```
Omrigan, line 70: I think you don't need to keep "successful" separate, you can just check responseType Below in the callback.

pkg/agent/execbridge.go
```diff
@@ -52,20 +67,26 @@ func (iface *execPluginInterface) Request(
...
 		if lastPermit != nil && *lastPermit != resp.Permit { // request is partially approved by the scheduler
-			return true
+			return true, scalingResponseTypePartiallyApproved
 		}
-		return false // scheduler denied the request
+		return false, scalingResponseTypeDenied // scheduler denied the request
 	}()
 	iface.runner.status.update(iface.runner.global, func(ps podStatus) podStatus {
 		if !successful {
+			switch responseType {
```
Omrigan, line 84: I don't think this needs to be inside the callback.

pkg/agent/execbridge.go
```diff
@@ -34,6 +34,21 @@ func makePluginInterface(r *Runner) *execPluginInterface {
...
+
+const (
+	scalingResponseTypeDenied            = "denied"
+	scalingResponseTypeApproved          = "approved"
+	scalingResponseTypePartiallyApproved = "partiallyApproved"
+	scalingResponseTypeFailed            = "failed"
+)
+
+const (
+	upscaleDirectionLabel   = "inc"
```
Omrigan, line 48: Can we somehow reuse the existing constants directionValueInc and directionValueDec?

pkg/agent/execbridge.go
mikhail-sakhnov: moved away from the callback

pkg/agent/execbridge.go
mikhail-sakhnov: done

pkg/agent/execbridge.go
mikhail-sakhnov: done

pkg/agent/runner.go
```diff
@@ -834,8 +834,11 @@ func (r *Runner) DoSchedulerRequest(
 		// Fatal because invalid JSON might also be semantically invalid
 		return nil, fmt.Errorf("Bad JSON response: %w", err)
 	}
-
-	logger.Debug("Received response from scheduler", zap.Any("response", respData))
+	if respData.Permit.HasFieldLessThan(resources) {
+		logger.Warn("Received response from scheduler", zap.Any("response", respData), zap.Any("requested", resources))
+	} else {
+		logger.Debug("Received response from scheduler", zap.Any("response", respData))
+	}
```
Omrigan, line 841:
```suggestion
       level := zap.DebugLevel
	if respData.Permit.HasFieldLessThan(resources) {
		level = zap.WarnLevel	
	}
	logger.Log(level, "Received response from scheduler", zap.Any("response", respData), zap.Any("requested", resources))
```
WDYT about this?

Omrigan (approved):
LGTM, pending the suggestion and changing the PR name to exclude "issue940".

mikhail-sakhnov:
Closed in favor of #1033 to rename the branch

Omrigan:
> Closed in favor of #1033 to rename the branch

Actually, branch rename is not needed. Subject I just changed.

sharnoff (approved):
LGTM, nice!

Issue #1030, opened 2024-08-09
Memory-heavy workloads may be scaled too high


sharnoff:
## Problem description / Motivation

Currently, the vm-monitor:

1. Reserves ~75% of memory for LFC
2. Asks for scale-up when postgres exceeds the remainder (_without looking at what is actually being used by the LFC_)

This works ok as a naive solution for most OLTP workloads, but happens to mean that certain memory-heavy workloads can be scaled higher than they need (note: _excluding cache usage by LFC_, so allocations are elsewhere — like in pgvector index build).

Meanwhile the autoscaler-agent triggers upscaling based on memory usage if postgres' memory usage exceeds 75% of memory... so it's basically always handled by the vm-monitor first in practice.

This came up in this thread: https://neondb.slack.com/archives/C03TN5G758R/p1723127762991289

## Feature idea(s) / DoD

We should be more careful about how we treat memory usage as a scaling signal, so that memory-heavy workloads are no longer scaled up beyond what's necessary, while also making sure that we don't harm performance for workloads that are memory-heavy _and also_ rely on LFC being in the OS page cache.

## Implementation ideas

See https://www.notion.so/neondatabase/0f75b15d47ad479094861302a99114af

```[tasklist]
- [ ] #1031
- [ ] neondatabase/neon#8668
```

sharnoff:
Now that neondatabase/neon#8668 has been merged, this will be fixed with the next compute release containing it.

sharnoff:
This has since been released, and should be fixed for new computes.

PR #1031, opened 2024-08-09
agent/core: Use cached LFC for memory scaling signal

sharnoff:
In short: In addition to scaling when there's a lot of memory used by postgres, we should also scale up to make sure that enough of the LFC is able to fit into the page cache alongside it.

To answer "how much is enough of the LFC", this PR takes the minimum of the estimated LFC working set size (from window size) and the cached memory (from the `Cached` field of `/proc/meminfo`, via vector's host metrics).

This PR also adds the `memoryTotalFractionTarget` field to the scaling config, serving a similar purpose to `memoryUsageFractionTarget`, but applying to the total of memory usage and cached data.

This PR is part of #1030 and must be deployed before the vm-monitor changes in order to make sure we don't regress performance for workloads that are both memory-heavy and rely on LFC being in the VM's page cache.

For more info, see: https://www.notion.so/neondatabase/0f75b15d47ad479094861302a99114af.

---

**Notes for review:**

1. This PR is broken into two commits -- the first is a refactor to make the second one easier to read.
1. Planning to test on staging with neondatabase/neon#8668 using some familiar workloads (in particular: LFC-aware scaling tests and pgvector index build)

### DIFF
--- a/pkg/agent/core/goalcu.go
+++ b/pkg/agent/core/goalcu.go
@@ -5,11 +5,11 @@ package core
 import (
 	"math"
 
+	"github.com/samber/lo"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
 	"github.com/neondatabase/autoscaling/pkg/api"
-	"github.com/neondatabase/autoscaling/pkg/util"
 )
 
 type scalingGoal struct {
@@ -29,25 +29,30 @@ func calculateGoalCU(
 		warn("Making scaling decision without all required metrics available")
 	}
 
-	var goalCU uint32
+	var lfcGoalCU, cpuGoalCU, memGoalCU, memTotalGoalCU uint32
 	var logFields []zap.Field
 
-	if systemMetrics != nil {
-		cpuGoalCU := calculateCPUGoalCU(cfg, computeUnit, *systemMetrics)
-		goalCU = max(goalCU, cpuGoalCU)
-
-		memGoalCU := calculateMemGoalCU(cfg, computeUnit, *systemMetrics)
-
-		goalCU = max(goalCU, memGoalCU)
-	}
+	var wss *api.Bytes // estimated working set size
 
 	if lfcMetrics != nil {
-		lfcGoalCU, lfcLogFunc := calculateLFCGoalCU(warn, cfg, computeUnit, *lfcMetrics)
-		goalCU = max(goalCU, lfcGoalCU)
+		var lfcLogFunc func(zapcore.ObjectEncoder) error
+		lfcGoalCU, wss, lfcLogFunc = calculateLFCGoalCU(warn, cfg, computeUnit, *lfcMetrics)
 		if lfcLogFunc != nil {
 			logFields = append(logFields, zap.Object("lfc", zapcore.ObjectMarshalerFunc(lfcLogFunc)))
 		}
 	}
+
+	if systemMetrics != nil {
+		cpuGoalCU = calculateCPUGoalCU(cfg, computeUnit, *systemMetrics)
+
+		memGoalCU = calculateMemGoalCU(cfg, computeUnit, *systemMetrics)
+	}
+
+	if systemMetrics != nil && wss != nil {
+		memTotalGoalCU = calculateMemTotalGoalCU(cfg, computeUnit, *systemMetrics, *wss)
+	}
+
+	goalCU := max(cpuGoalCU, memGoalCU, memTotalGoalCU, lfcGoalCU)
 
 	return scalingGoal{hasAllMetrics: hasAllMetrics, goalCU: goalCU}, logFields
 }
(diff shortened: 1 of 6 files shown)


pkg/agent/core/goalcu.go
```diff
@@ -120,6 +136,6 @@ func calculateLFCGoalCU(
 			return nil
 		}
 
-		return lfcGoalCU, lfcLogFields
+		return lfcGoalCU, lo.ToPtr(api.Bytes(estimateWssMem)), lfcLogFields
```
Omrigan, line 139: Perhaps split this function into two: one calculates estimateWssMem, the other turns it into lfcGoalCU?

pkg/api/vminfo.go
```diff
@@ -328,14 +328,27 @@ type ScalingConfig struct {
 	// this field is left out the settings will fall back on the global default.
 	LoadAverageFractionTarget *float64 `json:"loadAverageFractionTarget,omitempty"`
 
-	// MemoryUsageFractionTarget sets the desired fraction of current memory that
-	// we would like to be using. For example, with a value of 0.7, on a 4GB VM
-	// we'd like to be using 2.8GB of memory.
+	// MemoryUsageFractionTarget sets the maximum fraction of current memory that we'd like to allow
+	// to be allocated (exactly MemTotal - MemAvailable, or roughly the memory that postgres has
+	// allocated [notably excluding LFC]).
```
Omrigan, line 333:
```suggestion
	// MemoryUsageFractionTarget sets the maximum fraction of total memory that postgres allocations (MemoryUsage) must fit into. 
	// This doesn't count the LFC memory.
```

I think the current explanation is a bit overloaded.

pkg/api/vminfo.go
```diff
@@ -328,14 +328,27 @@ type ScalingConfig struct {
...
 	// this field is left out the settings will fall back on the global default.
 	MemoryUsageFractionTarget *float64 `json:"memoryUsageFractionTarget,omitempty"`
 
+	// MemoryTotalFractionTarget sets the maximum fraction of current memory that we'd like to allow
+	// to be consumed by allocated memory (MemTotal minus MemAvailable) plus short-term cached
+	// memory by the LFC.
+	//
+	// This is kind of parallel to MemoryUsageFractionTarget, where the difference is that the
+	// "usage" target mostly tries to makes sure there's plenty of memory for postgres, whereas the
+	// "total" target also takes LFC cached memory into account.
```
Omrigan, line 349:
```suggestion
	// MemoryTotalFractionTarget sets the maximum fraction of total memory that postgres allocations PLUS LFC memory (MemoryUsage + MemoryCached) must fit into. 
	//
	// Compared with MemoryUsageFractionTarget, this value can be set higher (e.g. 0.9 vs 0.75), because we can tolerate higher fraction of consumption for both in-VM memory consumers.
```

Omrigan (approved):
Mostly non-functional nits.

pkg/agent/core/goalcu.go
sharnoff: I started writing this, but the handling with log fields gets tricky when it's split across multiple functions... WDYT?

pkg/agent/core/goalcu.go
```diff
@@ -0,0 +1,141 @@
...
+	usageGoalBytes := api.Bytes(math.Round(systemMetrics.MemoryUsageBytes / *cfg.MemoryUsageFractionTarget))
+
+	// goal memory size, looking at allocated memory and min(page cache usage, LFC working set size)
+	var totalGoalBytes api.Bytes
+	if wss != nil {
+		lfcCached := min(float64(*wss), systemMetrics.MemoryCachedBytes)
+		totalGoalBytes = api.Bytes((lfcCached + systemMetrics.MemoryUsageBytes) / *cfg.MemoryTotalFractionTarget)
+	}
+
+	memGoalBytes := max(usageGoalBytes, totalGoalBytes)
```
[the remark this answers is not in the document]
sharnoff: Did this in fdb7f8be0ada89bea0868840d43c61ae459fdf65 -- lmk what you think.

pkg/agent/core/goalcu.go
Omrigan: No problem, let's leave as-is.

pkg/agent/core/goalcu.go
```diff
@@ -0,0 +1,141 @@
...
+	warn func(string),
+	cfg api.ScalingConfig,
+	computeUnit api.Resources,
+	systemMetrics *SystemMetrics,
+	lfcMetrics *LFCMetrics,
+) (scalingGoal, []zap.Field) {
+	hasAllMetrics := systemMetrics != nil && (!*cfg.EnableLFCMetrics || lfcMetrics != nil)
+	if !hasAllMetrics {
+		warn("Making scaling decision without all required metrics available")
+	}
```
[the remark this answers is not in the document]
Omrigan: Ok, sounds good.

Omrigan (approved):
Let's merge it?

PR #1034, opened 2024-08-13
neonvm: Add controller flag to disable runner cgroup

sharnoff:
This commit adds new flags to neonvm-controller and neonvm-runner:

* controller: `-disable-runner-cgroup`
* runner: `-enable-dummy-cpu-server`

If the controller is passed `-disable-runner-cgroup`, then it will pass both `-skip-cgroup-management` and `-enable-dummy-cpu-server` to new runner pods.

If neonvm-runner is passed `-enable-dummy-cpu-server` (requires the `-skip-cgroup-management` flag), it will *not* run QEMU in a cgroup, but still provide an HTTP server with endpoints that pretend as if the CPU limit was updated successfully.

Internally, the runner's "dummy" implementation still needs to store the most recently set value to provide it back to the controller, so that it doesn't infinitely loop trying to set the CPU.

ref https://neondb.slack.com/archives/C06SJG60FRB/p1723485730034399

### DIFF
--- a/neonvm/runner/main.go
+++ b/neonvm/runner/main.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -36,6 +37,7 @@ import (
 	"github.com/jpillora/backoff"
 	"github.com/kdomanski/iso9660"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/samber/lo"
 	"github.com/vishvananda/netlink"
 	"go.uber.org/zap"
 
@@ -594,6 +596,7 @@ type Config struct {
 	kernelPath           string
 	appendKernelCmdline  string
 	skipCgroupManagement bool
+	enableDummyCPUServer bool
 	diskCacheSettings    string
 	memoryProvider       vmv1.MemoryProvider
 	autoMovableRatio     string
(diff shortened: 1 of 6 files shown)


neonvm/runner/main.go
```diff
@@ -620,7 +624,10 @@ func newConfig(logger *zap.Logger) *Config {
 		cfg.appendKernelCmdline, "Additional kernel command line arguments")
 	flag.BoolVar(&cfg.skipCgroupManagement, "skip-cgroup-management",
 		cfg.skipCgroupManagement,
-		"Don't try to manage CPU (use if running alongside container-mgr)")
+		"Don't try to manage CPU (use if running alongside container-mgr, or if dummy CPU server is enabled)")
+	flag.BoolVar(&cfg.enableDummyCPUServer, "enable-dummy-cpu-server",
```
Omrigan, line 628: Do we actually plan to use runner without the enable-dummy-cpu-server?

I don't mind that it is there, but I think I'd run the server unconditionally, until we (likely) decide to abandon it overall.

Omrigan (approved):

neonvm/runner/main.go
sharnoff: > Do we actually plan to use runner without the enable-dummy-cpu-server?

For now, yes. We haven't come to a decision beyond that.

sharnoff:
Fyi @Omrigan -- merging this, will probably conflict with #1054.

PR #1037, opened 2024-08-19
plugin: add warning log message for the pods with assigned scheduler

### LINKED ISSUE #93: Should emit error/warning/etc when pod bypasses plugin
It's possible to bypass the scheduler plugin by explicitly setting the `.spec.node` field for a `Pod` — this messes up our bookkeeping, so it'd be good to have some kind of early warning signs if something's going on.

OTOH the main idea for this that comes to mind would be to modify the plugin's pod watcher so that - on Add - it checks that the pod doesn't (a) have a matching scheduler name, (b) already have a node assigned, and (c) not exist in the existing pods map.

mikhail-sakhnov:
plugin: add warning log message for the pods with assigned scheduler, nodeName or nodeSelector

closes #93

### DIFF
--- a/pkg/plugin/state.go
+++ b/pkg/plugin/state.go
@@ -572,7 +572,7 @@ func (e *AutoscaleEnforcer) handleNodeDeletion(logger *zap.Logger, nodeName stri
 // otherwise, we might (a) ignore resources from pods that weren't scheduled here, or (b) fail to
 // include pods that *were* scheduled here, but had spurious Unreserves.
 // (for more, see: https://github.com/neondatabase/autoscaling/pull/435)
-func (e *AutoscaleEnforcer) handleStarted(logger *zap.Logger, pod *corev1.Pod) {
+func (e *AutoscaleEnforcer) handleStarted(logger *zap.Logger, pod *corev1.Pod, preexisting bool) {
 	nodeName := pod.Spec.NodeName
 
 	logger = logger.With(
@@ -592,12 +592,14 @@ func (e *AutoscaleEnforcer) handleStarted(logger *zap.Logger, pod *corev1.Pod) {
 		// this may be a preexisting VM. If so, we should include it in "buffer" as long it's
 		// supposed to be handled by us (otherwise, the "buffer" will never be resolved)
 		includeBuffer: pod.Spec.SchedulerName == e.state.conf.SchedulerName,
+		preexisting:   preexisting,
 	})
 }
 
 type reserveOptions struct {
 	allowDeny     bool
 	includeBuffer bool
+	preexisting   bool
 }
 
 // reserveResources attempts to set aside resources on the node for the pod.
@@ -630,11 +632,23 @@ func (e *AutoscaleEnforcer) reserveResources(
 
 	e.state.lock.Lock()
 	defer e.state.lock.Unlock()
+	podName := util.GetNamespacedName(pod)
 
 	// If the pod already exists, nothing to do
-	if _, ok := e.state.pods[util.GetNamespacedName(pod)]; ok {
+	_, isPodInState := e.state.pods[podName]
+	if isPodInState {
 		logger.Info("Pod already exists in global state")
 		return true, &verdictSet{cpu: "", mem: ""}, nil
+	}
+
+	// If the following conditions are met, the pod has bypassed neon scheduler which might be a sign
+	// of a bug or misbehavior:
+	// - pod is assigned to autoscaler scheduler
+	// - pod not in the state
+	// - pod is not preexisting pod
+	// - pod has the node name
+	if !isPodInState && !opts.preexisting && pod.Spec.SchedulerName == e.state.conf.SchedulerName && pod.Spec.NodeName != "" {
+		logger.Warn("Pod has bypassed neon scheduler")
 	}
 
 	// Get information about the node
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -153,7 +153,7 @@ func makeAutoscaleEnforcerPlugin(
 				incEventCount()
 			}
 			pushToQueue(logger, pod.Name, func() {
-				p.handleStarted(hlogger, pod)
+				p.handleStarted(hlogger, pod, preexisting)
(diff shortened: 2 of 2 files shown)


pkg/plugin/state.go
```diff
@@ -630,13 +632,20 @@ func (e *AutoscaleEnforcer) reserveResources(
...
 	// If the pod already exists, nothing to do
-	if _, ok := e.state.pods[util.GetNamespacedName(pod)]; ok {
+	_, isPodInState := e.state.pods[podName]
+	if isPodInState {
 		logger.Info("Pod already exists in global state")
 		return true, &verdictSet{cpu: "", mem: ""}, nil
 	}
 
+	// if the pod doesn't exists, but has nodename attached log it
+	if !isPodInState && !opts.preexisting {
```
sharnoff, line 645: If `isPodInState`, we return, so this `!isPodInState` is redundant IIUC?

pkg/plugin/state.go
```diff
@@ -630,13 +632,20 @@ func (e *AutoscaleEnforcer) reserveResources(
...
+		logger.Warn("Pod is missing in the state and is not preexisting pod", zap.Object("pod", podName), zap.String("nodeName", pod.Spec.NodeName))
```
sharnoff, line 646: IIRC pod and node name are already included in the logger that's passed in

pkg/plugin/state.go
sharnoff, line 645: IIRC `reserveResources()` is also called for pods that aren't assigned to this scheduler -- in those cases, it's expected that the pod isn't already in the state.

sharnoff (commented):
Some thoughts -- partial review. Mostly small adjustments

pkg/plugin/state.go
mikhail-sakhnov: yeah, got it here mostly for explicit, but totally makes sense to drop it from the logical condition

mikhail-sakhnov:
@sharnoff may I ask you to have a look one more time after recent changes?

pkg/plugin/state.go
mikhail-sakhnov: Yes, I double-checked and the pod and node are included.

pkg/plugin/state.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/neondatabase/autoscaling/pkg/util/xact"
 )
 
+const autoscalerSchedulerName = "autoscale-scheduler"
```
sharnoff, line 27: We should already have this defined as a field in the config. Let's use that instead?

(see also `checkSchedulerName` in `plugin.go`)

pkg/plugin/state.go
```diff
@@ -630,13 +634,24 @@ func (e *AutoscaleEnforcer) reserveResources(
...
+	if isPodInState {
 		logger.Info("Pod already exists in global state")
 		return true, &verdictSet{cpu: "", mem: ""}, nil
 	}
 
+	// log the pods if:
+	// - pod is assigned to autoscaler scheduler
+	// - pod not in the state
+	// - pod is not preexisting pod
+	// - pod has the node name
```
sharnoff, line 650: Could you add a comment as to why this is an issue?

pkg/plugin/state.go
```diff
@@ -630,13 +634,24 @@ func (e *AutoscaleEnforcer) reserveResources(
...
+	if !isPodInState && !opts.preexisting && pod.Spec.SchedulerName == autoscalerSchedulerName && pod.Spec.NodeName != "" {
+		logger.Warn("Pod is not in the state and scheduling metainformation is inconsistent")
```
sharnoff, line 652: > scheduling metainformation is inconsistent

It's less about inconsistency and more that: If the pod doesn't exist in the state, _but_ we were supposed to schedule it, then it means we were bypassed for the scheduling of this pod.

pkg/plugin/state.go
```diff
@@ -630,13 +634,24 @@ func (e *AutoscaleEnforcer) reserveResources(
 
 	e.state.lock.Lock()
 	defer e.state.lock.Unlock()
+	podName := util.GetNamespacedName(pod)
 
 	// If the pod already exists, nothing to do
-	if _, ok := e.state.pods[util.GetNamespacedName(pod)]; ok {
+	_, isPodInState := e.state.pods[podName]
+	if isPodInState {
```
sharnoff, line 641: These diffs are now unnecessary, right?

sharnoff (commented):
brief comments, just about good otherwise

pkg/plugin/state.go
mikhail-sakhnov: They actually are, I use the `isPodInState` in the if branch right after this one original

pkg/plugin/state.go
```diff
@@ -630,13 +632,25 @@ func (e *AutoscaleEnforcer) reserveResources(
...
 		logger.Info("Pod already exists in global state")
 		return true, &verdictSet{cpu: "", mem: ""}, nil
 	}
 
+	// If the following conditions are met, the pod has bypassed neon scheduler which might be a sign
+	// of a bug or misbehavior:
+	// - pod is assigned to autoscaler scheduler
```
sharnoff, line 646:
```suggestion
	// - pod is assigned to autoscale-scheduler
```

pkg/plugin/state.go
```diff
@@ -630,13 +632,25 @@ func (e *AutoscaleEnforcer) reserveResources(
...
+	// - pod not in the state
+	// - pod is not preexisting pod
+	// - pod has the node name
+	if !isPodInState && !opts.preexisting && pod.Spec.SchedulerName == e.state.conf.SchedulerName && pod.Spec.NodeName != "" {
+		logger.Warn("Pod has bypassed neon scheduler")
```
sharnoff, line 651:
```suggestion
		logger.Warn("Pod has bypassed the scheduler plugin")
```

sharnoff (approved):
A couple minor touch-ups for consistency, otherwise LGTM!

sharnoff (approved):

PR #1042, opened 2024-08-21
Fix warnings in dockerfiles

bayandin:
Relatively recently, docker has introduced a "linter" for dockerfiles: https://docs.docker.com/reference/build-checks/
I went through all the files (using `for f in $(find . -name Dockerfile); do docker build --check -f $f . ; done`) and fixed found warnings. 
A newer version of ` docker/build-push-action` will report such warnings as build annotations (see https://github.com/neondatabase/autoscaling/pull/1041/files, scroll down to "Unchanged files with check annotations")

### DIFF
--- a/neonvm/Dockerfile
+++ b/neonvm/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.21 as builder
+FROM golang:1.21 AS builder
 ARG TARGETOS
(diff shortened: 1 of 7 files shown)


sharnoff (approved):
Awesome, thanks! buildx had been warning me about it locally for a while, hadn't gotten around to fixing it though :)

PR #1043, opened 2024-08-21
ci: add coverage report

mikhail-sakhnov:


.github/workflows/test.yaml
```diff
@@ -1,13 +1,18 @@
 name: test
 on:
   pull_request:
+    types: [opened, reopened, synchronize]
```
bayandin, line 4: Those are default types, we don't need to list them:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

.github/workflows/test.yaml
```diff
@@ -1,13 +1,18 @@
...
     branches:
       - main
 jobs:
   tests:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    permissions:
+      contents: read
+      actions: read
+      pull-requests: write
```
bayandin, line 15: I would add a comment that those permissions are required for `fgrosse/go-coverage-report` actions, like this https://github.com/neondatabase/autoscaling/blob/5c838f1509330d3bdb074178928950d4d56ec521/.github/workflows/approved-for-ci-run.yml#L101-L103

mikhail-sakhnov:
@Omrigan as we discussed, I added the comment with HTML report link

@bayandin I resolved your notices about adding comment and dropping default value for `pull_requests:` node.

Omrigan (approved):

bayandin:
>  I added the comment with HTML report link

Are you talking about this comment https://github.com/neondatabase/autoscaling/pull/1043#issuecomment-2317465793?
Should we upload this report to S3 to make it easier to view? And probably having a single comment would be a bit more tidy

.github/workflows/test.yaml
```diff
@@ -25,11 +30,66 @@ jobs:
...
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: newBody,
+              });
+            }
+
+      # create comment with coverage changes digest on PR and if tests ran
+      - name: "Code coverage report"
+        uses: fgrosse/go-coverage-report@v1.0.2
```
bayandin, line 91: Optional, there's newer version of the action:

```suggestion
        uses: fgrosse/go-coverage-report@v1.2.0
```

mikhail-sakhnov:
@Omrigan @bayandin I merged the comments into a single comment. Speaking about the S3, I created infra issue to set up the credentials for this repo, but it will be triaged next Monday and not clear how much time will it take for the actual work on the issue. What if we merge with the artifacts based report and once all the aws authorization bits and pieces are in place, change the workflow to use s3 again? Another option is to just leave PR here for a while in the open state

.github/workflows/test.yaml
```diff
@@ -25,11 +30,74 @@ jobs:
...
+              repo: context.repo.repo,
+            });
+
+            // Find the comment that contains the identifier
+            const existingComment = comments.find(comment => comment.body.includes(commentIdentifier));
+            const htmlCoverageTitleIdx = existingComment.body.indexOf(htmlCoverageTitle)
+            var originalCommentBody = existingComment.body 
+            
+            if (htmlCoverageTitleIdx != -1) {
+              originalCommentBody = existingComment.body.slice(0, htmlCoverageTitleIdx)
```
bayandin, line 83: A couple of trailing spaces:

```suggestion
            var originalCommentBody = existingComment.body

            if (htmlCoverageTitleIdx != -1) {
              originalCommentBody = existingComment.body.slice(0, htmlCoverageTitleIdx)
```

bayandin (approved):
> What if we merge with the artifacts based report and once all the aws authorization bits and pieces are in place, change the workflow to use s3 again?

Works for me!

LGTM
I've added some style-related comments to make the code look more consistent, most of them are optional, feel free to ignore.
Well done!

mikhail-sakhnov:
@Omrigan @bayandin folks, it stores the report in S3 now

PR #1044, opened 2024-08-21
ci: save cover.out profile as run artifact

mikhail-sakhnov:
ci: save cover.out profile-file as run artifact to enable using it as input for the reporter

Related to https://github.com/neondatabase/cloud/issues/16669

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,7 +6,6 @@ on:
       - "**.go"
     branches:
       - main
-
 jobs:
   tests:
     runs-on: ubuntu-latest
@@ -29,3 +28,8 @@ jobs:
         if: env.GIT_DIFF
         run: |
           make test
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v4
+        with:
+          name: code-coverage
+          path: cover.out


.github/workflows/test.yaml
```diff
@@ -6,7 +6,6 @@ on:
       - "**.go"
     branches:
       - main
-
```
sharnoff, line 9: intentional diff?

sharnoff (approved):
lgtm, nice!

.github/workflows/test.yaml
mikhail-sakhnov: lint formatter. Should I revert it?

sharnoff: i.e. it was made automatically? odd... IMO breaking up the major sections with a newline helps readability, but this is a small file -- I think it's ok to keep.

For the future I'd recommend either (a) disabling it, or (b) opening a PR to apply its recommendations (so we don't end up with more stray changes like this :smile:)

bayandin (commented):
Ah, late for the party.

What are the following actions about it?

For neon repo we upload coverage results to S3 (because it's faster and more reliable than GitHub cache for larger files). Also it allows to get coverage results for the base branch to create differential coverage report using lcov:
https://github.com/neondatabase/neon/blob/428b105dde089d402b1de035a8cb43ebea930583/.github/workflows/build_and_test.yml#L396-L443

mikhail-sakhnov:
@bayandin the following action is to set up github action to render diff between main branch coverage and a PR coverage report. Draft of the PR to check out the rendering step is here but it currently red because there are no corresponding artifact uploaded for the main branch
https://github.com/neondatabase/autoscaling/pull/1043

PR #1045, opened 2024-08-21
ci: temporary trigger test run on all main merges

mikhail-sakhnov:
I need this PR to be merged to trigger test run on main branch to get the artifacts uploaded. After the https://github.com/neondatabase/cloud/issues/16669 is closed, those artifacts will be used as baseline for coverage reports.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,8 +2,6 @@ name: test
 on:
   pull_request:
   push:
-    paths:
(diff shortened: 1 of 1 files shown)


Omrigan (approved):

sharnoff (approved):
Brief additional context from discussing in DMs: IMO it's ok to just permanently switch to always running test on main, and seems like it may be easier to work with for the coverage.

PR #1046, opened 2024-08-21
ci: trigger test run on all main branch merges

mikhail-sakhnov:
`technote-space/get-diff-action` prevents from running tests if no changes in the code files were pushed. We need to bypass that behavior on main branch to store artifacts with baseline coverage.


mikhail-sakhnov:
Follow up on the pr #1045

sharnoff (approved):

PR #1051, opened 2024-08-26
vm-builder: change sshd config to use internal-sftp subsystem

### LINKED ISSUE #930: Bug: scp into guest-vm doesn't work
## Steps to reproduce
```
> scp -v test guest-vm:/root/test
...
debug1: Sending subsystem: sftp
scp: Connection closed
...
```

## Expected result

Scp works out of the box

## Other logs, links

- ...

mikhail-sakhnov:
Change sshd_config to let scp command work outside of the box.
Add e2e test for scp command.

The root cause for non-working scp was config directive ```Subsystem	sftp	/usr/lib/ssh/sftp-server```

We don't have that binary in the guest vm. There are 2 ways to fix it - either add binary during the image build or change the configuration to use embedded realization of the sftp subsystem. I chose the second option since it is a bit easier.

More context: https://man.openbsd.org/sshd_config#Subsystem

Closes #930

### DIFF
--- a/neonvm/tools/vm-builder/files/sshd_config
+++ b/neonvm/tools/vm-builder/files/sshd_config
@@ -84,8 +84,10 @@ X11Forwarding no
 # no default banner path
 #Banner none
 
-# override default of no subsystems
-Subsystem	sftp	/usr/lib/ssh/sftp-server
+# Override default of no subsystems
+# Use internal-sftp for SFTP connections. This is the modern method and doesn't require additional binaries,
+# as the logic is implemented directly within the sshd binary.
+Subsystem	sftp	internal-sftp
 
 # Example of overriding settings on a per-user basis
 #Match User anoncvs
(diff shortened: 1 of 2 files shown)


neonvm/tools/vm-builder/files/sshd_config
```diff
@@ -85,7 +85,7 @@ X11Forwarding no
 #Banner none
 
 # override default of no subsystems
-Subsystem	sftp	/usr/lib/ssh/sftp-server
+Subsystem	sftp	internal-sftp
```
Omrigan, line 88: Could you please clarify in commit message what is the root cause of the issue, and how does this change fixes it?

mikhail-sakhnov: I added comment which explains what is internal-sftp and why it is in use. Also, we don't have `/usr/lib/ssh/sftp-server` binary in the guest VM, but I ain't sure if that should be mentioned in the pr

sharnoff: My 2 cents: good to mention in the PR ("it didn't work before because ..."), but comment is great as you've written.

mikhail-sakhnov: I updated the PR description to have more context in it

sharnoff (approved):

PR #1048, opened 2024-08-23
Update `controller-tools` from 0.13.0 to 0.14.0

bayandin:
This PR updates `controller-tools` to 0.14.0.

`controller-tools` 0.13 segfauls with `go` ≥ 1.22, which blocks upgrading `go` (see https://github.com/kubernetes-sigs/controller-tools/issues/880).

Extracted from https://github.com/neondatabase/autoscaling/pull/1035

PR #1053, opened 2024-08-29
neonvm: Replace deprecated generate-groups.sh with kube_codegen.sh

sharnoff:
The new script is a completely different format :(
See e.g. kubernetes/kubernetes#117262 for examples.

Changes required here:

1. Passing through the git directory, even when using worktrees, because kube_codegen.sh directly uses git for things (git grep, git ls-files)
2. Patching kube_codegen.sh because it expects that our types files are exactly `types.go`, but we have `virtualmachine_types.go` and `virtualmachinemigration_types.go`.

There are further breaking changes in later kubernetes versions, but we don't need to deal with those quite yet.

Extracted from #1048.

---

**Notes for review:** Please check it runs locally.
**Do not approve this PR unless you have checked that `make generate` works for you.**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,6 @@ generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 	set -x ; \
 	docker run --rm \
 		"$${volumes[@]}" \
-		--volume "$$volumes" \
 		--workdir /go/src/github.com/neondatabase/autoscaling \
 		--user $(shell id -u $(USER)):$(shell id -g $(USER)) \
 		$$(cat $$iidfile) \
--- a/Makefile
+++ b/Makefile
@@ -70,22 +70,31 @@ help: ## Display this help.
 .PHONY: generate
 generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 	# Use uid and gid of current user to avoid mismatched permissions
-	iidfile=$$(mktemp /tmp/iid-XXXXXX) && \
+	set -e ; \
+	iidfile=$$(mktemp /tmp/iid-XXXXXX) ; \
 	docker build \
 		--build-arg USER_ID=$(shell id -u $(USER)) \
 		--build-arg GROUP_ID=$(shell id -g $(USER)) \
 		--build-arg CONTROLLER_TOOLS_VERSION=$(CONTROLLER_TOOLS_VERSION) \
 		--build-arg CODE_GENERATOR_VERSION=$(CODE_GENERATOR_VERSION) \
 		--file neonvm/hack/Dockerfile.generate \
-		--iidfile $$iidfile . && \
+		--iidfile $$iidfile . ; \
+	volumes=('--volume' "$$PWD:/go/src/github.com/neondatabase/autoscaling") ; \
+	if [ -f .git ]; then \
+		gitdir="$$(git rev-parse --git-common-dir)" ; \
+		volumes+=('--volume') ; \
+		volumes+=("$$gitdir:$$gitdir") ; \
+	fi ; \
+	set -x ; \
 	docker run --rm \
-		--volume $$PWD:/go/src/github.com/neondatabase/autoscaling \
+		"$${volumes[@]}" \
+		--volume "$$volumes" \
 		--workdir /go/src/github.com/neondatabase/autoscaling \
 		--user $(shell id -u $(USER)):$(shell id -g $(USER)) \
 		$$(cat $$iidfile) \
-		./neonvm/hack/generate.sh && \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -70,22 +70,31 @@ help: ## Display this help.
...
+	if [ -f .git ]; then \
+		gitdir="$$(git rev-parse --git-common-dir)" ; \
+		volumes+=('--volume') ; \
+		volumes+=("$$gitdir:$$gitdir") ; \
+	fi ; \
+	set -x ; \
 	docker run --rm \
-		--volume $$PWD:/go/src/github.com/neondatabase/autoscaling \
+		"$${volumes[@]}" \
+		--volume "$$volumes" \
```
bayandin, line 91: Not sure if we need this line as we already have the line above.
It adds an empty `--volume` argument to the docker command:

```
docker run --rm --volume /home/runner/work/autoscaling/autoscaling:/go/src/github.com/neondatabase/autoscaling --volume --volume --workdir /go/src/github.com/neondatabase/autoscaling --user 1001:127 sha256:8f356c0be7cebf483fd219ae1e189b8ea5579ad666c613bee24ca19f9f10e7ef ./neonvm/hack/generate.sh
```

Ref https://github.com/neondatabase/autoscaling/actions/runs/10606700736/job/29397913790?pr=1053#step:3:203


```suggestion
```

Makefile
```diff
@@ -70,22 +70,31 @@ help: ## Display this help.
...
 		--build-arg CODE_GENERATOR_VERSION=$(CODE_GENERATOR_VERSION) \
 		--file neonvm/hack/Dockerfile.generate \
-		--iidfile $$iidfile . && \
+		--iidfile $$iidfile . ; \
+	volumes=('--volume' "$$PWD:/go/src/github.com/neondatabase/autoscaling") ; \
+	if [ -f .git ]; then \
+		gitdir="$$(git rev-parse --git-common-dir)" ; \
+		volumes+=('--volume') ; \
+		volumes+=("$$gitdir:$$gitdir") ; \
+	fi ; \
```
bayandin, line 87: I probably lack some git knowledge, so I don't really understand this part. 
- In which case is `.git` a file and not a directory? Something related to worktrees? 
- We mount `$PWD` as `/go/src/github.com/neondatabase/autoscaling` (and set workdir it also), won't it mount also `.git`?
- `git rev-parse --git-common-dir` may return a relative path (for a regular git repo, and an absolute one for a submodule), but a relative path can't be used for volume (I guess I might understand it once I get answers for  previous questions)

bayandin (commented):
Thanks for picking this up! 
I didn't mean to add more work by switching `generate-groups.sh` to `kube_codegen.sh` 🙇 

`make generate` works for my local setup on macOS!

I've noticed that on CI, it complains:
```
fatal: No names found, cannot describe anything.
```
(seems to be an error from `git describe`)
 
Ref https://github.com/neondatabase/autoscaling/actions/runs/10606700736/job/29397913790?pr=1053#step:3:5

But we do have the same message on the main branch ([link](https://github.com/neondatabase/autoscaling/actions/runs/10611128250/job/29409992591#step:3:5)), so it's not introduced in the PR

Makefile
sharnoff: Yeah...

> In which case is `.git` a file and not a directory? Something related to worktrees?

Yeah, if you're in a git worktree, `.git` is a file with contents like
```
gitdir: /path/to/common/dir/worktrees/name
```

> We mount `$PWD` as `/go/src/github.com/neondatabase/autoscaling` (and set workdir it also), won't it mount also `.git`?

So yeah, if you're using worktrees, then the git _directory_ (stored elsewhere) won't be included.

For example, my `main` worktree is located at `/home/em/neon/autoscaling/main`, with the `.git` _file_ at `/home/em/neon/autoscaling/main/.git`, but the git common dir (i.e. where the objects are stored) at `/home/em/neon/autoscaling/.git` (note: one directory up).

> `git rev-parse --git-common-dir` may return a relative path

ah crap, really? :disappointed:

bayandin: Ah, got it, thanks for the explanation!

> > `git rev-parse --git-common-dir` may return a relative path
> 
> ah crap, really? 😞

I've tried it only for a regular git repo (where `.git` is a directory): 
```
bayandin@mbp autoscaling % git rev-parse --git-common-dir
.git
```
Not sure about other cases

sharnoff: Ok, added some handling for relative paths in 4122115dad2f9434af1e2037f17d0ffb0ab99101

bayandin (approved):
Great work!

Omrigan (approved):
Works on my machine™

PR #1055, opened 2024-08-29
neonvm-controller: improve node failure reaction speed

Omrigan:
Multiple commits, their descriptions:

 - neonvm: recreate the pod for VM even if the older pod still exists

    We have previously assumed running multiple pods for one VM is
    dangerous.

    We now believe it should be fine, even if the old pod still tries to
    continue operating. What would happen is that the new pod would take
    over the safekeepers quorum, leaving the old one to disconnect and
    shutdown.

 - neonvm: check if deletion timestamp is in the past

    If we are passed the mark of the deletion timestamp, it means
    the deletion is stuck, and we should consider the pod failed anyway.

    Possible reasons for this are:
    1. Node is down.
    2. Pod is stuck pulling the image from the container registry.

 - neonvm: add explicit "node not ready" tolerations with 30s grace period

    By default they are 300s (5m), which is way too long.

Part of the https://github.com/neondatabase/cloud/issues/14114

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -939,6 +939,14 @@ const (
 // This is *similar* to the value of pod.Status.Phase, but we'd like to retain our own abstraction
 // to have more control over the semantics.
 func runnerStatus(pod *corev1.Pod) runnerStatusKind {
+	// Add 5 seconds to account for clock skew and k8s lagging behind.
+	deadline := metav1.NewTime(metav1.Now().Add(-5 * time.Second))
+
+	// If the pod is being deleted, we consider it failed. The deletion might be stalled
+	// because the node is shutting down, or the pod is stuck pulling an image.
+	if pod.DeletionTimestamp != nil && pod.DeletionTimestamp.Before(&deadline) {
+		return runnerFailed
+	}
 	switch pod.Status.Phase {
 	case "", corev1.PodPending:
 		return runnerPending
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -776,13 +776,12 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 			return err
 		}
 
-		// We must keep the VM status the same until we know the neonvm-runner container has been
-		// terminated, otherwise we could end up starting a new runner pod while the VM in the old
-		// one is still running.
+		// By default, we cleanup the VM, even if previous pod still exists. This behavior is for the case
+		// where the pod is stuck deleting, and we want to progress without waiting for it.
 		//
-		// Note that this is required because 'VmSucceeded' and 'VmFailed' are true if *at least
-		// one* container inside the runner pod has finished; the VM itself may still be running.
-		if apierrors.IsNotFound(err) || runnerContainerStopped(vmRunner) {
+		// However, this opens up a possibility for cascading failures where the pods would be constantly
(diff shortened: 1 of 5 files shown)


mikhail-sakhnov (approved):

sharnoff (commented):
_Broadly_ looks good, I think.

The one major item I'd like to see is to have this feature-gated behind some CLI flag. Reasoning is that this type of change is at higher risk of cascading failures (e.g., if restarting causes us to trigger even more restarts) — so we should have an escape hatch, just in case.

Other than that, a couple notes:

1. Some of the changes are internal to neonvm-controller ; could those commits be titled with `neonvm-controller:` as the prefix?
2. If you plan to rebase-and-merge, could you edit the commit titles to add the PR number before merging? (i.e. appending `(#1055)` so it looks similar to squash-and-merge)

Omrigan:
>     2. If you plan to rebase-and-merge, could you edit the commit titles to add the PR number before merging? (i.e. appending `(#1055)` so it looks similar to squash-and-merge)

I was thinking maybe it we don't really need it? 

From the [history](https://github.com/neondatabase/autoscaling/commits/main/), if you click on commit, then [the commit](https://github.com/neondatabase/autoscaling/commit/0f3917ab6a504c5389f6449d3aee64bde6000088) will have a PR link. 

So editing the commits saves us 1 click.

sharnoff:
> I was thinking maybe it we don't really need it?
>
> From the [history](https://github.com/neondatabase/autoscaling/commits/main/), if you click on commit, then [the commit](https://github.com/neondatabase/autoscaling/commit/0f3917ab6a504c5389f6449d3aee64bde6000088) will have a PR link.

It's very useful to have that information available when interacting with git locally

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -935,6 +934,12 @@ const (
 // This is *similar* to the value of pod.Status.Phase, but we'd like to retain our own abstraction
 // to have more control over the semantics.
 func runnerStatus(pod *corev1.Pod) runnerStatusKind {
+	// Add 5 seconds to account for clock skew and k8s lagging behind.
+	deadline := metav1.NewTime(metav1.Now().Add(-5 * time.Second))
+
+	if pod.DeletionTimestamp != nil && pod.DeletionTimestamp.Before(lo.ToPtr(deadline)) {
+		return runnerFailed
+	}
```
sharnoff, line 942: Could you add a brief comment here explaining what we're doing?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -935,6 +934,12 @@ const (
 // This is *similar* to the value of pod.Status.Phase, but we'd like to retain our own abstraction
 // to have more control over the semantics.
 func runnerStatus(pod *corev1.Pod) runnerStatusKind {
+	// Add 5 seconds to account for clock skew and k8s lagging behind.
+	deadline := metav1.NewTime(metav1.Now().Add(-5 * time.Second))
+
+	if pod.DeletionTimestamp != nil && pod.DeletionTimestamp.Before(lo.ToPtr(deadline)) {
```
sharnoff, line 940:
```suggestion
	if pod.DeletionTimestamp != nil && pod.DeletionTimestamp.Before(&deadline) {
```

? (up to you)

sharnoff (approved):
LGTM! two comments.

pkg/neonvm/controllers/vm_controller.go
Omrigan: Added!

pkg/neonvm/controllers/vm_controller.go
Omrigan: Done!

PR #1057, opened 2024-08-30
vxlan-controller: support dualstack 

chaporgin:
In Azure we need to run Dualstack AKS at some point because that is the way to provision a dualstack NLB for AKS. There is no "bring your NLB" for AKS right now.

This skips ipv6 addresses during initialization, because autoscaler does not need ipv6, and the following lines do not work with ipv6 resulting in an error.

https://github.com/neondatabase/cloud/issues/13924


sharnoff (approved):
LGTM, but @cicdteam knows much more about this than I do.

Before merging, could you rename the PR title to start with "vxlan-controller: " to match other PRs?

cicdteam (approved):

PR #1061, opened 2024-09-04
agent: Remove unused "error" exit status

sharnoff:
The tasks associated with a VM *originally* could exit with an error on startup, and we'd retry after a duration. This hasn't been possible for some time, so we're better off removing the machinery for it.

Note: This PR removes the following metric:

- `autoscaling_agent_runner_fatal_errors_total`

---

**Notes for review:** Given the metric is currently always zero, we should preemptively remove it from dashboards & alerts before merging.

Also worth noting this came up as part of neondatabase/docs#96.


Omrigan (approved):

sharnoff:
AFAICT, removed the metric from relevant places. Merging...

PR #1062, opened 2024-09-05
Support disk quotas

petuhovskiy:
This PR adds initial support for disk quotas into NeonVM:
- Add `enableQuotas` field to `EmptyDiskSource` struct. Setting it enables project quotas support on the ext4 filesystem.
- Add `/neonvm/bin/set-disk-quota` script to modify the quotas from inside the VM
```
# Usage can be like this:
/neonvm/bin/set-disk-quota <NEW_SIZE_IN_KB> <MOUNTPOINT>
```

It also enables `CONFIG_QFMT_V2=y` kernel option, because it's required for quotas support.

ref https://github.com/neondatabase/autoscaling/issues/1071


neonvm/tools/vm-builder/files/Dockerfile.img
```diff
@@ -65,6 +65,12 @@ RUN set -e \
 		openssh-server \
 	&& /helper.move-bins.sh sshd ssh-keygen
 
+# quota tools
+RUN set -e \
+	&& apk add --no-cache --no-progress --quiet \
+		quota-tools \
+	&& /helper.move-bins.sh quota edquota quotacheck quotaoff quotaon quotastats setquota repquota tune2fs
```
petuhovskiy, line 72: `tune2fs` – needed to tune ext4 fs to enable quotas
`setquota` – needed to actually set project quota
`repquota` – can be used to look at current quota usage

Other binaries seem useful, but I've actually not used them and don't know if they're necessary. In theory they can be removed, but may be useful in the future.

What do you think? Should we remove irrelevant binaries from `/helper.move-bins.sh` here? Or should we leave it as is for now?

neonvm/tools/vm-builder/files/set-disk-quota.sh
```diff
@@ -0,0 +1,21 @@
...
+    echo "expected 2 args, got $#" >&2
+    echo "$USAGE" >&2
+    exit 1
+fi
+
+echo "Setting disk quota for $mountpoint to $size KB"
+/neonvm/bin/setquota -P 0 0 "$size" 0 0 "$mountpoint"
+
+echo "DONE! Disk quota set"
+/neonvm/bin/repquota -P -a
```
petuhovskiy, line 20: This is not needed, it just shows current status. Probably useful for debug.

I'm not 100% we should leave it here, what do you all think?

Omrigan: I think we should leave it, if the command doesn't take any noticeable time to run.

Omrigan (commented):
LGTM, but better reviewed by @sharnoff, who did a simiar feature - swap resizing.

neonvm/runner/main.go
```diff
@@ -279,6 +279,11 @@ func createISO9660runtime(
 					opts = "-o discard"
 				}
 
+				if disk.EmptyDisk.EnableQuotas {
+					mounts = append(mounts, fmt.Sprintf(`tune2fs -Q prjquota $(/neonvm/bin/blkid -L %s)`, disk.Name))
+					mounts = append(mounts, fmt.Sprintf(`tune2fs -E mount_opts=prjquota $(/neonvm/bin/blkid -L %s)`, disk.Name))
```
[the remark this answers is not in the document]
petuhovskiy: `tune2fs -Q prjquota` – patches fs to enable support for project quotas
`tune2fs -E mount_opts=prjquota` – add project quotas in the default mount options, it means every (re)mount will enable quotas by default

This is what I found in the internet, there may be a better way though.
ref https://superuser.com/questions/1441509/what-are-project-ids-in-linux-as-mentioned-in-the-chattr-manual
ref https://stackoverflow.com/questions/52665377/project-quotas-in-ext4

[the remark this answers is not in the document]
sharnoff: If it's required to be in the mount options, maybe we can just add it there?

neonvm/hack/kernel/linux-config-6.1.92
```diff
@@ -2554,7 +2554,7 @@ CONFIG_QUOTA=y
 CONFIG_PRINT_QUOTA_WARNING=y
 # CONFIG_QUOTA_DEBUG is not set
 # CONFIG_QFMT_V1 is not set
-# CONFIG_QFMT_V2 is not set
+CONFIG_QFMT_V2=y
```
sharnoff, line 2557: Could you add the rationale for this to the PR description?

neonvm/tools/vm-builder/files/set-disk-quota.sh
sharnoff: IMO it should either have more context (e.g., "disk quota set to X") or be removed — otherwise people will be confused by it in the compute logs :sweat_smile:

sharnoff (commented):
High-level:

1. This PR in isolation seems fair enough
2. What's the plan for how to move from this change to autoscalable disks? (not that it'll necessarily happen soon, but want to avoid painting ourselves into a corner)

Plus some other questions :)

neonvm/runner/main.go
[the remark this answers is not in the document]
petuhovskiy: Em suggested to try to move all these options into `mkfs.ext4`, if it's possible. I'll try and get back here with the results.

neonvm/tools/vm-builder/files/set-disk-quota.sh
petuhovskiy: Example of repquota output:

```
root@vm-example-vxc8h:/var/db/postgres/compute# /neonvm/bin/repquota -P -a
*** Report for project quotas on device /dev/vdd
Block grace time: 7days; Inode grace time: 7days
                        Block limits                File limits
Project         used    soft    hard  grace    used  soft  hard  grace
----------------------------------------------------------------------
#0        --   25164       0       0              3     0     0       
```

petuhovskiy:
Added e2e test for disk quotas. It may help understanding usage and outputs of some binaries.

Here is a log of e2e test run: https://gist.github.com/petuhovskiy/26d90cb55d1cb866ce8e882b487b139c

neonvm/tools/vm-builder/files/set-disk-quota.sh
sharnoff: From the e2e test run:

```
*** Report for project quotas on device /dev/vdd
Block grace time: 7days; Inode grace time: 7days
                        Block limits                File limits
Project         used    soft    hard  grace    used  soft  hard  grace
----------------------------------------------------------------------
#0        --      20       0 1048576              2     0     0       
```

at first glance, I think this is kind of hard to read, and we could probably make do with much smaller output (e.g, just showing block used/hard limit and file used/hard limit).

Any ideas if that's possible?

neonvm/hack/kernel/linux-config-6.1.92
petuhovskiy: Quotas don't work without this.

neonvm/tools/vm-builder/files/set-disk-quota.sh
```diff
@@ -0,0 +1,21 @@
...
+    size="$1"
+    mountpoint="$2"
+else
+    echo "expected 2 args, got $#" >&2
+    echo "$USAGE" >&2
+    exit 1
+fi
+
+echo "Setting disk quota for $mountpoint to $size KB"
+/neonvm/bin/setquota -P 0 0 "$size" 0 0 "$mountpoint"
```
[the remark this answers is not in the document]
petuhovskiy: Added explanation for setquota usage.

neonvm/tools/vm-builder/files/set-disk-quota.sh
petuhovskiy: I don't think it's possible to make it much better. `repquota` has options to change display format to csv/xml, but then you will need to parse it to display something nice.

This is the full help output:
```
root@vm-example-zb8g4:~# /neonvm/bin/repquota -h
repquota: Utility for reporting quotas.
Usage:
repquota [-vugsi] [-c|C] [-t|n] [-F quotaformat] [-O (default | xml | csv)] (-a | mntpoint)

-v, --verbose                 display also users/groups without any usage
-u, --user                    display information about users
-g, --group                   display information about groups
-P, --project                 display information about projects
-s, --human-readable[=units]  display numbers in human friendly units (MB, GB,
                              ...). Units can be also specified explicitely by
                              an optional argument in format [kgt],[kgt] where
                              the first character specifies space units and the
                              second character specifies inode units
-t, --truncate-names          truncate names to 9 characters
-p, --raw-grace               print grace time in seconds since epoch
-n, --no-names                do not translate uid/gid to name
-i, --no-autofs               avoid autofs mountpoints
-c, --cache                   translate big number of ids at once
-C, --no-cache                translate ids one by one
-F, --format=formatname       report information for specific format
-O, --output=format           format output as xml or csv
-a, --all                     report information for all mount points with
                              quotas
-h, --help                    display this help message and exit
-V, --version                 display version information and exit

Bugs to jack@suse.cz
```

neonvm/runner/main.go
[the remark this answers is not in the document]
petuhovskiy: Tried to move all options into `mkfs.ext4`, failed to specify default mount options in `mkfs.ext4`. I think it's possible only with tune2fs, or in fstab.

As a side effect, made a PR with some comments https://github.com/neondatabase/autoscaling/pull/1085

neonvm/tools/vm-builder/files/set-disk-quota.sh
sharnoff: Ok. Let's leave it for now, and then maybe remove it later once things are proven stable?

petuhovskiy: Yes, sounds good. I can also remove it right now as I have no strong opinion on this, do you think it'll be better?

sharnoff: I think the additional observability is valuable enough in the short term that we should include it for now.

sharnoff (approved):
LGTM - discussed briefly in DMs

PR #1066, opened 2024-09-06
CI(test): fix auto-comment if there's no changes in code coverage

bayandin:
If there are no changes in code coverage `fgrosse/go-coverage-report` action doesn't post a GitHub comment. If there's no comment, we can't modify it with a link to the HTML report on S3[0].

This PR makes the workflow post a new comment if there's no previous one.

- [0] https://github.com/neondatabase/autoscaling/actions/runs/10729316896/job/29755904676


bayandin:
An alternative solution is not to post the link to the HTML Report if there are no changes in coverage, as  `grosse/go-coverage-report` does. What do you think, @mikhail-sakhnov?

.github/workflows/test.yaml
```diff
@@ -86,12 +87,18 @@ jobs:
 
             // Find the comment that contains the identifier
             const existingComment = comments.find(comment => comment.body.includes(commentIdentifier));
-            const htmlCoverageTitleIdx = existingComment.body.indexOf(htmlCoverageTitle)
-            var originalCommentBody = existingComment.body
+            // If there's no original comment, create a new one with hidden comment identifier
+            // to make it easier to find next time
+            let originalCommentBody = `<!-- ${commentIdentifier} -->`
```
Omrigan, line 92:
```suggestion
            let originalCommentBody = `<!-- ${commentIdentifier} -->
            
            No changes to the coverage.`
```

Omrigan (approved):

PR #1035, opened 2024-08-14
go 1.22

bayandin:
Go 1.23 was released yesterday https://go.dev/blog/go1.23, so it's time to catch up with the previous 1.22 release https://go.dev/blog/go1.22

The previous attempt: https://github.com/neondatabase/autoscaling/pull/938, probably, failed because of the older k8s version, let's see if it works now

PR #1067, opened 2024-09-06
CI(lint): golangci-lint 1.60.3

bayandin:
Update `golangci-lint` to 1.60.3 and fix new warnings

A prerequisite for go 1.23[0]

- [0] https://github.com/neondatabase/autoscaling/pull/1065

### DIFF
--- a/pkg/api/versionutils.go
+++ b/pkg/api/versionutils.go
@@ -34,10 +34,10 @@ func (r VersionRange[V]) String() string {
 // If either range is invalid, or no such version exists (i.e. the ranges are disjoint), then the
 // returned values will be (0, false).
 func (r VersionRange[V]) LatestSharedVersion(cmp VersionRange[V]) (_ V, ok bool) {
-	max := util.Min(r.Max, cmp.Max)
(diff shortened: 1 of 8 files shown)


Omrigan (approved):

pkg/api/versionutils.go
```diff
@@ -34,10 +34,10 @@ func (r VersionRange[V]) String() string {
 // If either range is invalid, or no such version exists (i.e. the ranges are disjoint), then the
 // returned values will be (0, false).
 func (r VersionRange[V]) LatestSharedVersion(cmp VersionRange[V]) (_ V, ok bool) {
-	max := util.Min(r.Max, cmp.Max)
```
sharnoff, line 37: Now that #1035 has been merged, we can also get rid of `util.{Min,Max}` -- I'll open a follow up :)

sharnoff (approved):

pkg/api/versionutils.go
bayandin: Thanks!

PR #1068, opened 2024-09-07
Specify minor go version in .mod file

Omrigan:
See also https://github.com/golang/go/issues/66175

Otherwise can't build:
  go: download go1.22 for linux/amd64: toolchain not available


sharnoff (approved):
Out of curiosity, in what context does it not build?
(works ok for me locally)

bayandin (approved):

Omrigan:
> Out of curiosity, in what context does it not build? (works ok for me locally)

On the dev server it doesn't build.

Issue #1060, opened 2024-09-03
Remove support for DIMM hotplug


stradig:
This code is no longer needed since we moved to virtio-mem on all nodes and tested this over a sufficient time. So we should remove it to reduce tech debt and clean up the code.

sharnoff:
Update: Waiting on me to rebase #1070.

sharnoff:
Status update: #1070 has been updated, now awaiting re-review.

sharnoff:
Status update: #1070 has been released; #1215 is now ready for review.

sharnoff:
With #1215 released, this is now complete

PR #1070, opened 2024-09-09
neonvm-controller: Remove support for DIMM slots

sharnoff:
This is part 1 of 2 for removing DIMM slots support and completing our transition to using virtio-mem instead (ref #1060).

With this change, neonvm-controller will always assume that VMs are using virtio-mem. The only checks exist at the top of the reconcile functions, making sure that this is indeed the case.

The fields still exist in the CRD (although they're marked as deprecated), so that rollback of this change is still sound.

neonvm-runner also retains support for both memory providers, so that custom neonvm-runner images continue to work for at least one version across the change.


neonvm/controllers/vm_controller.go
```diff
@@ -374,12 +374,15 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		vm.Status.Phase = vmv1.VmPending
 	case vmv1.VmPending:
 		// Generate runner pod name and set desired memory provider.
-		// Together with Status.MemoryProvider set for PodName != "" above,
-		// It is now guaranteed to have Status.MemoryProvider != nil
 		if len(vm.Status.PodName) == 0 {
 			vm.Status.PodName = names.SimpleNameGenerator.GenerateName(fmt.Sprintf("%s-", vm.Name))
+			//nolint:staticcheck // We know it's deprecated. That's why we're checking it.
```
Omrigan, line 379: Do we really need this check, or we can just leave MemoryProvider in whatever state, as long as it is not nil?

neonvm/controllers/vm_controller.go
```diff
@@ -881,47 +855,6 @@ func (r *VMReconciler) doVirtioMemScaling(vm *vmv1.VirtualMachine) (done bool, _
 	return done, nil
 }
 
-func (r *VMReconciler) doDIMMSlotsScaling(ctx context.Context, vm *vmv1.VirtualMachine) (done bool, _ error) {
-	log := log.FromContext(ctx)
-
-	memSlotsMin := vm.Spec.Guest.MemorySlots.Min
-	targetSlotCount := int(vm.Spec.Guest.MemorySlots.Use - memSlotsMin)
-
-	realSlots, err := QmpSetMemorySlots(ctx, vm, targetSlotCount, r.Recorder)
```
Omrigan, line 890: QmpSetMemorySlots (and maybe others) is dead code now?

tests/e2e/vm-migration/00-assert.yaml
Omrigan, line 1: Why vm-migration test?

neonvm/controllers/vm_controller.go
sharnoff: Good point, yeah. Fixed in a306ee16a7642c779367f90551d89cc5afc77de2

neonvm/controllers/vm_controller.go
sharnoff: Good catch. Yeah, lot of dead code now. Removed in d0d4865067159ebdaa1464ee95a3f93ff30cb36a

tests/e2e/vm-migration/00-assert.yaml
sharnoff: Not sure what you mean. Made some changes, resolving as this is probably out of date, but feel free to reopen.

sharnoff:
todo: Need to update this PR after recent changes.

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -221,6 +221,7 @@ type Guest struct {
 	MemorySlotSize resource.Quantity `json:"memorySlotSize"`
 	// +optional
 	MemorySlots MemorySlots `json:"memorySlots"`
+	// Deprecated: MemoryProvider is ignored and always interpreted as equal to VirtioMem.
```
mikhail-sakhnov, line 224: are we going to drop the variable eventually?

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -59,11 +59,13 @@ func (r *VirtualMachine) ValidateCreate() (admission.Warnings, error) {
 			r.Spec.Guest.CPUs.Max)
 	}
 
-	// validate .spec.guest.memorySlotSize w.r.t. .spec.guest.memoryProvider
-	if r.Spec.Guest.MemoryProvider != nil {
-		if err := r.Spec.Guest.ValidateForMemoryProvider(*r.Spec.Guest.MemoryProvider); err != nil {
-			return nil, fmt.Errorf(".spec.guest: %w", err)
-		}
+	// DIMMSlots memory provider is deprecated, and assumed to never be used.
+	if r.Spec.Guest.MemoryProvider != nil && *r.Spec.Guest.MemoryProvider == MemoryProviderDIMMSlots {
```
mikhail-sakhnov, line 63: I wonder, could we drop the enum value `MemoryProviderDIMMSlots` for the `MemoryProvider` type and get by the automatic validation?

neonvm/samples/vm-example.dimmslots.yaml
```diff
@@ -39,7 +39,7 @@ spec:
       min: 1
       max: 2
       use: 2
-    memoryProvider: VirtioMem
+    memoryProvider: DIMMSlots
```
mikhail-sakhnov, line 42: Do we still need this sample file, keeping in mind that we have explicit error return `errors.New("DIMMSlots memory provider is deprecated and disabled")` for the DIMMSlots value?

mikhail-sakhnov:
In general looks reasonable with couple of minor questions

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: Yeah -- the idea is that this PR i just the first half, making making this field not have an effect (but still set it to to VirtioMem). The second half will be to remove the field.

(We need to do it in two steps so that each one is safe to roll back)

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
sharnoff: I started implementing this, but realized I wasn't sure what happens if there _does_ happen to be some VM with `MemoryProvider = DIMMSlots`... I think k8s won't do anything special, and just leave it be?

Assuming that's true, I think it's probably better to have the explicit, loud failure on the deprecated value — does that make sense? Still could go either way though, curious for your thoughts

sharnoff:
Updates:

- Rebased onto main (8a076637e315f0b52c52425c388d69e1543a5348..4e833fd56158469a4409693d9ffd4ab176175f0f) ([diff](https://github.com/neondatabase/autoscaling/compare/8a076637e315f0b52c52425c388d69e1543a5348..4e833fd56158469a4409693d9ffd4ab176175f0f))
- Removed the dimm slots sample file (8a076637e315f0b52c52425c388d69e1543a5348..a28f6022cf955aa445c71b21e3c2ae7fdbadb927) ([diff](https://github.com/neondatabase/autoscaling/compare/8a076637e315f0b52c52425c388d69e1543a5348..a28f6022cf955aa445c71b21e3c2ae7fdbadb927))

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
mikhail-sakhnov: >I think k8s won't do anything special, and just leave it be?

yeah, I think that's correct

>does that make sense?

absolutely

mikhail-sakhnov (approved):
looks legit to me, if @Omrigan or @petuhovskiy have no further questions, let's proceed!

Omrigan (approved):

PR #1073, opened 2024-09-17
vm-builder: makefile targets to introduce cross-compilation of the linux kernel

mikhail-sakhnov:
#1079 

Introduces kernel arm 64 builds (not used anywhere). 

Signed-off-by: Misha Sakhnov <misha@neon.tech>


neonvm/hack/kernel/Dockerfile.kernel-builder
```diff
@@ -46,16 +48,31 @@ RUN set -e \
...
+RUN cd linux-${KERNEL_VERSION} && make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- -j `nproc`
 
+# Copy the kernel image to a separate step
 # Use alpine so that `cp` is available when loading custom kernels for the runner pod.
 # See the neonvm controller's pod creation logic for more detail.
-FROM alpine:3.19
+FROM alpine:3.18 AS kernel_arm64
 ARG KERNEL_VERSION
-COPY --from=build /build/linux-${KERNEL_VERSION}/arch/x86/boot/bzImage /vmlinuz
+COPY --from=build_arm64 /build/linux-${KERNEL_VERSION}/arch/arm64/boot/Image /vmlinuz
```
sharnoff, line 78: should we add a dummy target at the end that'll always fail? idea being that we should make usage without a target fail

neonvm/hack/kernel/Dockerfile.kernel-builder
```diff
@@ -46,16 +48,31 @@ RUN set -e \
...
+FROM build-deps AS build_arm64
+ARG KERNEL_VERSION
+ADD linux-config-${KERNEL_VERSION} linux-${KERNEL_VERSION}/.config
+RUN cd linux-${KERNEL_VERSION} && make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- -j `nproc`
 
+# Copy the kernel image to a separate step
 # Use alpine so that `cp` is available when loading custom kernels for the runner pod.
 # See the neonvm controller's pod creation logic for more detail.
-FROM alpine:3.19
+FROM alpine:3.18 AS kernel_arm64
```
sharnoff, line 76:
```suggestion
FROM alpine:3.19 AS kernel_arm64
```

sharnoff, line 76: A question for you: This `alpine` image is executed as-is on the host where we run the VM. Do we need to do anything special to make sure that we get the arm64 versions of packages?

sharnoff (commented):
broadly looks good; some questions

neonvm/hack/kernel/Dockerfile.kernel-builder
mikhail-sakhnov: It doesn't really matter at this step which architecture we run on, the compilation is going through cross-compiling toolchain and relies on the specific compiler packages (gcc-x86-64-linux-gnu and gcc-aarch64-linux-gnu)

sharnoff: What I mean is that the `cp` from that alpine image can be used directly:

https://github.com/neondatabase/autoscaling/blob/351b7b4d8c4ff13e5c97d5ae4a83ccc05efd5c26/neonvm/controllers/vm_controller.go#L1702

It's ok if this PR doesn't address that, but I want to make sure that's not forgotten :)

mikhail-sakhnov: Got you, looking into that 👀

sharnoff (approved):
From discussion -- having the arm kernel config not defined by some shared base with the x86 one is tech debt, but at this time, it's better to merge this for now to continue rapid iteration.

sharnoff:
For the future: note that this PR did not affect vm-builder, and instead just changed how we build the kernel.

neonvm/hack/kernel/linux-config-amd64-6.1.92
```diff
@@ -3418,6 +3418,7 @@ CONFIG_STRICT_DEVMEM=y
 #
 # x86 Debugging
 #
+CONFIG_X86_VERBOSE_BOOTUP=y
```
sharnoff, line 3421: @mikhail-sakhnov should this still be enabled?

PR #1076, opened 2024-09-20
Format with `gofumpt` and check it in CI

sharnoff:
This PR:

1. Formats the codebase with `gofumpt`
2. Changes the makefile `fmt` target to use `gofumpt` instead of `go fmt`
3. Adds a formatting job in CI to check that the codebase is formatted according to `gofumpt`

For more on `gofumpt`, see: https://github.com/mvdan/gofumpt

ref neondatabase/cloud#15747

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -48,6 +48,27 @@ jobs:
             exit 1
           fi
 
+  fmt:
+    name: make fmt
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: make fmt
+        run: |
+          make fmt
+      - name: check diff
+        run: |
+          if ! test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
+            git ls-files --exclude-standard --others .
+            git diff .
+            echo "ERROR: 'make fmt' modified the source tree."
+            exit 1
+          fi
+
   generated:
     name: check generated files
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,8 @@ endif
 # https://github.com/neondatabase/autoscaling/pull/130#issuecomment-1496276620
 export GOFLAGS=-buildvcs=false
 
+GOFUMPT_VERSION ?= v0.7.0
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
@@ -102,7 +104,7 @@ generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
-	go fmt ./...
+	go run mvdan.cc/gofumpt@${GOFUMPT_VERSION} -w .
 
 .PHONY: vet
 vet: ## Run go vet against code.
--- a/neonvm/apis/neonvm/v1/webhook_suite_test.go
+++ b/neonvm/apis/neonvm/v1/webhook_suite_test.go
@@ -44,11 +44,13 @@ import (
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
-var k8sClient client.Client
-var testEnv *envtest.Environment
-var ctx context.Context
-var cancel context.CancelFunc
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
 
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
@@ -127,7 +129,6 @@ var _ = BeforeSuite(func() {
 		conn.Close()
 		return nil
 	}).Should(Succeed())
-
 })
 
 var _ = AfterSuite(func() {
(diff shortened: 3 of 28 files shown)


.github/workflows/lint.yml
```diff
@@ -48,6 +48,27 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: make fmt
+        run: |
+          make fmt
+      - name: check diff
+        run: |
+          if ! test -z "$(git ls-files --exclude-standard --others .)$(git diff .)"; then
```
Omrigan, line 65: Would be cool to eventually have this as a precommit hook.

Omrigan (approved):
LGTM, but let's merge it carefully, so we don't have conflicts.

petuhovskiy:
FYI almost all useful golang linters/formatters are already included in golangci-lint :) 
This is the full list: https://golangci-lint.run/usage/linters/

That means that you can just add `gofumpt` to the config and it will work, like this:
```
arthur@neon-hetzner-dev-arthur:~/autoscaling$ git diff
diff --git a/.golangci.yml b/.golangci.yml
index ab66df9..bfa7d1a 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -41,6 +41,7 @@ linters:
     - nolintlint       # bad "nolint" directives
     - predeclared      # no identifiers in Go's list of predeclared identifiers, see <https://go.dev/ref/spec#Predeclared_identifiers>
     - unparam          # no unused function parameters or return values
+    - gofumpt
 
 linters-settings:
   # see: <https://golangci-lint.run/usage/linters/#dupword>, <https://github.com/Abirdcfly/dupword>
```

Then, when you run golangci-lint, it shows you the warnings if the files aren't formatted:
```
arthur@neon-hetzner-dev-arthur:~/autoscaling$ golangci-lint run
neonvm/apis/neonvm/v1/webhook_suite_test.go:47: File is not `gofumpt`-ed (gofumpt)
var cfg *rest.Config
var k8sClient client.Client
var testEnv *envtest.Environment
var ctx context.Context
var cancel context.CancelFunc
neonvm/apis/neonvm/v1/webhook_suite_test.go:130: File is not `gofumpt`-ed (gofumpt)

pkg/plugin/dumpstate.go:219: File is not `gofumpt`-ed (gofumpt)
```

And you can also run `--fix` to run all the formatters and linter's autofixers:
```
arthur@neon-hetzner-dev-arthur:~/autoscaling$ golangci-lint run --fix

arthur@neon-hetzner-dev-arthur:~/autoscaling$ git status
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .golangci.yml
        modified:   neonvm/apis/neonvm/v1/webhook_suite_test.go
        modified:   neonvm/controllers/functests/suite_test.go
        modified:   neonvm/controllers/functests/vm_controller_test.go
        modified:   neonvm/controllers/metrics.go
        modified:   neonvm/controllers/vm_controller.go
        modified:   neonvm/controllers/vm_qmp_queries.go
        modified:   neonvm/controllers/vmmigration_controller.go
        modified:   neonvm/pkg/ipam-demo.go
        modified:   neonvm/pkg/ipam/allocate.go
        modified:   neonvm/pkg/ipam/ipam.go
        modified:   neonvm/runner/main.go
        modified:   neonvm/tools/vm-builder/main.go
        modified:   neonvm/tools/vxlan/controller/main.go
        modified:   pkg/agent/billing/billing.go
        modified:   pkg/agent/billing/indexedstore.go
        modified:   pkg/agent/billing/prommetrics.go
        modified:   pkg/agent/core/state_test.go
        modified:   pkg/agent/core/testhelpers/construct.go
        modified:   pkg/agent/prommetrics.go
        modified:   pkg/agent/runner.go
        modified:   pkg/api/vminfo.go
        modified:   pkg/billing/client.go
        modified:   pkg/plugin/dumpstate.go
        modified:   pkg/plugin/plugin.go
        modified:   pkg/plugin/state.go
        modified:   pkg/util/watch/watch.go

```

petuhovskiy (commented):
LGTM, but I think I'd prefer to use `golangci-lint` for everything related to static code checks. The downside of `golangci-lint` is that it can be slower in case of heavy linters enabled, but provides unification between tools.

sharnoff:
Hm, my thinking about golangci-lint vs gofumpt directly was:

1. We already use golangci-lint, like you mentioned (positive)
2. `golangci-lint --fix` has a much broader scope than just formatting (negative)
3. Generally I think of "linting" and "formatting" as separate domains, and so I'd rather keep them separate in CI and our Makefile

WDYT?

(also, I probably should have put this in the PR description -- will remember that for next time)

petuhovskiy:
I think I still prefer `golangci-lint` for everything, but using `gofumpt` directly also doesn't have any objections from my side. So I'm ok with merging this PR as is.

> golangci-lint --fix has a much broader scope than just formatting (negative)
> Generally I think of "linting" and "formatting" as separate domains, and so I'd rather keep them separate in CI and our Makefile

IMO linting and formatting are mostly the same, the only difference is that linters are not always capable of fixing the code automatically.

Omrigan:
> Hm, my thinking about golangci-lint vs gofumpt directly was:
> 
>     1. We already use golangci-lint, like you mentioned (positive)
> 
>     2. `golangci-lint --fix` has a much broader scope than just formatting (negative)
> 
>     3. Generally I think of "linting" and "formatting" as separate domains, and so I'd rather keep them separate in CI and our Makefile

Can you elaborate a bit, why not treat `golangci-lint --fix` as our formatter, even though it had broader scope? My thinking is: broader scope -> less to fix by hand.

> 
> 
> WDYT?
> 
> (also, I probably should have put this in the PR description -- will remember that for next time)

How about use `gofumpt` in Makefile for fmt target, but enable it in golangci-lint config, and ditch the separate actions step?

petuhovskiy (approved):

PR #1088, opened 2024-09-25
Fix ECR access

cicdteam:
CI: use GitHub OIDC instead of static AWS credentials

>NOTE: merge after https://github.com/neondatabase/infra/pull/1950


sharnoff (approved):
Ah awesome, thanks!

Issue #1082, opened 2024-09-24
Replace hotplug cpu by online/offline CPU


mikhail-sakhnov:
## Problem description / Motivation
According to qemu documentation, hot plugging is not supported on arm. Marking cpu as online/offline is supported. We need to change acpi scripts used on x86_64 for hot plugging to support online/offline cpus on arm.

https://docs.kernel.org/next/arch/arm64/cpu-hotplug.html

https://www.qemu.org/docs/master/system/cpu-hotplug.html

stradig:
I suggest to move that to the start and also switch to online/offline on x86. I believe that will autoscaling also more robust on x86.

mikhail-sakhnov:
# CPU Scaling Feature via online/offline cpu state

## Current Implementation
In the existing NeonVM platform, CPU hotplugging is used to dynamically scale CPU resources up or down. The NeonVM controller (neonvm/controllers/vm_controller.go) sends commands via QMP (QEMU Monitor Protocol) to the QEMU hypervisor. When CPUs are added or removed, a udev rule (neonvm/tools/vm-builder/files/vminit) inside the VM automatically marks the newly added CPUs as online.

While this method works effectively on x86_64 architectures, the ARM architecture in QEMU does not support CPU hotplugging.

## Proposed Solution for ARM Architecture
To support CPU scaling on ARM, we will implement a different approach. Instead of hotplugging, all potential CPUs will be allocated to the VM instance during startup. Any CPUs beyond the current scaling limit will be marked as offline. The control of CPU states (online/offline) will no longer rely on the udev rule but will be handled through a VM control daemon running inside the VM.

### Key Aspects:

- Initial CPU Allocation: All possible CPUs will be assigned to the VM at startup. Only the CPUs required within the scaling limit will be marked as online, while the others remain offline.
- Daemon for CPU State Management: The introduction of a VM control daemon will enable direct control over the online/offline status of the CPUs. This daemon will expose an API to manage the number of active CPUs.

### Required Changes
- NeonVM Controller Modification: Update the NeonVM controller to communicate with the VM control daemon API for CPU scaling actions instead of using QMP directly to the QEMU monitor.
- VM Runner API: For enhanced security, the daemon will be exposed through the VM Runner API to ensure secure interaction.
- VM Control Daemon Implementation: Develop the VM control daemon that will run inside the VM to manage CPU states.
- Expose VM Control Daemon: Integrate the control daemon via the VM Runner interface
[...]

PR #1090, opened 2024-09-26
neonvm: Add new neonvm-daemon binary skeleton

sharnoff:
Extracted the minimal version from #1052, because it's clear that this daemon is required in multiple places, and we don't know which PR will merge first.

---

**Notes:** Not expected to pass CI! (images won't be in the right place). Haven't tested it since extracting from #1052.

cc @mikhail-sakhnov, as discussed :)


mikhail-sakhnov:
Fixed CI. Let's merge it and I use it in the #1082

sharnoff:
Looks like my original commit accidentally reverted some of the fixes from #1088. Fixing that...

tests/e2e/vm-neonvmd/00-assert.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 90
+commands:
+  - script: |
+      set -eux
+      pod="$(kubectl get neonvm -n "$NAMESPACE" example -o jsonpath='{.status.podName}')"
+      kubectl exec -n "$NAMESPACE" $pod -- ssh guest-vm ps aux | grep neonvmd | [ $(wc -l) -eq 1 ] || echo "neonvmd process is not found in the $pod"
```
sharnoff, line 8: AFAICT this will always succeed because `grep neonvmd` shows up in `ps aux` and matches itself.

Let's use `ps aux -o comm` to fix that?

mikhail-sakhnov: Changed to the ps aux -o comm.

.github/workflows/build-images.yaml
```diff
@@ -230,6 +236,17 @@ jobs:
...
+      # - name: Build and push neonvm-daemon image
+      #   uses: docker/build-push-action@v3
+      #   with:
+      #     context: .
+      #     platforms: linux/amd64
+      #     push: true
+      #     file: neonvm/daemon/Dockerfile
+      #     tags: ${{ needs.tags.outputs.daemon }}
+      #     build-args: |
+      #       GO_BASE_IMG=${{ env.GO_BASE_IMG }}
```
sharnoff, line 248: Should we remove this, and all the other references to `[neonvm-]daemon` in this workflow?

.github/workflows/build-images.yaml
```diff
@@ -102,7 +108,7 @@ jobs:
     # nb: use format(..) to catch both inputs.skip = true AND inputs.skip = 'true'.
     if: ${{ format('{0}', inputs.skip) != 'true' }}
     needs: [ tags, vm-kernel ]
-    runs-on: [ self-hosted, large ]
+    runs-on: [ self-hosted, gen3, large ]
```
sharnoff, line 111:
```suggestion
    runs-on: [ self-hosted, large ]
```

unintented change from rebasing, probably

neonvm/daemon/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG GO_BASE_IMG=autoscaling-go-base:dev
+FROM $GO_BASE_IMG AS builder
```
sharnoff, line 2: Worth considering: We probably _shouldn't_ use go-chef here (and as a dependency in the makefile), because otherwise the `build-test-vm` workflow will have to build all dependencies if there's a cache miss.

And, neonvm-daemon should be pretty quick to compile anyways :)

mikhail-sakhnov:
@sharnoff I resolved highlighted issues, dropped neonvmd mentions from the build-image, kept it only in the build-test-vm workflow. neonvmd doesn't use go-chef anymore. I also moved `neonvm/daemon` to `neonvm-daemon` to comply with new flat structure. I first thought to keep it where it was (`neonvm/daemon`) because it is a component which runs in the VM, but at the same time, vm-runner also runs in the VM but looks like it belongs to `neonvm-runner` now. 

@sharnoff @Omrigan could you take a look at the PR?

vm-builder/files/Dockerfile.img
```diff
@@ -97,7 +108,7 @@ RUN set -e \
     && /neonvm/bin/id -g sshd > /dev/null 2>&1 || /neonvm/bin/addgroup sshd \
     && /neonvm/bin/id -u sshd > /dev/null 2>&1 || /neonvm/bin/adduser -D -H -G sshd -g 'sshd privsep' -s /neonvm/bin/nologin sshd
 
-FROM alpine:3.19 AS builder
+FROM vm-runtime AS builder
```
sharnoff, line 111: This is a revert of a previous change, I think unintended?
(probably my bad, sorry!)

vm-builder/files/Dockerfile.img
```diff
@@ -65,6 +74,8 @@ RUN set -e \
 		quota-tools \
 	&& /helper.move-bins.sh quota edquota quotacheck quotaoff quotaon quotastats setquota repquota tune2fs
 
+COPY --from=vm-daemon-loader /neonvmd /neonvm/bin/neonvmd
```
sharnoff, line 77: I wonder, should this be named `neonvm-daemon-loader` instead? WDYT?

tests/e2e/vm-neonvmd/00-assert.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 90
+commands:
+  - script: |
+      set -eux
+      pod="$(kubectl get neonvm -n "$NAMESPACE" example -o jsonpath='{.status.podName}')"
+      kubectl exec -n "$NAMESPACE" $pod -- ssh guest-vm ps aux -o comm | grep neonvmd | [ $(wc -l) -eq 1 ] || echo "neonvmd process is not found in the $pod"
+---
+apiVersion: vm.neon.tech/v1
```
Omrigan, line 10: Can also spawn a service and check /cpu is accessible.

Maybe for later.

Makefile
```diff
@@ -137,9 +138,8 @@ build: fmt vet bin/vm-builder ## Build all neonvm binaries.
 	GOOS=linux go build -o bin/runner           neonvm/runner/*.go
 
 .PHONY: bin/vm-builder
-bin/vm-builder: ## Build vm-builder binary.
-	GOOS=linux CGO_ENABLED=0 go build -o bin/vm-builder -ldflags "-X main.Version=${GIT_INFO}" vm-builder/main.go
-
+bin/vm-builder: docker-build-daemon ## Build vm-builder binary.
```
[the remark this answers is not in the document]
Omrigan: Yeah, maybe it should be wherever we call vm-builder?

.github/workflows/build-test-vm.yaml
```diff
@@ -29,6 +29,8 @@ on:
 
 env:
   IMG_POSTGRES_16_BULLSEYE: "neondatabase/vm-postgres-16-bullseye"
+  # using locally build base image
+  IMG_DAEMON: "daemon:dev"
```
Omrigan, line 33: Asking, so I can understand better: how does daemon:dev gets build in the CI?

Omrigan (approved):
LGTM, with all comments resolved!

Omrigan (approved):
LGTM! (with all comments resolved)

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: The makefile has target 
```
docker-build-daemon:
	docker build \
		--tag $(IMG_DAEMON) \
		--file neonvm-daemon/Dockerfile \
``` which is called whenever we buld bin/vm-builder

File: .github/workflows/build-test-vm.yaml [hunk 10]
```diff
@@ -29,6 +29,8 @@ on:
 
 env:
   IMG_POSTGRES_16_BULLSEYE: "neondatabase/vm-postgres-16-bullseye"
+  # using locally build base image
```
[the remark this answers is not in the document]
mikhail-sakhnov: Leftover comment, should be dropped

PR #1096, opened 2024-10-03
neonvm-controller: disable cgroup management

mikhail-sakhnov:
neonvm-controller: change args in the deployment, add disable-runner-cgroup, drop enable-container-mgr as mutually exclusive


Omrigan:
I have a similar patch: https://github.com/neondatabase/autoscaling/pull/1054/commits/d7f526dc4df548e08f6e5147a86105006cf1a129

mikhail-sakhnov:
@Omrigan I think we yesterday discussed that we want to redeploy neonvm-controller with disabled cgroup management. Should we go with merging PR #1054 (which includes your patch) instead?

PR #1099, opened 2024-10-10
ci: report workflow stats 

fedordikarev:


sharnoff:
What is this change for?

fedordikarev:
@sharnoff we added workflow, that will store stats for all workflow runs (like jobs, steps, their duration, status, etc).
And based on that stats we build Dashboards with tracking runs duration evolution over time, detecting flaky tests and jobs, and usage of runners to better understand and forecast infrastructure usage and costs.
I will add same info to the description now.

sharnoff:
Thanks. Are there plans to move this workflow to `neondatabase` instead of your personal user?

PR #1102, opened 2024-10-10
neonvm-controller: Use JSON patch to update status

sharnoff:
The controller's final status update often fails due to conflict. This should be totally avoidable, if we switch to patching instead of trying to `Update()` with a potentially stale VM object.

Extracted this from #1009, modified to use JSON patch instead merge patch, as it means we don't have to do an extra `Get()` before updating the status.

---

**Notes for review:** When combined with #1103, I think we can potentially cause some consistency issues — specifically, we've previously seen that the controller can sometimes operate on out-of-date VM objects, which _normally_ would fail to update the status because they're outdated, but this would allow that to continue.

I'm not sure how big of an issue that is... WDYT?


sharnoff:
Closing after some discussion, because I don't think there's a sound way to implement this

PR #1103, opened 2024-10-10
neonvm-controller: Patch status even if reconcile fails

### LINKED ISSUE #889: neonvm-controller does not update VM status if reconcile fails

sharnoff:
Fixes #889. Extracted out from #1009.

Might be worthwhile to require #1102 because otherwise we'll end up double-failing when the VM object is out of date...

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -166,10 +166,10 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 	}
 
 	statusBefore := vm.Status.DeepCopy()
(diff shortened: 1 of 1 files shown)


sharnoff:
from review in the sync, IMO this approach is fundamentally flawed, given other features of the neonvm API. Closing, but we may revisit this in the future.

PR #1106, opened 2024-10-11
ci/vm-kernel: change CACHE_TAG calculation

bayandin:
In PRs, it is incorrect to use the last commit sha as a cache tag for a docker kernel image.
By default, on `pull_request`, `actions/checkout` pulls a fake merge commit (PR into the default branch), and potentially an outdated kernel image from the cache might be used in a PR (if `main` has some changes in the kernel, and PR doesn't include it yet).

This PR changes the logic from using the last commit sha to using a hash from `neonvm-kernel` directory content and the workflow file itself.


fedordikarev (approved):

sharnoff (approved):
nice, thanks!

PR #1107, opened 2024-10-12
agent: Add scaling event reporting

sharnoff:
This is part 2 of 2; see #1078 for the ground work and neondatabase/cloud#15939 for the full context.

In short, this PR:

* Adds a new package: `pkg/agent/scalingevents`
* Adds new callbacks to `core.State` to allow it to report on scaling events changes in desired CU.

---

**Notes for review:**

I'd like to add minio-based S3 tests to this, but it seemed like it'd be non-trivial, particularly because scaling events actually require that there's scaling that happens ­— unlike the existing billing tests.

So I figured I'd open this for review in the meantime.

~~**Also note: This PR builds on #1078 and must not be merged before it.**~~


pkg/agent/scalingevents/reporter.go
```diff
@@ -0,0 +1,133 @@
...
+	Clients ClientsConfig `json:"clients"`
+}
+
+type Reporter struct {
+	conf    *Config
+	sink    *reporting.EventSink[ScalingEvent]
+	metrics PromMetrics
+}
+
+type ScalingEvent struct {
```
sharnoff, line 35: todo: Need to add cluster information to this.

sharnoff:
Remaining items for me, on this:

1. Add more thorough e2e tests
2. Test on staging

pkg/api/vminfo.go
```diff
@@ -26,6 +26,10 @@ const (
 	AnnotationAutoscalingBounds   = "autoscaling.neon.tech/bounds"
 	AnnotationAutoscalingConfig   = "autoscaling.neon.tech/config"
 	AnnotationBillingEndpointID   = "autoscaling.neon.tech/billing-endpoint-id"
+
+	// ref cloud#15939; to be removed after rollout is complete.
```
Omrigan, line 30:
```suggestion
	// ref neondatabase/cloud#15939; to be removed after rollout is complete.
```

Question: why labels are needed?

pkg/agent/scalingevents/clients.go
```diff
@@ -0,0 +1,63 @@
...
+		})
+	}
+
+	return clients, nil
+}
+
+// Returns a function to generate keys for the placement of scaling events data into blob storage.
+//
+// Example: prefix/2024/10/31/23/events_{uuid}.ndjson.gz (11pm on halloween, UTC)
+func newBlobStorageKeyGenerator(prefix string) func() string {
```
Omrigan, line 51: Can we reuse the same key generator as we have for billing?

pkg/agent/runner.go
```diff
@@ -322,6 +333,102 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
+
+	rl.report(r.global.scalingReporter, timestamp, endpointID, currentCU, targetCU, parts)
+}
+
+type desiredScalingReportLimiter struct {
+	lastTarget *uint32
+	lastParts  *scalingevents.GoalCUComponents
+}
+
+func (rl *desiredScalingReportLimiter) report(
```
Omrigan, line 390: Potential alternative would be to have 
```go
type Skipper interface {
	Skip(event ScalingEvent) bool
}
```
injected into `Reporter`, and it would be called in `Submit()`.

This would allow to implement limiter in a more generic way, for any type of event. Plus no need to pass so many arguments, when we can pass only `ScalingEvent`.

pkg/agent/runner.go
```diff
@@ -322,6 +333,102 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
 	}
 }
 
+func (r *Runner) reportScalingEvent(timestamp time.Time, currentCU, targetCU uint32) {
```
Omrigan, line 336: Suggestion: use more consistent names, right now there is `Real` vs `Hypothetical` and `ScalingEvent` vs `DesiredScaling`. 

`reportRealScaling` and `reportHypotheticalScaling` would be fine, but it suggests as if hypothetical value cannot be real. Maybe `Actual` vs `Desired`?

pkg/agent/core/state.go
```diff
@@ -31,15 +31,24 @@ import (
...
 	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
 type ObservabilityCallbacks struct {
 	PluginLatency  revsource.ObserveCallback
 	MonitorLatency revsource.ObserveCallback
 	NeonVMLatency  revsource.ObserveCallback
+
+	ScalingEvent   ReportScalingEventCallback
+	DesiredScaling ReportDesiredScalingCallback
```
Omrigan, line 44: Suggestion: maybe instead of having those as callbacks, define a new adapter interface, and pass it like this?

Passing interface feels more idiomatic.

pkg/agent/core/state.go
```diff
@@ -727,8 +736,20 @@ func (s *state) desiredResourcesFromMetricsOrRequestedUpscaling(now time.Time) (
 	// 2. Cap the goal CU by min/max, etc
 	// 3. that's it!
 
+	reportGoals := func(goalCU uint32, parts scalingevents.GoalCUComponents) {
```
Omrigan, line 739: Suggestion: Instead of having callback here, we could merge `scalingevents.GoalCUComponents` into `scalingGoal`, define a method `GoalCU()` on it, which would return max, and we can put this object into DesiredScaling callback.

WDYT?

Omrigan (commented):
Looks good, some questions and suggestions.

pkg/agent/core/goalcu.go
```diff
@@ -29,30 +31,44 @@ func calculateGoalCU(
...
 	}
 
 	if systemMetrics != nil && wss != nil {
 		memTotalGoalCU = calculateMemTotalGoalCU(cfg, computeUnit, *systemMetrics, *wss)
+		reportedGoals.Mem = lo.ToPtr(max(*reportedGoals.Mem, memTotalGoalCU))
 	}
 
-	goalCU := max(cpuGoalCU, memGoalCU, memTotalGoalCU, lfcGoalCU)
+	goalCU := uint32(math.Ceil(max(
+		math.Round(cpuGoalCU), // for historical compatibility, use round() instead of ceil()
```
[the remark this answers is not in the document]
sharnoff: _Probably_ not, but I didn't want a scaling algorithm change to be a side-effect of this PR.

I'd be happy to change it in a separate PR?

pkg/agent/runner.go
```diff
@@ -322,6 +333,102 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger, vmInfoUpdated util
...
+) {
+	closeEnough := func(x *float64, y *float64) bool {
+		if (x != nil) != (y != nil) {
+			return false
+		} else if x == nil /* && y == nil */ {
+			return true
+		} else {
+			// true iff x and y are within the threshold of each other
+			return math.Abs(*x-*y) < 0.25
+		}
```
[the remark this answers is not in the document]
sharnoff: fair enough!

pkg/agent/core/state.go
sharnoff: Hm, how will that interact with things like #1129 / #1140 ? Otherwise I like this idea, I think it's a lot simpler.

pkg/agent/runner.go
sharnoff: +1 on more consistent naming. Let me explain where I'm coming from; I want your thoughts:

IIRC, the thing I was trying to distinguish here is that the "hypothetical"/"desired" scaling events can go far beyond the endpoint limits, and also contain the fractional CU values for each "part" (cpu/mem/lfc). We also change these events if the _components_ change significantly, even if the overall scaling is still the same (so: it's not like each "hypothetical"/"desired" scaling event constitutes _actual_ scaling)

I think I opted not to go fully for "desired" scaling also because IIRC "desired" is used elsewhere in the autoscaler-agent to mean "the scaling value we should be working towards", and _is_ restricted by endpoint CU limits.

Thoughts? Happy to go with `Actual` vs `Desired` if you think that makes more sense.

pkg/agent/config.go
```diff
@@ -193,6 +196,18 @@ func (c *Config) validate() error {
...
 		erc.Whenf(ec, c.Billing.Clients.S3.PrefixInBucket == "", emptyTmpl, ".billing.clients.s3.prefixInBucket")
 	}
+
+	erc.Whenf(ec, c.ScalingEvents.CUMultiplier == 0, zeroTmpl, ".scalingEvents.cuMultiplier")
+	erc.Whenf(ec, c.ScalingEvents.RereportThreshold == 0, zeroTmpl, ".scalingEvents.rereportThreshold")
+	erc.Whenf(ec, c.ScalingEvents.ClusterName == "", emptyTmpl, ".scalingEvents.clusterName")
+	erc.Whenf(ec, c.ScalingEvents.RegionName == "", emptyTmpl, ".scalingEvents.regionName")
+	if c.ScalingEvents.Clients.S3 != nil {
+		validateBaseReportingConfig(&c.ScalingEvents.Clients.S3.BaseClientConfig, "scalingEvents.clients.s3")
+		erc.Whenf(ec, c.ScalingEvents.Clients.S3.Bucket == "", emptyTmpl, ".scalingEvents.clients.s3.bucket")
```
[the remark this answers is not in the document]
sharnoff: Makes sense, done in 2ed34922c5253d580372fe752fdca6d7cde2902d

pkg/agent/scalingevents/reporter.go
```diff
@@ -0,0 +1,142 @@
...
+	// This exists because Neon allows fractional compute units, while the autoscaler-agent acts on
+	// integer multiples of a smaller compute unit.
+	CUMultiplier float64 `json:"cuMultiplier"`
+
+	// RereportThreshold sets the minimum amount of change in desired compute units required for us to
+	// re-report the desired scaling.
+	RereportThreshold float64 `json:"rereportThreshold"`
+
+	ClusterName string `json:"clusterName"`
+	RegionName  string `json:"regionName"`
```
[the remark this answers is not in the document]
sharnoff: Clarified in d8397d48a8ccae8f2a6f5d32d2f11559af6cc917. As for "why", refer to neondatabase/cloud#15939.

pkg/api/vminfo.go
sharnoff: Labels are there mostly because I'm expecting that there will _probably_ be some issues, being able to enable only for a single VM should make it easier to test, and labels being opt-in means the release process doesn't need to be blocked on testing scaling event reporting.

It's also in part because I'd like to get #1108 released without needing to coordinate testing of this PR :sweat_smile:

This is definitely not the only way to do it, though -- I'm curious for your thoughts regardless.

pkg/agent/scalingevents/clients.go
sharnoff: We could, but we'd end up changing the format. I figured basically:

1. It's a little annoying to make the same key generator available to both packages; and
2. It's _probably_ better to use the same key format as what proxy provides, rather than matching what we use for billing

Not 100% sure though. WDYT?

pkg/agent/runner.go
sharnoff: Hm, that's a fair point. I agree the current situation is overly complex; I'm not sure that adding a `Skipper` interface makes it simpler.

I'll think about it and get back you here...

pkg/agent/scalingevents/reporter.go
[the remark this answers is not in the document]
Omrigan: Neither explains the difference between them. Are they equal at all times? Maybe we need only one?

Omrigan (commented):
A few more questions + some previous are still open. Should be good to go soon.

pkg/agent/config.go
```diff
@@ -189,10 +196,20 @@ func (c *Config) validate() error {
...
 	}
+
+	erc.Whenf(ec, c.ScalingEvents.CUMultiplier == 0, zeroTmpl, ".scalingEvents.cuMultiplier")
+	erc.Whenf(ec, c.ScalingEvents.RereportThreshold == 0, zeroTmpl, ".scalingEvents.rereportThreshold")
+	erc.Whenf(ec, c.ScalingEvents.ClusterName == "", emptyTmpl, ".scalingEvents.clusterName")
+	erc.Whenf(ec, c.ScalingEvents.RegionName == "", emptyTmpl, ".scalingEvents.regionName")
+	if c.ScalingEvents.Clients.S3 != nil {
+		validateBaseReportingConfig(&c.ScalingEvents.Clients.S3.BaseClientConfig, "scalingEvents.clients.s3")
+		validateS3ReportingConfig(&c.ScalingEvents.Clients.S3.S3ClientConfig, ".scalingEvents.clients.s3")
+		erc.Whenf(ec, c.ScalingEvents.Clients.S3.PrefixInBucket == "", emptyTmpl, ".scalingEvents.clients.s3.prefixInBucket")
```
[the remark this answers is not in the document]
sharnoff: It can, but we'd have to pass it independently, because that `.S3` field is a different type between `Billing` and `ScalingEvents`. I thought that would be overly long :sweat_smile:

WDYT?

pkg/agent/scalingevents/prommetrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+type PromMetrics struct {
+	reporting  *reporting.EventSinkMetrics
+	totalCount prometheus.Gauge
+}
+
+func NewPromMetrics() PromMetrics {
+	return PromMetrics{
+		reporting: reporting.NewEventSinkMetrics("autoscaling_agent_events"),
+		totalCount: prometheus.NewGauge(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
sharnoff: Ah nope, good catch — I've now changed it to a gauge vec based on the event kind, and added usage in 2b9324cbcc45545097e6e75fb7df86747a76b009

pkg/agent/runner.go
sharnoff: Switched to `Actual` and `Hypothetical` in 2b9324cbcc45545097e6e75fb7df86747a76b009

pkg/agent/scalingevents/clients.go
sharnoff: Exactly, yeah

pkg/agent/scalingevents/reporter.go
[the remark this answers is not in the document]
sharnoff: Turns out we only need `RegionName` — removed `ClusterName` in def3d84ce9de0716b8fd11ed111eba9fc19512c3

pkg/api/vminfo.go
sharnoff: > But the global configuration is also opt-in, so we can always roll-out with scaling reporting disabled.

That's a fair point — removed the per-VM configuration in e15b277359bcdc23945db258ee87cea1482769a8

pkg/agent/runner.go
sharnoff: After thinking about this:

IMO it's much simpler to have a separate instance of the "skipper" for each VM, so that its state is naturally attached to that lifecycle, rather than trying to limit each VM independently on the reporter side.

If you go that route, then it ends up very similar to what's already implemented here — although I think it does make sense to tweak the interface a bit.

So I've changed the `desiredScalingReportLimiter` to take a `ScalingEvent` as the only argument (fd56bb0336423c8040a8ae15d5cbe40561fe92d3), but I'm inclined to leave the rest as is.

WDYT?

pkg/agent/core/state.go
sharnoff: Makes sense -- implemented in 980aa8538fa0efc6b46626c98cb0addd7d1f0eb4

pkg/agent/core/goalcu.go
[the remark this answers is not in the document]
sharnoff: Preserved this historical compatibility in 980aa8538fa0efc6b46626c98cb0addd7d1f0eb4; opened #1212 to track the follow-up.

Omrigan (approved):
LGTM, pending a few remaining things.

pkg/agent/config.go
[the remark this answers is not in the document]
sharnoff: > Can/should it be the same field?

It could be, but our options there are either:

1. Use the same outer `S3Config` for both billing & scaling events, which requires some shared package; or
2. Make `PrefixInBucket` part of `S3ClientConfig`, which is possible but means that the S3 "reporting" client has some involvement in file paths, instead of just accepting it from the caller

Because of those, I figured the solution with the least weirdness was to just duplicate (even though that is also kind of odd).

[going to leave as-is, just writing here for justification]

pkg/agent/scalingevents/clients.go
sharnoff: Makes sense -- added comments in 40cdc31bd44183d124defb786b4224b644b3559d.

pkg/agent/runner.go
sharnoff: Yup, makes sense - made that change in d7df4c737ff5cb481bab51b52ad05777a7d30b2d

pkg/agent/core/state.go
sharnoff: Hm, could you expand on this? (e.g., just these two? or the entire `ObservabilityCallbacks` struct?)

The reason I didn't already do that is because it seemed like a lot of boilerplate to just glue together different functionalities.

sharnoff:
One comment thread left, cc @Omrigan

pkg/agent/core/state.go
Omrigan: I mean just this two. 

I guess I am trying to figure out when it is better to have callbacks, and when it is better to introduce the adapter. Probably, what differentiates things I previously put into the `ObservabilityCallbacks` is that those are at the top of callstack, while you tend to use anonymous functions for DI, essentially gluing one piece of complex code to another piece of complex code. Plus, I tend to think about callbacks as "something people are forced to use in C, because they don't have an adequate method for defining methods on an object, and doing polymorphism"

But probably I am being too strict, the wiring is perfectly understandable either way. It is fine to use callbacks in Go, as long as we don't put massive amounts of business logic into it. And we even have anonymous functions, compared with C.

Omrigan (approved):

pkg/agent/core/state.go
sharnoff: That all makes sense. I think in general you're right to call out using an interface instead, especially because I have a tendency to make big anonymous functions :sweat_smile:

In this case, I would say it matches this:

> It is fine to use callbacks in Go, as long as we don't put massive amounts of business logic into it.

sharnoff:
Seems ready to merge; I'm planning on testing on staging first.

Two comments to unresolve after merging, for posterity: https://github.com/neondatabase/autoscaling/pull/1107#discussion_r1860796327 and https://github.com/neondatabase/autoscaling/pull/1107#discussion_r1860666752.

sharnoff:
Tested on staging; looks good!

PR #1108, opened 2024-10-12
agent: Add per-VM metric for desired CU(s)

sharnoff:
This commit adds a new per-VM metric: `autoscaling_vm_desired_cu`.

It's based on the same "desired CU" information exposed by the scaling event reporting, but updated continuously instead of being rate limited to avoid spamming our reporting.

The metric has the same base labels as the other per-VM metrics, with the addition of the "component" label, which is one of:

* `total` - the goal CU, after taking the maximum of the individual parts and rounding up to the next unit.
* `cpu` - goal CU size in order to fit the current CPU usage
* `mem` - goal CU size in order to fit the current memory usage (including some information derived from LFC, to make sure there's room for cache too)
* `lfc` - goal CU size in order to fit the estimated working set size

All of these values are also multiplied by the same Compute Unit factor as with the normal scaling event reporting, so that Neon's fractional compute units are exposed as such in the metrics, even as we use integer compute units in the autoscaler-agent.

Also note that all values except "total" are NOT rounded, and instead show the fractional amounts to allow better comparison.

**KNOWN LIMITATION:** If `ReportDesiredScaling` is disabled at runtime for a particular VM, the metrics will not be cleared, and instead will just cease to be updated. I figured this is a reasonable trade-off for simplicity.

---

**Notes for review:** Tested this locally with the following patch to vm-deploy.yaml:

```diff
diff --git a/vm-deploy.yaml b/vm-deploy.yaml
index 09588f60..dcf9f0f0 100644
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -10,6 +10,8 @@ metadata:
     autoscaling.neon.tech/enabled: "true"
     # Set to "true" to continuously migrate the VM (TESTING ONLY)
     autoscaling.neon.tech/testing-only-always-migrate: "false"
+    autoscaling.neon.tech/report-desired-scaling: "true"
+    neon/endpoint-id: "foobar"
 spec:
   schedulerName: autoscale-scheduler
   enableSSH: true
```

_AFAICT_ it works as intended, but metrics are sometimes tricky. I plan to test it on staging before merging.

~~**Also note: This PR builds on #1107 and must not be merged before it.**~~


pkg/agent/prommetrics.go
```diff
@@ -344,9 +347,22 @@ func WrapHistogramVec(hist *prometheus.HistogramVec) revsource.ObserveCallback {
 }
 
 type PerVMMetrics struct {
+	// activeMu and activeVMs exist to track the set of VMs currently represented in the metrics, so
+	// that when we set the desired CU from internal information, we can check whether the VM still
+	// exists.
+	// Otherwise it's not possible to prevent data races that would result in leaking metric labels.
+	activeMu  sync.Mutex
```
Omrigan, line 354: Why did this become needed only now, how `desiredCU` is different from other metrics we already had?

sharnoff: IIRC, it wasn't needed before because the metrics were all derived from the state of the VM, whereas `desiredCU` is derived from our internal state.

pkg/agent/watch.go
```diff
@@ -319,6 +319,14 @@ func setVMMetrics(perVMMetrics *PerVMMetrics, vm *vmv1.VirtualMachine, nodeName
 	for _, m := range restartCountMetrics {
 		perVMMetrics.restartCount.With(m.labels).Set(m.value)
 	}
+
+	// Add the VM to the internal tracker:
```
Omrigan, line 323: nit: I would abstract this as `perVMMetrcis.setActive(vm)` (and `setInactive`)

Omrigan (approved):
Only nits.

pkg/agent/watch.go
sharnoff: Makes sense — went with `updateActive` and `deleteActive`, in e9eb716175cb597821cd053228f7ec056688ece7

pkg/agent/prommetrics.go
```diff
@@ -426,3 +458,46 @@ type vmMetric struct {
 	labels prometheus.Labels
 	value  float64
 }
+
+func (m *PerVMMetrics) updateDesiredCU(
+	vm util.NamespacedName,
+	conversionFactor float64,
```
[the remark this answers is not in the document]
sharnoff: Makes sense -- fixed in cc3057efdd74a25d73331a4362a4a98823afea9a.

pkg/agent/prommetrics.go
```diff
@@ -426,3 +458,46 @@ type vmMetric struct {
...
+		{"lfc", parts.LFC},
+	}
+
+	for _, p := range pairs {
+		labels := prometheus.Labels{
+			"vm_namespace": vm.Namespace,
+			"vm_name":      vm.Name,
+			"endpoint_id":  info.endpointID,
+			"project_id":   info.projectID,
+			"reason":       p.key,
```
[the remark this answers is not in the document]
sharnoff: 4edc175af3c315e35262e44004ec6c4934b23e4f

sharnoff:
Tested on staging, looks good :D

ref https://neondb.slack.com/archives/C03TN5G758R/p1739221210836759

PR #1111, opened 2024-10-14
Misha/cpu scaling

mikhail-sakhnov:
Based on the non-merged PR #1104 

Introduce separate CPU scaling flow based on the vmSpec.cpuScalingMode

If vmSpec.cpuScalingMode is equal to `qmp_scaling` the logic of the
scaling is preserved as before:

- Scale, if required the amount of CPUs using qmp commands.
- If it is required to scale cgroups, call vm-runner /cpu_change endpoint

if vmSpec.cpuScalingMode is equal to `cpuSysfsStateScaling` all cpu
scaling requests go directly to the vm-runner /cpu_change, which in
that configuration goes to the neonvm-daemon to reconcile required
amount of online CPUs.

Value `cpuSysfsStateScaling` also modifies the qemu and the kernel
arguments to enable plug all CPUs but mark as online only first one.


neonvm-controller/cmd/main.go
```diff
@@ -105,6 +106,7 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&useCpuSysfsStateScaling, "use-cpu-sysfs-state-scaling", false, "Use sysfs cpu state scaling for CPU scaling")
```
sharnoff, line 109: nit: this would be better grouped further down, next to the other flags related to settings for the VM

Makefile
```diff
@@ -133,9 +133,10 @@ test: fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: fmt vet bin/vm-builder ## Build all neonvm binaries.
-	GOOS=linux go build -o bin/controller       neonvm/main.go
-	GOOS=linux go build -o bin/vxlan-controller neonvm/tools/vxlan/controller/main.go
-	GOOS=linux go build -o bin/runner           neonvm/runner/*.go
+	GOOS=linux go build -o bin/controller       neonvm-controller/cmd/main.go
+	GOOS=linux go build -o bin/vxlan-controller neonvm-vxlan-controller/cmd/main.go
+	GOOS=linux go build -o bin/daemon           neonvm-daemon/main.go
+	GOOS=linux go build -o bin/runner           neonvm-runner/cmd/main.go
```
sharnoff, line 139: Are these required for this PR? (if no: let's break that out into a separate change? will be easier to discuss without derailing this PR)

neonvm-controller/deployment.yaml
```diff
@@ -55,6 +55,7 @@ spec:
         - "--concurrency-limit=128"
         - "--skip-update-validation-for="
         - "--disable-runner-cgroup"
+        - "--use-cpu-sysfs-state-scaling"
```
sharnoff, line 58: We probably shouldn't enable by default on initial rollout, right? otherwise rollback will result in new computes being unable to be handled by the old controller (?)

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
```diff
@@ -0,0 +1,196 @@
+package cpuscaling
```
sharnoff, line 1: nits:

* Underscores in filenames have semantic meaning in Go, should we avoid that here?
* The current file structure has a centralized `pkg/` directory but individual `component/cmd/` directories. Can we move this to the top-level `pkg/` directory?

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
```diff
@@ -0,0 +1,196 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// CPU directory path
+const cpuPath = "/sys/devices/system/cpu/"
+
+type CPUSysFsStateScaler struct {
+}
```
sharnoff, line 15: What's the intent behind making an empty type with methods, instead of just public functions?

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (c *CPUSysFsStateScaler) EnsureOnlineCPUs(X int) error {
```
sharnoff, line 17:
```suggestion
func (c *CPUSysFsStateScaler) EnsureOnlineCPUs(x int) error {
```

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	onlineCount, err := c.GetActiveCPUsCount()
+	if err != nil {
+		return err
+	}
+
+	if onlineCount < uint32(X) {
+		for _, cpu := range cpus {
+			if cpu == 0 {
```
sharnoff, line 30: nit: this is a lot of nested control flow. Maybe worth separating into >1 function?

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
```diff
@@ -0,0 +1,196 @@
...
+	data, err := os.ReadFile(filepath.Join(cpuPath, "possible"))
+	if err != nil {
+		return nil, err
+	}
+
+	return parseCPURange(string(data))
+}
+
+// parseCPURange parses the CPU range string (e.g., "0-3") and returns a list of CPUs.
+func parseCPURange(cpuRange string) ([]int, error) {
```
sharnoff, line 135: Would it make more sense to return `(start ID, end ID, error)` rather than `([]IDs, error)` ?

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,76 @@ func main() {
...
+	if _, err := w.Write([]byte(fmt.Sprintf("%d", activeCPUs*1000))); err != nil {
+		s.logger.Error("could not write response", zap.Error(err))
+	}
 }
 
-type cpuServer struct{}
+func (s *cpuServer) handleSetCPUStatus(w http.ResponseWriter, r *http.Request) {
+	// TODO: should the call to this method be conditional, only if the statefs cpu scaling is enabled?
+	// on the other hand, currently this endpoint is called by runner only if the statefs scaling is enabled
+	// and it is a bit tricky to pass vmSpec here
```
sharnoff, line 70: commenting on the TODO: it's not worth it to double-check that here. Let's trust what the runner says.

neonvm-runner/cmd/main.go
```diff
@@ -890,12 +889,23 @@ func buildQEMUCmd(
 		logger.Warn("not using KVM acceleration")
 	}
 	qemuCmd = append(qemuCmd, "-cpu", "max")
-	qemuCmd = append(qemuCmd, "-smp", fmt.Sprintf(
-		"cpus=%d,maxcpus=%d,sockets=1,cores=%d,threads=1",
-		vmSpec.Guest.CPUs.Min.RoundedUp(),
-		vmSpec.Guest.CPUs.Max.RoundedUp(),
-		vmSpec.Guest.CPUs.Max.RoundedUp(),
-	))
+	if vmSpec.CpuScalingMode != nil && *vmSpec.CpuScalingMode == vmv1.CpuScalingModeCpuSysfsState {
```
sharnoff, line 892: IMO: we should be moving away from relying on the VM spec in `neonvm-runner`. WDYT about making this a CLI flag instead?

neonvm-runner/cmd/main.go
```diff
@@ -890,12 +889,23 @@ func buildQEMUCmd(
...
+			vmSpec.Guest.CPUs.Max.RoundedUp(),
+			vmSpec.Guest.CPUs.Max.RoundedUp(),
+		))
+	} else {
+		// if we use hotplug we specify start cpus equal to min cpus and scale using udev rules for cpu plug events
+		qemuCmd = append(qemuCmd, "-smp", fmt.Sprintf(
+			"cpus=%d,maxcpus=%d,sockets=1,cores=%d,threads=1",
+			vmSpec.Guest.CPUs.Min.RoundedUp(),
+			vmSpec.Guest.CPUs.Max.RoundedUp(),
+			vmSpec.Guest.CPUs.Max.RoundedUp(),
```
sharnoff, line 906: nit: this is hard to read, particularly when comparing with the section above it (because "min" vs "max" looks quite similar if the rest of the line is the same :sweat_smile:)

Maybe let's use some helper variables above it?

neonvm-runner/cmd/main.go
```diff
@@ -983,6 +993,10 @@ func makeKernelCmdline(cfg *Config, vmSpec *vmv1.VirtualMachineSpec, vmStatus *v
 	if cfg.appendKernelCmdline != "" {
 		cmdlineParts = append(cmdlineParts, cfg.appendKernelCmdline)
 	}
+	if vmSpec.CpuScalingMode != nil && *vmSpec.CpuScalingMode == vmv1.CpuScalingModeCpuSysfsState {
+		// if we use sysfs based scaling we need to specify the start cpus as min CPUs
+		cmdlineParts = append(cmdlineParts, fmt.Sprintf("maxcpus=%d", vmSpec.Guest.CPUs.Min.RoundedUp()))
```
sharnoff, line 998: Why is this required in the kernel command line when it wasn't before?

neonvm-runner/cmd/main.go
```diff
@@ -1029,8 +1043,12 @@ func runQEMU(
 	wg := sync.WaitGroup{}
 
 	wg.Add(1)
+	useCpuSysfsStateScaling := false
+	if vmSpec.CpuScalingMode != nil && *vmSpec.CpuScalingMode == vmv1.CpuScalingModeCpuSysfsState {
+		useCpuSysfsStateScaling = true
+	}
 	go terminateQemuOnSigterm(ctx, logger, &wg)
-	if !cfg.skipCgroupManagement || cfg.enableDummyCPUServer {
+	if !cfg.skipCgroupManagement || cfg.enableDummyCPUServer || useCpuSysfsStateScaling {
```
sharnoff, line 1051: not blocking: @Omrigan with the revert of the crictl patches, this `if` should now be unconditional, right?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -605,6 +618,7 @@ func (p VmPhase) IsAlive() bool {
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="Node",type=string,priority=1,JSONPath=`.status.node`
 // +kubebuilder:printcolumn:name="Image",type=string,priority=1,JSONPath=`.spec.guest.rootDisk.image`
+// +kubebuilder:printcolumn:name="CPUScalingMode",type=string,priority=1,JSONPath=`.spec.cpuScalingMode`
```
sharnoff, line 621: Is this worth including in `kubectl get neonvm -o wide` ?

pkg/api/types.go
```diff
@@ -476,12 +476,17 @@ type RunnerProtoVersion uint32
 
 const (
 	RunnerProtoV1 RunnerProtoVersion = iota + 1
+	RunnerProtoV2
```
sharnoff, line 479: Is this used?

sharnoff (commented):
Partial review, haven't yet gotten to the neonvm-controller changes

neonvm-runner/cmd/main.go
mikhail-sakhnov: Because without that change all cpu would be marked as online on start, while probably we want to have minimum amount of CPUs marked as online on start?

Makefile
mikhail-sakhnov: Not really but I use build target locally for fast builds and run linter

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
mikhail-sakhnov: Honestly, not a lot of reason to have it struct, but I like less coupled code in the wrapping cpuServer structure much more than just a calls to public functions. 

Plus I was originally going to add unittests here and forget :D

neonvm-runner/cmd/main.go
mikhail-sakhnov: so it is what is done for QMP hotplugs by `cpus=minCpus`, has the same semantic

neonvm-runner/cmd/main.go
mikhail-sakhnov: Introduced variables, should be easier to read now

neonvm-daemon/pkg/cpuscaling/sys_fs_state_scaling.go
mikhail-sakhnov: Moved, renamed

neonvm-runner/cmd/main.go
mikhail-sakhnov: Fun fact, that's how it was in the first iteration of the branch 😄 

Sure, will do.

neonvm-runner/cmd/main.go
```diff
@@ -1812,3 +1842,33 @@ func overlayNetwork(iface string) (mac.MAC, error) {
 
 	return mac, nil
 }
+
+func setNeonvmDaemonCPU(cpu vmv1.MilliCPU) error {
```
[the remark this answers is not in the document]
mikhail-sakhnov: both options (direct call to vm daemon or through runner) works for me and I don't have strong opinion. @sharnoff I remember you had some points, may I ask you to share them here?

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,76 @@ func main() {
...
+	if err != nil {
+		s.logger.Error("could not read request body", zap.Error(err))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	updateInt, err := strconv.Atoi(string(body))
+	update := vmv1.MilliCPU(updateInt)
+	if err != nil {
+		s.logger.Error("could not unmarshal request body", zap.Error(err))
+		w.WriteHeader(http.StatusBadRequest)
```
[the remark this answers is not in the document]
mikhail-sakhnov: It doesn't really change a lot where we log it - here or in the runner, in a both ways log goes to the pod log

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -52,6 +52,14 @@ const (
...
 	// The value of this annotation is always a JSON-encoded VirtualMachineResources object.
 	VirtualMachineResourcesAnnotation string = "vm.neon.tech/resources"
+
+	// CpuScalingModeQMP is the value of the VirtualMachineSpec.CpuScalingMode field that indicates
+	// that the VM should use QMP to scale CPUs.
+	CpuScalingModeQMP string = "qmpScaling"
+
+	// CpuScalingModeCpuSysfsState is the value of the VirtualMachineSpec.CpuScalingMode field that
+	// indicates that the VM should use the CPU sysfs state interface to scale CPUs.
+	CpuScalingModeCpuSysfsState string = "sysfsScaling"
```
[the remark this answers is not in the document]
mikhail-sakhnov: Renamed the variable, but I don't like changing the value, for the sake of easy grepping

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
sharnoff: Yeah, broadly my view is that neonvm-daemon is privileged inside the VM, and we shouldn't expose it beyond the pod. With neonvm-runner, it should be _relatively_ easy for us to add mTLS to verify that connections are coming from neonvm-controller, but I'm hesitant to do that in neonvm-daemon because getting secrets into VMs is a little convoluted and has potentially unexpected properties (e.g., they won't be refreshed).

Similarly, in the worst-case scenario of a DoS attack on the HTTP server, I would rather that happens to neonvm-runner than neonvm-daemon so that the internals of the VM are unaffected. There's occasionally attacks on HTTP implementations that require unusual traffic, and having neonvm-runner act as an intermediary guards against that.

Makefile
sharnoff: Hm, that sounds like we should potentially improve our local setup there. I always just manually run `go build ./...` or `golangci-lint run` (approximately — the actual commands are much more convoluted, but I have them in my shell history...)

If you also think there's room for improvement, could you open an issue?

neonvm-runner/cmd/main.go
sharnoff: Ahhhh I see. Could you add a comment saying this?

mikhail-sakhnov: it actually has the comment 🤔 I tried to rephrase it to be a more clarifying, https://github.com/neondatabase/autoscaling/pull/1111/files#diff-7eb577493ab5186c29d09f0791bee695a98db31cd3619c95e834a7924744823fR1003

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
mikhail-sakhnov: Actually I just realized that even from perspective of the ownership going through runner is a bit more clear model. Let's say theoretically some operations can be performed only inside vm (throug daemon) but some only outside (through qmp). Why should controller bother to know the details what to use in every particular case?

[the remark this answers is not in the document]
sharnoff: > Why should controller bother to know the details what to use in every particular case?

Great point, yeah. This is exactly the issue with having the controller directly connect to QEMU for QMP.

Makefile
sharnoff: We discussed this 1-1 — I'm ok to leave this change in the PR, but maybe we should just remove `make build` entirely (we already have `make lint`, etc.)

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,73 @@ func main() {
...
+type cpuServer struct {
+	// Protects CPU operations from concurrent access to prevent multiple ensureOnlineCPUs calls from running concurrently
+	// and ensure that status response is always actual
+	cpuOperationsMutex  *sync.Mutex
+	cpuSystemWideScaler *cpuscaling.CPUSysFsStateScaler
+	logger              *zap.Logger
+}
+
+func (s *cpuServer) handleGetCPUStatus(w http.ResponseWriter) {
+	// should be replaced with cgroups milliseconds to be exposed instead of having CPU
```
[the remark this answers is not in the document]
mikhail-sakhnov: leftover

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,73 @@ func main() {
...
+	s.cpuOperationsMutex.Lock()
+	defer s.cpuOperationsMutex.Unlock()
+	activeCPUs, err := s.cpuSystemWideScaler.GetActiveCPUsCount()
```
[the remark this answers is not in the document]
mikhail-sakhnov: Will change, but it is not exactly getter

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,73 @@ func main() {
...
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
```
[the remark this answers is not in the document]
mikhail-sakhnov: great catch, than you, should be logged

neonvm-runner/cmd/main.go
```diff
@@ -641,15 +640,12 @@ func newConfig(logger *zap.Logger) *Config {
 	flag.BoolVar(&cfg.skipCgroupManagement, "skip-cgroup-management",
 		cfg.skipCgroupManagement,
 		"Don't try to manage CPU")
-	flag.BoolVar(&cfg.enableDummyCPUServer, "enable-dummy-cpu-server",
```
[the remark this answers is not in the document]
mikhail-sakhnov: It will fail with undefined argument. How much is it an issue since we control manifests to deploy?

pkg/neonvm/controllers/vm_qmp_test.go
```diff
@@ -1,10 +1,8 @@
-package controllers_test
+package controllers
```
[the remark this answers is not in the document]
mikhail-sakhnov: leftover from moving files during development, thanks for catching

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,73 @@ func main() {
 	defer logger.Sync() //nolint:errcheck // what are we gonna do, log something about it?
 
 	logger.Info("Starting neonvm-daemon", zap.String("addr", *addr))
+	cpuHotPlugController := &cpuscaling.CPUSysFsStateScaler{}
+	srv := cpuServer{
+		cpuOperationsMutex:  &sync.Mutex{},
+		cpuSystemWideScaler: cpuHotPlugController,
```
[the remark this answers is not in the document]
mikhail-sakhnov: I agree to rename cpuHotPlugController to something (or actually drop intermediate variable here at all), but I don't want to rename "cpuSystemWideScaler" to "cpuSysfsScaler", may be "cpuScaler"? 

because later there would be also something to ensure cgroup limits and my idea was that in the code it would something like:

```pseudocode
s.cpuScaler.EnsureOnlineCPUs(cpuN) // n is non fractional, may be could be EnsureCPUs
s.cgroupsScaler.SetCGroupLimit(cgroupsN) // cgroupsN is fractional and by design lower or equal to cpuN
```

neonvm-runner/cmd/main.go
```diff
@@ -1030,37 +1050,31 @@ func runQEMU(
...
-					return setCgroupLimit(logger, cpu, cgroupPath)
-				},
-			}
-		}
 
-		wg.Add(1)
-		go listenForCPUChanges(ctx, logger, vmSpec.RunnerPort, callbacks, &wg)
+	var callbacks cpuServerCallbacks
+
+	lastValue := &atomic.Uint32{}
```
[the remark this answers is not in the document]
mikhail-sakhnov: Discussed on 1o1, we need it until we have fractional cgroups here implemented since controller sends the fractional cgroup limit which is rounded up to the amount of cpu

tests/e2e/tmp-default-cpu-scaling-mode/00-assert.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 90
+---
+apiVersion: vm.neon.tech/v1
+kind: VirtualMachine
+metadata:
+  name: example
+# temporary added this check to validate that defaulting cpuScalingMode works as expected
```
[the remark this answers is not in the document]
mikhail-sakhnov: >Why is it temporary? To me it seems it should be around as long as the field exists.

I think original idea was to eventually we drop the field, when we will be sure that sys fs scaling works

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	r.updateVMStatusCPU(ctx, vm, vmRunner, pluggedCPU, cgroupUsage)
+	return hotPlugCPUScaled, nil
+}
+
+func (r *VMReconciler) handleCPUScalingSysfs(ctx context.Context, vm *vmv1.VirtualMachine, vmRunner *corev1.Pod) (bool, error) {
+	log := log.FromContext(ctx)
+	specCPU := vm.Spec.Guest.CPUs.Use
+
+	cgroupUsage, err := getRunnerCgroup(ctx, vm)
```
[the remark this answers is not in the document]
mikhail-sakhnov: moved to a separate file and renamed to setRunnerCPULimits/getRunnerCPULimits

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -539,21 +548,25 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
-
 			// get cgroups CPU details from runner pod
 			cgroupUsage, err := getRunnerCgroup(ctx, vm)
 			if err != nil {
 				log.Error(err, "Failed to get CPU details from runner", "VirtualMachine", vm.Name)
 				return err
 			}
-
+			var pluggedCPU uint32
+			if vm.Spec.CpuScalingMode != nil && *vm.Spec.CpuScalingMode == vmv1.CpuScalingModeCpuSysfsState {
```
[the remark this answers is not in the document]
mikhail-sakhnov: > What's the conclusion of the spec and/or status/annotation discussion, are we leaving this value in spec?

We talked with @sharnoff and decided to leave it in spec and make it immutable if it is not nil

mikhail-sakhnov:
@sharnoff @Omrigan I noticed we don't have readiness probe for the vm pod. Wdyt about adding readiness probe to check neonvm-daemon socket availability?

neonvm-daemon/cmd/main.go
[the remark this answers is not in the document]
Omrigan: Yes, `cpuScaler` sounds good, I just wanted `SystemWide` out of the way, to not introduce new concepts.

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
Omrigan: I guess it doesn't matter, it was set by neonvm-controller, and you've removed it in this PR as well.

And neonvm-runner image is tied to neovm-controller version, so there is no chance older controller would run newer images.

@sharnoff am I correct?

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
Omrigan: Yes, right, maybe add a comment here.

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,121 @@
...
+
+	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
+	"github.com/neondatabase/autoscaling/pkg/api"
+)
+
+// handleCPUScaling is extracted from doReconcile and encapsulates the logic to handle CPU scaling.
+// if vm scaling mode is set to CpuSysfsState, the scaling is delegated to neonvm-daemon
+// otherwise the scaling is first done by scaling amount of cores in the VM using QMP and then by updating the cgroup
+func (r *VMReconciler) handleCPUScaling(ctx context.Context, vm *vmv1.VirtualMachine, vmRunner *corev1.Pod) (bool, error) {
+
```
Omrigan, line 19:
```suggestion
```

tests/e2e/tmp-default-cpu-scaling-mode/00-assert.yaml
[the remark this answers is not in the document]
Omrigan: > not forgetting to delete it

The test would break - impossible to forget. 

I'm just saying explicitly marking it as temporary gives the impression that the test is "more temporary", than the field.

Anyway, up to you.

Omrigan (commented):
Didn't do another full pass yet, but a couple of notes:
1. There are couple of unaddressed comments, maybe they slipped from you because they are on now-outdated code. They are visible in "Conversation" tab.
2. Do you plan to merge it with rebase or with squash?

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
mikhail-sakhnov: added

tests/e2e/tmp-default-cpu-scaling-mode/00-assert.yaml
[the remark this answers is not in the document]
mikhail-sakhnov: kept it as is

mikhail-sakhnov:
@Omrigan 

> Didn't do another full pass yet, but a couple of notes:
> There are couple of unaddressed comments, maybe they slipped from you because they are on now-outdated code. They are visible in "Conversation" tab.

As I mentined, I haven't found any more unaddressed comments.


> Do you plan to merge it with rebase or with squash?

I don't have any specific preference here, what do you suggest?

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,119 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	corev1 "k8s.io/api/core/v1"
+
+	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
+	"github.com/neondatabase/autoscaling/pkg/api"
+)
+
+// handleCPUScaling is extracted from doReconcile and encapsulates the logic to handle CPU scaling.
```
sharnoff, line 15: suggestion:

> handleCPUScaling is extracted from doReconcile

IIUC, that's not relevant to what this function is / does, right? that's just the history?

If so, I'd just remove that part, so it reads

> handleCPUScaling encapsulates the logic to handle CPU scaling.

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if err != nil {
+		log.Error(err, "Failed to scale CPU", "VirtualMachine", vm.Name, "CpuScalingMode", vm.Spec.CpuScalingMode)
+		return false, err
+	}
+
+	return scaled, nil
+}
+
+// handleCPUScalingQMP handles CPU scaling using QMP, extracted as is from doReconcile
```
sharnoff, line 38: Same note here

pkg/neonvm/controllers/vm_qmp_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	"github.com/neondatabase/autoscaling/pkg/neonvm/controllers"
+	. "github.com/neondatabase/autoscaling/pkg/neonvm/controllers"
```
sharnoff, line 7: suggestion: IIUC this change is unrelated? if so, let's remove it and the other changes to this file?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1474,10 +1382,9 @@ func podSpec(
 					Command: func() []string {
 						cmd := []string{"runner"}
 						if config.DisableRunnerCgroup {
-							cmd = append(cmd, "-skip-cgroup-management")
 							// cgroup management disabled, but we still need something to provide
 							// the server, so the runner will just provide a dummy implementation.
```
sharnoff, line 1386: question: Is this comment still valid now that `-enable-dummy-cpu-server` was removed?

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -145,6 +145,11 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) (admission.Warnings,
 		}
 	}
 
+	// allow to change CPU scaling mode only if it's not set
+	if r.Spec.CpuScalingMode != nil && before.Spec.CpuScalingMode != nil && *r.Spec.CpuScalingMode != *before.Spec.CpuScalingMode {
+		return nil, fmt.Errorf(".spec.cpuScalingMode is immutable")
```
sharnoff, line 150: suggestion: not quite "immutable" per se, but can only be set once, right? If so, should this message be different?

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
```diff
@@ -145,6 +145,11 @@ func (r *VirtualMachine) ValidateUpdate(old runtime.Object) (admission.Warnings,
 		}
 	}
 
+	// allow to change CPU scaling mode only if it's not set
+	if r.Spec.CpuScalingMode != nil && before.Spec.CpuScalingMode != nil && *r.Spec.CpuScalingMode != *before.Spec.CpuScalingMode {
```
sharnoff, line 149: question: This still allows (non-nil) -> (nil), right? is that intentional?

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -150,6 +158,11 @@ type VirtualMachineSpec struct {
 	// propagated to the VM.
 	// +optional
 	TargetRevision *RevisionWithTime `json:"targetRevision,omitempty"`
+
+	// Controls how CPU scaling is performed, either hotplug new CPUs with QMP, or enable them in sysfs.
+	// +kubebuilder:validation:Enum=qmpScaling;sysfsScaling
+	// +optional
+	CpuScalingMode *string `json:"cpuScalingMode,omitempty"`
```
sharnoff, line 165: suggestion: I would rather use a distinct type here, instead of `string`. IIRC you can then do the `kubebuilder:validation:Enum` on the type definition (see e.g. `RestartPolicy` in this file)

(with that, a small nit: the values should probably be upper camel case)

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	onlineCount, err := c.ActiveCPUsCount()
+	if err != nil {
+		return err
+	}
+
+	if onlineCount < uint32(targetCount) {
+		for _, cpu := range cpus {
+			if cpu == 0 {
```
sharnoff, line 29: suggestion: this is a lot of nested control flow. Maybe worth separating into >1 function?

*(recreated from https://github.com/neondatabase/autoscaling/pull/1111#discussion_r1804084387)*

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,195 @@
...
+	data, err := os.ReadFile(filepath.Join(cpuPath, "possible"))
+	if err != nil {
+		return nil, err
+	}
+
+	return parseCPURange(string(data))
+}
+
+// parseCPURange parses the CPU range string (e.g., "0-3") and returns a list of CPUs.
+func parseCPURange(cpuRange string) ([]int, error) {
```
sharnoff, line 134: suggestion: Should this return (start ID, end ID, error) rather than ([]IDs, error) ?

*(recreated from https://github.com/neondatabase/autoscaling/pull/1111#discussion_r1804084900)*

neonvm-runner/cmd/main.go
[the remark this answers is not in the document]
sharnoff: That's correct, yeah. Each neonvm-controller instance only creates runners with the same version.

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,119 @@
...
+func (r *VMReconciler) handleCPUScalingQMP(ctx context.Context, vm *vmv1.VirtualMachine, vmRunner *corev1.Pod) (bool, error) {
```
sharnoff, line 39: suggestion: This function is very visually dense. Could you break it up more by adding newlines between sections?

sharnoff (commented):
Some comments — looks mostly there!

A couple higher level things I'd note:

* There are some unrelated changes in here that make it harder to review and unnecessarily delay those changes (see e.g. `Makefile` changes, `pkg/neonvm/controllers/vm_controller_unt_test.go`) — if it's not too much effort, it'd be useful to separate those out. Not strictly required to merge though.

* For things like "moving the CPU QMP scaling logic out of `doReconcile`", it would be helpful to be able to view that as a separate commit. This is of course much more involved, so I don't think it's required to merge, but for the future — I often find that keeping refactoring separate from other changes is really helpful for reviewing :sweat_smile:

pkg/neonvm/cpuscaling/sysfs.go
mikhail-sakhnov: It is not always possible, input might be for example "0-13,15,17-19". The logic of this particular code is that we always enable/disable in order, but in general the CPU state is shared resource and there are no guarantees that it is kept in order.

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
mikhail-sakhnov: Nope, it is a great catch in the logic flaw, thanks for pointing out

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: droped comment

neonvm/apis/neonvm/v1/virtualmachine_webhook.go
mikhail-sakhnov: changed message

pkg/neonvm/controllers/vm_qmp_test.go
mikhail-sakhnov: reverted that change

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
mikhail-sakhnov: rephrased comment

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
mikhail-sakhnov: rephrased comment

neonvm/apis/neonvm/v1/virtualmachine_types.go
mikhail-sakhnov: changed it to be type + used kubebuilder:validation:Enum at the type level

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
mikhail-sakhnov: tried to improve readability here.

pkg/neonvm/cpuscaling/sysfs.go
mikhail-sakhnov: Changed the code in this layer:
- split into two structures, one that reconciles the amount of CPUs and the other one works with sysfs files
- added unit tests for the reconcilation logic
- changed functions to work with ranges instead of slices.

pkg/neonvm/cpuscaling/sysfs.go
mikhail-sakhnov: changed

pkg/neonvm/cpuscaling/sysfsscaling.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return err
+	}
+
+	onlineCount, err := c.GetActiveCPUsCount()
+	if err != nil {
+		return err
+	}
+
+	if onlineCount < uint32(X) {
```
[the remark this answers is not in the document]
mikhail-sakhnov: Is it still applicable keeping in mind changes from the other comment https://github.com/neondatabase/autoscaling/pull/1111#discussion_r1840573281?

pkg/neonvm/cpuscaling/cpuscaler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if onlineCount != targetCount {
+		return fmt.Errorf("failed to ensure %d CPUs are online, current online CPUs: %d", targetCount, onlineCount)
+	}
+
+	return nil
+}
+
+// ActiveCPUsCount() returns the count of online CPUs.
+func (c *CPUScaler) ActiveCPUsCount() (int, error) {
```
sharnoff, line 96: Just discovered `/sys/devices/system/cpu/online` — would that work here?

mikhail-sakhnov: it coulds be used as well, that's correct. I found that checking each cpu is more straightforward approach then parsing ranges in this file because format is a bit of a mix of ranges and values and could be contrintuitive. For example, value like `0,3,7-10,13-15,17-19` is totally legit.

mikhail-sakhnov: on the other hand we still use range parsing to get list of available CPUs, so might be used just the same here. Let me double-check tomorrow, why I don't use it at the moment, I remember that originally I used this file, but switched to the per-cpu files during testing.

pkg/neonvm/cpuscaling/cpuscaler.go
```diff
@@ -0,0 +1,98 @@
+package cpuscaling
```
sharnoff, line 1: question: IIUC this package only works with sysfs scaling, is there a reason for that?

(asking in part because it seems like our sysfs usage could be a lot more efficient, if not for the definition of the `CPUStater` interface)

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	end, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return -1, -1, err
+	}
+	return start, end, nil
+}
+
+// parseMultipleCPURange parses the multiple CPU range string (e.g., "0-3,5-7") and returns a list of CPUs.
+func (cs *cpuSysfsState) parseMultipleCPURange(cpuRange string) ([]int, error) {
```
sharnoff, line 89:
```suggestion
func (cs *cpuSysfsState) parseMultipleCPURange(cpuRanges string) ([]int, error) {
```

(because it's not just one range ?)

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,127 @@
...
+		for cpu := start; cpu <= end; cpu++ {
+			cpus = append(cpus, cpu)
+		}
+	}
+
+	return cpus, nil
+}
+
+// ActiveCPUsCount() returns the count of online CPUs.
+func (c *cpuSysfsState) ActiveCPUsCount() (int, error) {
```
sharnoff, line 109: Here, can we just
```go
ranges, err := os.ReadFile("/sys/devices/system/cpu/online")
if err != nil {
    ...
}

cpus, err := c.parseMultipleCPURange(ranges)
if err != nil {
    ...
}

return len(cpus), nil
```

?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -81,8 +78,7 @@ type VMReconciler struct {
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 	Config   *ReconcilerConfig
-
-	Metrics ReconcilerMetrics `exhaustruct:"optional"`
+	Metrics  ReconcilerMetrics `exhaustruct:"optional"`
```
sharnoff, line 81: nit: unrelated diff?

neonvm-daemon/cmd/main.go
```diff
@@ -26,24 +32,71 @@ func main() {
...
+type cpuServer struct {
+	// Protects CPU operations from concurrent access to prevent multiple ensureOnlineCPUs calls from running concurrently
+	// and ensure that status response is always actual
+	cpuOperationsMutex *sync.Mutex
+	cpuScaler          *cpuscaling.CPUScaler
+	logger             *zap.Logger
 }
 
-type cpuServer struct{}
+func (s *cpuServer) handleGetCPUStatus(w http.ResponseWriter) {
```
sharnoff, line 51: same comment re: whitespace on `handleGetCPUStatus` and `handleSetCPUStatus` — just a couple blank lines to provide visual distinction to the different parts should suffice

neonvm/apis/neonvm/v1/virtualmachine_types.go
sharnoff: nice, looks good!

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,127 @@
...
+func (cs *cpuSysfsState) PossibleCPUs() (int, int, error) {
+	data, err := os.ReadFile(filepath.Join(cpuPath, "possible"))
+	if err != nil {
+		return -1, -1, err
+	}
+
+	return cs.parseCPURange(string(data))
+}
+
+// parseCPURange parses the CPU range string (e.g., "0-3") and returns a list of CPUs.
```
sharnoff, line 62: > returns a list of CPUs.

question: This is no longer true, right? Let's change it?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -165,6 +161,17 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	// examine cpuScalingMode and set it to the default value if it is not set
+	if vm.Spec.CpuScalingMode == nil || *vm.Spec.CpuScalingMode == "" {
+		log.Info("Setting default CPU scaling mode", "default", r.Config.DefaultCPUScalingMode)
+		vm.Spec.CpuScalingMode = lo.ToPtr(vmv1.CpuScalingMode(r.Config.DefaultCPUScalingMode))
```
sharnoff, line 167: question: Is the `vmv1.CpuScalingMode(...)` cast necessary here?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -165,6 +161,17 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	// examine cpuScalingMode and set it to the default value if it is not set
+	if vm.Spec.CpuScalingMode == nil || *vm.Spec.CpuScalingMode == "" {
```
sharnoff, line 165: question: `*vm.Spec.CpuScalingMode == ""` can't happen, right?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -553,18 +560,35 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
-				log.Error(err, "Failed to get CPU details from runner", "VirtualMachine", vm.Name)
+			if vm.Spec.CpuScalingMode == nil { // should not happen
+				err := fmt.Errorf("CPU scaling mode is not set")
+				log.Error(err, "Unknown CPU scaling mode", "VirtualMachine", vm.Name)
+				return err
+			}
+
+			switch *vm.Spec.CpuScalingMode {
+			case vmv1.CpuScalingModeSysfs:
+				log.Info("CPU usage check based on cgroups", "CpuScalingMode", *vm.Spec.CpuScalingMode)
```
sharnoff, line 579: question: What's going on here? Saw "sysfs" + "based on cgroups" and was confused :sweat_smile:

sharnoff (commented):
One more round. I think after this it's likely to be good to go.

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: Yes, a leftover from the field not being valiated

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: dropped

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: yeah, it is a consequence of the fact that all helper functions to call runner used to be called with Cgroups in it, like setRunnerGroups/getRunnerGroups, etc and eventually, it reached to the current state. Changed the log message.

mikhail-sakhnov: In fact, I don't think we really need those logs at all

pkg/neonvm/controllers/vm_controller_cpu_scaling.go
```diff
@@ -0,0 +1,127 @@
...
+				vm.Name))
+		return false, nil
+	} else if specCPU != cgroupUsage.VCPUs {
+		_, err := r.handleCgroupCPUUpdate(ctx, vm, cgroupUsage)
+		if err != nil {
+			log.Error(err, "Failed to update cgroup CPU", "VirtualMachine", vm.Name)
+			return false, err
+		}
+	} else {
+		log.Info("No need to plug or unplug CPU")
```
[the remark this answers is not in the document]
mikhail-sakhnov: dropped

pkg/neonvm/cpuscaling/cpuscaler.go
mikhail-sakhnov: The current package name came from the interface rather than implementation, it feels more or less natural to me to name packages after what they do, not how they do. I can rename the package, if that is not reasonable for you.

>asking in part because it seems like our sysfs usage could be a lot more efficient, if not for the definition of the CPUStater interface

I'm not sure I understand the statement. Could you please clarify? The interface is defined as a separate part that actually works with the filesystem to allow unit testing of the scaling logic.

pkg/neonvm/cpuscaling/sysfs.go
mikhail-sakhnov: changed

pkg/neonvm/cpuscaling/sysfs.go
mikhail-sakhnov: changed to use intermediate getAllOnlineCPUs which is used in GetState

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: reverted

pkg/neonvm/cpuscaling/cpuscaler.go
sharnoff: For some context, the reason I'm pushing on this is that I'm wary about extra complexity — we have a whole setup where we're looping through CPUs, checking their states, and keeping a running counter while we maybe update them, but we already have an easy way of finding just the CPUs we want to update — why not just use that!

pkg/neonvm/cpuscaling/cpuscaler.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if len(online) == targetCount {
+		return nil
+	}
+
+	if len(online) > targetCount {
+		diff := len(online) - targetCount
+		// offline 'diff' CPUs that are currently online
+		// reverse online slice
```
sharnoff, line 44: > reverse online slice

The comment says what this is doing, but not why — could you add that?

sharnoff (approved):
Looks great! :rocket:

Left just a couple final comments

pkg/neonvm/cpuscaling/sysfs.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (cs *cpuSysfsState) OnlineCPUs() ([]int, error) {
+	data, err := os.ReadFile(filepath.Join(cpuPath, "online"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read online CPUs: %w", err)
+	}
+	cpuIDs, err := cs.parseMultipleCPURange(string(data))
+	if err != nil {
+		// log value of the file in case we can't parse to help debugging
+		return nil, fmt.Errorf("failed to parse online CPUs `%s`: %w", string(data), err)
```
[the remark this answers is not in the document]
mikhail-sakhnov: changed

neonvm-runner/cmd/main.go
```diff
@@ -890,12 +886,32 @@ func buildQEMUCmd(
...
-	))
+
+	// cpu scaling details
+	maxCPUs := vmSpec.Guest.CPUs.Max.RoundedUp()
+	minCPUs := vmSpec.Guest.CPUs.Min.RoundedUp()
+
+	switch cfg.cpuScalingMode {
+	case vmv1.CpuScalingModeSysfs:
+		qemuCmd = append(qemuCmd, "-smp", fmt.Sprintf(
+			// Boot with all CPUs plugged, we will online them on-demand
```
Omrigan, line 897:
```suggestion
		// Boot with all CPUs plugged, we will online them on-demand
		qemuCmd = append(qemuCmd, "-smp", fmt.Sprintf(
```

To match the format below.

Omrigan (approved):
LGTM, one suggestion.

Issue #1081, opened 2024-09-24
Modify vm-builder to provide arm64 vm images


mikhail-sakhnov:
## Problem description / Motivation
vm-buider builds vm image for the neonvm-runner. Build images should be modified to support both arm64 and x86. 

## Feature idea(s) / DoD
Probably, multiarch images could be helpful here.

Issue #1089, opened 2024-09-26
Build multiplatform images for autoscaling components


mikhail-sakhnov:
To fully support ARM we need to have CI able to build artifacts for arm architecture

Issue #1084, opened 2024-09-24
Support ARM arch in CI for e2e tests and unittests


mikhail-sakhnov:
## Problem description / Motivation
To fully support ARM instances we need to have proper CI pipelines which are able to run e2e tests on arm

PR #1116, opened 2024-10-17
neonvm-builder: arm images

mikhail-sakhnov:
Modify neonvm-builder to support building images with both amd64 or arm64 architectures.
Change busybox loader to use public image busybox:1.35.0-musl instead of downloading busybox from busybox.net
Parametrize vector url to be architecture specific.
Has no cross-build support, produced image architecture is the same as host machine architecture.

#1081

### DIFF
--- a/.github/workflows/build-test-vm.yaml
+++ b/.github/workflows/build-test-vm.yaml
@@ -31,6 +31,7 @@ env:
   IMG_POSTGRES_16_BULLSEYE: "neondatabase/vm-postgres-16-bullseye"
   # using image built in the same workflow
   IMG_DAEMON: "neondatabase/neonvm-daemon"
+  TARGET_ARCH: "amd64"
 
 defaults:
   run:
@@ -87,7 +88,7 @@ jobs:
 
       - name: build ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
         run: |
-          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ needs.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ needs.tags.outputs.daemon }}
+          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ needs.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ needs.tags.outputs.daemon }} -target-arch linux/${TARGET_ARCH}
       - name: docker push ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
         run: |
           docker push ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
--- a/.github/workflows/vm-example.yaml
+++ b/.github/workflows/vm-example.yaml
@@ -8,6 +8,9 @@ on:
       - ".github/workflows/vm-example.yaml"
   workflow_dispatch: # adds ability to run this manually
 
+env:
+  TARGET_ARCH: amd64
+
 jobs:
   vm-example:
     runs-on: ubuntu-latest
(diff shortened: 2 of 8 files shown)


.github/workflows/vm-example.yaml
```diff
@@ -5,6 +5,9 @@ on:
     - cron:  '42 4 * * 2'  # run once a week
   workflow_dispatch: # adds ability to run this manually
 
+env:
+  TARGET_ARCH: amd64
```
sharnoff, line 9: Is it worth adding this as a workflow parameter, to allow running for ARM? No pressure to do so in this PR — always possible to make the change later if not immediately useful.

Makefile
```diff
@@ -20,7 +20,7 @@ GOOS ?= $(shell go env GOOS)
 # The target architecture for linux kernel. Possible values: amd64 or arm64. 
 # Any other supported by linux kernel architecture could be added by introducing new build step into neonvm/hack/kernel/Dockerfile.kernel-builder
 KERNEL_TARGET_ARCH ?= amd64
-
+TARGET_ARCH ?= amd64
```
sharnoff, line 23: Should we auto-detect this?

Makefile
```diff
@@ -184,6 +184,7 @@ docker-build-controller: docker-build-go-base ## Build docker image for NeonVM c
 docker-build-runner: docker-build-go-base ## Build docker image for NeonVM runner
 	docker build \
 		--tag $(IMG_RUNNER) \
+		--build-arg TARGET_ARCH=$(TARGET_ARCH) \
```
sharnoff, line 187: IIRC `neonvm-runner` doesn't yet support ARM, right? Should this be left for later?

(unless I'm misremembering, in which case, please disregard :smile:)

vm-builder/main.go
```diff
@@ -129,18 +139,28 @@ type inittabCommand struct {
...
 		log.Println("neonvm-daemon image not set, needs to be explicitly passed in, or compiled with -ldflags '-X main.NeonvmDaemonImage=...'")
 		flag.PrintDefaults()
 		os.Exit(1)
 	}
 
+	if targetArch == nil || *targetArch == "" {
+		log.Println("Target architecture not set, see usage info:")
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
```
sharnoff, line 156: IIUC, this will mean that we start requiring a new flag to be set — WDYT about alternately auto-detecting it here? (I'm personally undecided, want to hear what you think)

vm-builder/main.go
```diff
@@ -291,7 +311,8 @@ func main() {
 		SpecBuild:       "",  // overridden below if spec != nil
 		SpecMerge:       "",  // overridden below if spec != nil
 		InittabCommands: nil, // overridden below if spec != nil
-		ShutdownHook:    "",  // overridden below if spec != nil
+		AgettyInitLine:  getAgettyInitLine(*targetArch),
+		ShutdownHook:    "", // overridden below if spec != nil
```
sharnoff, line 315: super tiny nit: maybe let's move `AgettyInitLine` into a blog above, so it doesn't cause an unnecessary diff?

vm-builder/main.go
```diff
@@ -348,6 +369,8 @@ func main() {
 		{"vmstart", scriptVmStart},
 		{"vmshutdown", scriptVmShutdown},
 		{"inittab", scriptInitTab},
+		{"agetty-init-amd64", scriptAgettyInitAmd64},
+		{"agetty-init-arm64", scriptAgettyInitArm64},
```
sharnoff, line 373: IIUC these are not used in the dockerfile, right?

sharnoff (commented):
Broadly looks good. Left some comments, nothing major.

.github/workflows/vm-example.yaml
mikhail-sakhnov: I have a separate issue to get everything working in CI #1089 (this one more dedicated to actual builds) and #1084 (this one more dedicated to have working e2e tests in CI)

Makefile
mikhail-sakhnov: probably a leftover - I originally have one branch which was split into this one and #1119

vm-builder/main.go
mikhail-sakhnov: that's correct, thanks for pointing this out!

Makefile
mikhail-sakhnov: Yes, I for sure will add as a part of the CI related issues (I mentioned them in a previous comment) while setting up arm builder!

vm-builder/main.go
mikhail-sakhnov: done

sharnoff (approved):
lgtm! Given that #1111 is broader than just ARM, let's keep this PR separate until that's merged?

.github/workflows/build-test-vm.yaml
```diff
@@ -85,7 +86,7 @@ jobs:
 
       - name: build ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
         run: |
-          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ needs.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ needs.tags.outputs.daemon }}
+          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ needs.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ needs.tags.outputs.daemon }} -target-arch linux/${TARGET_ARCH}
```
bayandin, line 89: Out of curiosity, why do we need `TARGET_ARCH` / `-target-arch` if

> Has no cross-build support, produced image architecture is the same as host machine architecture.

(sorry if I missed this in the comments/discusstions)

vm-builder/main.go
```diff
@@ -541,3 +564,15 @@ func (f file) validate() []error {
 
 	return errs
 }
+
+func getAgettyInitLine(targetArch string) string {
+	switch targetArch {
+	case targetArchLinuxAmd64:
+		return scriptAgettyInitAmd64
```
Omrigan, line 571: Why did you put it into a separate file, instead of constructing the line here? The bulk of it can be put as a string literal here, and the final string gets formatetted based on the architecture. 

A description or a comment explaining why agetty line needs to be different would be helpful also.

Omrigan (approved):
LGTM, except for one comment.

One more thing is that it would be helpful to either enumerate the changes in a commit message (generic busybox, generic kuttl, custom init line and parateers propagations), or, have them as separate commit after merge.

Makefile
```diff
@@ -491,8 +496,14 @@ CODE_GENERATOR_VERSION ?= v0.28.12
 
 KUTTL ?= $(LOCALBIN)/kuttl
 # k8s deps @ 1.28.3
-KUTTL_VERSION ?= v0.16.0
-
+KUTTL_VERSION ?= v0.19.0
```
sharnoff, line 499: Please do not bump the kuttl version here; kuttl v0.19.0 uses k8s deps at version 1.31.0

vm-builder/main.go
mikhail-sakhnov: I didn't want to use string literal, because the line was quite long. Though this change actually comes from the time I wasn't sure how much difference we might want between agetty for arm and amd, currently it can be replaced by passing only tty variable, since it is the only difference.

Makefile
mikhail-sakhnov: reverted

Issue #1080, opened 2024-09-24
Modify neonvm-runner to support running VMs with arm architecture


mikhail-sakhnov:
## Problem description / Motivation

Modify neonvm-runner to use arm64 qemu binary in arm64 environments and build correct arguments set. 

## Feature idea(s) / DoD

DoD: arm64 instances are possible to run.

Issue #1083, opened 2024-09-24
Prevent live migrations to be executed across different architectures


mikhail-sakhnov:
## Problem description / Motivation
Implement cordons or guards against VM migration from one architecture to another.

stradig:
Not necessary anymore with the dedicated architecture flag.

PR #1119, opened 2024-10-21
neonvm-runner: arm support

mikhail-sakhnov:
Modify qemu arguments to run VM on arm machine

#1080

### DIFF
--- a/neonvm-runner/Dockerfile
+++ b/neonvm-runner/Dockerfile
@@ -5,7 +5,6 @@ COPY . .
 # Build
 RUN CGO_ENABLED=0 go build -o /runner neonvm-runner/cmd/*.go
 
-
 FROM alpine:3.19
 
 RUN apk add --no-cache \
@@ -21,6 +20,7 @@ RUN apk add --no-cache \
     busybox-extras \
     e2fsprogs \
     qemu-system-x86_64 \
+    qemu-system-aarch64 \
     qemu-img \
     cgroup-tools \
     openssh
--- a/neonvm-runner/Dockerfile
+++ b/neonvm-runner/Dockerfile
@@ -25,8 +25,11 @@ RUN apk add --no-cache \
     cgroup-tools \
     openssh
 
+
 COPY --from=builder /runner /usr/bin/runner
 COPY neonvm-kernel/vmlinuz /vm/kernel/vmlinuz
 COPY neonvm-runner/ssh_config /etc/ssh/ssh_config
+# QEMU_EFI used only by runner running on the arm architecture
+RUN wget https://releases.linaro.org/components/kernel/uefi-linaro/16.02/release/qemu64/QEMU_EFI.fd -O /vm/QEMU_EFI_ARM.fd
 
 ENTRYPOINT ["/sbin/tini", "--", "runner"]
--- a/neonvm-runner/cmd/main.go
+++ b/neonvm-runner/cmd/main.go
@@ -623,10 +623,14 @@ type Config struct {
 	appendKernelCmdline  string
 	skipCgroupManagement bool
 	diskCacheSettings    string
-	memoryProvider       vmv1.MemoryProvider
-	autoMovableRatio     string
-	cpuScalingMode       vmv1.CpuScalingMode
-	architecture         string
+	// memoryProvider is a memory provider to use. Validated in newConfig.
+	memoryProvider vmv1.MemoryProvider
+	// autoMovableRatio value for VirtioMem provider. Validated in newConfig.
+	autoMovableRatio string
+	// cpuScalingMode is a mode to use for CPU scaling. Validated in newConfig.
+	cpuScalingMode vmv1.CpuScalingMode
+	// System CPU architecture. Set automatically equal to runtime.GOARCH.
+	architecture string
 }
 
 func newConfig(logger *zap.Logger) *Config {
@@ -637,10 +641,10 @@ func newConfig(logger *zap.Logger) *Config {
 		appendKernelCmdline:  "",
 		skipCgroupManagement: false,
 		diskCacheSettings:    "cache=none",
-		memoryProvider:       "",             // Require that this is explicitly set. We'll check later.
-		autoMovableRatio:     "",             // Require that this is explicitly set IFF memoryProvider is VirtioMem. We'll check later.
-		cpuScalingMode:       "",             // Require that this is explicitly set. We'll check later.
-		architecture:         runtime.GOARCH, // arm64, amd64
+		memoryProvider:       "",
+		autoMovableRatio:     "",
+		cpuScalingMode:       "",
+		architecture:         runtime.GOARCH,
 	}
 	flag.StringVar(&cfg.vmSpecDump, "vmspec", cfg.vmSpecDump,
 		"Base64 encoded VirtualMachine json specification")
--- a/neonvm-runner/cmd/main.go
+++ b/neonvm-runner/cmd/main.go
@@ -1253,8 +1253,10 @@ func getQemuBinaryName(architecture string) string {
 func getMachineType(architecture string) string {
 	switch architecture {
 	case architectureArm64:
+		// virt is the most up to date and generic ARM machine architecture
 		return "virt"
 	case architectureAmd64:
+		// q35 is the most up to date and generic x86_64 machine architecture
 		return "q35"
 	default:
 		panic(fmt.Errorf("unknown architecture %s", architecture))
--- a/neonvm-runner/cmd/main.go
+++ b/neonvm-runner/cmd/main.go
@@ -922,7 +922,6 @@ func buildQEMUCmd(
 		"-only-migratable",
 		"-audiodev", "none,id=noaudio",
 		"-serial", "pty",
-		"-serial", "stdio",
 		"-msg", "timestamp=on",
 		"-qmp", fmt.Sprintf("tcp:0.0.0.0:%d,server,wait=off", vmSpec.QMP),
 		"-qmp", fmt.Sprintf("tcp:0.0.0.0:%d,server,wait=off", vmSpec.QMPManual),
@@ -954,7 +953,16 @@ func buildQEMUCmd(
 	}
 
 	if cfg.architecture == architectureArm64 {
+		// add custom firmware to have ACPI working
 		qemuCmd = append(qemuCmd, "-bios", "/vm/QEMU_EFI_ARM.fd")
+		// arm virt has only one UART, setup virtio-serial to add more /dev/hvcX
+		qemuCmd = append(qemuCmd,
+			"-chardev", "stdio,id=virtio-console",
+			"-device", "virtconsole,chardev=virtio-console",
+		)
+	} else {
+		// on amd we have multiple UART ports so we can just use serial stdio
+		qemuCmd = append(qemuCmd, "-serial", "stdio")
 	}
 
 	for _, disk := range vmSpec.Disks {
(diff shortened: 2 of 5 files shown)


neonvm-runner/cmd/main.go
```diff
@@ -625,9 +631,10 @@ func newConfig(logger *zap.Logger) *Config {
 		appendKernelCmdline:  "",
 		skipCgroupManagement: false,
 		diskCacheSettings:    "cache=none",
-		memoryProvider:       "", // Require that this is explicitly set. We'll check later.
-		autoMovableRatio:     "", // Require that this is explicitly set IFF memoryProvider is VirtioMem. We'll check later.
-		cpuScalingMode:       "", // Require that this is explicitly set. We'll check later.
+		memoryProvider:       "",             // Require that this is explicitly set. We'll check later.
+		autoMovableRatio:     "",             // Require that this is explicitly set IFF memoryProvider is VirtioMem. We'll check later.
+		cpuScalingMode:       "",             // Require that this is explicitly set. We'll check later.
+		architecture:         runtime.GOARCH, // arm64, amd64
```
sharnoff, line 637: super tiny nit: can we put this elsewhere in the struct so it doesn't expand the diff?

neonvm-runner/cmd/main.go
```diff
@@ -1104,6 +1111,28 @@ func runQEMU(
...
+	case architectureArm64:
+		return qemuBinArm64
+	case architectureAmd64:
+		return qemuBinX8664
+	default:
+		panic(fmt.Errorf("unknown architecture %s", architecture))
+	}
+}
+
+func getMachineType(architecture string) string {
```
sharnoff, line 1125: It'd be good to have a comment explaining the situation here -- why these particular machine types?

tests/e2e/autoscaling.cpu-sys-fs-state-scaling/00-create-vm.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       min: 1
       max: 5
       use: 1
-    memoryProvider: DIMMSlots
+    memoryProvider: VirtioMem
```
sharnoff, line 40: Should this change be merged into the PR this is based on?

(+ same question for the other changes to this test)

sharnoff (commented):
Looks good, AFAICT

neonvm-runner/cmd/main.go
mikhail-sakhnov: I put it here as all other fields have for the sake of similarity.  Happy to move all three comments to the structure definition though.

neonvm-runner/cmd/main.go
mikhail-sakhnov: tbh there are no particular reason. for q35 it is probably historical reason and virt is the most generic arm platform which we use in runners already.

sharnoff: perhaps @cicdteam can add some context on why q35?

tests/e2e/autoscaling.cpu-sys-fs-state-scaling/00-create-vm.yaml
sharnoff: Fair enough, makes sense

neonvm-runner/cmd/main.go
cicdteam: Sorry, I just saw this discussion.

In the x86_64 architecture, QEMU has two groups of emulated machines - `pc-q35` and `pc-i440fx` (yes, there are a couple of others, like `microvm` or `none`, but this is not our case). The `q35` series is more recent, so it was chosen. In fact, there are no other options for x86_64. 

FYI: `-machine q35` is alias of `-machine pc-q35-9.0` (in QEMU v9.0.1),by other words - most recent hardware.

mikhail-sakhnov: added some clarifications on that

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	sysruntime "runtime"
```
Omrigan, line 29: Why non-standard name?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1222,7 +1222,7 @@ func affinityForVirtualMachine(vm *vmv1.VirtualMachine) *corev1.Affinity {
 					{
 						Key:      "kubernetes.io/arch",
 						Operator: "In",
-						Values:   []string{"amd64"},
```
Omrigan, line 1225: What's the reasoning behind this change,. why does VM's affinity must match controller's architecture?

If we might have heterogeneous clusters, maybe we need to specify the architecture explicitly in the spec?

neonvm-vxlan-controller/Dockerfile
Omrigan, line 1: vxlan-controller changes should be a separate commit.

Omrigan (commented):
Looking good!

Perhaps, PR should be named something like "enable e2e tests on arm", with multiple commits, 1 per each component?

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: We already can specify it through the VM spec, by passing affinity. 

I think at the moment using by default the same arch as the controller might be a safe assumption. Do you have any other options in your mind?

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: The same file also has `"k8s.io/apimachinery/pkg/runtime"` import

mikhail-sakhnov:
> Perhaps, PR should be named something like "enable e2e tests on arm", with multiple commits, 1 per each component?

@Omrigan I am hesitant towards that title, because the PR does what it says - enables vm-runner to run ARM vms. There is one test modified to reflect that change. Plus there actually separate issue and PR about e2e tests on ARM.

neonvm-vxlan-controller/Dockerfile
mikhail-sakhnov: I have feeling that it is a separate commit at the moment https://github.com/neondatabase/autoscaling/commit/59ccc8965da891958f09606a5f2fe6888f351b23

Omrigan: Agreed to extract it: https://github.com/neondatabase/autoscaling/pull/1160.

pkg/neonvm/controllers/vm_controller.go
Omrigan: > Maybe it worth to discuss whether we want to have heterogeneous clusters.

The existence of https://github.com/neondatabase/autoscaling/pull/1123 hints toward yes. In that case, we need to align on how it should be determined where to schedule the VM.

mikhail-sakhnov: @Omrigan do you have any suggestions?

mikhail-sakhnov:
@Omrigan @sharnoff what is the current state of review?

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: we agreed verbally with @sharnoff to keep it as current PR has, but change PR #1083 with explicit spec.arch field

Omrigan (approved):

neonvm-runner/Dockerfile
```diff
@@ -21,12 +20,16 @@ RUN apk add --no-cache \
     busybox-extras \
     e2fsprogs \
     qemu-system-x86_64 \
+    qemu-system-aarch64 \
```
sharnoff, line 23: question: What is the expected behavior when installing either of these on the other architecture?

(is it worth only installing one, depending on target arch?)

neonvm-runner/cmd/main.go
```diff
@@ -851,6 +861,19 @@ func buildQEMUCmd(
...
+		qemuCmd = append(qemuCmd, "-bios", "/vm/QEMU_EFI_ARM.fd")
+		// arm virt has only one UART, setup virtio-serial to add more /dev/hvcX
+		qemuCmd = append(qemuCmd,
+			"-chardev", "stdio,id=virtio-console",
+			"-device", "virtconsole,chardev=virtio-console",
+		)
+	} else {
+		// on amd we have multiple UART ports so we can just use serial stdio
+		qemuCmd = append(qemuCmd, "-serial", "stdio")
+	}
```
sharnoff, line 875: suggestion: let's use `switch` for these and panic in the default case, so it's more clear which architecture gets which behavior?

(+ same suggestion for other occurrences of this)

neonvm-runner/cmd/main.go
```diff
@@ -625,9 +635,10 @@ func newConfig(logger *zap.Logger) *Config {
 		appendKernelCmdline:  "",
 		skipCgroupManagement: false,
 		diskCacheSettings:    "cache=none",
-		memoryProvider:       "", // Require that this is explicitly set. We'll check later.
-		autoMovableRatio:     "", // Require that this is explicitly set IFF memoryProvider is VirtioMem. We'll check later.
-		cpuScalingMode:       "", // Require that this is explicitly set. We'll check later.
+		memoryProvider:       "",
+		autoMovableRatio:     "",
+		cpuScalingMode:       "",
+		architecture:         runtime.GOARCH,
```
sharnoff, line 641: question: Is this only ever equal to `runtime.GOARCH` ? Do we plan to keep it that way? (if so: can we make assumptions that e.g. we're only ever executing `qemu-system-x86_64` from `amd64` go code?)

sharnoff (commented):
Left a couple comments. I may have some follow-up questions pending the outcome of the `runtime.GOARCH` one, but should be nothing major -- broadly looks good

neonvm-runner/cmd/main.go
mikhail-sakhnov: Yes, and I think I got where are you heading to - why not to use `runtime.GOARCH` everywhere?

mikhail-sakhnov: >But also — why install both qemu-system-x86_64 and qemu-system-aarch64 if we're only going to need one? (and then if we like, we can set up some symlink to refer to it, so we don't need to match the name inside the go code)



Basically to simplify the image builds, since this PR was introduced way before crossplatform images. I can update PR with crossplatform images to use only the correct version of the qemu.

But in general, cfg.architecture is used in a couple of places (kernel arguments, qemu arguments (not binary name), adding firmware) and I don't want to make it using `runtime.GOARCH` everywhere instead. 

so: 
- I will update the PR with #1138 to have only one qemu binary
- I keep cfg.architecture here, for the sake of using it in different places

How does it sound?

neonvm-runner/Dockerfile
mikhail-sakhnov: As mentioned in another thread, it was done this way to simplify builds. There is a PR to make proper cross-platform images, which might be extended with changes to install only platform-specific qemu package.

neonvm-runner/cmd/main.go
mikhail-sakhnov: changed to switch

neonvm-runner/cmd/main.go
sharnoff: Fair enough, sounds like a plan :)

sharnoff (approved):

PR #1122, opened 2024-10-23
neonvm-controller: Remove unused auth proxy rbac

sharnoff:
We used to run kube-auth-proxy as a sidecar alongside neonvm-controller. That was removed in #825 - I guess the RBAC for it was forgotten.

This showed up because the v0.36.0 release YAMLs have a regression where the RBAC for this unused proxy refers to 'system' and 'controller' instead of 'neonvm-system' and 'neonvm-controller'.


mikhail-sakhnov (approved):
looks good, none of the deleted entities are used in the manifests

PR #1128, opened 2024-10-31
agent/revsource: fix out-of-bounds error

Omrigan:
Index equal to length should also result in an early return.


sharnoff (approved):
nice catch!

PR #1129, opened 2024-11-04
agent/core: Allow setting goal CU in tests

sharnoff:
This is kind of a second take on #737, and a pre-req to #729 so that we can freely change how metrics are interpreted without needing to rewrite our unit tests in 'pkg/agent/core/state_test.go'.


pkg/agent/core/state.go
```diff
@@ -103,12 +102,12 @@ type LogConfig struct {
...
 // decisions
-type State struct {
-	internal state
+type State[A AlgorithmState] struct {
+	internal state[A]
 }
 
 // one level of indirection below State so that the fields can be public, and JSON-serializable
-type state struct {
+type state[A AlgorithmState] struct {
```
Omrigan, line 110: Is there a reason why this has to be a type parameter, compared with just having an interface-typed field?

pkg/agent/core/state.go
```diff
@@ -130,9 +129,7 @@ type state struct {
 	// NeonVM records all state relevant to the NeonVM k8s API
 	NeonVM neonvmState
 
-	Metrics *SystemMetrics
-
-	LFCMetrics *LFCMetrics
+	Metrics A
```
Omrigan, line 132: Shouldn't we rename it here as well?

sharnoff: Yup, good catch, thanks. Did that rename in 8d053b1cc3c57202ed5e1fcceb1e2ed26d67dc2f

pkg/agent/core/state.go
sharnoff: Yeah, otherwise callers of `(*core.State).UpdateAlgorithmState()` would have to downcast to the specific type in order to do the updates — possible, but I figured why both downcasting when we can just use the exact type

Omrigan: Can you take a quick look at https://github.com/neondatabase/autoscaling/pull/1140? I wanted to try to implement the same change, but without introducing generics, so that the impact is minimal.

sharnoff: Ah, I see, the idea there is that the algorithm doesn't store the metrics? So, the "manual CU" algorithm would have shared state that's updated externally?

Omrigan: > Ah, I see, the idea there is that the algorithm doesn't store the metrics? So, the "manual CU" algorithm would have shared state that's updated externally?

Yes and yes. These are the consequences of doing the minimal possible changes.

sharnoff: Ok, I see. My two cents: I think the changes here aren't _too_ bad (it's mostly in tests), and having shared mutable state makes me nervous (specifically: changing things out from underneath another object) — but maybe that's just a holdover from my Rust programming :sweat_smile:

PR #1131, opened 2024-11-05
neonvm-controller: refactor doReconcile

Omrigan:
Best viewed by-commit. 

Will be merged with rebase.

Part of the #733.

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -311,6 +311,42 @@ func (r *VMReconciler) updateVMStatusMemory(
 	}
 }
 
+func (r *VMReconciler) acquireOverlayIP(ctx context.Context, vm *vmv1.VirtualMachine) error {
+	if vm.Spec.ExtraNetwork == nil || !vm.Spec.ExtraNetwork.Enable || len(vm.Status.ExtraNetIP) != 0 {
+		// If the VM has extra network disabled or already has an IP, do nothing.
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -311,6 +311,41 @@ func (r *VMReconciler) updateVMStatusMemory(
 	}
 }
 
+func (r *VMReconciler) acquireOverlayIP(ctx context.Context, vm *vmv1.VirtualMachine) error {
+	if vm.Spec.ExtraNetwork == nil || !vm.Spec.ExtraNetwork.Enable || len(vm.Status.ExtraNetIP) != 0 {
```
sharnoff, line 316: IIUC, this is equivalent to: "if the extra network is disabled OR we already have an extra IP address, nothing to do" , right?

(worth adding that as a comment?)

sharnoff (approved):
Looks good, but I think d5defa1d7ff24f7f07c94bd31f6e7eefc97f72f9 ("extract createPod") may leak ssh secrets in edge cases. Can give a more thorough example of what I mean later, but in the meantime maybe let's merge the other commits?

sharnoff:
@Omrigan before merging, can you rename the commits to append the PR number?

Omrigan:
> @Omrigan if you're going to rebase-and-merge, could you rename the commits to append the PR number?

Right, totally forgot! I guess it is 3 strikes to automate it, first take is at https://github.com/neondatabase/autoscaling/pull/1152/files.

PR #1135, opened 2024-11-07
Upgrade k8s deps to 1.29

edude03:
It's that time again, where we upgrade our kubernetes clusters from v1.28 => v1.29 (with the eventual goal of being at 1.31). Since the version of the neon deps have to be within 1 version of the kubernetes version, I'm upgrading the matching dependencies. 

[See epic](https://github.com/neondatabase/cloud/issues/18098)


autoscale-scheduler/cmd/main.go
```diff
@@ -65,7 +67,9 @@ func runProgram(logger *zap.Logger) (err error) {
 	// everything fit nicely, we'll redirect it to zap as well.
 	redirectKlog(logger.Named("klog"))
 
-	constructor := plugin.NewAutoscaleEnforcerPlugin(ctx, logger, conf)
+	constructor := func(ctx context.Context, configuration runtime.Object, f framework.Handle) (framework.Plugin, error) {
+		return plugin.NewAutoscaleEnforcerPlugin(ctx, logger, conf)(configuration, f)
+	}
 	command := app.NewSchedulerCommand(app.WithPlugin(plugin.Name, constructor))
```
edude03, line 73: So, apparently the function signature changed somewhere with the upgrade - such that if I'm understanding correctly, `withPlugin` is expected to take a constructor that takes a `ctx`, not just `configuration` and `framework.Handle`. Not sure what the right fix for that is, so I'm trying wrapping it. This is also the solution copilot suggested, but it's probable we're both dumb 😆

edude03: Yeah! Looks like you're right! And since it's only used in this one spot it was easy to clean up (and even remove the unused `_obj` parameter)

go.mod
```diff
@@ -165,64 +164,65 @@ require (
...
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.1.2 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.28.0 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
-	sigs.k8s.io/yaml v1.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+require github.com/matttproud/golang_protobuf_extensions/v2 v2.0.0 // indirect
```
kszafran, line 228: Not sure why another `require` section got created. Please move this into the block above.

edude03: Seems like the go.mod file requires more hands on mantainence than I expected. Any guidelines I should check out?

kszafran: Most of the time `go get <something>` just works. Idk why this ended up in a separate section.

sharnoff (commented):
Broadly LGTM, thanks! Any idea why there's so many removed dependencies in `go.sum`?

edude03:
> Broadly LGTM, thanks! Any idea why there's so many removed dependencies in `go.sum`?

LOTS of duplication. I don't know if that's normal, but it references 39 versions of go?

kszafran:
My guess it that nobody ran `go mod tidy` before? Sounds strange, but it could explain it. I think `go.sum` just keeps growing if you don't run `go mod tidy`, keeping old versions and what not.

go.mod
```diff
@@ -165,64 +163,65 @@ require (
...
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.1.2 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.28.0 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
-	sigs.k8s.io/yaml v1.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+require github.com/mattn/go-isatty v0.0.16 // indirect
```
kszafran, line 227: Still a rogue dependency outside of the above `require` section 😉

sharnoff:
A few more things, sorry:

1. If you plan to cut a release from this branch _before merging to main_, could you rebase this PR onto the previous release so that we don't accidentally get changes we're not expecting?
2. This PR basically affects all components, right? Could you remove "agent:" from the PR title to reflect that?
3. Could you add a description to this PR? e.g., linking to a relevant discussion, issue, etc.

Other than those and the rogue dependency that @kszafran highlighted, seems good to me.

edude03:
> If you plan to cut a release from this branch before merging to main

I don't _think_ I need to, I assume one 1-3 are resolved we can merge this? Or do we need to stick to 1.28 deps for awhile?

Omrigan (approved):

PR #1138, opened 2024-11-12
ci: build images for arm and amd

mikhail-sakhnov:
Introduce matrix strategy builds to build images for arm and amd64. 
After matrix execution merge artefacts using docker manifest tool

#1089


.github/workflows/build-images.yaml
```diff
@@ -30,26 +30,39 @@ on:
...
+        description: 'Architecture to build for'
+        type: string
+        required: false
+        default: 'amd64'
+      matrix-step-name:
+        required: false
+        type: string
+      matrix-key:
+        required: false
+        type: string
```
bayandin, line 43: I see it's used in a workaround later, could you please add some more information about it? (maybe we can do it without that)

.github/workflows/build-images.yaml
```diff
@@ -221,12 +252,12 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: .
-          platforms: linux/amd64
           push: true
+          platforms: linux/${{ inputs.arch }}
```
bayandin, line 256: I think we don't need to set platform, as we build images on native platforms

.github/workflows/build-images.yaml
```diff
@@ -245,41 +276,41 @@ jobs:
...
         with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/${{ inputs.arch }}
           push: true
           file: autoscale-scheduler/Dockerfile
           cache-from: type=registry,ref=cache.neon.build/autoscale-scheduler:cache
           cache-to: ${{ github.ref_name == 'main' && 'type=registry,ref=cache.neon.build/autoscale-scheduler:cache,mode=max' || '' }}
-          tags: ${{ needs.tags.outputs.scheduler }}
+          tags: ${{ steps.tags.outputs.scheduler}}
```
bayandin, line 313:
```suggestion
          tags: ${{ steps.tags.outputs.scheduler }}
```

.github/workflows/build-test-vm.yaml
```diff
@@ -38,25 +49,19 @@ defaults:
...
-      IMG_DAEMON: ${{ needs.tags.outputs.daemon }}
+    runs-on: ${{ fromJson(format('["self-hosted", "{0}"]', inputs.arch == 'arm64' && 'large-arm64' || 'large')) }}
+    outputs:
+      vm-postgres-16-bullseye: ${{ steps.tags.outputs.vm-postgres-16-bullseye }}
+      daemon: ${{ steps.tags.outputs.daemon }}
     steps:
+      # tags converted to be a step and moved here to be in the same strategy contextt
+      - id: tags
+        run: |
+          echo "vm-postgres-16-bullseye=${{ env.IMG_POSTGRES_16_BULLSEYE }}-${{inputs.arch}}:${{ inputs.tag }}" | tee -a $GITHUB_OUTPUT
```
bayandin, line 63: Usually, we prefer this style (one day, we'll have a linter for that)

```suggestion
          echo "vm-postgres-16-bullseye=${{ env.IMG_POSTGRES_16_BULLSEYE }}-${{ inputs.arch }}:${{ inputs.tag }}" | tee -a $GITHUB_OUTPUT
```

Could you also check the other places in the PR, please?

.github/workflows/build-test-vm.yaml
```diff
@@ -65,15 +70,34 @@ jobs:
...
+      #   with:
+      #     context: .
+      #     push: false
+      #     platforms: linux/${{ inputs.arch }}
+      #     file: neonvm-daemon/Dockerfile
+      #     cache-from: type=registry,ref=cache.neon.build/neonvm-daemon:cache
+      #     cache-to: ${{ github.ref_name == 'main' && 'type=registry,ref=cache.neon.build/neonvm-daemon:cache,mode=max' || '' }}
+      #     tags: ${{ steps.tags.outputs.daemon }}
+      #     build-args: |
+      #       GO_BASE_IMG=${{ env.GO_BASE_IMG }}
```
bayandin, line 90: Can we remove that?

.github/workflows/vm-kernel.yaml
```diff
@@ -185,14 +190,15 @@ jobs:
...
+          target: ${{format('kernel_{0}', inputs.arch)}}
           # Push kernel image only for scheduled builds or if workflow_dispatch/workflow_call input is true
           push: true
           pull: true
           file: neonvm-kernel/Dockerfile.kernel-builder
-          cache-from: type=registry,ref=cache.neon.build/vm-kernel:cache
-          cache-to: ${{ github.ref_name == 'main' && 'type=registry,ref=cache.neon.build/vm-kernel:cache,mode=max' || '' }}
+          cache-from: ${{format('type=registry,ref=cache.neon.build/vm-kernel:{0}cache', inputs.arch)}}
+          # Cache the kernel image for future builds, temporary from all branches
+          cache-to: ${{ format('type=registry,ref=cache.neon.build/vm-kernel:{0}cache,mode=max', inputs.arch) || '' }}
```
bayandin, line 203: We usually add "arch" part to image name itself:

```suggestion
          cache-from: ${{format('type=registry,ref=cache.neon.build/vm-kernel-{0}:cache', inputs.arch)}}
          # Cache the kernel image for future builds, temporary from all branches
          cache-to: ${{ format('type=registry,ref=cache.neon.build/vm-kernel-{0}:cache,mode=max', inputs.arch) || '' }}
```

cluster-autoscaler/Dockerfile
```diff
@@ -30,7 +30,15 @@ RUN cd autoscaler/cluster-autoscaler \
 # This is adapted from CA's Dockerfile.amd64, here:
 # https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.24.1/cluster-autoscaler/Dockerfile.amd64
 
-FROM gcr.io/distroless/static:nonroot-amd64
+
+# NB: two build stages, one for each architecture, because I wasn't able to use variable substitution in FROM statements
```
bayandin, line 34: Hmm, I wonder if we do something different here, usually it works like this:

```dockerfile
ARG ARCH
FROM gcr.io/distroless/static:nonroot-${ARCH}
```

mikhail-sakhnov: TIL, that Dockerfile args can have different scopes. Thanks for pointing out.

>Build arguments declared in the global scope of a Dockerfile aren't automatically inherited into the build stages. >They're only accessible in the global scope.

https://docs.docker.com/build/building/variables/#scoping

.github/workflows/build-images.yaml
```diff
@@ -334,6 +340,92 @@ jobs:
...
+          TAG: ${{ inputs.tag }}
+          BUILD_CLUSTER_AUTOSCALER: ${{ format('{0}', inputs.build-cluster-autoscaler) }}
+          UPLOAD_TO_ECR: ${{ format('{0}', inputs.upload-to-ecr) }}
+        run: |
+          images=("${{ env.IMG_CONTROLLER }}" \
+          "${{ env.IMG_VXLAN_CONTROLLER }}" \
+          "${{ env.IMG_RUNNER }}" \
+          "${{ env.IMG_SCHEDULER }}" \
+          "${{ env.IMG_AUTOSCALER_AGENT }}" \
+          "${{ env.IMG_DAEMON }}")
```
bayandin, line 389: In `run:` you can use regular `${VAR}` instead of `${{ env.VAR }}` — env variables are added as env variables 😄 
I know we don't usually do it here, so jfyi, and we can keep it as is for now

.github/workflows/report-workflow-stats.yml
```diff
@@ -18,6 +18,7 @@ on:
 
 jobs:
   gh-workflow-stats:
+    if: false # drop me, reduce noise while experimenting
```
bayandin, line 21: Don't forget to revert it back 😉

cluster-autoscaler/Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN cd autoscaler/cluster-autoscaler \
 # This is adapted from CA's Dockerfile.amd64, here:
 # https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.24.1/cluster-autoscaler/Dockerfile.amd64
```
bayandin, line 31: Should we adjust the comment?

bayandin (approved):
The approach LGTM, I left several comments and nits that would be nice to fix before merging the PR itself.

Nevertheless, I didn't dive too deep into details, if it works — it works

.github/workflows/build-images.yaml
```diff
@@ -334,6 +340,92 @@ jobs:
...
+        with:
+          username: ${{ secrets.NEON_DOCKERHUB_USERNAME }}
+          password: ${{ secrets.NEON_DOCKERHUB_PASSWORD }}
+
+      - name: Create multi-arch image
+        env:
+          ARCHS: ${{ join(fromJson(inputs.archs), ' ') }}
+          TAG: ${{ inputs.tag }}
+          BUILD_CLUSTER_AUTOSCALER: ${{ format('{0}', inputs.build-cluster-autoscaler) }}
+          UPLOAD_TO_ECR: ${{ format('{0}', inputs.upload-to-ecr) }}
```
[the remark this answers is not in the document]
mikhail-sakhnov: I think it is done like that in other places, there is a comment

>nb: use format(..) to catch both inputs.skip = true AND inputs.skip = 'true'.

.github/workflows/build-images.yaml
```diff
@@ -245,41 +250,41 @@ jobs:
         uses: docker/build-push-action@v6
```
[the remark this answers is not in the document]
mikhail-sakhnov: done

.github/workflows/build-images.yaml
mikhail-sakhnov: done

.github/workflows/build-images.yaml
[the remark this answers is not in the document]
bayandin: We do it when if have an `if`-condition in yaml; in shell script, it's going to be a string anyway.
Btw, didn't find the usage of `UPLOAD_TO_ECR`

.github/workflows/build-images.yaml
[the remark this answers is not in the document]
sharnoff: @mikhail-sakhnov can we do this in a separate PR? It's worth calling out explicitly, and AFIACT doesn't have any inter-dependencies with this work

.github/workflows/build-test-vm.yaml
```diff
@@ -22,10 +22,15 @@ on:
         type: boolean
         required: false
         default: false
+      archs:
+        description: 'Architecture to build for'
+        type: string
+        required: false
+        default: '["amd64","arm64"]'
```
sharnoff, line 29: question: Should these match the format of `build-images.yaml`

.github/workflows/build-test-vm.yaml
```diff
@@ -38,25 +43,22 @@ defaults:
...
-      IMG_DAEMON: ${{ needs.tags.outputs.daemon }}
+    strategy:
+      matrix:
+        arch: ${{ fromJson(inputs.archs) }}
+    runs-on: ${{ fromJson(format('["self-hosted", "{0}"]', matrix.arch == 'arm64' && 'huge-arm64' || 'large')) }}
+    outputs:
+      vm-postgres-16-bullseye: ${{ steps.tags.outputs.vm-postgres-16-bullseye }}
+      daemon: ${{ steps.tags.outputs.daemon }}
     steps:
+      # tags converted to be a step and moved here to be in the same strategy contextt
```
sharnoff, line 57: > converted to be a step and moved here to be in the same strategy contextt

suggestion: Let's remove this comment? I don't think it provides any context outside the process of reviewing this PR (unless I'm missing something -- happy to be convinced otherwise!)

.github/workflows/build-test-vm.yaml
```diff
@@ -65,15 +67,21 @@ jobs:
...
+        env:
+          IMG_DAEMON: ${{ steps.tags.outputs.daemon }}
 
       - name: upload vm-builder
         if: ${{ format('{0}', inputs.upload-vm-builder) == 'true' }}
         uses: actions/upload-artifact@v4
         with:
           name: vm-builder
-          path: bin/vm-builder
+          path: ${{ format('bin/vm-builder-{0}', matrix.arch) }}
```
sharnoff, line 84: question: Does this mean we need to change `release.yaml`'s handling of vm-builder?

.github/workflows/build-test-vm.yaml
```diff
@@ -86,9 +94,55 @@ jobs:
           username: ${{ secrets.NEON_DOCKERHUB_USERNAME }}
           password: ${{ secrets.NEON_DOCKERHUB_PASSWORD }}
 
-      - name: build ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
+      - name: build ${{ steps.tags.outputs.vm-postgres-16-bullseye  }}
```
sharnoff, line 97:
```suggestion
      - name: build ${{ steps.tags.outputs.vm-postgres-16-bullseye }}
```

.github/workflows/build-test-vm.yaml
```diff
@@ -86,9 +94,55 @@ jobs:
...
         run: |
-          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ needs.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ needs.tags.outputs.daemon }} -target-arch linux/${TARGET_ARCH}
-      - name: docker push ${{ needs.tags.outputs.vm-postgres-16-bullseye }}
+          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ steps.tags.outputs.vm-postgres-16-bullseye  }} -daemon-image ${{ steps.tags.outputs.daemon }} -target-arch linux/${{ matrix.arch }}
```
sharnoff, line 99:
```suggestion
          ./bin/vm-builder -src postgres:16-bullseye -spec tests/e2e/image-spec.yaml -dst ${{ steps.tags.outputs.vm-postgres-16-bullseye }} -daemon-image ${{ steps.tags.outputs.daemon }} -target-arch linux/${{ matrix.arch }}
```

.github/workflows/build-test-vm.yaml
```diff
@@ -86,9 +94,55 @@ jobs:
...
+
+      # vm-builder doesn't attach platform annotation to the image, so we need to do it manually
```
sharnoff, line 101: question: Can we "just" change vm-builder to do that? seems silly to work around our own code :sweat_smile:

.github/workflows/build-test-vm.yaml
```diff
@@ -86,9 +94,55 @@ jobs:
...
+      vm-postgres-16-bullseye: ${{ steps.merge-vm-images.outputs.vm-postgres-16-bullseye }}
+    env:
+      ARCHS: ${{ join(fromJson(inputs.archs), ' ') }}
+      TAG: ${{ inputs.tag }}
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v4
+
+      - name: set custom docker config directory
+        uses: neondatabase/dev-actions/set-docker-config-dir@6094485bf440001c94a94a3f9e221e81ff6b6193
```
sharnoff, line 125: question: Should we remove this? (cc @bayandin)

Saw the following warning while reviewing:

> Unnecessary usage of `.github/actions/set-docker-config-dir`
>
> No need to use `.github/actions/set-docker-config-dir` action on GitHub-hosted runners

.github/workflows/e2e-test.yaml
```diff
@@ -79,7 +81,11 @@ jobs:
       matrix:
         cluster:
           - ${{ inputs.cluster || 'k3d' }}
-    runs-on: [ self-hosted, gen3, large ]
+        # run tests on amd64 only, since scope of the PR is to build images, there is separate issue for e2e tests
```
sharnoff, line 84: > since scope of the PR is to build images

suggestion: Same thing as before — let's tune this comment so it makes more sense to someone reading the code without the context of this PR? (e.g., "run tests on amd64 only; see \<PR number\> to enable for arm64 as well")

.github/workflows/test.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
         id: html-artifact
         env:
           COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
-        if: ${{ !cancelled() && (steps.changed-files.outputs.any_changed == 'true' && github.event_name == 'pull_request')}}
+        if: ${{ !cancelled() && (steps.changed-files.outputs.any_changed == 'true' && github.event_name == 'pull_request') }}
```
sharnoff, line 61: Unrelated diff? +1 on the change, let's do it in a separate PR?

.github/workflows/vm-kernel.yaml
```diff
@@ -28,24 +28,33 @@ on:
         type: boolean
         required: false
         default: false
+      archs:
+        description: 'Architectures to build for'
+        type: string
+        required: true
```
sharnoff, line 34: question: The other workflows have a default here -- why not this one?

.github/workflows/vm-kernel.yaml
```diff
@@ -28,24 +28,33 @@ on:
...
+  VM_KERNEL_IMAGE: neondatabase/vm-kernel
+
 
 defaults:
   run:
     shell: bash -euo pipefail {0}
 
 jobs:
   setup-build-vm-kernel-image:
+    # todo: looks redundant, check with manual combinations of inputs and delete strategy if it's not needed
```
sharnoff, line 51: todo before merging? or will this be an issue somewhere?

.github/workflows/vm-kernel.yaml
```diff
@@ -126,13 +135,56 @@ jobs:
           docker tag ${CACHED_IMAGE} ${VM_KERNEL_IMAGE}:${NEW_TAG}
           docker push ${VM_KERNEL_IMAGE}:${NEW_TAG}
 
-  build-vm-kernel-image:
+      - name: debug output
```
sharnoff, line 138: should be removed before merging?

sharnoff (commented):
Lots of comments, sorry :sweat_smile:

Got most of the way through, didn't quite finish looking at `vm-kernel.yaml`

Please also add a high-level description of the changes to the PR description :pray: (e.g., "changes XYZ workflows so that A was split into BC and ...") — there's a lot of moving pieces here!

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: Yes, it should be dropped here (should be kept only for self-hosted runners if I recall correctly)

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: https://github.com/neondatabase/autoscaling/issues/1250

.github/workflows/test.yaml
mikhail-sakhnov: #1250

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: fixed

.github/workflows/build-images.yaml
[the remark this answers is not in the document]
mikhail-sakhnov: #1251

.github/workflows/build-images.yaml
```diff
@@ -244,84 +255,92 @@ jobs:
...
         with:
+          provenance: false
+          sbom: false
           context: cluster-autoscaler
-          platforms: linux/amd64
           push: true
-          tags: ${{ needs.tags.outputs.cluster-autoscaler }}
-          cache-from: type=registry,ref=cache.neon.build/cluster-autoscaler-neonvm:cache
-          cache-to: ${{ github.ref_name == 'main' && 'type=registry,ref=cache.neon.build/cluster-autoscaler-neonvm:cache,mode=max' || '' }}
+          target: ${{ format('cluster_autoscaler_{0}', matrix.arch) }}
```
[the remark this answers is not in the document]
mikhail-sakhnov: dropped since this PR has changed in this area.

.github/workflows/e2e-test.yaml
mikhail-sakhnov: droped

.github/workflows/vm-kernel.yaml
mikhail-sakhnov: added

.github/workflows/vm-kernel.yaml
mikhail-sakhnov: droped

.github/workflows/build-images.yaml
```diff
@@ -334,6 +353,91 @@ jobs:
...
+      cluster-autoscaler: ${{ steps.tags.outputs.cluster-autoscaler }}
+    needs: [build]
+    runs-on: ubuntu-latest
+    steps:
+      - name: set custom docker config directory
+        uses: neondatabase/dev-actions/set-docker-config-dir@6094485bf440001c94a94a3f9e221e81ff6b6193
+
+      - uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: network=host
```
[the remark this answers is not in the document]
mikhail-sakhnov: looks like it is needed to use localhost:5000, which is used in the go-base image.

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: Can we do it as a separate thing? The current vm-builder implementation knows nothing about manifests and it would take some time to rewrite it to act like `docker manifest` command.

.github/workflows/build-test-vm.yaml
mikhail-sakhnov: fixed

go.mod
```diff
@@ -48,7 +48,7 @@ require (
 	github.com/containernetworking/cni v1.1.1
 	github.com/coreos/go-iptables v0.6.0
 	github.com/digitalocean/go-qemu v0.0.0-20220826173844-d5f5e3ceed89
-	github.com/docker/cli v25.0.3+incompatible
+	github.com/docker/cli v24.0.9+incompatible
```
sharnoff, line 51: question: Out of curiosity, why was this downgraded?

.github/workflows/vm-kernel.yaml
```diff
@@ -153,46 +201,76 @@ jobs:
...
-          KERNEL_VERSION_TAG: ${{ inputs.tag || steps.get-kernel-version.outputs.VM_KERNEL_VERSION }}
-          CACHE_TAG: ${{ needs.setup-build-vm-kernel-image.outputs.kernel-cache-tag }}
-        run: |
-          # A comma-separated list of tags
-          TAGS="${VM_KERNEL_IMAGE}:${KERNEL_VERSION_TAG}"
-          TAGS="${VM_KERNEL_IMAGE}:${CACHE_TAG},${TAGS}"
-          TAGS="${VM_KERNEL_IMAGE}:${GITHUB_RUN_ID},${TAGS}"
-
-          echo "tags=${TAGS}" >> $GITHUB_OUTPUT
+          VM_KERNEL_IMAGE: neondatabase/vm-kernel
```
sharnoff, line 219: nit: let's place `env` above `run` ?

.github/workflows/vm-kernel.yaml
```diff
@@ -153,46 +201,76 @@ jobs:
...
+      - name: Process kernel image tags
+        id: generate-full-image-tags
         run: |
-          linux_config=$(ls neonvm-kernel/linux-config-*)  # returns something like "neonvm-kernel/linux-config-amd64-6.1.63"
-          kernel_version=${linux_config##*-}               # returns something like "amd64-6.1.63"
-
-          echo VM_KERNEL_VERSION=$kernel_version >> $GITHUB_OUTPUT
+          PROCESSED_TAGS=""
+          for tag in ${TAGS//,/ }; do
+            full_image_tag="$VM_KERNEL_IMAGE-${{ matrix.arch }}:$(echo "$tag" | xargs)"
```
sharnoff, line 209: suggestion: I spent a couple minutes trying to figure out what these two lines do. I'm sure I could spend more, but perhaps that's an indication of complexity we should either reduce or explain. Can you write a comment explaining what this does? or if you have any ideas to simplify, that'd be welcome, too :sweat_smile:

For reference, things I wasn't sure about:

1. what `${TAGS//,/ }` does
2. why `| xargs`
3. why we need to process these in the first place

sharnoff (commented):
Ok, finished review of `vm-kernel.yaml` and remaining items :)

go.mod
mikhail-sakhnov: Good catch, by mistake!

mikhail-sakhnov:
@sharnoff do we have anything left unreviewed here?

sharnoff (approved):
looks good!

bayandin (approved):
I just noticed that GitHub doesn't allow merging the PR because it is missing the job `e2e-tests (k3d)`. 
It happened because now we have `e2e-tests (k3d, amd64)` and `e2e-tests (k3d, arm64)` jobs 🤷 

I propose the following:
- In another PR create `conclusion` job that depends on `e2e-tests` (like [this](https://github.com/neondatabase/neon/blob/da9c101939eded0063e78f32ea53f0a7e6a44aa1/.github/workflows/build_and_test.yml#L1347-L1384))
- Merge that PR
- Change branch protection rules (I can do that for you)
- Add `conclusion` job to this PR (or just pull the fresh main into it)

PR #1141, opened 2024-11-15
WIP: e2e in ci

mikhail-sakhnov:
tmp branch to trigger e2e tests.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -81,7 +81,7 @@ jobs:
       matrix:
         cluster:
           - ${{ inputs.cluster || 'k3d' }}
-        arch: [ amd64 ]
+        arch: [ amd64, arm64 ]
 
     runs-on: ${{ fromJson(format('["self-hosted", "{0}"]', matrix.arch == 'arm64' && 'huge-arm64' || 'large')) }}
 
@@ -199,6 +199,9 @@ jobs:
           docker pull "$TEST_IMAGE"
           docker image tag "$TEST_IMAGE" "$IMG_E2E_TEST"
           make load-example-vms
+
+      - run: make arm_patch_e2e
+        if: matrix.arch == 'arm64'
 
       - run: make e2e
         timeout-minutes: 15
--- a/Makefile
+++ b/Makefile
@@ -505,6 +505,8 @@ endif
 KUBECTL ?= $(LOCALBIN)/kubectl
 KUBECTL_VERSION ?= v1.30.7
 
+
+YQ ?= $(LOCALBIN)/yq
 ETCD ?= $(LOCALBIN)/etcd
 
 # Use the same version kuberentes is tested against, see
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -596,3 +600,15 @@ $(ETCD): $(LOCALBIN)
...
+yq: $(YQ) ## Download yq locally if necessary.
+$(YQ): $(LOCALBIN)
+	test -s $(LOCALBIN)/yq || { curl -sfSLo $(YQ) https://github.com/mikefarah/yq/releases/download/$(YQ_VERSION)/yq_linux_$(TARGET_ARCH) && chmod +x $(YQ); }
+
+# modify suites to work on arm64
+# Set cpuScalingMode to SysfsScaling
+# Set targetArchitecture to arm64
+arm_patch_e2e: yq
+	@find neonvm/samples/*yaml tests/e2e -name "*.yaml" | xargs -I{} ./bin/yq eval '(select(.kind == "VirtualMachine") | .spec.cpuScalingMode = "SysfsScaling") // .' -i {}
+	@find neonvm/samples/*yaml tests/e2e -name "*.yaml" | xargs -I{} ./bin/yq eval '(select(.kind == "VirtualMachine") | .spec.targetArchitecture = "arm64") // .' -i {}
```
sharnoff, line 614: Oof, that's kind of unfortunate, but I see why it's necessary.

`targetArchitecture` I guess will need to be like that until we redefine `nil` to be "no preference", right?

But for `cpuScalingMode`, I wonder if `kuttl` can run only a subset of the tests..? Otherwise I guess that'll be needed until we fully deprecate ACPI CPU hotplug :(

.github/workflows/e2e-test.yaml
```diff
@@ -91,6 +91,7 @@ jobs:
           - arch: amd64
             # nb: use format(..) to catch both inputs.push-yamls = true AND inputs.push-yamls = 'true'.
             push-yamls: ${{ inputs.push-yamls }}
+        arch: [ amd64, arm64 ]
```
sharnoff, line 94: IIUC this is no longer correct, as of 5bd47d245642b58d9f3c28617fc19bf03376c6b7; see the "TODO: when adding arm64 ..." above

sharnoff (commented):
One question, one minor fix - then should be good to go

Makefile
mikhail-sakhnov: With some scripting it is possible to run only a set of tests - kuttl accepts test name to run.
However, I think it is more beneficial for us to cover a wider set of tests on arm until the moment drop ACPI hot plugging, otherwise we just will have no coverage for some of the suites, while the scenario is absolutely possible to happen on staging/production. 

For example, suite for pod restarts. Without that patching the suite would never run on arm. We can skip it, but where is the guarantee that counting logic will work? It has already happened before that counting logic was broken on ARM, but not on AMD.

mikhail-sakhnov: >However, I think it is more beneficial for us to cover a wider set of tests on arm until the moment drop ACPI hot plugging, otherwise we just will have no coverage for some of the suites, while the scenario is absolutely possible to happen on staging/production.

Plus actually, IF I understand correctly, when we drop ACPI hot plug, we will automatically have all those test run on ARM as well, why not to emulate same behavior right now if it provides more safety.

sharnoff: Oh, for sure -- I'm not thinking that we should only run the tests that currently will work on ARM, just that if we could run a subset, we can exclude the ones that will never run on ARM (e.g, autoscaling/live migration with QMP scaling mode)

sharnoff (approved):
lgtm!

.github/workflows/e2e-test.yaml
```diff
@@ -91,6 +91,10 @@ jobs:
           - arch: amd64
```
sharnoff, line 91: Can we also remove the TODO immediately above this?

mikhail-sakhnov: yep, rephrased to describe the current state rather then intention

Makefile
mikhail-sakhnov: I don't think we have any tests that require QMP in particular (I know it is an example!), it is rather a default mode for amd, and default is different for arm. I guess, if we imply it in the code (eg, always use sysfs scaling on arm), there will be no neccessity for the patching at all.

PR #1144, opened 2024-11-18
api: Remove VERSIONING.md

sharnoff:
It hasn't been updated for a while, and doesn't seem to be useful.

Issue #995, opened 2024-06-26
Internal feature: Stateless scheduler


Omrigan:
## Problem description / Motivation

Currently, the scheduler acts as the source of truth for the cluster state. We'd like to change that because of the following reasons:
1. Makes scheduler replication hard, as it is unclear how to reliably replicate that state.
2. Every scheduler restart results in a state reset, thus a period of warm-up is needed to repopulate the state by the autoscaler-agent. This mechanism is responsible for additional complication of the components logic and protocol, namely:
  a. The existence of the `Buffer` values.
  b. The notion of `lastPermit`.
3. Potentially, the upscaling communication diagram can be simplified to bypass the autoscaler-agent after the upscaling was approved by scheduler. 

## Feature idea(s) / DoD

1. A scheduler can be replicated.
2. A scheduler restart doesn't cause the sub-optimal scheduling decisions. 
3. We no longer have the concepts of `Buffered` resources and `lastPermit`.

## Implementation ideas

The scheduler's state and autoscaler-agent ↔ scheduler protocol can be moved entirely to annotations on the associated VirtualMachine object.

## Related work

The implementation of https://github.com/neondatabase/neon/pull/8111 might allow us to significantly simplify autoscaler-agent (or even merge it's remaining functionality into other components).

## Further information

There are more details in the [internal RFC](https://www.notion.so/neondatabase/DRAFT-RFC-Stateless-scheduler-e9d3253836124f9f98e681f89c146a19?pvs=4).

sharnoff:
Status update: I've written a draft implementation, doesn't yet pass tests.

This week, I'll finish up that initial implementation, and write up an RFC/design doc.

sharnoff:
Status update: #1163 is in review, with a number of changes left for me to make. It's also been thoroughly tested on staging.

sharnoff:
Status update: IIUC, final review is in. Next step is to let it bake on staging for a while (1 week minimum) to give any remaining issues a final opportunity to surface.

sharnoff:
This has now been released and rolled out 🎉

PR #1145, opened 2024-11-18
agent,plugin: Remove metrics from AgentRequests

sharnoff:
AFAICT is a requirement for #995 — otherwise we need to continue updating every pod regularly, which may overload etcd. In the longer term, we may wish to have the scheduler fetch its own metrics, e.g. via the metrics API [1] [2]. In the meantime, this PR implements a metric-less approach for prioritizing live migrating by preferring to migrate the older pods on a node, with the intent being that this will naturally be _somewhat_ fair so that individual VMs are not migrated too much.

[1]: https://pkg.go.dev/k8s.io/metrics
[2]: https://kubernetes.io/docs/tasks/debug/debug-cluster/resource-metrics-pipeline/

**Notes for review:** ~~This PR currently builds upon #1143 and #1144. Only the final commit in this PR (e385e408dbb72fe42a9be4829c9b921cae95aaa6) is the actual intended change here.~~

~~Also, I think the other PRs are probably more clearly worthwhile than this one~~ — I'm happy to leave this as a draft until I have something building on it.

### DIFF
--- a/pkg/agent/core/action.go
+++ b/pkg/agent/core/action.go
@@ -24,7 +24,6 @@ type ActionWait struct {
 type ActionPluginRequest struct {
 	LastPermit     *api.Resources        `json:"current"`
 	Target         api.Resources         `json:"target"`
-	Metrics        *api.Metrics          `json:"metrics"`
 	TargetRevision vmv1.RevisionWithTime `json:"targetRevision"`
 }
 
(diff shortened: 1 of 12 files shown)


sharnoff:
I thought this was required for #995, but seems like it's not (see #1163). Closing.

PR #1148, opened 2024-11-18
agent/goalcu: implement ZoneLoadAverage 

Omrigan:
ZoneLoadAverage compares abs(load1 - load5) against two thresholds, to determine if compute is in one of 3 zones:
 - "stable zone", where load5 is used to inform the target CU value
 - "scaling zone", where load1 is used to inform the target CU value
 - "mixed zone", where load1 and load5 is blended to get an intermediate value

The thresholds are computed as a percentage of a load5, the percentage is set by two new values: `cpuStableZoneRatio` and `cpuMixedZoneRatio`. 

For example, for the `cpuStableZoneRatio = 0.25`,  `cpuMixedZoneRatio = 0.25` and `load5 = 8`, we'd get:
 - stable zone when `abs(load5 - load1) < 2` <=> `6 < load1 < 10`
 - mixed zone when `2 < abs(load5 - load1) < 4` <=> `4 < load1 < 6` or `10 < load1 < 12`.
 - scaling zone otherwise

Similar algorithm is described in https://www.notion.so/neondatabase/RFC-SwitchLoadAverage-13ef189e0047804f9297e8772739508e?pvs=4

Fixes https://github.com/neondatabase/autoscaling/issues/729

### DIFF
--- a/autoscaler-agent/config_map.yaml
+++ b/autoscaler-agent/config_map.yaml
@@ -17,8 +17,8 @@ data:
           "lfcToMemoryRatio": 0.75,
           "lfcWindowSizeMinutes": 5,
           "lfcMinWaitBeforeDownscaleMinutes": 5,
-          "cpuStableZoneRatio": 0.25,
-          "cpuMixedZoneRatio": 0.25
+          "cpuStableZoneRatio": 0,
+          "cpuMixedZoneRatio": 0
         }
       },
       "billing": {
--- a/autoscaler-agent/config_map.yaml
+++ b/autoscaler-agent/config_map.yaml
@@ -16,7 +16,9 @@ data:
           "enableLFCMetrics": false,
           "lfcToMemoryRatio": 0.75,
           "lfcWindowSizeMinutes": 5,
-          "lfcMinWaitBeforeDownscaleMinutes": 5
+          "lfcMinWaitBeforeDownscaleMinutes": 5,
+          "cpuStableZoneRatio": 0.25,
+          "cpuMixedZoneRatio": 0.25
         }
       },
       "billing": {
--- a/pkg/agent/core/goalcu.go
+++ b/pkg/agent/core/goalcu.go
@@ -82,13 +82,14 @@ func calculateCPUGoalCU(
 }
 
 func blendingFactor[T constraints.Float](value, t1, t2 T) T {
-	if value < t1 {
+	if value <= t1 {
 		return 0
-	} else if value >= t2 {
-		return 1
-	} else {
-		return (value - t1) / (t2 - t1 + 1e-6)
 	}
+	if value >= t2 {
+		return 1
+	}
+	// 1e-6 is just a precaution, if t1==t2, we'd return earlier.
+	return (value - t1) / (t2 - t1 + 1e-6)
 }
 
 // For Mem:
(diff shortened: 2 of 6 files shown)


pkg/agent/core/goalcu.go
```diff
@@ -65,11 +66,29 @@ func calculateCPUGoalCU(
...
 	return cpuGoalCU
 }
 
+func blendingFactor[T constraints.Float](value, t1, t2 T) T {
+	if value < t1 {
+		return 0
+	} else if value >= t2 {
+		return 1
+	} else {
+		return (value - t1) / (t2 - t1 + 1e-6)
```
sharnoff, line 88: suggestion: It took me a while to understand that we won't get `value/1e-6` in practice because if `t2 - t1 == 0` then we'll almost certianly hit `value < t1` or `value >= t2`.

Could you add a comment explaining why dividing by a very small value is ok here (and won't produce very large values :sweat_smile:) ?

pkg/agent/core/goalcu.go
```diff
@@ -65,11 +66,29 @@ func calculateCPUGoalCU(
 	computeUnit api.Resources,
 	systemMetrics SystemMetrics,
 ) uint32 {
-	goalCPUs := systemMetrics.LoadAverage1Min / *cfg.LoadAverageFractionTarget
+	stableThreshold := *cfg.CPUStableZoneRatio * systemMetrics.LoadAverage5Min
+	mixedThreshold := stableThreshold + *cfg.CPUMixedZoneRatio*systemMetrics.LoadAverage5Min
+
+	diff := math.Abs(systemMetrics.LoadAverage1Min - systemMetrics.LoadAverage5Min)
+	load1Weight := blendingFactor(diff, stableThreshold, mixedThreshold)
```
sharnoff, line 73:
```suggestion
	// load1Weight is 0 when diff < stableThreshold, and 1 when diff > mixedThreshold.
	// If diff is between the thresholds, it'll be between 0 and 1.
	load1Weight := blendingFactor(diff, stableThreshold, mixedThreshold)
```

I had a hard time figuring out what was happening here the first time I read it.

Wording isn't strict -- you probably have a better idea about how to explain mixedThreshold :sweat_smile:

pkg/agent/core/goalcu.go
```diff
@@ -65,11 +66,29 @@ func calculateCPUGoalCU(
 	computeUnit api.Resources,
 	systemMetrics SystemMetrics,
 ) uint32 {
-	goalCPUs := systemMetrics.LoadAverage1Min / *cfg.LoadAverageFractionTarget
+	stableThreshold := *cfg.CPUStableZoneRatio * systemMetrics.LoadAverage5Min
+	mixedThreshold := stableThreshold + *cfg.CPUMixedZoneRatio*systemMetrics.LoadAverage5Min
```
sharnoff, line 70: question: What happens if load5 is zero?

autoscaler-agent/config_map.yaml
```diff
@@ -16,7 +16,9 @@ data:
           "enableLFCMetrics": false,
           "lfcToMemoryRatio": 0.75,
           "lfcWindowSizeMinutes": 5,
-          "lfcMinWaitBeforeDownscaleMinutes": 5
+          "lfcMinWaitBeforeDownscaleMinutes": 5,
+          "cpuStableZoneRatio": 0,
+          "cpuMixedZoneRatio": 0
```
sharnoff, line 21: Should we set these to something other than zero by default?

sharnoff (approved):
Looks great, awesome work on this :rocket:

Left a few small comments; feel free to merge once addressed.

PR #1151, opened 2024-11-19
cluster-autoscaler: Upgrade to k8s 1.29

edude03:
This patch was essentially a rewrite of the original. 

Context being, upstream, the cluster autoscaler team unified the watch and lister interfaces using informers. So, instead of patching both of those methods, only the `List()` method on the `allPodLister` is modified. Assuming my understanding is correct it'll have the same behaviour but with much less code (as evidenced by the removed lines in the original patch) 

I'm not sure how to validate it works as expected other than allowing the e2e tests to run, so if there is anything I can do, let me know :)


cluster-autoscaler/ca.patch
```diff
@@ -1,137 +1,66 @@
...
+ 	"k8s.io/apimachinery/pkg/fields"
+ 	"k8s.io/apimachinery/pkg/labels"
+ 	"k8s.io/client-go/informers"
+@@ -46,6 +48,11 @@ type ListerRegistry interface {
+ 	StatefulSetLister() v1appslister.StatefulSetLister
  }
  
-+// copied from github.com/neondatabase/autoscaling, neonvm/apis/neonvm/v1/virtualmachine_types.go.
-+//
-+// this is duplicated so we're not *also* managing an additional dependency.
```
sharnoff, line 38: Let's also copy over the comment?

cluster-autoscaler/ca.patch
```diff
@@ -1,137 +1,66 @@
...
- type NodeLister interface {
- 	List() ([]*apiv1.Node, error)
+ // List returns all scheduled pods.
+ func (lister *AllPodLister) List() ([]*apiv1.Pod, error) {
+ 	var pods []*apiv1.Pod
+@@ -229,9 +252,12 @@ func (lister *AllPodLister) List() ([]*apiv1.Pod, error) {
+ 	if err != nil {
+ 		return pods, err
+ 	}
++
```
sharnoff, line 57: unneeded change? (ordinarily it'd be fine, but it's worth reducing the number of lines adjacent to the diff)

cluster-autoscaler/ca.patch
```diff
@@ -1,137 +1,66 @@
...
+ 	for _, p := range allPods {
+ 		if p.Status.Phase != apiv1.PodSucceeded && p.Status.Phase != apiv1.PodFailed {
+-			pods = append(pods, p)
++			podCopy := p.DeepCopy()
++			updatePodRequestsFromNeonVMAnnotation(podCopy)
++			pods = append(pods, podCopy)
```
sharnoff, line 63: What's the thinking behind the copy here? (I wouldn't have thought it's necessary, curious for your perspective)

sharnoff (commented):
broadly looks good! left some comments

cluster-autoscaler/ca.patch
edude03: apparently p is a pointer to the object in the informer cache, so instead of modifying the "real" object, clone it, modify the clone and return that

sharnoff: Fair enough, makes sense

Omrigan: Can we add a comment here? I had the same question :-)

Omrigan (commented):
Some comments are left, and it needs a rebase, after that should be good to go.

PR #1153, opened 2024-11-21
vm-runner: external traffic Prometheus metrics endpoint

myrrc:
Add /metrics Prometheus endpoint to neonvm-runner
exposing following metrics:
runner_vm_egress_bytes
runner_vm_ingress_bytes
runner_vm_network_fetch_errors_total
_bytes metrics use iptables and query rules configured in 
https://github.com/neondatabase/cloud/blob/main/compute-init/compute-init.sh#L143
 
Resolves: https://github.com/neondatabase/neon/issues/4704


neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -150,6 +161,11 @@ type VirtualMachineSpec struct {
 	// propagated to the VM.
 	// +optional
 	TargetRevision *RevisionWithTime `json:"targetRevision,omitempty"`
+
+	// Enable network monitoring on the VM
+	// +kubebuilder:default:=false
+	// +optional
+	EnableNetworkMonitoring *bool `json:"enableNetworkMonitoring,omitempty"`
```
sharnoff, line 168: idle thought: If we're concerned about mismatched iptables rules, maybe instead of a boolean, this should explicitly give the rules to track (and, potentially a mapping for how they should be exposed?)

Not sure, don't know enough about iptables to say either way -- just putting it out there as an idea.

sharnoff (commented):
Nicely done! Here's some initial review, leaving out the autoscaler-agent changes for now, pending discussion on how we actually want to consume the network usage data.

neonvm-runner/cmd/main.go
```diff
@@ -1194,14 +1208,65 @@ func listenForCPUChanges(
...
+
+func incNetBytesCounter(iptables *iptables.IPTables, chain string, cnt *vmv1.NetworkBytes) error {
+	// TODO(myrrc) neon_private_cidr?
+	rules, err := iptables.StructuredStats("filter", chain)
+	if err != nil {
+		return err
+	}
+	for _, stat := range rules {
+		if stat.Protocol == "tcp" {
+			*cnt += vmv1.NetworkBytes(stat.Bytes)
```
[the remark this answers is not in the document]
myrrc: Yep

neonvm/apis/neonvm/v1/virtualmachine_types.go
myrrc: This would increase complexity of the code (and tie it to the cloud/ repo) without any benefits - we'd still need to query iptables, so just querying all the rules seems simpler. The mismatch issue was not the rules being wrong, just the rules being absent.

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -68,6 +71,14 @@ type VirtualMachineResources struct {
 	MemorySlotSize resource.Quantity `json:"memorySlotSize"`
 }
 
+type (
+	NetworkBytes               uint64
+	VirtualMachineNetworkUsage struct {
+		IngressBytes NetworkBytes `json:"ingress_bytes"`
+		EgressBytes  NetworkBytes `json:"egress_bytes"`
+	}
+)
```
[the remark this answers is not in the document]
sharnoff: I know -- it won't auto-format that way if you add comments to the types.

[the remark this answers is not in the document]
myrrc: Fixed

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,85 @@ func main() {
...
+		return
+	}
+
+	if ingress, err := getNetworkBytesCounter(iptables, "INPUT"); err != nil {
+		m.Errors.WithLabelValues(util.RootError(err).Error()).Inc()
+		return
+	} else {
+		m.IngressBytes.Add(float64(ingress - m.IngressBytesRaw))
+		m.IngressBytesRaw = ingress
+	}
```
sharnoff, line 744: Here, I would recommend rewriting this to put `ingress, err :=` on its own line, and then move the contents of the `else { ... }` block in the main body of the function.

That way the `if err != nil { ... }` is easier to visually skip over to focus reading on just the happy-path.

\+ same comment for the ones below

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,85 @@ func main() {
...
+
+	if egress, err := getNetworkBytesCounter(iptables, "OUTPUT"); err != nil {
+		m.Errors.WithLabelValues(util.RootError(err).Error()).Inc()
+		return
+	} else {
+		m.EgressBytes.Add(float64(egress - m.EgressBytesRaw))
+		m.EgressBytesRaw = egress
```
sharnoff, line 751: IIUC, here and above, we do this `Add(previous value - current value)` — am I understanding this correctly, and if so, can we just `Set(current value)` instead?

And further: If we _do_ change it that way, then we should probably rename this method from `increment()` to something that more closely matches the content of what it's doing: setting the metric values based on the current iptables counts.

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,85 @@ func main() {
...
+			prometheus.CounterOpts{
+				Name: "runner_vm_network_fetch_errors_total",
+				Help: "Number of errors while fetching network monitoring data",
+			},
+			[]string{"cause"},
+		),
+	}
+	reg.MustRegister(m.IngressBytes)
+	reg.MustRegister(m.EgressBytes)
+	reg.MustRegister(m.Errors)
```
sharnoff, line 708: Suggestion: Try wrapping each `prometheus.NewFoo(...)` with `util.RegisterMetric(reg, ...)` — we added that to avoid the series of detached `reg.MustRegister(...)` calls that we'd otherwise have to have.

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,85 @@ func main() {
...
+	}
+	for _, rawStat := range rules {
+		stat, err := iptables.ParseStat(rawStat)
+		if err != nil {
+			return cnt, err
+		}
+		if stat.Protocol == "6" { // tcp
+			cnt += stat.Bytes
+		}
+	}
```
sharnoff, line 726: Does this include _all_ traffic, not just external traffic? If so, is that desired?

sharnoff (commented):
Partial review, looking good.

neonvm-runner/cmd/main.go
myrrc: No, because this is a Counter, not a Gauge, thus we are incrementing

neonvm-runner/cmd/main.go
myrrc: My personal preference is this style as it's more terse. If we don't have this in the style guide I'd prefer to leave this as is

neonvm-runner/cmd/main.go
sharnoff: Should we make it a gauge then, to make this simpler?

myrrc: Technically it's a Counter as bytes can't go down (only if iptable rules are reset which doesn't happend with us) so I'd better expose the Counter predicate

neonvm-runner/cmd/main.go
myrrc: Fixed

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,91 @@ func main() {
...
+			},
+		)),
+		IngressBytesRaw: 0,
+		EgressBytesRaw:  0,
+		Errors: util.RegisterMetric(reg, prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "runner_vm_network_fetch_errors_total",
+				Help: "Number of errors while fetching network monitoring data",
+			},
+			[]string{"cause"},
```
sharnoff, line 703: I'm worried about label explosion here, especially because this metric will already be collected per-pod. Let's just make this a regular `Counter`, not `CounterVec` ?

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,91 @@ func main() {
...
+		if err != nil {
+			return cnt, err
+		}
+		if stat.Protocol != "6" { // count tcp only
+			continue
+		}
+		src, dest := stat.Source.IP, stat.Destination.IP
+		if src.IsUnspecified() || dest.IsUnspecified() ||
+			src.IsLoopback() || dest.IsLoopback() ||
+			src.IsPrivate() || dest.IsPrivate() {
```
sharnoff, line 728: Lots of duplication here, can we extract the `!ip.IsUnspecified() && !ip.IsLoopback() && !ip.IsPrivate()` into some shared function?

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,91 @@ func main() {
...
+	// We need to measure only external traffic to/from vm, so we filter internal traffic
+	for _, rawStat := range rules {
+		stat, err := iptables.ParseStat(rawStat)
+		if err != nil {
+			return cnt, err
+		}
+		if stat.Protocol != "6" { // count tcp only
+			continue
+		}
+		src, dest := stat.Source.IP, stat.Destination.IP
```
sharnoff, line 725: Clarifying question here: what are these source/destination IPs? are they the source/destination of the iptables rules? if so, will that actually catch everything?

sharnoff (commented):
One last review, after this it should be good

neonvm-runner/cmd/main.go
sharnoff: > If we rely on rules to monitor traffic then we should be ok filtering over them

Right so currently we don't, they're just for limiting connectivity.

But if this works as is (including the most recent change to filter out internal traffic), or if you're already planning to change the iptables rules to make sure this works, then all good from me.

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,93 @@ func main() {
...
+
+	// We need to measure only external traffic to/from vm, so we filter internal traffic
+	for _, rawStat := range rules {
+		stat, err := iptables.ParseStat(rawStat)
+		if err != nil {
+			return cnt, err
+		}
+		if stat.Protocol != "6" { // count tcp only
+			continue
+		}
```
sharnoff, line 728: Shouldn't this be `if stat.Protocol == <something>` rather than `!=`, if we only want to count TCP traffic? unless I'm missing something...

neonvm-runner/cmd/main.go
```diff
@@ -669,6 +673,93 @@ func main() {
...
+		cnt += stat.Bytes
+	}
+	return cnt, nil
+}
+
+func (m *NetworkMonitoringMetrics) increment() {
+	// Rules configured at github.com/neondatabase/cloud/blob/main/compute-init/compute-init.sh#L98
+	iptables, err := iptables.New()
+	if err != nil {
+		m.Errors.Inc()
```
sharnoff, line 742: One last change request, sorry: if we get errors, we should at least log them so it's not impossible to debug.

sharnoff (approved):
Approved, pending resolution of the comments

neonvm-runner/cmd/main.go
myrrc: As we continue on non-tcp (non-6) traffic, all left is tcp traffic

Omrigan: Can you use a named constant here from https://pkg.go.dev/golang.org/x/net/internal/iana#pkg-constants?

Or at least make a comment where does "6" comes from.

neonvm-runner/cmd/main.go
myrrc: I re-checked locally and corrected tracking rules. When we count traffic from unspecified or to unspecified locations (note we have rules for internal traffic ACCEPT), bytes are counted correctly. I've amended the PR.

PR #1157, opened 2024-11-22
Use 'vmv1' for neonvm API import everywhere

sharnoff:
Within neonvm-controller, we're consistently using `vmv1` to refer to the `.../apis/neonvm/v1` import, but elsewhere we're using `vmapi` instead.

`vmv1` is closer to the typical kubernetes imports like `corev1` and `metav1`, even if it doesn't exactly match the import path.

---

**Notes for review:** Low priority, been on my todo list for a while and I figured that #995 was a good enough excuse to do it. Also, this is probably a good candidate to add to `.git-blame-ignore-revs` after merging, because otherwise it's a lot of noise.

### DIFF
--- a/autoscaler-agent/cmd/main.go
+++ b/autoscaler-agent/cmd/main.go
@@ -13,7 +13,7 @@ import (
 	scheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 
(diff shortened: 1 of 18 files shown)


Omrigan (approved):
Nice, wanted to do the same!

Inconsistent imports are annoying when working across several components.

PR #1161, opened 2024-12-02
cluster-autoscaler: Bump cluster autoscaler to 1.30

edude03:
Bump underlying version of cluster-autoscaler to 1.30


sharnoff (approved):
Broadly looks fine. Please note the current PR title is misleading — it should be prefixed with `cluster-autoscaler:`, not `plugin:` (the latter indicates this is related to the scheduler plugin, which this is not)

PR #1165, opened 2024-12-03
neonvm-daemon: Use go-chef in Dockerfile

sharnoff:
Probably an inadvertent merge conflict between #1090 and #989 meaning we accidentally weren't using go-chef for neonvm-daemon.

Noticed this while working on #1163 locally and saw that it was re-downloading all of the dependencies for neonvm-daemon every time, even though I was making changes in the scheduler and the dependencies hadn't changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ docker-build-runner: docker-build-go-base ## Build docker image for NeonVM runne
 		.
 
 .PHONY: docker-build-daemon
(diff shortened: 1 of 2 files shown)


mikhail-sakhnov (approved):

sharnoff:
From DMs with @mikhail-sakhnov -- apparently I originally suggested that it should be this way. I've changed my mind after developing locally, because it takes _so long_ to recompile each time :sweat_smile:

PR #1169, opened 2024-12-06
neonvm: Bump max mem slots to 512

sharnoff:
In practice this is limiting the maximum size of our computes, even though memory slots aren't physically used since we switched to virtio-mem.

ref neondatabase/cloud#18281


mtyazici:
should we not generate the CRD?

sharnoff:
yup, CI complained because I didn't also regenerate the CRD :sweat_smile:

mtyazici (approved):
👍🏼

PR #1159, opened 2024-12-02
neonvm-kernel: update ubuntu builder image to LTS

mikhail-sakhnov:
Update ubuntu:23.10 to the LTS version ubuntu:24.04

Closes #1154

PR #1172, opened 2024-12-09
neonvm-kernel: Fix upgrade instruction

petuhovskiy:
Fix the instruction for kernel upgrades, add separate commands for arm64 and amd64 config upgrades.

Rerun the commands with current version. It has produced some diff, most likely due to ubuntu upgrade in https://github.com/neondatabase/autoscaling/pull/1159.

Split off from https://github.com/neondatabase/autoscaling/pull/1170


mikhail-sakhnov (approved):

neonvm-kernel/README.md
```diff
@@ -27,17 +29,43 @@ with the following sequence of actions:
...
+   ```sh
+   # compare the two versions
+   diff linux-config-amd64-6.1.92 linux-config-amd64-$NEW_VERSION
+   # If all looks good, delete the old version. This is required so auto-selection works.
+   rm linux-config-amd64-6.1.92
+   ```
+
+### On arm64 (aarch64 ARM)
+
+Note: you need to run this on ARM64 machine. It worked on my MacBook M2, but didn't on x64.
```
mikhail-sakhnov, line 47: @petuhovskiy could you try running make menuconfig with explicit arch? Something like
Also --platform argument should be dropped for config steps.

```
make menuconfig ARCH=x86_64
```
or 
```
make menuconfig ARCH=arm64
```

should work. 


For example, I tried on my x86_64 devbox:

```
➜  neonvm-kernel git:(main) ✗ docker build --build-arg KERNEL_VERSION=6.1.92 --target build-deps -t kernel-build-deps -f Dockerfile.kernel-builder .
[+] Building 0.9s (10/10) FINISHED                                                                                                                                                                                                                                                              
 => => writing image sha256:a3f1f9fd37e7af58018a862f5f57f09dada959db87a2ef62df88690317b3c655                                                                                                                                                                                                              0.0s
 => => naming to docker.io/library/kernel-build-deps                                                                                                                                                                                                                                                      0.0s
➜  neonvm-kernel git:(main) ✗ docker run --rm -v $PWD:/host --name kernel-build -it kernel-build-deps bash

root@8d6d5478e19d:/build# cd linux-6.1.92
root@8d6d5478e19d:/build/linux-6.1.92# make menuconfig ARCH=arm64

<skipped output>

# using defaults found in arch/arm64/configs/defconfig

```


gives me arm configurator

```
Linux/arm64 6.1.92 Kernel Configuration
```

petuhovskiy: Thanks, it works. But there is a small diff, I made a commit with a new config. Does it mean that building the config always can produce different results, depending on the platform?

petuhovskiy:
Merging it to unblock #1170, feel free to follow up async and ping me if you notice any issue with this PR.

PR #1173, opened 2024-12-09
agent/billing: reset metrics right before setting it

Omrigan:
Right now we sometimes see gaps in metrics, which can be explained by too much time between Reset() and the next Set().

See example: https://neondb.slack.com/archives/C03TN5G758R/p1731596259801509


sharnoff (approved):
nice, thanks!

Issue #850, opened 2024-03-08
Epic: Support ARM AWS instances


stradig:
## Motivation
ARM instances in AWS are in most cases cheaper than x86 instances. Thus to reduce COGS it makes sense to switch to ARM instances.

There are a couple of aspects which need to be taken into account when switching to ARM instances:
- A different method for adding and removing CPUs needs to be used because on ARM hotplugging CPUs is not supported
- Since at least at the moment our K8s clusters are used for running not only compute but also other kinds of services we must ensure that all those work on ARM. Or we need to use mixed clusters and make sure that only those services which are running fine on ARM are running there.
- Performance of ARM CPUs will most probably differ from x86 nodes. Since we use the notion of a CU we need to calibrate the actual amount of resources one CU is getting when running on ARM so that customers get a similar performance when switching to ARM.

## DoD
- Computes can run on ARM instances without negative effects on the rest of the deployment, this includes autoscaling.
- Customers get similar performance when running on ARM than when running on x86

PR #1179, opened 2024-12-13
neonvm: handle sysfs scaling mode in migrations

mikhail-sakhnov:
We need to sync CPUs only if we have QMP mode active. For sysfs scaling CPU state should sync automatically.

Related to #850 and #1111

### DIFF
--- a/pkg/neonvm/controllers/vmmigration_controller.go
+++ b/pkg/neonvm/controllers/vmmigration_controller.go
@@ -317,13 +317,14 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 			migration.Status.SourcePodIP = vm.Status.PodIP
 			migration.Status.TargetPodIP = targetRunner.Status.PodIP
 
-			// do hotplugCPU in targetRunner before migration
-			log.Info("Syncing CPUs in Target runner", "TargetPod.Name", migration.Status.TargetPodName)
-			if err := QmpSyncCpuToTarget(vm, migration); err != nil {
-				return ctrl.Result{}, err
+			if *vm.Spec.CpuScalingMode == vmv1.CpuScalingModeQMP {
+				// do hotplugCPU in targetRunner before migration
+				log.Info("Syncing CPUs in Target runner", "TargetPod.Name", migration.Status.TargetPodName)
+				if err := QmpSyncCpuToTarget(vm, migration); err != nil {
+					return ctrl.Result{}, err
+				}
+				log.Info("CPUs in Target runner synced", "TargetPod.Name", migration.Status.TargetPodName)
 			}
-			log.Info("CPUs in Target runner synced", "TargetPod.Name", migration.Status.TargetPodName)
-
 			// Migrate only running VMs to target with plugged devices
 			if vm.Status.Phase == vmv1.VmPreMigrating {
 				// update VM status


pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -309,11 +309,16 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 			migration.Status.SourcePodIP = vm.Status.PodIP
 			migration.Status.TargetPodIP = targetRunner.Status.PodIP
 
-			// do hotplugCPU in targetRunner before migration
+			// do cpu hot plug in targetRunner before migration
+			// in case of QMP mode, we need to sync CPUs before migration
+			// in case of Sysfs mode, we need to sync CPUs during migration
```
sharnoff, line 314: > in case of Sysfs mode, we need to sync CPUs during migration

Why? I would have thought it just works, because all of the CPUs are already hotplugged in both places?

pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -357,10 +362,22 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 						Reason:  "Reconciling",
 						Message: message,
 					})
-				// finally update migration phase to Running
+				return r.updateMigrationStatus(ctx, migration)
+			case vmv1.VmMigrating:
+				// migration is in progress so we can scale CPU using sysfs
```
sharnoff, line 367: > migration is in progress so we can scale CPU using sysfs

This should be forbidden because we only scale CPUs when the VM is in the `Scaling` phase (although eventually we may change that...)

pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -357,10 +362,22 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
+				if *vm.Spec.CpuScalingMode == vmv1.CpuScalingModeSysfs {
+					if err := setRunnerCPULimits(ctx,
+						vm,
+						targetRunner.Status.PodIP,
+						vm.Spec.Guest.CPUs.Use); err != nil {
+						return ctrl.Result{}, err
+					}
+				}
+				// if cpu scaling is not sysfs based we just update VM status to Running, since migration is done at the moment
 				migration.Status.Phase = vmv1.VmmRunning
```
Omrigan, line 377: It is not the VM status which is updated, but a migration status. Migration is not done, and, I believe, it is just starting. 

Does this code works? From just looking at the code, I'd expect it is not safe to connect to the target pod until the migration is completed. Maybe, our tests don't catch it, because migration completes very quickly.

Is the handling for sysfs mode actually required? I think, we can just adjust number of CPUs from vm_controller, and don't change it during migration. If VM spec is changed during the migration, the number of CPUs will be adjusted during the first reconciliation loop after the completion of migration.

If that's true, than this PR can be a one-liner: just the `*vm.Spec.CpuScalingMode == vmv1.CpuScalingModeQMP` check before `QmpSyncCpuToTarget`.

WDYT?

pkg/neonvm/controllers/vmmigration_controller.go
mikhail-sakhnov: Mostly because at that moment machine is not booted yet, so it's kind chicken-and-egg thing.

pkg/neonvm/controllers/vmmigration_controller.go
mikhail-sakhnov: I think here I meant that we can call the neonvm-daemon to enable all required cpus, because before that machine wasn't yet booted and the neonvm was not accessible

pkg/neonvm/controllers/vmmigration_controller.go
mikhail-sakhnov: >It is not the VM status which is updated, but a migration status. Migration is not done, and, I believe, it is just starting.

to clarify - it is about the comment text?

>Does this code works? From just looking at the code, I'd expect it is not safe to connect to the target pod until the migration is completed. Maybe, our tests don't catch it, because migration completes very quickly.

yes, without the change the migration stucks for an infinite amount of time.

Basically, my train of thoughts was like:
- we have pod A
- we have target pod B
- let's say cpu range is from X to Y and the current value of online CPUs is somewhere in between for pod A
- we start target pod B with qemu argument to wait for the migration
- we need (may be that's a false assumption?) to say to the target pod B that it should have currently online CPUs equal to the amount of online CPUs in source pod A, using neonvm-daemon
- but since the qemu is waiting for migration be done, the VM in pod B is not running
- so to call neonvm-daemon we first need to boot VM

However, with you suggestion 
> I think, we can just adjust number of CPUs from vm_controller, and don't change it during migration. If VM spec is changed during the migration, the number of CPUs will be adjusted during the first reconciliation loop after the completion of migration.

it might be actually not necessary. Let me check the approach you suggested and get back to you

Omrigan: There is no booting process involved in the live migration. VM is booted long before the live migration is initiated. During live migration, the state of memory is transferred from one host to another, VM is paused on source, and resumed from the same place on target. 

Therefore, after the VM is resumed on target, we are supposed to have the same CPUs enabled as we had on source.

mikhail-sakhnov: > VM is booted long before the live migration is initiated

do you mean on the target pod? 

I see in the runner code 
```
        // should runner receive migration ?
        if os.Getenv("RECEIVE_MIGRATION") == "true" {
                qemuCmd = append(qemuCmd, "-incoming", fmt.Sprintf("tcp:0:%d", vmv1.MigrationPort))
        }
```

and if I understand correctly that prevents the machine from booting normally, because it first waits for the migration data.

mikhail-sakhnov: >Therefore, after the VM is resumed on target, we are supposed to have the same CPUs enabled as we had on source. 

with that in mind we don't really need to manually call neonvm-daemon, as you mentioned.

Omrigan: > > VM is booted long before the live migration is initiated
> 
> do you mean on the target pod?

I mean on source pod.

> 
> I see in the runner code
> 
> ```
>         // should runner receive migration ?
>         if os.Getenv("RECEIVE_MIGRATION") == "true" {
>                 qemuCmd = append(qemuCmd, "-incoming", fmt.Sprintf("tcp:0:%d", vmv1.MigrationPort))
>         }
> ```
> 
> and if I understand correctly that prevents the machine from booting normally, because it first waits for the migration data.

That's correct.

I guess, I understand "booting" as the process by which the operating system initializes it's state, starts userspace processes, mounts filesystem, etc. All of this doesn't happen during live migration. 

QEMU with `-incoming` starts an empty shell which waits for the VM state to arrive from the other QEMU instance.

mikhail-sakhnov:
@sharnoff I resolved all discussions as @Omrigan was indeed right and the whole change could be replaced with basically dropping qmpsync operation

PR #1158, opened 2024-11-29
Bump k8s dependencies to 1.30 

edude03:
Continuing the epic to upgrade from kubernetes 1.28 -> 1.31, bump the k8s dependencies

PR #1181, opened 2024-12-17
make generate: Remove generated files beforehand

sharnoff:
Spawned out of conversation with @edude03 for #1158 — basically, it's hard to be sure that updates to controller-gen are working correctly, so making sure that all old files are removed before re-generating can help provide assurances that all files continue to be generated after a change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ help: ## Display this help.
 generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 	# Use uid and gid of current user to avoid mismatched permissions
 	set -e ; \
+	rm -rf neonvm/client neonvm/apis/neonvm/v1/zz_generated.deepcopy.go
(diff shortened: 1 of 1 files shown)


petuhovskiy (approved):

Makefile
```diff
@@ -83,6 +83,7 @@ help: ## Display this help.
 generate: ## Generate boilerplate DeepCopy methods, manifests, and Go client
 	# Use uid and gid of current user to avoid mismatched permissions
 	set -e ; \
+	rm -rf neonvm/client neonvm/apis/neonvm/v1/zz_generated.deepcopy.go
```
mikhail-sakhnov, line 86: suggestion: would be nice in general have a traditional "make clean" target that drops generated and compiled things, because we already do clean ups here and there (kernel, generated files).

Omrigan: Created https://github.com/neondatabase/autoscaling/issues/1195#issue-2771089408

Omrigan (approved):

Issue #1184, opened 2024-12-23
neonvm-runner's main.go is unweildy


sharnoff:
It's gradually grown to be a very long file that's somewhat hard to navigate, and this makes it more difficult for new contributors.

We should break it up into multiple files at the very least, possibly with some other refactoring in there as well.

sharnoff:
This was closed by #1185.

PR #1185, opened 2024-12-23
neonvm-runner: Small cleanups

sharnoff:
I was looking at trying to implement some changes to what neonvm-runner puts in the ISO 9660 filesystem, and quickly began hitting #1184, so decided to make some changes to fix that.

The commits are basically just what the subject messages say.

---

It's worth noting that these changes are kind of only surface-level; I think there's deeper structural changes that we should make so neonvm-runner can be easier to maintain and extend, but I'm not entirely sure what that should look like, and just breaking it up into separate files seemed less objectionable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,12 +136,12 @@ test: vet envtest ## Run tests.
 
 .PHONY: build
 build: vet bin/vm-builder ## Build all neonvm binaries.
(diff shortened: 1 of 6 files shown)


Omrigan (approved):

sharnoff:
Rebased onto main, appended PR number to each commit, and moved the docker `libnetwork` commit to the end, so it's feasible to revert later if we want to.

PR #1186, opened 2024-12-23
agent: Add support for simplified LFC scaling

sharnoff:
We have an ongoing issue where the LFC goal size algorithm is hitting degenerate edge-case behavior on a large workload — but because of that bug, it's hard to tell whether the impacted workload *can* be feasibly autoscaled with the information we have available.

For large workloads, it's reasonable enough to just look at the 1-hour working set window, and testing with that will indicate whether we need to look at a longer time horizon to effectively scale LFC on such a workload — or if the information we already have should be sufficient.

See also neondatabase/cloud#22214 for more information.


pkg/api/vminfo.go
```diff
@@ -353,6 +353,13 @@ type ScalingConfig struct {
 	// default.
 	EnableLFCMetrics *bool `json:"enableLFCMetrics,omitempty"`
 
+	// UseSimpleLFCScaling, if true, calculates goal LFC size only based on the largest available
+	// working set size window, instead of trying to allow downscaling at earlier opportunities.
+	//
+	// This is not fit for general use. It's meant as a temproary escape hatch to let us assess the
+	// upper bound of potential improvements to our LFC goal size heuristics.
+	UseSimpleLFCScaling *bool `json:"useSimpleLFCScaling,omitempty"`
```
Omrigan, line 361:
```suggestion
	LFCUseLargestWindow *bool `json:"lfcUseLargestWindow,omitempty"`
```

IMO explains what it does right in the name.

Omrigan (approved):
Naming proposal, otherwise LGTM.

pkg/api/vminfo.go
sharnoff: Fair enough, yeah -- renamed in 19b5c47603b6cb883026ed526c02b5daee0c0227

PR #1189, opened 2024-12-27
neonvm: mv root disk from init container, not cp

sharnoff:
This commit is relatively straightforward: In the neonvm-runner init container, switch from copying the root disk into `/vm/images` to *moving* it there.

Copying necessarily causes a lot of disk writes, but moving may not. If the init container's `disk.qcow2` is on the same underlying filesystem as the `/vm/images` volume, then the kernel MAY do the move very efficiently.

To be clear: I'm not totally sure whether this will help — it may be that moving files out of an overlayfs still requires copying. But if it helps, it could dramatically reduce our disk IO usage when many computes are started at once. And even if it doesn't help, it's probably a better default to move instead of copying.

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -1321,7 +1321,7 @@ func podSpec(
 					}},
 					Command: []string{
 						"sh", "-c",
-						"cp /disk.qcow2 /vm/images/rootdisk.qcow2 && " +
+						"mv /disk.qcow2 /vm/images/rootdisk.qcow2 && " +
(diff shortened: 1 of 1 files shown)


Omrigan (approved):
> But if it helps, it could dramatically reduce our disk IO usage when many computes are started at once. 

Will we be able to this in metrics, once this change is rolled out? I am curious to know, and it would be good idea to drop a note in this PR, if we learn if it helps or not.

sharnoff:
I _think_ we should be able to see it in the node IO metrics pretty quickly if it helps, yeah

PR #1190, opened 2024-12-30
neonvm: add readiness probe for runner container

### LINKED ISSUE #1147: Add readiness probe for the vm-runner process
We don't have any readiness probe for the pod with VM. 
Some time ago there was readiness probe in case if pod was running through container manager. 

I suggest we add readiness probe to avoid having Ready VMs with in fact non-ready runner.

For the readiness probe we could use TCP connection check to vm-daemon and/or postgres process.

mikhail-sakhnov:
Call runner /ready which, in sysfs scaling mode, proxifies to daemon's /cpu to check if runner pod and vm is ready. Runner's endpoint /ready does nothing in case of qmp scaling model.
Move neonvm-daemon line in the inittab to have it starting later since we use it for the health checks.

It is a blocker for #1141 because working without ready probe introduces race in some e2e tests and kuttl tries to execute shell command `kill` while VM is not fully booted and acpid rules are not in action yet.

It closes #1147 which is in the backlog but since it is a blocker I extracted it from the PR #1141 

#1147


vm-builder/files/inittab
```diff
@@ -1,13 +1,13 @@
...
 ::once:/neonvm/bin/touch /neonvm/vmstart.allowed
-::respawn:/neonvm/bin/neonvmd --addr=0.0.0.0:25183
 ::respawn:/neonvm/bin/udhcpc -t 1 -T 1 -A 1 -f -i eth0 -O 121 -O 119 -s /neonvm/bin/udhcpc.script
 ::respawn:/neonvm/bin/udevd
 ::wait:/neonvm/bin/udev-init.sh
 ::respawn:/neonvm/bin/acpid -f -c /neonvm/acpi
 ::respawn:/neonvm/bin/vector -c /neonvm/config/vector.yaml --config-dir /etc/vector --color never
 ::respawn:/neonvm/bin/chronyd -n -f /neonvm/config/chrony.conf -l /var/log/chrony/chrony.log
 ::respawn:/neonvm/bin/sshd -E /var/log/ssh.log -f /neonvm/config/sshd_config
+::respawn:/neonvm/bin/neonvmd --addr=0.0.0.0:25183
```
sharnoff, line 10: question: Why this change?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -932,12 +933,26 @@ func runnerStatus(pod *corev1.Pod) runnerStatusKind {
...
-		return runnerRunning
+		return isRunnerPodReady(pod)
 	default:
 		panic(fmt.Errorf("unknown pod phase: %q", pod.Status.Phase))
 	}
 }
 
+// isRunnerPodReady returns whether the runner pod is ready respecting the readiness probe of its containers.
+func isRunnerPodReady(pod *corev1.Pod) runnerStatusKind {
+	if pod.Status.ContainerStatuses == nil {
```
sharnoff, line 944: question: Is there a difference between `== nil` vs `len(...) == 0` ? I was a little surprised to see this check on a slice

sharnoff (commented):
Mostly LGTM, left a couple questions

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: You right, I don't think there is any difference for a slice if we use `== nil` or `len(...) == 0`

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1443,6 +1463,21 @@ func podSpec(
 		},
 	}
 
+	if withReadinessProbe {
```
mikhail-sakhnov, line 1466: this is for the live migration, but I don't like this approach. I pushed that version to discuss what should be the correct approach for the live migrations. The thing is that qemu in target pod doesn't boot until migration is performed. But the migration is not performed unless pod is ready. There are 2 ways to tackle the issue:
- either change ready probe in target pods (the line I left comment for does that) but we can't change readyiness probe on running pod so this pod will be forever without readiness probe. I found this suboptimal tbh - some pods will have probe, some won'y
- change the logic in live migration controller and don't wait for the pod readiness

@sharnoff do you see any better options ?

sharnoff: Hm, readiness probes (a) are valid beyond startup, and (b) are used to route service traffic, right?

I would imagine that VM migration doesn't work well with services because we either have extended time with >1 pod ready or a brief moment when the old pod is "ready" but shouldn't receive traffic.

But it'd be a little weird to have readiness probes at runtime only working if the VM hasn't been migrated yet...

Given that, I would probably recommend the second option you gave: use the readiness probe on both pods, and change the logic in the migration controller. That way, we have a bit more information about the migration state visible from the pods, services should work a _little_ better, and we get consistent readiness probes across the lifetime of a VM.

vm-builder/files/inittab
sharnoff: IIUC then, the idea behind moving neonvmd down is to give other daemons a chance to start first, right?

Is that actually guaranteed?

pkg/neonvm/controllers/vm_controller.go
Omrigan: nil slices and non-nil empty slices are two different things: https://go.dev/play/p/UtIWK670wxW

I always use `len(...) == 0` to cover both cases.

mikhail-sakhnov: I usually tend use  != nil or len(slice) depending on the semantics of a particular code (like if I am interested in length or emptiness for sure) but I am ok with using any option here.

neonvm-runner/cmd/main.go
```diff
@@ -497,6 +497,19 @@ func runQEMU(
...
+			if cfg.cpuScalingMode == vmv1.CpuScalingModeSysfs {
+				err := checkNeonvmDaemonCPU()
+				if err != nil {
+					logger.Warn("neonvm-daemon ready probe failed", zap.Error(err))
+					return false
+				}
+				return true
+			}
+			// immediately return ready status for QMP mode
+			return true
```
sharnoff, line 511: > for QMP mode

suggestion: This comment implies something different from the code — basically, there's a hidden assumption that the only two scaling modes are sysfs and QMP. Maybe let's make that explicit via if/else or switch?

(very small suggestion, I know, sorry.)

neonvm-runner/cmd/main.go
```diff
@@ -753,3 +766,27 @@ func setNeonvmDaemonCPU(cpu vmv1.MilliCPU) error {
 
 	return nil
 }
+
+// checkNeonvmDaemonCPU sends a GET request to the NeonVM Daemon to get the current CPU limit for the sake of readiness probe.
+func checkNeonvmDaemonCPU() error {
```
sharnoff, line 771: suggestion: Maybe let's extract this into a function to fetch the CPU from neonvm-daemon, and also use it in the `get` callback?

(only just now noticed that we don't use it -- maybe we should? WDYT?)

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -864,12 +865,45 @@ func runnerStatus(pod *corev1.Pod) runnerStatusKind {
...
 		return runnerFailed
 	case corev1.PodRunning:
-		return runnerRunning
+		return isRunnerPodReady(pod)
 	default:
 		panic(fmt.Errorf("unknown pod phase: %q", pod.Status.Phase))
 	}
 }
 
+// isRunnerPodReady returns whether the runner pod is ready respecting the readiness probe of its containers.
```
sharnoff, line 874:
```suggestion
// isRunnerPodReady returns whether the runner pod is ready, respecting the readiness probe of its containers.
```

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1178,6 +1212,10 @@ func imageForVmRunner() (string, error) {
 	return image, nil
 }
 
+// podSpec returns a VirtualMachine Pod object
+// withReadinessProbe - if true, adds a readiness probe to the container
```
sharnoff, line 1216: question: What does `withReadinessProbe` refer to?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1378,6 +1416,19 @@ func podSpec(
 		},
 	}
 
+	pod.Spec.Containers[0].ReadinessProbe = &corev1.Probe{
```
sharnoff, line 1419: question: Any reason to add it here instead of as we construct the container above?

sharnoff (commented):
Left some questions — high level, looks fine, just fussing over the details.

neonvm-runner/cmd/main.go
mikhail-sakhnov: clarification: what should be the behavior in case of unknown scaling mode? always false, always true or panic?

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: leftover from previous iteration, forget to clean up, thanks for pointing out

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: and again, left over from the version when it was conditional

neonvm-runner/cmd/main.go
mikhail-sakhnov: >suggestion: Maybe let's extract this into a function to fetch the CPU from neonvm-daemon, and also use it in the get callback?

I don't have an opinion here tbh, I think in general neonvm-daemon client should be a bit refactored to avoid duplications but I kept it out of the scope of the PR

>(only just now noticed that we don't use it -- maybe we should? WDYT?)
Hm, tbh I don't remember the PARTICULAR reason why it is not used now, probably because from the controller we send fractional value, and from neonvm-daemon we always receive rounded value, so if we use the value from neonvm-daemon it would be endless reconciliation loop (controller send 0.75, we round up to 1 and enable 1 cpu, controller checks cpu from neonvm-daemon, sees 1 != 0.75, triggers reconciliation once again). That would be changed once we have fractional limits implemented in the neonvm-daemon.

mikhail-sakhnov: Discussed in DM: extracted an issue #1243

neonvm-runner/cmd/main.go
sharnoff: I would say let's panic, yeah

vm-builder/files/inittab
mikhail-sakhnov: We discussed in the DM, that no guarantees for a particular order of starting, but reduced flakiness at least locally.

neonvm-runner/cmd/main.go
mikhail-sakhnov: Added panic for unknown scaling mode, as discussed in DM/during sync.

pkg/neonvm/controllers/vm_controller.go
sharnoff: This resolved without comment or accepting the change, was that intentional?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -864,12 +865,45 @@ func runnerStatus(pod *corev1.Pod) runnerStatusKind {
...
 	case corev1.PodRunning:
-		return runnerRunning
+		return isRunnerPodReady(pod)
 	default:
 		panic(fmt.Errorf("unknown pod phase: %q", pod.Status.Phase))
 	}
 }
 
+// isRunnerPodReady returns whether the runner pod is ready respecting the readiness probe of its containers.
+func isRunnerPodReady(pod *corev1.Pod) runnerStatusKind {
```
sharnoff, line 875: nit: functions titled `isFoo` should return `bool`

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -864,12 +865,45 @@ func runnerStatus(pod *corev1.Pod) runnerStatusKind {
...
+
+	// first we check if the pod is a target pod for the migration
+	// in that case we don't wait for the readiness probe
+	// from neonvm-daemon because qemu started
+	// in incoming migration mode
+	for _, containerSpec := range pod.Spec.Containers {
+		if containerSpec.Name == "neonvm-runner" {
+			for _, env := range containerSpec.Env {
+				if env.Name == "RECEIVE_MIGRATION" && env.Value == "true" {
+					return runnerRunning
```
sharnoff, line 889: question(blocking): Does this mean that we'll return `runnerRunning` even when the neonvm-runner container is not yet running? (or alternately: can this return ready even when the pod is still pending?)

sharnoff, line 889: suggestion(non-blocking): Maybe extract the "is this a target pod" check into a separate function?

pkg/neonvm/controllers/vm_controller.go
sharnoff: Unresolving this - why keep it that way, then?

pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -713,6 +713,7 @@ func (r *VirtualMachineMigrationReconciler) targetPodForVirtualMachine(
 	pod.Name = migration.Status.TargetPodName
 
 	// add env variable to turn on migration receiver
+	// TODO: make it false or empty after the migration is done to enable correct readiness probe
```
sharnoff, line 716: question(non-blocking): what's the plan here? I don't think we can modify container environments at runtime... And, is this TODO tracked somewhere?

sharnoff (commented):
Looks good. Some blocking questions then will be good to merge

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: renamed to `runnerContainerStatus(..)`

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: oh, human factor mistake by me, accidentally did second round of code review fixes on top of wrong revision

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: the same wrong revision thing

pkg/neonvm/controllers/vmmigration_controller.go
mikhail-sakhnov: I think we discussed it verbally with @Omrigan and I promised to create an issue for that,  since I am a bit afraid of the scenario like:
- we start VM
- we migrate VM to a new pod
- new pod has `RECEIVE_MIGRATION` eq true which means that in some hypothetical scenario of restarting pod (which tbh I am not even sure is possible in our lifecycle) qemu will not boot, but will wait for some migration (which is not actually happening at the moment). 
Since we have already an issue to audit vmmigration controller, I will add it as a context there.

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: done

mikhail-sakhnov: Fair point. I introduce the new status and change the vmmigration controller to use it instead of running.

mikhail-sakhnov: After the discussion with you and @Omrigan I changed the code here to check the pod references to determine if it is a target pod for migration, which looks like the most reliable way we can currently afford.

sharnoff (approved):
lgtm!

PR #1194, opened 2025-01-02
neonvm: Small changes for stateless scheduler

sharnoff:
Extracted out of #1163, as suggested by https://github.com/neondatabase/autoscaling/pull/1163#discussion_r1890406226 and https://github.com/neondatabase/autoscaling/pull/1163#issuecomment-2556855781.

These are _mostly_ three unrelated commits among the broader theme of making stronger guarantees about what the Pods look like.

I'm intending to merge via rebase-and-merge to keep these commits separate.


Omrigan (approved):
LGTM

PR #1193, opened 2025-01-02
neonvm-kernel: Switch to version 6.12.8

petuhovskiy:
This commit updates the kernel and switches from 6.6.x to 6.12.x LTS branch.

As of 2025-01-02, www.kernel.org says this is the most recent version of the 6.12.x LTS branch.

The upgrade is done according to the instruction in neonvm-kernel/README.md.

ref https://github.com/neondatabase/cloud/issues/20195

PR #1205, opened 2025-01-15
Revert "neonvm-kernel: Switch to version 6.12.8 (#1193)"

petuhovskiy:
This reverts commit 85350adfb34dc80b842d32ef0ba3ef87a24d59c7.

We started having issues with tainted kernels after upgrade, rolling back the update for the fix.
Example kernel logs for these issues:
```
[  128.464639] WARNING: CPU: 1 PID: 0 at arch/x86/kernel/cpu/cpuid-deps.c:117 do_clear_cpu_cap+0xdf/0x140
...stacktrace...
[  128.576531] kworker/0:1: vmemmap alloc failure: order:9, mode:0x4cc0(GFP_KERNEL|__GFP_RETRY_MAYFAIL), nodemask=(null),cpuset=/,mems_allowed=0
[  128.577673] CPU: 0 UID: 0 PID: 9 Comm: kworker/0:1 Tainted: G        W          6.12.8 #1	
[  128.578099] Tainted: [W]=WARN
```

We also had a few cases with NULL pointer exceptions, which we didn't see ever before, see e.g.:
```
[22570.178384] BUG: kernel NULL pointer dereference, address: 0000000000000000
```

Overall, it seems we better rollback now, and debug those issues in staging.


Omrigan (approved):
LGTM

PR #1213, opened 2025-01-21
neonvm: create a Certificate object for each VM

conradludgate:
https://github.com/neondatabase/cloud/issues/22997

Initially, this PR set out to create a `cert-manager.io/Certificate` object for each VM. However, a design decision for cert-manager has the associated secrets produced as orphans. This caused secrets to accumulate without cleanup.

Instead, I have augmented the flow to emulate `Certificate`.

1. Generate a private key, and store it in a temporary Secret.
2. Generate a CertificateRequest based on the associated public key.
3. Wait for the CertificateRequest to be approved
4. Create a tls Secret based on the CertificateRequest and private key.
5. Delete the CertificateRequest and the temporary secret.
6. Wait 23 hours
7. Return to step 1.

This only takes place if the VM spec lists a TlsCertificateIssuer value


edude03:
If a vm runs longer than 23 hours does the cert reload?

conradludgate:
> If a vm runs longer than 23 hours does the cert reload

Yeah - when paired with the other PR, yes it does reload as expected. Still needs work to make postgres to recognise the change, but yes it should work

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -786,6 +835,134 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 	return nil
 }
 
+func (r *VMReconciler) doReconcileCertificateSecret(ctx context.Context, vm *vmv1.VirtualMachine, certSecret *corev1.Secret) (*corev1.Secret, error) {
```
sharnoff, line 838: suggestion: It seems like:

1. This function has some mixing of high-level logic ("pass the secret key we stored into creating a signing request") with lower level operations ("generate a key, generate a secret, create the secret in k8s").
2. We're doing a few similar operations ("check if X exists, if not create it")
3. Some of the operations we're doing are mirrored in the place this function is called ("check if TLSSecret exists, if not call `doReconcileCertificateSecret`")

WDYT about:

- Unifying all the certificate secret operations into a single function
- Splitting out the individual "create X if not exists" operations into separate functions, so the high level logic is easier to follow? (still valuable to have separation of object generation and k8s operations -- in this case, I'm asking to have glue functions that do "check if object exists; if not, generate it with existing function, and create in k8s")

?

---

(of course, `doReconcile` already commits these sins in abundance. We'd like to refactor that, but haven't gotten around to it — just trying to make sure that at least new code is up to higher standards :smile:)

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1568,6 +1820,160 @@ func podSpec(
 	return pod, nil
 }
 
+func certSpec(
+	vm *vmv1.VirtualMachine,
+) *certv1.Certificate {
```
sharnoff, line 1825: nit: Can be all on one line?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1568,6 +1820,160 @@ func podSpec(
...
+		Spec: certSpec,
+	}
+}
+
+func tmpKeySecretSpec(
+	vm *vmv1.VirtualMachine,
+	key crypto.PrivateKey,
+) (*corev1.Secret, error) {
+	runnerVersion := api.RunnerProtoV1
+	labels := labelsForVirtualMachine(vm, &runnerVersion)
```
sharnoff, line 1876: suggestion: Do we have labels on the SSH secret that we create? if no, I propose we leave them out here, as it's added complexity and we're unlikely to uphold the guarantees we typically make (that labels are continuously propagated from the VM to the object — which _is_ what we do for pods)

\+ Same thing re: certificate request, certificate, and the TLS secret

sharnoff (commented):
Initial review -- lots of comments, but overall looks ok enough.

I didn't look at `doReconcileCertificateSecret` very deeply, figured I'd wait until after refactoring

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -142,6 +142,20 @@ type VirtualMachineSpec struct {
 	// +optional
 	EnableSSH *bool `json:"enableSSH,omitempty"`
 
+	// The CertificateIssuer for the certificates issued to this VM
+	// +optional
+	TlsCertificateIssuer *string `json:"tlsCertificateIssuer,omitempty"`
+
+	// If a TlsCertificateIssuer is provided, this is required to set the duration that the certificate should
+	// be valid for before expiring
```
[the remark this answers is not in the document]
conradludgate: Ah, smart

pkg/neonvm/controllers/vm_controller.go
conradludgate: >  just trying to make sure that at least new code is up to higher standards

aww man, and there was me thinking I could get away with copy-pasting everything :D

pkg/neonvm/controllers/vm_controller.go
conradludgate: `gofmt` is weird 🤔

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -142,6 +142,20 @@ type VirtualMachineSpec struct {
 	// +optional
 	EnableSSH *bool `json:"enableSSH,omitempty"`
 
+	// The CertificateIssuer for the certificates issued to this VM
+	// +optional
+	TlsCertificateIssuer *string `json:"tlsCertificateIssuer,omitempty"`
```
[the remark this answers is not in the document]
sharnoff: ~~(tbf, the rust styling is much better...)~~

pkg/neonvm/controllers/vm_controller.go
conradludgate: Yeah, I see no labels.

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -380,6 +390,41 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 		return errors.New("DIMMSlots memory provider is deprecated and disabled")
 	}
 
+	enableTLS := vm.Spec.TlsCertificateIssuer != nil
+
+	certSecret := &corev1.Secret{}
+
+	if enableTLS {
```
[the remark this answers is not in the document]
conradludgate: The annoying part here is that we need to reconcile TLS when the VM is either pending or running (to handle renewals). I guess it can be refactored into a function that both cases call into.

go.sum
```diff
@@ -3,7 +3,7 @@ cloud.google.com/go/compute v1.23.0 h1:tP41Zoavr8ptEqaW6j+LQOnyBBhO7OkOMAGrgLopT
 cloud.google.com/go/compute v1.23.0/go.mod h1:4tCnrn48xsqlwSAiLf1HXMQk8CONslYbdiEZc9FEIbM=
 cloud.google.com/go/compute/metadata v0.2.3 h1:mg4jlk7mCAj6xXp9UJ4fjI9VUI5rubuGBW5aJ7UnBMY=
 cloud.google.com/go/compute/metadata v0.2.3/go.mod h1:VAV5nSsACxMJvgaAuX6Pk2AawlZn8kiOGuCv6gTkwuA=
-github.com/Azure/azure-sdk-for-go v66.0.0+incompatible h1:bmmC38SlE8/E81nNADlgmVGurPWMHDX2YNXVQMrBpEE=
+github.com/Azure/azure-sdk-for-go v67.3.0+incompatible h1:QEvenaO+Y9ShPeCWsSAtolzVUcb0T0tPeek5TDsovuM=
```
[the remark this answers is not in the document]
conradludgate: I guess? go mod tidy keeps adding it back no matter what I do

pkg/neonvm/controllers/vm_controller.go
conradludgate: Same for annotations? 👀

sharnoff: Should be the same for annotations, yeah (i.e. don't try to match the annotations we use for the pod)

conradludgate:
> Sorry, may be it was previously discussed, but are there more context on that statement? It's a bit unclear which design decision is in question

A design decision _by cert-manager.io_

go.sum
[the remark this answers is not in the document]
sharnoff: Fair enough :shrug:

go.mod
```diff
@@ -214,12 +217,14 @@ require (
 	k8s.io/csi-translation-lib v0.0.0 // indirect
 	k8s.io/dynamic-resource-allocation v0.30.7 // indirect
 	k8s.io/kms v0.30.7 // indirect
+	k8s.io/kube-aggregator v0.27.1 // indirect
```
sharnoff, line 220: This kube-aggregator dependency should be v0.30.x, not v0.27.1 (to match our other deps). Checking with `go mod why`, shows:

```sh-session
$ go mod why -m k8s.io/kube-aggregator
# k8s.io/kube-aggregator
github.com/neondatabase/autoscaling/pkg/neonvm/controllers
github.com/cert-manager/cert-manager/pkg/util/pki
github.com/cert-manager/cert-manager/pkg/api/util
github.com/cert-manager/cert-manager/pkg/logs
github.com/cert-manager/cert-manager/pkg/api
k8s.io/kube-aggregator/pkg/apis/apiregistration/v1
```

So let's update cert-manager to a version that matches our k8s deps? Looks like `v.1.15.x` has the versoin we want — see e.g. https://github.com/cert-manager/cert-manager/blob/v1.15.0/go.mod#L52-L57

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -642,6 +670,13 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 					"Memory in spec", memorySizeFromSpec)
 				vm.Status.Phase = vmv1.VmScaling
 			}
+
+			// check if the certificate needs renewal for this running VM.
+			if enableTLS {
+				if _, err := r.reconcileCertificateSecretIfNeeded(ctx, vm); err != nil {
+					return err
+				}
+			}
```
sharnoff, line 679: This is currently in the `VmRunning` case, right? IIUC, that means we wouldn't update the certificate while the VM is migrating or if stuck scaling for a long time.

For now I think it does need to be outside big switch statement, like it was originally (and, a single function is much cleaner — I think my comment here no longer holds true https://github.com/neondatabase/autoscaling/pull/1213#discussion_r1952412125)

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "Failed to get vm-runner certificate Secret")
+		return nil, err
+	} else if err != nil /* not found */ {
+		msg := fmt.Sprintf("VirtualMachine %s certificate secret %s not found", vm.Name, vm.Status.TLSSecretName)
+		r.Recorder.Event(vm, "Normal", "SigningCertificate", msg)
+		certSecret, err = r.doReconcileCertificateSecret(ctx, vm, nil)
+		if err != nil {
+			return nil, err
+		}
```
sharnoff, line 856: suggestion: Currently, the logic here is such that we'll continue to make progress until we hit an error. Doing many steps can result in later reconcile operations working on stale data — maybe for things like this, let's actually just `return nil, err` unconditionally?

(see here for a recent thread: https://neondb.slack.com/archives/C039YKBRZB4/p1739445269901289?thread_ts=1739395085.275459)

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 	return nil
 }
 
+func (r *VMReconciler) reconcileCertificateSecretIfNeeded(ctx context.Context, vm *vmv1.VirtualMachine) (*corev1.Secret, error) {
```
sharnoff, line 840: random idea: WDYT about putting all of these certificate related functions in a separate file? There's a lot here, and I figure it might be a good opportunity to not let this humongous file grow any larger :sweat_smile:

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+		certSecret, err = r.doReconcileCertificateSecret(ctx, vm, certSecret)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return certSecret, nil
+}
+
+func (r *VMReconciler) doReconcileCertificateSecret(ctx context.Context, vm *vmv1.VirtualMachine, certSecret *corev1.Secret) (*corev1.Secret, error) {
```
sharnoff, line 872: on the naming of this function:

_In general_, reconciling is expected to be idempotent. I think it makes sense to have an idempotent function responsible for managing the certificate, but IIUC that's not quite what this does.

I _think_ if we merge `reconcileCertificateIfNeeded` with this function, we'll get something closer to that — and in this case, I think it makes sense (even if it may be a bit unwieldy), because it encapsulates the entire certificate lifecycle and update process into a single stream of logic.

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+	if len(certificateReq.Status.Certificate) == 0 {
+		// we cannot yet update the cert secret.
+		// return it untouched.
+		return certSecret, nil
+	}
+
+	// we have a certificate and the corresponding private key
+	// create/update the proper certificate secret and delete the tmp secret
+	if certSecret == nil {
+		certSecret, err = r.createTlsSecret(ctx, vm, key, certificateReq)
```
sharnoff, line 915: question(blocking): Thinking about idempotency — IIUC, if we get a spurious error on secret creation here (e.g., client-side timeout, but it actually went through), then we wouldn't (ever?) set `TLSRenewAt` or delete the temp secret, because the certificate already exists

Is that an accurate understanding? if so, any ideas to resolve it?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+	return certSecret, nil
+}
+
+func (r *VMReconciler) createTlsTmpSecret(ctx context.Context, vm *vmv1.VirtualMachine) (*corev1.Secret, error) {
+	log := log.FromContext(ctx)
+
+	// create a new key for this VM
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		log.Error(err, "Failed to generate private key for VirtualMachine")
```
sharnoff, line 944: suggestion: We have multiple places where we generate private keys (both here, and for SSH keys). Let's have some more context in this log line?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -802,6 +837,217 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
...
+		return nil, err
+	}
+
+	// Define the secret
+	tmpKeySecret, err := r.tmpKeySecretForVirtualMachine(vm, key)
+	if err != nil {
+		log.Error(err, "Failed to define new Secret resource for VirtualMachine")
```
sharnoff, line 951: suggestion: Similar to above, we have multiple places where we create secrets for VMs (both here and for SSH keys) - let's have some more context in this log line?

mikhail-sakhnov:
>Enabling this flag cluster wide would fix things for neonvm, but the outcome is not ideal for our other services that rely on cert-manager.


I think one of the options as well is to have a controller for secrets to clean them up, if they orpaned, instead of re-wrtiting cert manager logic.

pkg/neonvm/controllers/vm_controller.go
conradludgate: >  then we wouldn't (ever?) set TLSRenewAt

we ignore the deletion errors at the very least, so they don't block the remainder of the flow

pkg/neonvm/controllers/vm_controller.go
conradludgate: I made some changes, I am not entirely sure if that's what you had in mind. I like it less like it is now

pkg/neonvm/controllers/vm_controller.go
conradludgate: Do you mean that when we renew the certificate, we should force a reconcile failure to trigger a re-reconcile with fresher data?

pkg/neonvm/controllers/vm_cert_controller.go
```diff
@@ -0,0 +1,341 @@
+package controllers
```
sharnoff, line 1: suggestion: `vm_controller.go` and `vmmigration_controller.go` are named as such because they define the `Controller` for these types. Given it's not quite the same for certificates, maybe let's drop `controller` from the filename?

pkg/neonvm/controllers/vm_cert_controller.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil /* not found */ {
+		msg := fmt.Sprintf("VirtualMachine %s TLS secret %s not found", vm.Name, vm.Status.TLSSecretName)
+		r.Recorder.Event(vm, "Normal", "SigningCertificate", msg)
+
+		notfound = true
+	} else if time.Now().After(vm.Status.TLSRenewAt.Time) /* due for renewal */ {
+		msg := fmt.Sprintf("VirtualMachine %s TLS secret %s is due for renewal", vm.Name, vm.Status.TLSSecretName)
+		r.Recorder.Event(vm, "Normal", "SigningCertificate", msg)
+	} else /* certificate exists and does not need renewal */ {
+		return certSecret, nil
```
sharnoff, line 52: question(blocking): If we early-return here, we don't clean up the temp secret, right?

pkg/neonvm/controllers/vm_cert_controller.go
```diff
@@ -0,0 +1,341 @@
...
+	} else if err != nil /* not found */ {
+		tmpKeySecret, err = r.createTlsTmpSecret(ctx, vm)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	key, err := pki.DecodePrivateKeyBytes(tmpKeySecret.Data[corev1.TLSPrivateKeyKey])
+	if err != nil {
+		log.Error(err, "Failed to decode TLS private key")
```
sharnoff, line 70: question(blocking): is not returning here intentional? If so, what do we expect to happen by falling through?

pkg/neonvm/controllers/vm_cert_controller.go
```diff
@@ -0,0 +1,341 @@
...
+	msg = fmt.Sprintf("VirtualMachine %s CertificateRequest %s was deleted", vm.Name, certificateReq.Name)
+	r.Recorder.Event(vm, "Normal", "Deleted", msg)
+}
+
+func certSpecCSR(vm *vmv1.VirtualMachine) (*x509.CertificateRequest, error) {
+	certSpec := certv1.CertificateSpec{
+		CommonName: vm.Spec.TLS.ServerName,
+		DNSNames:   []string{vm.Spec.TLS.ServerName},
+		PrivateKey: &certv1.CertificatePrivateKey{
+			// TODO: can we support Ed25519?
```
sharnoff, line 236: question: What's the context on this TODO? will it be logged as an issue somewhere? if no, let's remove?

sharnoff (commented):
A couple small things left. Should be good to go once addressed, I think

pkg/neonvm/controllers/vm_cert_controller.go
conradludgate: now returning an error

pkg/neonvm/controllers/vm_cert_controller.go
conradludgate: Removed

sharnoff (approved):
lgtm - nice!

Issue #1220, opened 2025-01-27
Epic: Scaling event reporting follow-ups


sharnoff:
ref https://neondb.slack.com/archives/C03TN5G758R/p1737753872393229?thread_ts=1737751496.835499

Basically, there are some follow-up changes to #1107 we should make to improve the efficiency of our scaling event reporting.

- [x] #1221
- [x] #1226
- [x] #1227

sharnoff:
With #1227 merged, I think that's the last of it, and so this issue is done :)

PR #1221, opened 2025-01-27
agent: Send last billing/scaling events before shutdown

sharnoff:
Follow-up to https://github.com/neondatabase/autoscaling/pull/1107, ref https://github.com/neondatabase/autoscaling/issues/1220.

In short, this change:

1. Moves `reporting.EventSink`'s event sender thread management into a new `Run()` method
2. Calls `(EventSink).Run()` from `(billing.MetricsCollector).Run()`, appropriately canceling when done generating events, and waiting for it to finish before returning.
3. Adds `(scalingevents.Reporter).Run()` that just calls the underlying `(EventSink).Run()`, and calls *that* in the entrypoint's taskgroup.

Or in other words, exactly what was suggested here: https://github.com/neondatabase/autoscaling/pull/1221#discussion_r1931155991


pkg/reporting/sink.go
```diff
@@ -3,21 +3,31 @@ package reporting
...
+	ctx context.Context,
+	logger *zap.Logger,
+	tg taskgroup.Group,
+	metrics *EventSinkMetrics,
+	clients ...Client[E],
+) *EventSink[E] {
 	var queueWriters []eventQueuePusher[E]
-	signalDone := make(chan struct{})
+
+	senderCtx, cancelSenders := context.WithCancel(ctx)
```
Omrigan, line 30: It seems to me you don't really need `Finish()`, `done` and `senderCtx`. You can just pass `ctx` to `senderLoop` and have it the only way to stop sender loop.

pkg/reporting/sink.go
```diff
@@ -28,15 +38,18 @@ func NewEventSink[E any](logger *zap.Logger, metrics *EventSinkMetrics, clients
 			client:           c,
 			metrics:          metrics,
 			queue:            qr,
-			done:             signalDone,
 			lastSendDuration: 0,
 		}
-		go sender.senderLoop(logger.Named(fmt.Sprintf("send-%s", c.Name)))
+		taskName := fmt.Sprintf("send-%s", c.Name)
+		tg.Go(taskName, func(_ *zap.Logger) error {
```
Omrigan, line 44: Already discussed, that my approach would be is to make functions blocking:
1. Add a blocking `Run(ctx)` method on `EventSink`.
2. Call it from `Run` method of `MetricsCollector`.
3. Add a `Run(ctx)` to `Reporter`, and call `EventSink.Run()` from it.
4. Call `Reporter.Run` in top-level taskgroup in entrypoint.go.

That's what pretty much described in Google's style guide: [goroutine lifetime](https://google.github.io/styleguide/go/decisions#goroutine-lifetimes) and [synchronous functions](https://google.github.io/styleguide/go/decisions#synchronous-functions). Uber styleguide only contains the provision about [controlling goroutine's lifetime](https://github.com/uber-go/guide/blob/master/style.md#dont-fire-and-forget-goroutines), but doesn't say anything about blocking vs non-blocking design.

Side note: WDYT about adopting one of this styleguides for our repo?

sharnoff: Thought about this for a bit -- makes a lot of sense. I've updated this PR with basically exactly what you've described.

re: adopting as a style guide, I think that makes sense, but let's get buy-in from the rest of the team.

pkg/reporting/sink.go
sharnoff: Resolving as this is no longer relevant because I changed the approach.

pkg/agent/entrypoint.go
```diff
@@ -88,6 +88,10 @@ func (r MainRunner) Run(logger *zap.Logger, ctx context.Context) error {
 	}
 
 	tg := taskgroup.NewGroup(logger, taskgroup.WithParentContext(ctx))
+	tg.Go("scalingevents-run", func(logger *zap.Logger) error {
```
sharnoff, line 91: One thought I had: how do we handle panics? ideally, a panic from this task (or, any of the tasks in its own taskgroup) triggers shutdown of the autoscaler-agent.

We can set the top-level taskgroup's panic handler to cancel the parent context, but doing that from the child taskgroup is a little trickier.

Going to think about this, but cc @Omrigan in case you have thoughts.

pkg/agent/billing/billing.go
```diff
@@ -96,22 +97,33 @@ func (mc *MetricsCollector) Run(
...
-	logger = logger.Named("collect")
-
 	collectTicker := time.NewTicker(time.Second * time.Duration(mc.conf.CollectEverySeconds))
 	defer collectTicker.Stop()
 	// Offset by half a second, so it's a bit more deterministic.
 	time.Sleep(500 * time.Millisecond)
 	accumulateTicker := time.NewTicker(time.Second * time.Duration(mc.conf.AccumulateEverySeconds))
 	defer accumulateTicker.Stop()
 
+	// Create a new taskgroup to manage mc.sink.Run() -- we want to run the event senders in the
```
Omrigan, line 107: I would go one step further here:
 - Extract a new function `.runCollector()`
 - In this function do 
```
tg := taskgroup.NewGroup(logger)

tg.Go("collector-run", func(logger *zap.Logger) error {
	err := mc.collectorRun(ctx)
	cancelSink()
	return err
})

tg.Go("sink-run", func(logger *zap.Logger) error {
	mc.sink.Run(tg.Ctx())
	return nil
})

return tg.Wait()
```

See also https://github.com/neondatabase/autoscaling/pull/1235.

pkg/agent/entrypoint.go
Omrigan: UPD: `tg.Ctx()` can be used here as well - just return the error from the task when you want for the entire group to finish. See also https://github.com/neondatabase/autoscaling/pull/1235

pkg/reporting/sink.go
```diff
@@ -3,43 +3,84 @@ package reporting
...
+
+			for _, sender := range senders {
+				taskName := fmt.Sprintf("send-%s", sender.client.Name)
+				tg.Go(taskName, func(logger *zap.Logger) error {
+					sender.senderLoop(tg.Ctx(), logger)
+					return nil
+				})
+			}
+
+			_ = tg.Wait() // no need to check error, they all return nil.
```
Omrigan, line 56: Strictly speaking panic can be returned from here.

pkg/reporting/sink.go
```diff
@@ -3,43 +3,84 @@ package reporting
...
+		}
+	} else {
+		// Special case when there's no clients -- we want our run function to just wait until the
+		// context is complete, matching what the behavior *would* be if there were actually sender
+		// threads we were waiting on.
+		runSenders = func(ctx context.Context) {
+			<-ctx.Done()
```
Omrigan, line 63: What happens if you don't wait here, but immediately return?

pkg/reporting/sink.go
```diff
@@ -3,43 +3,84 @@ package reporting
...
 	}
 }
 
+// Run executes the client threads responsible for actually pushing enqueued events to the
+// appropriate places.
+//
+// The clients will periodically push events until the context expires, at which point they will
+// push any remaining events. Run() only completes after these final events have been pushed.
+//
+// Calling Run() more than once is unsound.
```
Omrigan, line 79: Side note: does it have to be? I imagine semantics can be here that on any object with `Run(ctx)` we could call it, and it would run as long as ctx is not cancelled. And then we can run it again with another ctx.

Probably not much of a real usecase. One example is that we could have had a separate control over long-running processes, something like "start/stop/restart events export" knobs.

sharnoff: I think in theory it doesn't have to be, but "can't be called more than once" was a nice assumption to have available in the implementation.

I'd be open to changing that in the future if we have a use case for it, though

pkg/reporting/sink.go
sharnoff: I _think_ nothing bad happens, but I figured it's worthwhile to preserve the behavior of waiting until context completion. Might not be worth the complexity though... WDYT?

pkg/agent/billing/billing.go
sharnoff: Makes sense, implemented in f0ba926ca848f93dffb9f6916ea7de137c9e3959

pkg/agent/entrypoint.go
sharnoff: Makes sense -- Seems like switching to "just" propagating errors is sufficient. Did that with ed73d72b9d1405cc6a99a66cd12e3513f3bf44cd

PR #1226, opened 2025-01-29
reporting: Continuously send full batches of events

sharnoff:
Part of #1220.

In cases where (a) we expect batches to be very large, and (b) we expect that some autoscaler-agent instances may end up with many batches of events between reporting periods, we might see excessive memory usage from those events.

So instead of pushing events exactly every push period, we should push *at least* every push period, and exactly when the next batch is full if it's full before the push period is reached.

### Implementation

The new implementation replaces the event queue between `reporting.EventSink` and the event sender threads with a queue-like construct that collects events into discrete batches. This new construct is `eventBatcher`, in the new `batcher.go` file.

Then, in all of the places where the event sender previously dealt with collecting chunks of events, it instead operates on a single "batch" of events at a time.

The gauge metric showing the number of events in the queue still works in the same way it did before, counting both events in batches that haven't been finalized, and finalized batches that haven't been sent yet.

---

~~**Note: This PR builds on #1221, and must not be merged before it.**~~


pkg/reporting/batcher_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	var notified bool
+	notify := func() {
+		notified = true
+	}
+	gauge := prometheus.NewGauge(prometheus.GaugeOpts{})
+
+	batcher := newEventBatcher[string](targetBatchSize, notify, gauge)
+
+	// First batch:
```
mikhail-sakhnov, line 21: Do those batches rely on each other? if no, I suggest to wrap them in t.Run calls, it would be a bit nicer to read in the unittest run report

pkg/reporting/send.go
```diff
@@ -37,8 +41,10 @@ type eventSender[E any] struct {
 }
 
 func (s eventSender[E]) senderLoop(ctx context.Context, logger *zap.Logger) {
-	ticker := time.NewTicker(time.Second * time.Duration(s.client.BaseConfig.PushEverySeconds))
-	defer ticker.Stop()
+	heartbeat := time.Second * time.Duration(s.client.BaseConfig.PushEverySeconds)
```
mikhail-sakhnov, line 44: nitpick (out of context): BaseConfig could have those fields defines as time.Duration which supports json marshaling

pkg/reporting/send.go
```diff
@@ -47,10 +53,27 @@ func (s eventSender[E]) senderLoop(ctx context.Context, logger *zap.Logger) {
 		case <-ctx.Done():
 			logger.Info("Received notification that events submission is done")
 			final = true
-		case <-ticker.C:
+			// finish up any in-progress batch, so that we can send it before we exit.
+			s.queue.finishOngoing()
+		case <-s.batchComplete.Wait():
+			s.batchComplete.Awake() // consume this notification
```
mikhail-sakhnov, line 59: question: why is that needed? 

sidenote, nitpick: comment is a great example of "what we do" statement while it would be more useful to have "why we do" statement

pkg/reporting/send.go
```diff
@@ -37,8 +41,10 @@ type eventSender[E any] struct {
 }
 
 func (s eventSender[E]) senderLoop(ctx context.Context, logger *zap.Logger) {
-	ticker := time.NewTicker(time.Second * time.Duration(s.client.BaseConfig.PushEverySeconds))
-	defer ticker.Stop()
+	heartbeat := time.Second * time.Duration(s.client.BaseConfig.PushEverySeconds)
+
+	timer := time.NewTimer(heartbeat)
```
mikhail-sakhnov, line 46: question, suggestion: Since we are effectively have a loop here, should it be ticker instead of timer?

pkg/reporting/batcher.go
```diff
@@ -0,0 +1,124 @@
...
+// the batcher.
+//
+// Once done with these batches, you should call (*eventBatcher[E]).dropCompleted() to remove them
+// from future consideration.
+func (b *eventBatcher[E]) peekCompleted() []batch[E] {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	tmp := make([]batch[E], len(b.completed))
+	copy(tmp, b.completed)
```
mikhail-sakhnov, line 89: question: why do we copy slice every time we peek into it?

pkg/reporting/send.go
```diff
@@ -59,51 +82,54 @@ func (s eventSender[E]) senderLoop(ctx context.Context, logger *zap.Logger) {
...
+	// terminate. For the most part, that's ok: worst-case, we miss that the parent context has
+	// expired, which isn't the end of the world (eventually the autoscaler-agent will just be
+	// force-killed). Any long-running call to this function will be reported by
+	// s.metrics.lastSendDuration as we go (provided the request timeout isn't too long), so we
+	// should get observability for it either way.
 	for {
-		if size := s.queue.size(); size != 0 {
-			logger.Info("Current queue size is non-zero", zap.Int("queueSize", size))
-		}
+		batches := s.queue.peekCompleted()
```
mikhail-sakhnov, line 108: question: It looks like we never use batches as a collection of batches, rather sticking with the first completed batch only. Should the queue API be changed to something like "peekLatestCompleted" and "peekCompletedCount"?

mikhail-sakhnov (approved):
looks legit, left few questions, none of them I consider as a blocker for merging.

pkg/reporting/send.go
sharnoff: IIUC `time.Duration` doesn't have fancy JSON marshaling (it's just an int64 of nanoseconds), unless I'm missing something?

sharnoff:
Updates

- Rebased onto main —b6d3c656112d154e13c8976e2cfc403e0440e9bf..97adf1d26cf531feaee756816d28f13b37363ddc ([diff](https://github.com/neondatabase/autoscaling/compare/b6d3c656112d154e13c8976e2cfc403e0440e9bf..97adf1d26cf531feaee756816d28f13b37363ddc))
- Some review changes — 97adf1d26cf531feaee756816d28f13b37363ddc..a8e7dfeddf67149981a07b122957f68071ca4e05 ([diff](https://github.com/neondatabase/autoscaling/compare/97adf1d26cf531feaee756816d28f13b37363ddc..a8e7dfeddf67149981a07b122957f68071ca4e05))

pkg/reporting/send.go
sharnoff: Good point. Fixed in 6425e140142c2365ee3be7bf60957d9808bcf7da

pkg/reporting/batcher.go
sharnoff: Honestly can't remember. Made obsolete by 6425e140142c2365ee3be7bf60957d9808bcf7da as part of https://github.com/neondatabase/autoscaling/pull/1226#discussion_r1963531490

pkg/reporting/send.go
sharnoff: It's needed because that's the interface for that notification type.

Although, thinking about it again, I think we can use something simpler and just read from a buffered channel.

Implemented that fix in 50923447360fda103dddd566472f5880963c7412.

pkg/reporting/batcher_test.go
sharnoff: Hm... the first batch is isolated, but with the others I've intentionally mixed the reading of batch N with the writing of batch N+1.

I took a stab at implementing this — pushed it as 707a9990d1234549186fd7cf5c4594ed5b61cb49 before reverting. Can go with that if you think it's better, but at first glance seems not worth it to me.

pkg/reporting/send.go
mikhail-sakhnov: omg, you are right, I messed it with the time.ParseDuration which I always used with custom type implementing Unmarshal method on it.

pkg/reporting/batcher_test.go
mikhail-sakhnov: >Hm... the first batch is isolated, but with the others I've intentionally mixed the reading of batch N with the writing of batch N+1.

If we rely on the order, let's make it a single case then. 

>but at first glance seems not worth it to me.

Mostly the benefit is usually coming from the fact you can run each t.Run as a separate test, plus a bit better reporting imo.

But since we have mixed read N and write N+1, doesn't make sense, I guess?

PR #1228, opened 2025-01-29
agent: Add azure blob support for scaling events

sharnoff:
Must have forgotten to include handling for azure blob storage clients as part of #1107; this change adds it.

While we're at it, unify the config checking for azure blob between billing and scaling events (like we have for S3), and also start checking `prefixInContainer` where we weren't before.

### DIFF
--- a/pkg/agent/scalingevents/clients.go
+++ b/pkg/agent/scalingevents/clients.go
@@ -12,7 +12,8 @@ import (
 )
 
 type ClientsConfig struct {
-	S3 *S3ClientConfig `json:"s3"`
+	AzureBlob *AzureBlobStorageClientConfig `json:"azureBlob"`
+	S3        *S3ClientConfig               `json:"s3"`
 }
 
 type S3ClientConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/agent/scalingevents/clients.go
```diff
@@ -21,11 +22,32 @@ type S3ClientConfig struct {
...
 type eventsClient = reporting.Client[ScalingEvent]
 
 func createClients(ctx context.Context, logger *zap.Logger, cfg ClientsConfig) ([]eventsClient, error) {
 	var clients []eventsClient
 
+	if c := cfg.AzureBlob; c != nil {
+		generateKey := newBlobStorageKeyGenerator(c.PrefixInContainer)
+		client, err := reporting.NewAzureBlobStorageClient(c.AzureBlobStorageClientConfig, generateKey)
+		if err != nil {
+			return nil, fmt.Errorf("error creating S3 client: %w", err)
```
petuhovskiy, line 40:
```suggestion
			return nil, fmt.Errorf("error creating azureblob client: %w", err)
```

petuhovskiy (approved):

pkg/agent/scalingevents/clients.go
sharnoff: Oops, good catch -- thanks!

PR #1235, opened 2025-01-31
util: clarify semantics for Ctx() in TaskGroup

Omrigan:


sharnoff (approved):
Makes sense - thanks!

PR #1239, opened 2025-02-03
ci: this is forks check

yatheo-me:
Check how we run workflows on Forks


fedordikarev (approved):
I see that some jobs failed due to they can't access to secrets.
Lets check if "Approve" will help there

fedordikarev (changes requested):
I checked it works, but we shouldn't merge it actually

Issue #1139, opened 2024-11-13
Move pkg/billing into autoscaler-agent


sharnoff:
*Spun off from https://github.com/neondatabase/autoscaling/pull/1078#discussion_r1795250515*

> @Omrigan 
>
> > Does it make sense to get rid of [pkg/billing] and move all the stuff to either agent/billing or reporting?
>
> @sharnoff 
>
> > I'd like to do that, yeah. Wanted to leave it outside the scope of this change for now though — will keep this comment open as a reminder.

Basically what the comment says — much of the functionality has been moved out of `pkg/billing`. All that's left is just moving the type definitions and some final utilities.

PR #1242, opened 2025-02-03
agent,billing: move pkg/billing

mikhail-sakhnov:
Move event definitions to pkg/reporting.
Move hostname helper to the pkg/agent/billing by place of usage. 
Move schema file billing-v1.yaml to s3 e2e test by place of single usage.

#1139


pkg/agent/billing/hostname.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+)
+
+var hostname string
+
+func init() {
+	var err error
+	hostname, err = os.Hostname()
+	if err != nil {
+		hostname = fmt.Sprintf("unknown-%d", rand.Intn(1000))
```
mikhail-sakhnov, line 15: I am a bit concerned about this code, what will happen in a very rare but still possible case if we have the same generated hostname for 2 different machines? should it be uuid instead?

mikhail-sakhnov: @sharnoff @Omrigan

sharnoff:
Hm, `pkg/reporting` is meant to be a more general "pushing events [of some kind] to HTTP/S3/Azure blob". Maybe we should move `pkg/{billing => agent/billing}` ?

pkg/agent/billing/hostname.go
mikhail-sakhnov: created an issue out of that.

pkg/agent/billing/billing.go
```diff
@@ -276,10 +275,10 @@ func (s *metricsState) drainEnqueue(
 		history.finalizeCurrentTimeSlice()
 
 		countInBatch += 1
-		enqueue(logAddedEvent(logger, billing.Enrich(now, hostname, countInBatch, batchSize, &billing.IncrementalEvent{
+		enqueue(logAddedEvent(logger, enrich(now, hostname, countInBatch, batchSize, &IncrementalEvent{
 			MetricName:     conf.CPUMetricName,
-			Type:           "", // set by billing.Enrich
-			IdempotencyKey: "", // set by billing.Enrich
+			Type:           "", // set by reporting.Enrich
+			IdempotencyKey: "", // set by reporting.Enrich
```
sharnoff, line 281: nit: remove reference to `reporting`? (+ same below)

pkg/agent/billing/events.go
```diff
@@ -69,3 +70,20 @@ func (e *IncrementalEvent) setType() {
 func (e *IncrementalEvent) getIdempotencyKey() *string {
 	return &e.IdempotencyKey
 }
+
+// уnrich sets the event's Type and IdempotencyKey fields, so that users of this API don't need to
+// manually set them
+func enrich[E Event](now time.Time, hostname string, countInBatch, batchSize int, event E) E {
```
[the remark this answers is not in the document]
mikhail-sakhnov: done

pkg/agent/billing/billing.go
mikhail-sakhnov: dropped

PR #1244, opened 2025-02-04
neonvm: add architecture field to vm.spec

mikhail-sakhnov:
Add spec.cpuArchitecture to use for scheduling pods instead of relying on spec.affinity.
Use spec.cpuArchitecture value while building the vm pod spec affinity.

Extracted from #1123


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1145,7 +1146,17 @@ func affinityForVirtualMachine(vm *vmv1.VirtualMachine) *corev1.Affinity {
 	if a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
 		a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
 	}
-	// if NodeSelectorTerms list is empty - add default values (arch==amd64 or os==linux)
+
+	// if kubernetes.io/arch is set via vm.Spec.Affinity, override it with the target architecture
+	for _, term := range a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
```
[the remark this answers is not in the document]
sharnoff: > there might be instances in storage with spec.Affinity and spec.CPUArchitecture has different values

I think this shouldn't be possible because there's no computes running on ARM currently, right?

Assuming that's the case: I would prefer actually to not bother checking against `kubernetes.io/arch`, and instead expect control plane to set the architecture field -- that way, we ge ta simpler API, and if they misuse it that's on them :sweat_smile:

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -107,6 +107,9 @@ type VirtualMachineSpec struct {
 
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
+	// +kubebuilder:default:=amd64
+	CPUArchitecture CPUArchitecture `json:"cpuArchitecture,omitempty"`
```
sharnoff, line 111: A couple thoughts on this field:

1. OTOH `CPUArchitecture` feels like a non-standard name. WDYT about `Arch` / `Architecture` / `TargetArch` / `TargetArchitecture` ?
2. Automatically setting this value to match the current machine may be difficult. WDYT about e.g. making this an array of acceptable options, and having a "status" field on the object to show what architecture it's currently running on?

(genuinely asking your opinion — not sure either way :sweat_smile:)

sharnoff (commented):
Just a couple thoughts

pkg/neonvm/controllers/vm_controller.go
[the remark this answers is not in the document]
mikhail-sakhnov: >I think this shouldn't be possible because there's no computes running on ARM currently, right?

I mean technically, nobody prevents you try to create VM with 
```
spec:
  affinity:
    nodeSelector:
      kubernetes.io/arch: "arm64"
  cpuArchitecture: amd64
```

with the approach as in the current PR at the moment it will be created and stored like this, but the actual affinity in the pod will always be value from spec.cpuArchitecture.


If we move logic from here to webhook and override affinity there, trying to apply the manifest above will result in saving manifest like
```
spec:
  enableNetworkMonitoring: true
  enableSSH: true
  affinity:
    nodeSelector:
      kubernetes.io/arch: "amd64" # overriden with the value from cpuArchitecture
  cpuArchitecture: amd64
```

I like the second approach more, but it means having more logic in the webhook.

Ideally I would love to have an approach with returning a error from webhook if affinity section has selectors for `kubernetes.io/arch` but that might be an issue if we have any VMs in cluster that already have set spec.affinity (due to some reason).


>that way, we ge ta simpler API, and if they misuse it that's on them 😅

Can't say it doesn't work :smile:

neonvm/apis/neonvm/v1/virtualmachine_types.go
mikhail-sakhnov: >Automatically setting this value to match the current machine may be difficult. WDYT about e.g. making this an array of acceptable options, 

I think it is already an array of options, because it is defined as enum https://github.com/neondatabase/autoscaling/pull/1244/files#diff-5360cd1b7d5cabb8afe6cbad91a67289c46afd21dd930b8f51a1ec80833a4ceaR999? Did I understand the question correctly? 


>and having a "status" field on the object to show what architecture it's currently running on?

Good idea, let me add 👍

mikhail-sakhnov: Renamed the field, and added it to the `kubectl get -owide` output.

pkg/neonvm/controllers/vm_controller.go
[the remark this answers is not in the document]
sharnoff: Honestly affinity seems really complicated, and I'd rather avoid inspecting what the user provided if we don't have to :sweat_smile: — especially because of the edge cases we'd get from _starting_ to care about it.

But I agree that all else being equal, it'd be nice for the creation webhook to check for a weird configuration :)

neonvm/apis/neonvm/v1/virtualmachine_types.go
mikhail-sakhnov: Hm-hm. I am thinking about the next scenario/

VM starts with that spec, pod is created for arm64
Later compute is halted for the endpoint.
Later2, a customer connects to the endpoint, we spin a new pod, now it is scheduled on amd64.

Will it work correctly with the data stored while the arm pod was active? If no, we will need to store the used architecture in 3rd place (like, status or metainformation) and basically always schedule this compute on the first architecture.

sharnoff: From discussion offline: Multiple options for architecture are probably only needed for local development -- maybe let's just make it a single nullable value, and have null imply "I don't care which architecture it runs on"

(IIRC?)

mikhail-sakhnov: sounds reasonable to me

mikhail-sakhnov: I changed field to be nullabe. If it is not null, we add architecture affinity match. If it is null, we don't add the match.
I also dropped attempts to override affinity in spec.affinity.

sharnoff:
Sorry, a couple higher level things I realized w.r.t. rollout:

1. What happens to old VMs? Will k8s automatically default them to the new behavior? How will we make sure that they continue to get x86 if they restart?
2. What happens on rollback?

Given these: Maybe worthwhile to (a) remove the CRD default to x86, and (b) interpret nil as x86, until all clients have been updated? And then we can update all VM objects to set `TargetArchitecture` to x86 and change the controller to have nil handling in the way that we want.

mikhail-sakhnov:
@sharnoff 

>What happens to old VMs? Will k8s automatically default them to the new behavior? How will we make sure that they continue to get x86 if they restart?


So in general, the old VMs have no materialized value for the field in the kube-api storage. When deploying a version with spec.targetArchitecture annotated with "kubebuilder:default:amd64", the lack of the set value will be interpreted as "amd64". 

Indeed, if we rollback to the version without spec.targetArchitecture, the behaviour would be to ignore spec.targetArchutecture even if it is materialized in the kube-api storage. In practice, that would lead to always create a pod with nodeArchitecture affinity equal controller architecture.

If it feels safer, I changed the default value for spec.targetArchitecture to be nil and add code to explicitly set it to amd64 in vm_controller.go, by this we will have materialized values for all VMs after at least reconciliation. 

However, in case of rollback, it doesn't change anything, since if we rollback, we rollback to the version which has no spec.targetArchitecture and function `affinityForVirtualMachine` will behave based on the controller architecture.

Did I answer your question, or should I elaborate deeper on something?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1136,6 +1152,7 @@ func annotationsForVirtualMachine(vm *vmv1.VirtualMachine) map[string]string {
 
 func affinityForVirtualMachine(vm *vmv1.VirtualMachine) *corev1.Affinity {
 	a := vm.Spec.Affinity
+
```
sharnoff, line 1155: nit: unintended diff?

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1145,23 +1162,29 @@ func affinityForVirtualMachine(vm *vmv1.VirtualMachine) *corev1.Affinity {
 	if a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
 		a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
 	}
-	// if NodeSelectorTerms list is empty - add default values (arch==amd64 or os==linux)
+
+	// if NodeSelectorTerms list is empty - add default values (arch==vm.Spec.Affinity or os==linux)
```
sharnoff, line 1166: question: I know it's the previous behavior, but isn't this not correct? Not sure how I'm only just noticing it, but that means we won't respect the architecture if there's any node selector terms, right?

Can we change this to always add ours?

sharnoff (approved):
Looks good, modulo two comments

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: probably automatically added it for readability

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: But didn't we decided to drop it last week? Or did I understand you incorrectly?

https://github.com/neondatabase/autoscaling/pull/1244#discussion_r1941227562

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1145,25 +1162,30 @@ func affinityForVirtualMachine(vm *vmv1.VirtualMachine) *corev1.Affinity {
...
+					Key:      "kubernetes.io/arch",
+					Operator: "In",
+					// vm.Spec.TargetArchitecture is guaranteed to be set by reconciler loop
+					Values: []string{string(*vm.Spec.TargetArchitecture)},
+				},
+			},
+		},
+	)
+
+	a.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = nodeSelectorTerms
```
mikhail-sakhnov, line 1187: after append slice might be re-allocated so should be overridden in the structure

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: Discussed in 1o1/DM, changed accordingly.

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: rolled back

pkg/neonvm/controllers/vm_controller.go
mikhail-sakhnov: Tbh, I don't see a significant difference in terms of readability or functionality between any of the last 2 or 3 iterations so let's go with it.

PR #1246, opened 2025-02-04
cluster-autoscaler: Try disabling anti affinity check

edude03:
After changing the over-provisioning pod deployment to have a zonal anti affinity, we noted that cluster autoscaler fails to create a node for the un-schedulable over provisioning pod. This is a known issue (see below, the comment right about the change mentions this) however this is of course not the behaviour we want. 

While this isn't a proper fix, it should "force" cluster-autoscaler to provision new nodes giving us the extra capacity we hope to gain from the over-provisioning pods (which will then of course scheduling the over-provisioning pods)


cluster-autoscaler/ca.patch
```diff
@@ -1,3 +1,103 @@
...
+ 
+@@ -195,10 +203,12 @@ func (e *BinpackingNodeEstimator) tryToScheduleOnNewNodes(
+ 			// Note that this may still fail (ex. if topology spreading with zonal topologyKey is used);
+ 			// in this case we can't help the pending pod. We keep the node in clusterSnapshot to avoid
+ 			// adding and removing node to snapshot for each such pod.
+-			if err := e.predicateChecker.CheckPredicates(e.clusterSnapshot, pod, estimationState.lastNodeName); err != nil {
++			err := e.predicateChecker.CheckPredicates(e.clusterSnapshot, pod, estimationState.lastNodeName)
++			if err != nil && err.PredicateName() != "InterPodAffinity" {
+ 				break
+ 			}
```
edude03, line 72: This is basically the essential part (even with delve I needed a bunch of printf debugging 😅 ).

So - cluster autoscaler works by taking a cluster snapshot, simulating changing the cluster state, then checks if the state changes after passing the simulated state to the kube scheduler. 

For testing adding nodes, it "samples" the node pool by finding a node in the target node group (the node group be evaluated for scale up) and duplicates it in the node list. 

This doesn't work if you have a single group that spans zones, plus an interpod(anti)affinity because the new simulated node will be in the same zone as the existing node and thus, the zone anti affinity will fail - so it decides that adding a new node would be useless. 

Patching out that particular check (or in this case saying, ignore that particular error) allows the node to be scheduled as expected, however, it's a gamble - if the node(s) created isn't in a useful zone it'll be reaped on the next iteration of the loop (technically, it's tainted pendingDeletion so it's extra useless, computes won't use it as extra capacity) 

This is obviously not a great fix, but I think until we finish our cluster layout migration it should prevent the pending over-provisioning pod issue.

sharnoff: for posterity, some discussion here: https://neondb.slack.com/archives/C03TN5G758R/p1738855560629109

cluster-autoscaler/ca.patch
```diff
@@ -1,3 +1,103 @@
...
++++ b/cluster-autoscaler/core/scaleup/orchestrator/orchestrator.go
+@@ -604,7 +604,10 @@ func (o *ScaleUpOrchestrator) SchedulablePodGroups(
+ 	var schedulablePodGroups []estimator.PodEquivalenceGroup
+ 	for _, eg := range podEquivalenceGroups {
+ 		samplePod := eg.Pods[0]
+-		if err := o.autoscalingContext.PredicateChecker.CheckPredicates(o.autoscalingContext.ClusterSnapshot, samplePod, nodeInfo.Node().Name); err == nil {
++		err := o.autoscalingContext.PredicateChecker.CheckPredicates(o.autoscalingContext.ClusterSnapshot, samplePod, nodeInfo.Node().Name)
++		// Ignore inter-pod affinity since this is not usually a reason to fail the ASG
++		// (unless the anti-affinity conflict is with a DaemonSet pod, but there's no way to tell)
++		if err == nil || err.PredicateName() == "InterPodAffinity" {
```
sharnoff, line 13: question(blocking): Took some time to dig into this — I think we can constrain this further, by ignoring the error only if `err.PredicateName() == "InterPodAffinity"` _and_ that the specific error message from the InterPodAffinity plugin is `interpodaffinity.ErrReasonAntiAffinityRulesNotMatch`

See more here: https://github.com/kubernetes/autoscaler/blob/c5199d8bc558f438db675df9a2519df38597631e/cluster-autoscaler/vendor/k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L377-L379

WDYT?

cluster-autoscaler/ca.patch
```diff
@@ -1,3 +1,103 @@
...
+ 				Pods: eg.Pods,
+diff --git a/cluster-autoscaler/estimator/binpacking_estimator.go b/cluster-autoscaler/estimator/binpacking_estimator.go
+index 229a2e264..f145a218f 100644
+--- a/cluster-autoscaler/estimator/binpacking_estimator.go
++++ b/cluster-autoscaler/estimator/binpacking_estimator.go
+@@ -108,13 +108,18 @@ func (e *BinpackingNodeEstimator) Estimate(
+ 		var err error
+ 		var remainingPods []*apiv1.Pod
+ 
++		klog.V(4).Infof("trying to schedule %d pods on existing nodes", len(podsEquivalenceGroup.Pods))
```
sharnoff, line 25: Let's remove the print debugging?

cluster-autoscaler/ca.patch
```diff
@@ -1,3 +1,103 @@
...
+ 				break
+ 			}
+ 
+@@ -195,10 +203,12 @@ func (e *BinpackingNodeEstimator) tryToScheduleOnNewNodes(
+ 			// Note that this may still fail (ex. if topology spreading with zonal topologyKey is used);
+ 			// in this case we can't help the pending pod. We keep the node in clusterSnapshot to avoid
+ 			// adding and removing node to snapshot for each such pod.
+-			if err := e.predicateChecker.CheckPredicates(e.clusterSnapshot, pod, estimationState.lastNodeName); err != nil {
++			err := e.predicateChecker.CheckPredicates(e.clusterSnapshot, pod, estimationState.lastNodeName)
++			if err != nil && err.PredicateName() != "InterPodAffinity" {
```
sharnoff, line 70: question(blocking): Should this be `if err != nil || err.PredicateName() == "InterPodAffinity"` ? or if not, what's the purpose of this change?

\+ similar modification w.r.t. also checking against the error message

cluster-autoscaler/ca.patch
sharnoff: aside: when hacking on this locally, I figured this was a reasonable approach:

```go
ignoreErr := err != nil &&
    err.PredicateName() == "InterPodAffinity" &&
    err.Message() == interpodaffinity.ErrReasonAntiAffinityRulesNotMatch
if err == nil || ignoreErr {
    // ...
```

sharnoff (commented):
Left a few thoughts, general approach is ok, as discussed.

Please also prefix your PR title with `cluster-autoscaler:`

cluster-autoscaler/ca.patch
edude03: Honestly, I'd prefer to keep it in unless you strongly think we should take it out. If we run into issue with this again it'd be a lot easier to debug with the debugging messages than of course, adding them back and the back and forth that entails. Though - we default to v=4 I believe so I'd be happy to give it a higher verbosity (5?)

cluster-autoscaler/ca.patch
edude03: Yeah, good point.

cluster-autoscaler/ca.patch
edude03: Going to lift that 👍

cluster-autoscaler/ca.patch
sharnoff: In general, I'd like to minimize the size of our patch, to avoid conflicts when updating.

So in this case, if you think the debugging is useful, let's keep it in but revisit in 1-3 months to see if we can get rid of it?

sharnoff (approved):
lgtm, pending final unresolved comment

cluster-autoscaler/ca.patch
```diff
@@ -1,3 +1,137 @@
...
++	interpodaffinity "k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity"
+ )
+ 
+ // ScaleUpOrchestrator implements scaleup.Orchestrator interface.
+@@ -604,7 +605,17 @@ func (o *ScaleUpOrchestrator) SchedulablePodGroups(
+ 	var schedulablePodGroups []estimator.PodEquivalenceGroup
+ 	for _, eg := range podEquivalenceGroups {
+ 		samplePod := eg.Pods[0]
+-		if err := o.autoscalingContext.PredicateChecker.CheckPredicates(o.autoscalingContext.ClusterSnapshot, samplePod, nodeInfo.Node().Name); err == nil {
++		err := o.autoscalingContext.PredicateChecker.CheckPredicates(o.autoscalingContext.ClusterSnapshot, samplePod, nodeInfo.Node().Name)
```
sharnoff, line 18: Let's break this into multiple lines?

sharnoff (approved):

PR #1260, opened 2025-02-13
ci/release: Fix vm-builder artifact paths

sharnoff:
Follow-up to #1259 :disappointed:

Previous release workflow attempt failed here: https://github.com/neondatabase/autoscaling/actions/runs/13294819272/job/37125168593

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -82,8 +82,8 @@ jobs:
         run: |
           # add the arch-specific suffix to the binary name, so they don't conflict when used as
           # assets attached to the release.
-          mv vm-builder-amd64/vm-builder{,-amd64}
-          mv vm-builder-arm64/vm-builder{,-arm64}
+          mv vm-builder-amd64/vm-builder vm-builder-amd64/vm-builder-amd64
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -80,14 +78,25 @@ jobs:
           # See e2e-test.yaml: the individual yamls are flattened inside the artifact.
           path: rendered_manifests
 
+      - name: rename vm-builder files
+        run: |
+          # FOR DEBUGGING ONLY:
+          ls *
+          # add the arch-specific suffix to the binary name, so they don't conflict when used as
+          # assets attached to the release.
+          mv vm-builder-amd64/vm-builder{,-amd64}
```
mikhail-sakhnov, line 87: what does "mv vm-builder-amd64/vm-builder{,-amd64}" do (without a second operand)?

mikhail-sakhnov: Could it be simplified to an expanded version? Doesn't look like it is much longer, but so much easier to read

sharnoff: Fair enough, fixed in 1984ba7b82e6d02c6a999c230609a4fd7cdeb594

mikhail-sakhnov (approved):

PR #1263, opened 2025-02-14
Remove exp/slices

conradludgate:
exp/slices has been merged into slices. it has some breaking changes though, notably it switches from an `isLess -> bool` function to a `compare -> int` function.


sharnoff (approved):
Looks good, thanks!

PR #1271, opened 2025-02-17
neonvm: Add Guest.MemhpAutoMovableRatio

petuhovskiy:
Allow to override default movable ratio that is set in neonvm-controller.

related to https://github.com/neondatabase/cloud/issues/23865


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -1342,10 +1342,16 @@ func podSpec(
 						if config.DisableRunnerCgroup {
 							cmd = append(cmd, "-skip-cgroup-management")
 						}
+
+						memhpAutoMovableRatio := config.MemhpAutoMovableRatio
+						if specValue := vm.Spec.Guest.MemhpAutoMovableRatio; specValue != nil {
+							memhpAutoMovableRatio = *specValue
+						}
```
sharnoff, line 1349: Worth using `lo.FromPtrOr` ? (from `github.com/samber/lo`)

Up to you

petuhovskiy: Tried it, the line was too long, decided not to use it.

PR #1274, opened 2025-02-18
CI(lint): golangci-lint v1.64.5

bayandin:
Bump `golangci-lint` to the latest version


Omrigan (approved):

sharnoff (approved):
thanks for bumping! any ideas why the `exhaustruct` exemption is no longer necessary for those azure sdk types?

PR #1277, opened 2025-02-21
plugin/metrics: Update metrics in-place when possible

### LINKED ISSUE #1276: Bug: Gaps under load in scheduler's node resource metrics
## Environment

Production

## Steps to reproduce

Put the scheduler under load (e.g., sustaining >60 reconcile operations per second).

## Expected result

The scheduler plugin's node resource metrics should have no gaps.

## Actual result

There's occasional, small gaps in the metrics. [For example](https://neonprod.grafana.net/goto/RiSPGycHR?orgId=1):

screenshot of grafana panel titled "percent CPU reserved", showing a single series with a gap in the middle where one datapoint is missing

## Other logs, links

- ...

sharnoff:
Fixes #1276.

Currently, the way we update these node metrics is by removing all the old ones, then adding back the current values. We do it that way so that the old values can be cleaned up when there's label changes.

However: if metrics are scraped in between removing the old and adding the new, we can end up with single-datapoint gaps for one node at a time.

So to fix this, we should avoid removing the old metrics if and only if the labels are unchanged -- which we can check just by storing the previous labels we used.

### DIFF
--- a/pkg/plugin/metrics/node.go
+++ b/pkg/plugin/metrics/node.go
@@ -1,6 +1,8 @@
 package metrics
 
 import (
+	"slices"
+
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/neondatabase/autoscaling/pkg/plugin/state"
@@ -10,6 +12,8 @@ import (
 type Node struct {
 	// InheritedLabels are the labels on the node that are directly used as part of the metrics
 	InheritedLabels []string
+
+	lastLabels []string
 
 	cpu *prometheus.GaugeVec
 	mem *prometheus.GaugeVec
@@ -22,6 +26,8 @@ func buildNodeMetrics(labels nodeLabeling, reg prometheus.Registerer) Node {
 
 	return Node{
 		InheritedLabels: labels.k8sLabelNames,
+
+		lastLabels: []string{}, // will be set when (*Node).Update() is first called
 
 		cpu: util.RegisterMetric(reg, prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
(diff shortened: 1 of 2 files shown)


mikhail-sakhnov:
question (to gain more project context): why do we delete metrics for previous node?

sharnoff:
> why do we delete metrics for previous node?

We include some node labels in the metrics, and if those node labels change, the metrics update needs to also remove the series with the old labels.

pkg/plugin/metrics/node.go
```diff
@@ -47,9 +53,10 @@ func (m *Node) Update(node *state.Node) {
 		commonLabels = append(commonLabels, value)
 	}
 
-	// Remove old metrics before setting the new ones, because otherwise we may end up with
-	// un-updated metrics if node labels change.
-	m.Remove(node)
+	if !slices.Equal(commonLabels, m.lastLabels) {
```
petuhovskiy, line 56: Hmm, I'm not sure it will work...

To me it looks like `metrics.Node` holds metrics for all the nodes, and `m.Update()` is called for each node. Nodes can have different names and labels, so `commonLabels` will be overwritten each time by a different node.

Also, modifying `m.lastLabels` concurrently can be unsafe.

petuhovskiy: As a fix, I can suggest to make a map from nodeName to nodeLabels, and check that labels haven't changed for each node separately.

sharnoff: Oh right :facepalm:

Great catch, thanks — will do that!

sharnoff: Fixed in 40028f24e155dadebf3a8f26b6eac0a559bd4e29, LMK what you think!

petuhovskiy (approved):

PR #1278, opened 2025-02-21
neonvm-controller: overhaul IPAM

Omrigan:
TBD

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -258,7 +258,7 @@ func (r *VMReconciler) doFinalizerOperationsForVirtualMachine(ctx context.Contex
 			return
 		}
 		defer ipam.Close()
-		ip, err := ipam.ReleaseIP(ctx, vm.Name, vm.Namespace)
+		ip, err := ipam.ReleaseIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
 		if err != nil {
 			// ignore error
 			log.Error(err, "fail to release IP, error ignored")
@@ -363,7 +363,7 @@ func (r *VMReconciler) acquireOverlayIP(ctx context.Context, vm *vmv1.VirtualMac
 		return err
 	}
 	defer ipam.Close()
-	ip, err := ipam.AcquireIP(ctx, vm.Name, vm.Namespace)
+	ip, err := ipam.AcquireIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
 	if err != nil {
 		log.Error(err, "fail to acquire IP")
 		return err
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -416,7 +416,9 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 
 	case "":
 		if err := r.acquireOverlayIP(ctx, vm); err != nil {
-			return err
+			log.Error(err, "Failed to acquire overlay IP", "VirtualMachine", vm.Name)
+			r.Recorder.Event(vm, "Warning", "OverlayNet", "Failed to acquire overlay IP")
+			// Don't return error, we will keep running without an overlay IP
 		}
 		// VirtualMachine just created, change Phase to "Pending"
 		vm.Status.Phase = vmv1.VmPending
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -81,6 +81,7 @@ type VMReconciler struct {
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 	Config   *ReconcilerConfig
+	IPAM     *ipam.IPAM
 
 	Metrics ReconcilerMetrics `exhaustruct:"optional"`
 }
@@ -238,27 +239,7 @@ func (r *VMReconciler) doFinalizerOperationsForVirtualMachine(ctx context.Contex
 
 	// Release overlay IP address
 	if vm.Spec.ExtraNetwork != nil {
-		// Create IPAM object
-		nadName, err := nadIpamName()
-		if err != nil {
-			// ignore error
-			log.Error(err, "ignored error")
-			return
-		}
-		nadNamespace, err := nadIpamNamespace()
-		if err != nil {
-			// ignore error
-			log.Error(err, "ignored error")
-			return
-		}
-		ipam, err := ipam.New(ctx, nadName, nadNamespace)
-		if err != nil {
-			// ignore error
-			log.Error(err, "ignored error")
-			return
-		}
-		defer ipam.Close()
-		ip, err := ipam.ReleaseIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
+		ip, err := r.IPAM.ReleaseIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
 		if err != nil {
 			// ignore error
 			log.Error(err, "fail to release IP, error ignored")
@@ -347,23 +328,7 @@ func (r *VMReconciler) acquireOverlayIP(ctx context.Context, vm *vmv1.VirtualMac
 	}
 
 	log := log.FromContext(ctx)
-
-	// Create IPAM object
-	nadName, err := nadIpamName()
-	if err != nil {
-		return err
-	}
-	nadNamespace, err := nadIpamNamespace()
-	if err != nil {
-		return err
-	}
-	ipam, err := ipam.New(ctx, nadName, nadNamespace)
-	if err != nil {
-		log.Error(err, "failed to create IPAM")
-		return err
-	}
-	defer ipam.Close()
-	ip, err := ipam.AcquireIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
+	ip, err := r.IPAM.AcquireIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
 	if err != nil {
 		log.Error(err, "fail to acquire IP")
 		return err
@@ -1712,15 +1677,7 @@ func podSpec(
 		if len(vm.Spec.ExtraNetwork.MultusNetwork) > 0 { // network specified in spec
 			nadNetwork = vm.Spec.ExtraNetwork.MultusNetwork
 		} else { // get network from env variables
-			nadName, err := nadRunnerName()
-			if err != nil {
-				return nil, err
-			}
-			nadNamespace, err := nadRunnerNamespace()
-			if err != nil {
-				return nil, err
-			}
-			nadNetwork = fmt.Sprintf("%s/%s", nadNamespace, nadName)
+			nadNetwork = fmt.Sprintf("%s/%s", config.NADConfig.IPAMNamespace, config.NADConfig.IPAMName)
 		}
 		pod.ObjectMeta.Annotations[nadapiv1.NetworkAttachmentAnnot] = fmt.Sprintf("%s@%s", nadNetwork, vm.Spec.ExtraNetwork.Interface)
 	}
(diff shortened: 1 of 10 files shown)


cicdteam:
I haven't gone into the changes in depth, but I just want to ask what issues are being addressed?

Omrigan:
> I haven't gone into the changes in depth, but I just want to ask what issues are being addressed?

Key issues are:
1. Leader election times out when the rate of allocations is significant. 
2. When a release of an IP fails, it is not retried, so we have 13k stale allocations.
3. k8s client is recreated on every request.

Plus a few other things.

pkg/neonvm/ipam/ipam.go
```diff
@@ -82,9 +111,12 @@ func New(ctx context.Context, nadName string, nadNamespace string) (*IPAM, error
 	if err != nil {
 		return nil, fmt.Errorf("error creating kubernetes client: %w", err)
 	}
+	return NewWithClient(kClient, nadName, nadNamespace)
+}
 
+func NewWithClient(kClient *Client, nadName string, nadNamespace string) (*IPAM, error) {
 	// read network-attachment-definition from Kubernetes
-	nad, err := kClient.nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(nadNamespace).Get(ctx, nadName, metav1.GetOptions{})
+	nad, err := kClient.NADClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(nadNamespace).Get(context.Background(), nadName, metav1.GetOptions{})
```
sharnoff, line 119: suggestion: We should not make a network call without a timeout, otherwise this can potentially block forever

sharnoff (commented):
Partial review, just skimming through -- broadly looks good, need to dig a little deeper into the new core of the ipam actions.

One thing that I'm not sure about: What happens if we get a lot of reconcile operations simultaneously waiting on ipam operations? Should we have some mechanism to immediately bail if there's too many other threads also waiting? (so that, under load, other operations don't completely grind to a halt as well)

pkg/neonvm/ipam/ipam.go
```diff
@@ -47,6 +41,8 @@ const (
 	DefaultLeaderLeaseDurationMs = 3000
```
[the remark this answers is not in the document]
Omrigan: Done

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -416,7 +381,9 @@ func (r *VMReconciler) doReconcile(ctx context.Context, vm *vmv1.VirtualMachine)
 
 	case "":
 		if err := r.acquireOverlayIP(ctx, vm); err != nil {
-			return err
+			log.Error(err, "Failed to acquire overlay IP", "VirtualMachine", vm.Name)
+			r.Recorder.Event(vm, "Warning", "OverlayNet", "Failed to acquire overlay IP")
+			// Don't return error, we will keep running without an overlay IP
```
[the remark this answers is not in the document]
sharnoff: I don't quite follow — we already have this in `acquireOverlayIP` meaning that it'll handle old VMs just fine anyways, right?

https://github.com/neondatabase/autoscaling/blob/0176fbe62040a07de44ee115a8ac6f9d4dd2b85f/pkg/neonvm/controllers/vm_controller.go#L324-L328

(won't this just sometimes violate our guarantee of providing an overlay IP when `ExtraNetwork` is enabled?)

sharnoff:
(assigned you for now @Omrigan, please feel free to assign back to me after resolving comments if you think it's ready for another round of review)

pkg/neonvm/controllers/vm_controller.go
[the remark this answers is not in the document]
Omrigan: > it'll handle old VMs just fine anyways, right?

Right. What I am trying to say is given that we already have old VMs, we can afford to have these kind of VMs, where we failed to allocate an IP.

> violate our guarantee of providing an overlay IP when ExtraNetwork is enabled

Yes, it would. They'd have `ExtraNetwork.Enable = true` and `ExtraNetwork.ExtraNetIP == []`. Do you see any problems with that?

I can double-check cplane would treat it as a compute without an overlay network.

Alternatively, we can decide this is too much precaution and I can drop this patch.

pkg/neonvm/ipam/allocate.go
```diff
@@ -41,16 +63,14 @@ func doAcquire(
...
 	reservation []whereaboutstypes.IPReservation,
-	vmName string,
-	vmNamespace string,
+	vmName types.NamespacedName,
 ) (net.IPNet, []whereaboutstypes.IPReservation, error) {
 	// reduce whereabouts logging
 	whereaboutslogging.SetLogLevel("error")
 
-	vmID := fmt.Sprintf("%s/%s", vmNamespace, vmName)
+	vmID := fmt.Sprintf("%s/%s", vmName.Namespace, vmName.Name)
```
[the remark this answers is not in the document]
Omrigan: Fixed

pkg/neonvm/ipam/ipam.go
```diff
@@ -42,32 +37,85 @@ const (
...
+	for {
+		vms, err := i.VMClient.NeonvmV1().VirtualMachines(namespace).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return fmt.Errorf("error listing virtual machines: %w", err)
+		}
+		vmsMap := make(map[string]bool)
+		for _, vm := range vms.Items {
+			vmsMap[fmt.Sprintf("%s/%s", vm.Namespace, vm.Name)] = true
+		}
+		_, err = i.runIPAM(ctx, getCleanupAction(vmsMap))
```
sharnoff, line 87: question: What happens if the set of VMs that have been assigned extra IPs changes between the call to `List()` and running the cleanup action?

pkg/neonvm/ipam/ipam.go
```diff
@@ -42,32 +37,85 @@ const (
...
+
+func (i *IPAM) ReleaseIP(ctx context.Context, vmName types.NamespacedName) (net.IPNet, error) {
+	ip, err := i.runIPAM(ctx, getReleaseAction(vmName))
+	if err != nil {
+		return net.IPNet{}, fmt.Errorf("failed to release IP: %w", err)
+	}
+	return ip, nil
+}
+
+func (i *IPAM) runCleanup(ctx context.Context, namespace string) error {
```
sharnoff, line 77: question: Do we actually need the cleanup action? If it's to handle previously leaked IP addresses from earlier versions of the code, could we just disable overlay IPs and manually reset the IP pool object? If we're worried about rare bugs in the future, we could add an HTTP endpoint to manually run `ReleaseIP(...)` for a particular VM.

pkg/neonvm/ipam/ipam.go
```diff
@@ -168,174 +229,84 @@ func LoadFromNad(nadConfig string, nadNamespace string) (*IPAMConfig, error) {
...
+	log := log.FromContext(ctx)
 
-	// check action
-	switch action {
-	case Acquire, Release:
-	default:
-		return ip, fmt.Errorf("got an unknown action: %v", action)
+	ok := i.concurrencyLimiter.TryAcquire(1)
+	if !ok {
+		return net.IPNet{}, ErrAgain
```
sharnoff, line 240: suggestion: Can you add a comment here explaining why we have this concurrency limiter _and_ the mutex below it?

pkg/neonvm/ipam/ipam.go
```diff
@@ -168,174 +229,84 @@ func LoadFromNad(nadConfig string, nadNamespace string) (*IPAMConfig, error) {
...
 	}
+	defer i.concurrencyLimiter.Release(1)
 
-	ctxWithTimeout, ctxCancel := context.WithTimeout(ctx, IpamRequestTimeout)
-	defer ctxCancel()
+	i.mu.Lock()
+	defer i.mu.Unlock()
```
sharnoff, line 245: Logging something I mentioned via DMs: I still worry that this approach will have throughput issues — current tolerances mean we'd need to be _comfortably_ <100ms per operation, which I'm not sure is feasible with this kind of design.

Alternately, worth considering implementing batching if we see that load tests validate that we have throughput issues.

sharnoff (commented):
More thorough review, lots of comments -- broadly looks ok. If there's smaller commits that I haven't commented on, I'm happy to approve them separately if that would help (reducing surface area of the diff, etc.)

pkg/neonvm/ipam/allocate.go
```diff
@@ -1,38 +1,58 @@
...
+	return func(ipRange RangeConfiguration, reservation []whereaboutstypes.IPReservation) (net.IPNet, []whereaboutstypes.IPReservation, error) {
+		return doRelease(ipRange, reservation, vmName)
+	}
+}
+
+func getCleanupAction(vms map[string]bool) ipamAction {
+	return func(ipRange RangeConfiguration, reservation []whereaboutstypes.IPReservation) (net.IPNet, []whereaboutstypes.IPReservation, error) {
+		return doCleanup(ipRange, reservation, vms)
+	}
+}
```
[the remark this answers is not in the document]
Omrigan: Makes sens, done!

pkg/neonvm/ipam/ipam.go
Omrigan: Yeah, that's fair. I have an idea for a fix, but I agree we should drop it.

pkg/neonvm/ipam/ipam.go
Omrigan: Done.

pkg/neonvm/ipam/ipam.go
Omrigan: Yeah, let's see in staging. If it is not good enough, then we will see.

go.mod
```diff
@@ -192,7 +192,7 @@ require (
 	golang.org/x/mod v0.17.0 // indirect
 	golang.org/x/net v0.33.0 // indirect
 	golang.org/x/oauth2 v0.20.0 // indirect
-	golang.org/x/sync v0.10.0 // indirect
+	golang.org/x/sync v0.10.0
```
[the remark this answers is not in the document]
Omrigan: Fixed.

pkg/neonvm/controllers/vm_controller.go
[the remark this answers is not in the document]
Omrigan: Fair enough, I dropped the patch. See also https://github.com/neondatabase/cloud/issues/25330

pkg/neonvm/ipam/ipam.go
Omrigan: Fair enough, replaced it with [`04be1a6` (#1278)](https://github.com/neondatabase/autoscaling/pull/1278/commits/04be1a67a66a72b6d21bca254e41f62bee8d4982)

pkg/neonvm/ipam/allocate.go
```diff
@@ -1,38 +1,54 @@
...
+) (net.IPNet, []whereaboutstypes.IPReservation, error)
+
+// makeAcquireAction creates a callback which changes IPPool state to include a new IP reservation.
+func makeAcquireAction(vmName types.NamespacedName) ipamAction {
+	return func(ipRange RangeConfiguration, reservation []whereaboutstypes.IPReservation) (net.IPNet, []whereaboutstypes.IPReservation, error) {
+		return doAcquire(ipRange, reservation, vmName)
+	}
+}
+
+// makeReleaseAction creates a callback which changes IPPool state to deallocate an IP reservation.
```
sharnoff, line 25: These comments are great - succinct, implies how to use it while only telling you what it does :)

pkg/neonvm/ipam/allocate.go
```diff
@@ -41,20 +57,17 @@ func doAcquire(
...
 ) (net.IPNet, []whereaboutstypes.IPReservation, error) {
 	// reduce whereabouts logging
 	whereaboutslogging.SetLogLevel("error")
 
-	vmID := fmt.Sprintf("%s/%s", vmNamespace, vmName)
 	_, ipnet, _ := net.ParseCIDR(ipRange.Range)
 
 	// try to release IP for given VM
-	newReservation, ip, err := whereaboutsallocate.IterateForDeallocation(reservation, vmID, getMatchingIPReservationIndex)
+	newReservation, ip, err := whereaboutsallocate.IterateForDeallocation(reservation, vmName.String(), getMatchingIPReservationIndex)
```
sharnoff, line 70: Just double-checking, this is idempotent, right? i.e., trying to release an IP that's already been released will not return error?

(thinking about failure modes for https://github.com/neondatabase/autoscaling/pull/1278/commits/04be1a67a66a72b6d21bca254e41f62bee8d4982)

sharnoff (approved):
Looks good, nice!

pkg/neonvm/ipam/allocate.go
Omrigan: Yeah, added a test for this.

Omrigan: But it was actually failing, fixed it now in the latest commit.

PR #1282, opened 2025-02-25
neonvm: ensure the TLS mount directory is available in the VM

conradludgate:
While testing, I noticed that the certificates were not getting copied inside the VM, but they were available in the pod.

I remember thinking about it, but it seems like I forgot that the `mounts.sh` still needs to `mkdir -p ...` the directory for the secrets.

In the process, I decided to make the mount dir configurable and add an additional test to check the file exists in the VM.

### DIFF
--- a/neonvm-runner/cmd/main.go
+++ b/neonvm-runner/cmd/main.go
@@ -230,11 +230,10 @@ func run(logger *zap.Logger) error {
 		// add the tls path.
 		// this is needed to just `mkdir` the mounting directory.
 		if vmSpec.TLS != nil {
-			watch := true
 			disks = append(disks, vmv1.Disk{
 				Name:      "tls-keys",
 				MountPath: vmSpec.TLS.MountPath,
-				Watch:     &watch,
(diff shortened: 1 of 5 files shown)


neonvm-runner/cmd/main.go
```diff
@@ -225,13 +225,33 @@ func run(logger *zap.Logger) error {
...
+		disks := vmSpec.Disks
+
+		// add the tls path.
+		// this is needed to just `mkdir` the mounting directory.
+		if vmSpec.TLS != nil {
+			watch := true
+			disks = append(disks, vmv1.Disk{
+				Name:      "tls-keys",
+				MountPath: vmSpec.TLS.MountPath,
+				Watch:     &watch,
```
sharnoff, line 237:
```suggestion
				Watch:     lo.ToPtr(true),
```

(we tend to use `github.com/samber/lo` to help make this sort of thing cleaner)

sharnoff (approved):
lgtm!

PR #1286, opened 2025-02-25
neonvm-runner: Reduce startup logs

petuhovskiy:
This PR disables verbose logging for repetitive startup logs, such as kernel debug boot logs.

It sets default kernel loglevel as INFO (https://docs.kernel.org/core-api/printk-basics.html) and vminit script sets loglevel back to DEBUG after booting has completed.

The full list of changes can be seen in commit messages.

Startup logs before this PR (559 lines): https://gist.github.com/petuhovskiy/cfa0e7654cd25b37247694a33a9d556b
Startup logs after this PR (123 lines): https://gist.github.com/petuhovskiy/a85a7dad7bb4015d2f1f33fcd6cce9fd

Closes https://github.com/neondatabase/cloud/issues/19678

### DIFF
--- a/neonvm-runner/cmd/net.go
+++ b/neonvm-runner/cmd/net.go
@@ -163,7 +163,7 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 	// pass incoming traffic to .Guest.Spec.Ports into VM
 	var iptablesArgs []string
 	for _, port := range ports {
-		logger.Info(fmt.Sprintf("setup DNAT rule for incoming traffic to port %d", port.Port))
+		logger.Debug(fmt.Sprintf("setup DNAT rule for incoming traffic to port %d", port.Port))
 		iptablesArgs = []string{
 			"-t", "nat", "-A", "PREROUTING",
 			"-i", "eth0", "-p", fmt.Sprint(port.Protocol), "--dport", fmt.Sprint(port.Port),
@@ -173,7 +173,7 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 			logger.Error("could not set up DNAT rule for incoming traffic", zap.Error(err))
 			return nil, err
 		}
-		logger.Info(fmt.Sprintf("setup DNAT rule for traffic originating from localhost to port %d", port.Port))
+		logger.Debug(fmt.Sprintf("setup DNAT rule for traffic originating from localhost to port %d", port.Port))
 		iptablesArgs = []string{
 			"-t", "nat", "-A", "OUTPUT",
 			"-m", "addrtype", "--src-type", "LOCAL", "--dst-type", "LOCAL",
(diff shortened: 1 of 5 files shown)


neonvm-runner/cmd/net.go
```diff
@@ -163,7 +163,7 @@ func defaultNetwork(logger *zap.Logger, cidr string, ports []vmv1.Port) (mac.MAC
 	// pass incoming traffic to .Guest.Spec.Ports into VM
 	var iptablesArgs []string
 	for _, port := range ports {
-		logger.Info(fmt.Sprintf("setup DNAT rule for incoming traffic to port %d", port.Port))
+		logger.Debug(fmt.Sprintf("setup DNAT rule for incoming traffic to port %d", port.Port))
```
sharnoff, line 166: I don't think we have a way of setting neonvm-runner's log level. Should we just delete these?

vm-builder/files/set-disk-quota.sh
```diff
@@ -18,6 +18,7 @@ echo "Setting disk quota for $mountpoint to $size KB"
 # setquota -P <project_id> <block-softlimit> <block-hardlimit> <inode-softlimit> <inode-hardlimit> <filesystem>
 /neonvm/bin/setquota -P 0 0 "$size" 0 0 "$mountpoint"
 
-echo "DONE! Disk quota set"
-/neonvm/bin/repquota -P -a
+# Don't print the quota report, it's too verbose
+# echo "DONE! Disk quota set"
+# /neonvm/bin/repquota -P -a
```
sharnoff, line 23: Maybe let's just remove, rather than commenting out?

vm-builder/files/vminit
```diff
@@ -25,6 +25,9 @@ mount -t proc  proc  /proc
 mount -t sysfs sysfs /sys
 mount -t cgroup2 cgroup2 /sys/fs/cgroup
 
+# change kernel log level to DEBUG
+echo 7 > /proc/sys/kernel/printk
```
sharnoff, line 29: Can we do this earlier on? Maybe nitpicking, but I worry about missing logs due to earlier changes (e.g., udev rules) taking effect before this change

sharnoff (approved):
LGTM, nice!

neonvm-runner/cmd/net.go
petuhovskiy: Even if we don't have a way to change log level yet, debug/trace logs can serve the same role as comments, so I think we can keep them.

vm-builder/files/vminit
petuhovskiy: No we can't, procfs is mounted right before :)

vm-builder/files/set-disk-quota.sh
petuhovskiy: I think it's useful as info, to remind how to look at the quotas (repquota flags are hard to remember). So I'd keep it.

vm-builder/files/vminit
sharnoff: Ah! Fair enough, I guess :sweat_smile:

lassizci (approved):
Nice, looks good to me!

PR #1292, opened 2025-03-03
agent: add ExtraIP per-VM metric 

Omrigan:
Plus a couple of refactorings.

To be merged with rebase.

Part of #1294.

### DIFF
--- a/pkg/agent/watch.go
+++ b/pkg/agent/watch.go
@@ -300,24 +300,45 @@ func makeVMRestartMetrics(vm *vmv1.VirtualMachine) []vmMetric {
 	}
 }
 
+// gaugeSpec groups a source of metrics (maker) with a destination (gauge).
+type gaugeSpec struct {
+	maker func(*vmv1.VirtualMachine) []vmMetric
+	gauge *prometheus.GaugeVec
+}
+
+// getGaugeSpecs constructs our list of metrics to expose.
+//
+// It is used for creating, updating, and deleting metrics.
+func getGaugeSpecs(perVMMetrics *PerVMMetrics) []gaugeSpec {
+	return []gaugeSpec{
+		{
+			maker: makeVMCPUMetrics,
+			gauge: perVMMetrics.cpu,
+		},
+		{
+			maker: makeVMMemMetrics,
+			gauge: perVMMetrics.memory,
+		},
+		{
+			maker: makeVMRestartMetrics,
+			gauge: perVMMetrics.restartCount,
+		},
+	}
+}
+
 func setVMMetrics(perVMMetrics *PerVMMetrics, vm *vmv1.VirtualMachine, nodeName string) {
 	if vm.Status.Node != nodeName {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/agent/watch.go
```diff
@@ -300,24 +300,42 @@ func makeVMRestartMetrics(vm *vmv1.VirtualMachine) []vmMetric {
 	}
 }
 
+type gaugeSpec struct {
```
sharnoff, line 320: suggestion: Could you add some comments explaining what `gaugeSpec` does / is for?

I'm having a little trouble wrapping my head around it :sweat_smile:

sharnoff (commented):
Refactorings seem mostly fair enough, one question.

re: Adding the metric — I wonder if we actually need this? What could you do with such a metric that you couldn't do before? (I'm imagining that you can always cross-reference with cplane's view of the VM if you want to check whether it has the overlay network enabled ­— but maybe I'm missing something!)

pkg/agent/watch.go
Omrigan: Expanded comments around `gaugeSpec` and `getGaugeSpecs`

Omrigan:
> Adding the metric — I wonder if we actually need this? What could you do with such a metric that you couldn't do before? 

I thought of two use cases:
1. View it in "per-VM metrics" dashboard, if we need to debug computes and network around it.
2. Aggregate it to see the share of computes in the overlay network.

> (I'm imagining that you can always cross-reference with cplane's view of the VM if you want to check whether it has the overlay network enabled ­— but maybe I'm missing something!)

Yeah, that's fair, but cplane info is not in grafana.

pkg/agent/watch.go
```diff
@@ -376,6 +397,10 @@ func updateVMMetrics(perVMMetrics *PerVMMetrics, oldVM, newVM *vmv1.VirtualMachi
 		updateMetrics(spec.gauge, oldMetrics, newMetrics)
 	}
 
+	oldExtraIPMetrics := makeVMExtraIPMetrics(oldVM)
+	newExtraIPMetrics := makeVMExtraIPMetrics(newVM)
+	updateMetrics(perVMMetrics.extraIP, oldExtraIPMetrics, newExtraIPMetrics)
```
sharnoff, line 402: question(blocking): Aren't these already included in `getGaugeSpecs` ?

pkg/agent/watch.go
```diff
@@ -392,6 +417,11 @@ func deleteVMMetrics(perVMMetrics *PerVMMetrics, vm *vmv1.VirtualMachine, nodeNa
 		}
 	}
 
+	extraIPMetrics := makeVMExtraIPMetrics(vm)
+	for _, m := range extraIPMetrics {
+		perVMMetrics.extraIP.Delete(m.labels)
+	}
```
sharnoff, line 423: + same here

pkg/agent/watch.go
Omrigan: Oops, sorry about that, I changed the order of the commits, so the old way and new way to add metrics got mixed up.

PR #1293, opened 2025-03-03
neonvm-controller/ipam: add metrics

Omrigan:
Two metrics:
 - Number of ongoing IPAM operations
 - Duration histogram

Part of https://github.com/neondatabase/autoscaling/issues/1294.

### DIFF
--- a/neonvm-controller/cmd/main.go
+++ b/neonvm-controller/cmd/main.go
@@ -37,6 +37,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -201,9 +202,16 @@ func main() {
 		NADConfig:               controllers.GetNADConfig(),
 	}
 
-	// Let's not have more than a quater of reconcilliation workers stuck
-	// at IPAM mutex.
-	ipam, err := ipam.New(rc.NADConfig.IPAMName, rc.NADConfig.IPAMNamespace, max(1, concurrencyLimit/4))
+	ipam, err := ipam.New(&ipam.IPAMParams{
+		NadName:      rc.NADConfig.IPAMName,
+		NadNamespace: rc.NADConfig.IPAMNamespace,
+
+		// Let's not have more than a quater of reconcilliation workers stuck
+		// at IPAM mutex.
+		ConcurrencyLimit: max(1, concurrencyLimit/4),
+
+		MetricsReg: metrics.Registry,
+	})
 	if err != nil {
 		setupLog.Error(err, "unable to create ipam")
 		panic(err)
--- a/neonvm-controller/cmd/main.go
+++ b/neonvm-controller/cmd/main.go
@@ -206,7 +206,7 @@ func main() {
 		NadName:      rc.NADConfig.IPAMName,
 		NadNamespace: rc.NADConfig.IPAMNamespace,
 
-		// Let's not have more than a quater of reconcilliation workers stuck
+		// Let's not have more than a quarter of reconcilliation workers stuck
 		// at IPAM mutex.
 		ConcurrencyLimit: max(1, concurrencyLimit/4),
 
--- a/neonvm-controller/cmd/main.go
+++ b/neonvm-controller/cmd/main.go
@@ -202,7 +202,7 @@ func main() {
 		NADConfig:               controllers.GetNADConfig(),
 	}
 
-	ipam, err := ipam.New(&ipam.IPAMParams{
+	ipam, err := ipam.New(ipam.IPAMParams{
 		NadName:      rc.NADConfig.IPAMName,
 		NadNamespace: rc.NADConfig.IPAMNamespace,
 
--- a/pkg/neonvm/ipam/ipam.go
+++ b/pkg/neonvm/ipam/ipam.go
@@ -13,6 +13,7 @@ import (
 
 	whereaboutsallocate "github.com/k8snetworkplumbingwg/whereabouts/pkg/allocate"
 	whereaboutstypes "github.com/k8snetworkplumbingwg/whereabouts/pkg/types"
+	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/semaphore"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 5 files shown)


pkg/neonvm/ipam/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+		startTime: time.Now(),
+		metrics:   m,
+		action:    action,
+
+		finished: false,
+	}
+}
+
+func (t *metricTimer) Finish(outcome string) {
+	if t.finished {
```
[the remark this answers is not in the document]
Omrigan: We don't have any concurrency here, `finished bool` does as well of a job 😄

neonvm-controller/cmd/main.go
```diff
@@ -201,9 +202,16 @@ func main() {
...
 	}
 
-	// Let's not have more than a quater of reconcilliation workers stuck
-	// at IPAM mutex.
-	ipam, err := ipam.New(rc.NADConfig.IPAMName, rc.NADConfig.IPAMNamespace, max(1, concurrencyLimit/4))
+	ipam, err := ipam.New(&ipam.IPAMParams{
+		NadName:      rc.NADConfig.IPAMName,
+		NadNamespace: rc.NADConfig.IPAMNamespace,
+
+		// Let's not have more than a quater of reconcilliation workers stuck
```
sharnoff, line 209:
```suggestion
		// Let's not have more than a quarter of reconcilliation workers stuck
```

pkg/neonvm/ipam/ipam.go
```diff
@@ -49,30 +50,38 @@ type Temporary interface {
...
+type IPAMParams struct {
+	NadName          string
+	NadNamespace     string
+	ConcurrencyLimit int
+	MetricsReg       prometheus.Registerer
+}
+
 func (i *IPAM) AcquireIP(ctx context.Context, vmName types.NamespacedName) (net.IPNet, error) {
-	ip, err := i.runIPAM(ctx, makeAcquireAction(ctx, vmName))
+	ip, err := i.runIPAMWithMetrics(ctx, makeAcquireAction(ctx, vmName), IPAMAcquire)
```
sharnoff, line 68: suggestion(non-blocking): Could we de-duplicate the `makeFooAction` + `IPAMFoo` in the args, so that metric label is provided by the action?

(maybe e.g., `IPAMAction` should be a struct or interface, so it can self-describe?)

pkg/neonvm/ipam/ipam.go
```diff
@@ -49,30 +50,38 @@ type Temporary interface {
...
+	ip, err := i.runIPAMWithMetrics(ctx, makeReleaseAction(ctx, vmName), IPAMRelease)
 	if err != nil {
 		return net.IPNet{}, fmt.Errorf("failed to release IP: %w", err)
 	}
 	return ip, nil
 }
 
 // New returns a new IPAM object with ipam config and k8s/crd clients
-func New(nadName string, nadNamespace string, concurrencyLimit int) (*IPAM, error) {
+func New(params *IPAMParams) (*IPAM, error) {
```
sharnoff, line 84: question(non-blocking): Does params need to be a pointer?

pkg/neonvm/ipam/ipam.go
```diff
@@ -186,6 +196,20 @@ func LoadFromNad(nadConfig string, nadNamespace string) (*IPAMConfig, error) {
 	return n.IPAM, nil
 }
 
+func (i *IPAM) runIPAMWithMetrics(ctx context.Context, action ipamAction, actionName string) (net.IPNet, error) {
+	timer := i.metrics.StartTimer(actionName)
+	// This is if we get a panic
+	defer timer.Finish(IPAMUnknown)
```
sharnoff, line 202: question: If this is only when we get a panic, should we indicate it as such?

pkg/neonvm/ipam/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+	IPAMUnknown = "unknown"
+)
+
+func NewIPAMMetrics(reg prometheus.Registerer) *IPAMMetrics {
+	// Copied bucket values from controller runtime latency metric. We can
+	// adjust them in the future if needed.
+	buckets := []float64{
+		0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
+		1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60,
+	}
```
sharnoff, line 34: suggestion: Let's tune this a little bit? It's 40 buckets × 3 operations × 3 results, and I bet a ton of the values are unneeded. E.g. what if we just do powers of 10 for now, and iterate on the bucket values from tests on staging?

(thinking about this because e.g. VM startup time buckets are wildly off but haven't been fixed)

Not a huge deal, you're welcome to ignore if you like. Just thought I'd mention it

pkg/neonvm/ipam/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+	return &IPAMMetrics{
+		ongoing: *util.RegisterMetric(reg, prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "ipam_ongoing_requests",
+			Help: "Number of ongoing IPAM requests",
+		}, []string{"action"})),
+		duration: *util.RegisterMetric(reg, prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "ipam_request_duration_seconds",
+			Help:    "Duration of IPAM requests",
+			Buckets: buckets,
+		}, []string{"action", "outcome"})),
```
sharnoff, line 45: suggestion: Let's not dereference the metrics here? I figured that `NewGaugeVec` / `NewHistogramVec` return pointers for a reason

sharnoff (commented):
Some comments -- broadly looks good; some ideas for improvement, but nothing critical.

pkg/neonvm/ipam/metrics.go
Omrigan: Changed to this:
```
buckets := []float64{
		0.01, 0.05, 0.1, 0.25, 0.5, 1, 2, 3, 5, 10, 30,
}
```

pkg/neonvm/ipam/ipam.go
Omrigan: That's fair, let me keep it this way for now, and I'll do that next time I work with this code.

pkg/neonvm/ipam/ipam.go
Omrigan: No, but why is one option is better than the other?

pkg/neonvm/ipam/ipam.go
Omrigan: Done.

pkg/neonvm/ipam/metrics.go
Omrigan: Yes, make sense. But we have other places where it is like that.

pkg/neonvm/ipam/ipam.go
sharnoff: Just thinking it's worth avoiding pointers if they don't provide any value, and it looked like that might be the case here? If not, feel free to leave as is

pkg/neonvm/ipam/ipam.go
sharnoff: Sounds good -- worth opening an issue for it?

sharnoff (approved):
LGTM, modulo existing comment threads

pkg/neonvm/ipam/ipam.go
Omrigan: I'll write it down for now.

pkg/neonvm/ipam/ipam.go
Omrigan: I've already changed it, I'll follow-up later on that topic.

PR #1295, opened 2025-03-03
neonvm-controller: implement migration webhooks

### LINKED ISSUE #719: Implement VirtualMachineMigration webhook validators
AFAICT, in theory this means that a VirtualMachineMigration can be changed out from underneath the controller in ways we might not expect, which can lead to hard-to-debug errors (especially if we didn't realize we were changing the migration in the first place).

Omrigan:
Not much we need to check for, but it guards us from creating a migration for the wrong VM.

Closes #719.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
+++ b/neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
@@ -45,9 +45,6 @@ var _ webhook.Validator = &VirtualMachineMigration{}
 //
 // The controller wraps this logic so it can inject extra control in the webhook.
 func (r *VirtualMachineMigration) ValidateCreate() (admission.Warnings, error) {
-	if len(r.Spec.VmName) == 0 {
-		return nil, fmt.Errorf("vmName is required")
-	}
 	return nil, nil
 }
 
--- a/neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
+++ b/neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1
 
 import (
+	"fmt"
+
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -42,7 +44,6 @@ var _ webhook.Validator = &VirtualMachineMigration{}
 //
 // The controller wraps this logic so it can inject extra control in the webhook.
 func (r *VirtualMachineMigration) ValidateCreate() (admission.Warnings, error) {
-	// TODO: implement creation validation webhook (?)
 	return nil, nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/neonvm/controllers/webhook.go
```diff
@@ -135,6 +137,16 @@ var _ webhook.CustomValidator = (*VMWebhook)(nil)
 // ValidateCreate implements webhook.CustomValidator
 func (w *VMMigrationWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
 	vmm := obj.(*vmv1.VirtualMachineMigration)
+	var vm vmv1.VirtualMachine
+	err := w.Client.Get(ctx, types.NamespacedName{Name: vmm.Spec.VmName, Namespace: vmm.Namespace}, &vm)
+	if err != nil {
+		return nil, fmt.Errorf("vm %s not found: %w", vmm.Spec.VmName, err)
```
sharnoff, line 143: question(blocking): This can fail for other reasons than "not found", right?

pkg/neonvm/controllers/webhook.go
```diff
@@ -135,6 +137,16 @@ var _ webhook.CustomValidator = (*VMWebhook)(nil)
 // ValidateCreate implements webhook.CustomValidator
 func (w *VMMigrationWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
sharnoff, line 138: Higher level question here: With #1300, the migration will already fail if the VM doesn't exist. Why do we need the same check in `ValidateCreate` ?

Separately, I'm not sure it's a good idea to perform k8s operations inside a webhook, because the webhook itself blocks other k8s operations. I think _in theory_ it should be fine, but that possibility of feedback makes me nervous.

sharnoff (commented):
Some comments about `ValidateCreate`, otherwise looks good

pkg/neonvm/controllers/webhook.go
Omrigan: Well... even if API call failed, we VM is still kind of "not found" 😄 

Not sure if this is the right logic.

sharnoff: I guess this is obsolete now, but for the future: There's a handful of unrelated reasons that a request can fail -- e.g., auth, API server restarted, etcd timeout :sweat_smile:

neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
```diff
@@ -42,22 +45,27 @@ var _ webhook.Validator = &VirtualMachineMigration{}
 //
 // The controller wraps this logic so it can inject extra control in the webhook.
 func (r *VirtualMachineMigration) ValidateCreate() (admission.Warnings, error) {
-	// TODO: implement creation validation webhook (?)
+	if len(r.Spec.VmName) == 0 {
+		return nil, fmt.Errorf("vmName is required")
+	}
```
sharnoff, line 50: Huh, won't this already be required by the CRD? (if not: can we have it handled there? IIRC there's tons of validation options, I would assume "non-empty string" is one of them :sweat_smile:)

sharnoff (approved):
LGTM, modulo one comment

neonvm/apis/neonvm/v1/virtualmachinemigration_webhook.go
Omrigan: You are right, I did it with `kubebuilder:validation:MinLength` in the last commit.

However, it is trickier to test.

pkg/neonvm/controllers/webhook.go
Omrigan: Right, "vm not found" usually perceived as "vm does not exist", even though _technically_ it can mean "etcd timouted, so we couldn't find the vm, ergo it is _not found_". 

But I agree it would be confusing.

PR #1296, opened 2025-03-04
use .Dockerfile as file ext, following best practices from Docker

fedordikarev:
That will allow syntax highlight for the files, and number of utilities expect `.Dockerfile` to be an ext for the files.

```
Some projects may need distinct Dockerfiles for specific purposes. A common convention is to name these <something>.Dockerfile. You can specify the Dockerfile filename using the --file flag for the docker build command. Refer to the [docker build CLI reference](https://docs.docker.com/reference/cli/docker/buildx/build/#file) to learn about the --file flag.
```

https://docs.docker.com/build/concepts/dockerfile/#filename


sharnoff (approved):
Looks good - thanks!

sharnoff:
As a follow-up: @mikhail-sakhnov WDYT about renaming `neonvm-kernel/{kernel-builder.Dockerfile => Dockerfile}` ? I'm realizing it's a little redundant :sweat_smile:

mikhail-sakhnov:
@sharnoff I think it can be just Dockerfile

fedordikarev:
Let me add that rename `neonvm-kernel/{kernel-builder.Dockerfile => Dockerfile}` here as well

mikhail-sakhnov (approved):

PR #1298, opened 2025-03-04
neonvm-controller: fix the deletion logic for tmp secrets

conradludgate:
Looks like I copy-pasted the wrong logic for tmp secret clean up :D


sharnoff (approved):

conradludgate:
Looks like `autoscaling.cpu-sys-fs-state-scaling` is suddenly very slow.

PR #1299, opened 2025-03-04
feat: automatically set gomaxprocs for cfs

itspooya:
see https://github.com/golang/go/issues/33803 for more info

### DIFF
--- a/autoscale-scheduler/cmd/main.go
+++ b/autoscale-scheduler/cmd/main.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/neondatabase/autoscaling/pkg/plugin"
 	"github.com/neondatabase/autoscaling/pkg/util"
+	_ "go.uber.org/automaxprocs"
 )
 
 // all of the juicy bits are defined in pkg/plugin/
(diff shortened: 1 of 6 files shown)


sharnoff:
Hi @itspooya, thanks for the PR! This may have unexpected impacts, so we'll discuss within the team to decide how we want to go about it.

itspooya:
​I understand your concerns about the potential impacts of this change. In my experience, incorporating the automaxprocs library can be advantageous, especially in large clusters or on instances with numerous CPU cores. I've observed positive outcomes in other projects after implementing this library. However, as you mentioned, it's always prudent to benchmark and assess the implications of such changes.​

Upon reviewing your repository, I noticed that you're setting Kubernetes resources, particularly Requests and Limits. Integrating automaxprocs, in conjunction with these settings, could enhance the performance of your Go applications and services. Additionally, I recommend considering the addition of GOMEMLIMIT, given that you've set memory limits alongside CPU limits. This adjustment could further optimize resource utilization.​

Monitoring metrics in Kubernetes would likely demonstrate the benefits of these changes. Another approach is to set GOMAXPROCS to one more than the number of CPU cores specified in the request. While this can be beneficial, many projects find that starting with the automaxprocs library yields positive results, with the option to fine-tune settings later as needed.

sharnoff:
Hi @itspooya, sorry for the late reply — we discussed this internally and decided not to go with this approach at this time.

That said, thank you for opening the PR! It spawned good discussion, and we will likely end up doing something similar, albeit with a different approach.

PR #1300, opened 2025-03-05
neonvm-controller: fix state transitions

Omrigan:
Previous version was unsound.


pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -150,25 +150,35 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
+					Status:  metav1.ConditionTrue,
+					Reason:  "Reconciling",
+					Message: message,
+				})
+			migration.Status.Phase = vmv1.VmmFailed
+			return r.updateMigrationStatus(ctx, migration)
 		}
-		return &vm, nil
+		// return err and try reconcile again
+		return ctrl.Result{}, err
```
sharnoff, line 174: question(blocking): Given that `if err != nil` always ends in returning here, does that mean that we never remove the finalizer, and so the migration object cannot be deleted if the VM doesn't exist?

pkg/neonvm/controllers/vmmigration_controller.go
```diff
@@ -254,21 +242,31 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
...
-			if err := r.Status().Update(ctx, vm); err != nil {
-				log.Error(err, "Failed to update VM status to PreMigrating", "Status", vm.Status.Phase)
-				return ctrl.Result{}, err
-			}
-			// Migration just created, change Phase to "Pending"
-			migration.Status.Phase = vmv1.VmmPending
-			return r.updateMigrationStatus(ctx, migration)
+	if vm.Status.MigrationName == "" {
+		if vm.Status.Phase != vmv1.VmRunning {
+			// VM must be scaling, so we will try again in a second
```
sharnoff, line 247: question:

> VM must be scaling

Why?

sharnoff (commented):
A couple questions

pkg/neonvm/controllers/vmmigration_controller.go
Omrigan: That's true, and this problem existed before this PR. Let me think about a solution.

pkg/neonvm/controllers/vmmigration_controller.go
Omrigan: That's an evolution of the `// some other VM status (Scaling may be), requeue after second`

But I agree, "may be" should not have turned into `must be`.

PR #1302, opened 2025-03-05
neonvm: add RTC to arm kernel config

mikhail-sakhnov:
CONFIG_RTC_CLASS to enable RTC subsystem.
CONFIG_RTC_DRV_EFI to use RTC from efi firmware.

With this combination of settings time in guest vm is valid. Otherwise it always starts with 1970- (0 timestamp).


sharnoff:
Hm, any idea why it's needed on arm and not x86?

mikhail-sakhnov:
>Hm, any idea why it's needed on arm and not x86?

There might be almost infinitive amount of reasons and I haven't invest time into that. 
x86 might use CMOS time, or x86 might have some [legacy code to handle time](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/time.c#L76) without external rtc device.

Plus on arm we use uefi on arm, which might give us different behaviour on such a base level.

Omrigan (approved):

Omrigan:
See also https://github.com/neondatabase/autoscaling/issues/712

Issue #1243, opened 2025-02-04
neonvm-daemon: use GET /cpu as source of information for sysfs scaling mode


mikhail-sakhnov:
We currently never use neonvm-daemon /cpu GET endpoint, since it has no support for fractional CPU limits. Instead we rely on the value stored in the runner memory via atomic variable, which is updated on each scaling request.

We should use it as a source of information for sysfs scaling mode.

Reasons:
- as said, first step towards proper fractional limits
- clearer ownership - all cpu related state will be kept in a single component (neonvm-daemon) rather than in 2 components as now (neonvm-daemon, neonvm-runner)

Context:
Discussion in https://github.com/neondatabase/autoscaling/pull/1190#discussion_r1941035937_

sharnoff:
This should also be a dependency for #850

PR #1308, opened 2025-03-07
neonvm: use neonvm-daemon/cpu as source of information in runner

mikhail-sakhnov:
neonvm: use neonvm-daemon/cpu as source of information for current CPU state in the runner. Should affect only sysfs scaling mode.

#1243


neonvm-runner/cmd/main.go
```diff
@@ -487,17 +488,34 @@ func runQEMU(
...
 		get: func(logger *zap.Logger) (*vmv1.MilliCPU, error) {
-			return lo.ToPtr(vmv1.MilliCPU(lastValue.Load())), nil
+			switch cfg.cpuScalingMode {
+			case vmv1.CpuScalingModeSysfs:
+				cpu, err := getNeonvmDaemonCPU()
+				if err != nil {
+					logger.Error("failed to get CPU from NeonVM Daemon", zap.Error(err))
+					return nil, err
+				}
+				return &cpu, nil
```
sharnoff, line 498: As discussed, I think we still need to compare against `lastValue` so that we can translate from integer CPU → externally requested fractional CPU

mikhail-sakhnov: Sorry, that s a draft to trigger e2e, didn’t expect it being reviewed

mikhail-sakhnov: But thanks for the catch - but the pr isn’t finished yet, might be some shenanigans in other parts of the diff

mikhail-sakhnov: >can translate from integer CPU → externally requested fractional CPU —

why do we need it? I thought it could be done in the controller by rounding up requested fractional CPU before comparing with the value from daemon?

mikhail-sakhnov: Discussed in DM, both ways would work, moving to the runner.

petuhovskiy (approved):
LGTM, but I'm not very familiar with sysfs scaling yet.

neonvm-runner/cmd/main.go
```diff
@@ -487,17 +488,42 @@ func runQEMU(
...
+				if err != nil {
+					logger.Error("failed to get CPU from NeonVM Daemon", zap.Error(err))
+					return nil, err
+				}
+				storedCpu := vmv1.MilliCPU(lastValue.Load())
+				if storedCpu.RoundedUp() != cpu.RoundedUp() {
+					logger.Warn("CPU from NeonVM Daemon does not match stored value, returning daemon value to let controller reconcile correct state", zap.Any("stored", storedCpu), zap.Any("current", cpu))
+					return &cpu, nil
+				}
+				logger.Debug("Returning stored CPU value to controller", zap.Any("cpu", storedCpu))
```
sharnoff, line 503: question: Do we output debug logs? If no, is it worth keeping this?

neonvm-runner/cmd/main.go
```diff
@@ -487,17 +488,42 @@ func runQEMU(
...
+				return &storedCpu, nil
+
+			case vmv1.CpuScalingModeQMP:
+				return lo.ToPtr(vmv1.MilliCPU(lastValue.Load())), nil
+			default:
+				panic(fmt.Errorf("unknown CPU scaling mode %s", cfg.cpuScalingMode))
```
sharnoff, line 509: nit: Let's use `%q` instead of `%s`? It makes it easier to identify unusual strings (e.g., empty string, etc.)

neonvm-runner/cmd/main.go
```diff
@@ -487,17 +488,42 @@ func runQEMU(
...
 				if err != nil {
 					logger.Error("setting CPU through NeonVM Daemon failed", zap.Any("cpu", cpu), zap.Error(err))
 					return err
 				}
+				lastValue.Store(uint32(cpu))
+			case vmv1.CpuScalingModeQMP:
+				lastValue.Store(uint32(cpu))
+				return nil
+			default:
+				panic(fmt.Errorf("unknown CPU scaling mode %s", cfg.cpuScalingMode))
```
sharnoff, line 525: \+ same here

neonvm-runner/cmd/main.go
```diff
@@ -865,6 +891,45 @@ func setNeonvmDaemonCPU(cpu vmv1.MilliCPU) error {
 	return nil
 }
 
+func getNeonvmDaemonCPU() (vmv1.MilliCPU, error) {
```
sharnoff, line 894: suggestion: We already have `checkNeonvmDaemonCPU()` — maybe let's just adapt that function instead of creating a new one? (or, replace it with this one?)

sharnoff (commented):
Some comments -- broadly looks good, should be good to go after these.

neonvm-runner/cmd/main.go
mikhail-sakhnov: done here and in other panics in the file

neonvm-runner/cmd/main.go
mikhail-sakhnov: I think this part should be refactored - all neonvm daemon related functions are very similar (as well as functions we use in controller to call runner) but I intentionally left it out of the scope. I think it is better to keep as is now (to keep reasonable get/set/check interface) and extract them into a client as a separate issue.

neonvm-runner/cmd/main.go
mikhail-sakhnov: I think it is wort having it for a potential debugs, right no there is no need, but I think is easier to have it logged at debug level and enable debug level if it is required, comparing to the case that if we need it, we need to add logging, cut release and redeploy. Let me know, if you don't like the approach, I can drop it.

Omrigan: I agree with Misha, I think even if we can't enable it now, but it still serves as documentation, and in case we will have an ability to enable debug logs.

sharnoff: Ack, fair enough - resolving

neonvm-runner/cmd/main.go
```diff
@@ -487,17 +488,42 @@ func runQEMU(
 
 	callbacks = cpuServerCallbacks{
 		get: func(logger *zap.Logger) (*vmv1.MilliCPU, error) {
-			return lo.ToPtr(vmv1.MilliCPU(lastValue.Load())), nil
+			switch cfg.cpuScalingMode {
```
Omrigan, line 491: I think nesting here is a bit too much, IMO we should change the structure of code here.

neonvm-runner/cmd/main.go
mikhail-sakhnov: Yes, like the `setNeonvmDaemonCPU` is copy-paste with changing request type to post and adding body 😄 

I pushed the changes but in general I disagree with that approach. 
We anyway need to change it, but with having separate functions changes in the PR are limited - they affect only new functionality. 

However, with the dropping checkNeonvmDaemonCPU we also start touching separate functionality - healthchecks. 

Not a critical in this particular case, but in general I'd rather prefer to have PRs scoped to a single blast radius.

neonvm-runner/cmd/main.go
mikhail-sakhnov: dropped

neonvm-runner/cmd/main.go
mikhail-sakhnov: Agreed in DM to proceed with that later

PR #1313, opened 2025-03-11
neonvm: drop generated default value in CRD

mikhail-sakhnov:
We use kubebuilder:default annotation for scaling mode fields, which effiectively means we never allow it to be non set and that bypasses the logic with --default-cpu-scaling-mode controller argument


sharnoff (approved):
LGTM - maybe PR title should say `neonvm:` and not `neonvm-controller:`, as it only changes the CRD? Up to you.

PR #1283, opened 2025-02-25
CI: GHCR as the main registry

bayandin:
We're shifting the main workload from Docker Hub to GHCR.

This PR makes CI in the repo use GHCR as the main container registry.
For releases (`inputs.upload-to-ecr-acr` is set to `true`) we still upload images to Docker Hub for compatibility.

This PR also replaces `_push-to-acr.yaml` workflow with `_push-to-container-registry.yml` (which we already use in `neon` repo to spread docker images across registries)

PR #1315, opened 2025-03-11
ci/build-images: fix push to ECR

bayandin:
Push to ECR started to fail after https://github.com/neondatabase/autoscaling/pull/1312 changes:

```
Copy neondatabase/neonvm-controller:testrelease-20250311.1c7bb48.13791913752 to dev ECR
ERROR: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```
Spotted in [link](https://github.com/neondatabase/autoscaling/actions/runs/13791913752/job/38574226461?pr=1311#step:11:36)

It tries to push images to `<ECR>/neondatabase/<IMAGE_NAME>` instead of `<ECR>/<IMAGE_NAME>`

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -413,9 +413,12 @@ jobs:
           "${IMG_CLUSTER_AUTOSCALER}")
 
           for image in "${images[@]}"; do
+            # Remove "neondatabase/" prefix from image name
+            image_without_prefix=${image#neondatabase/}
+
             image_from=${image}:${TAG}
-            image_ecr_dev=${ECR_DEV}/${image#neondatabase/}:${TAG}
-            image_ecr_prod=${ECR_PROD}/${image#neondatabase/}:${TAG}
+            image_ecr_dev="${ECR_DEV}/${image_without_prefix}:${TAG}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,7 +1,7 @@
 name: release
 on:
   # # UNCOMMENT TO ALLOW TESTING:
-  # pull_request:
+  pull_request:
```
bayandin, line 4: TODO: revert before merge

sharnoff (approved):
Oh, whoops - thanks! Now that you mention it, I remember originally using the hard-coded names instead of env vars exactly because of this issue, in #1259 :facepalm:

I like your solution better — maybe worth commenting what's going on with removing the `neondatabase/` prefix from `$image`? The prefix removal syntax in bash is one of those small things I often forget what it means :sweat_smile:

bayandin:
> I like your solution better — maybe worth commenting what's going on with removing the neondatabase/ prefix from $image? 

Sure will add a comment!

Also, probably in https://github.com/neondatabase/autoscaling/pull/1283 I'm going to drop `neondatabase/` part from the image name to start using GHCR as the default registry.

.github/workflows/release.yaml
bayandin: Reverted

PR #1316, opened 2025-03-11
neonvm: add support for mounting service account tokens

conradludgate:


conradludgate:
Hmm. This doesn't yet work for accessing the aws-iam-token volume mount which is automatically created for us. I'll have to think more for a potential solution. Ideally we'd have some way to encode that "this volumemount already exists, but please watch it anyway". Maybe a "none" disk?

Additionally, if the VM wants access to the kubernetes API (why??) then we would need to somehow get the local kubernetes API endpoint into the VM - I imagine we don't need this

mikhail-sakhnov:
@conradludgate may I ask to provide more context for the PR like what the issue is it related to and what do we want to achieve? We struggled a bit with this one during team sync, because apparently none of us has any context.

conradludgate:
> @conradludgate may I ask to provide more context for the PR like what the issue is it related to and what do we want to achieve? We struggled a bit with this one during team sync, because apparently none of us has any context.

As far as I understand, the migration assistant project will require access to S3. Their current access setup is not sufficient as they are limited to 1 hour, and it doesn't work on Azure (cc @NanoBjorn for corrections).

This project runs `pg_dump`/`pg_restore` within compute VMs provisioned by neonvm.

They would rather switch to a service account model, where their migration assistant VMs have a service account assigned, and this service account grants them access to assume an AWS role. Kubernetes does this with service account tokens, [mounted into the file system using a projected volume](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#launch-a-pod-using-service-account-token-projection). These tokens are not static, they are JWT based and will eventually expire. Before expiration, kubernetes will update the projected volume.

Exactly the same problem as with TLS keys (#1213), the VM must be able to observe changes to this file in order to refresh the AWS or Azure permissions. This is achieved by adding ProjectedVolume support to the synchronisation routine that is currently supported for both secrets and config maps.

PR #1317, opened 2025-03-12
reporting/batch: Remove newlines in JSONArrayBuilder

sharnoff:
The sink we're sending it to reads input as JSONL, which means that it expects each line to have a complete JSON object. Breaking up the object across lines causes issues.

ref https://neondb.slack.com/archives/C08HDEK2WFL/p1741787069351919?thread_ts=1741786042.532179

### DIFF
--- a/pkg/reporting/batch_jsonarray.go
+++ b/pkg/reporting/batch_jsonarray.go
@@ -51,7 +51,7 @@ func NewJSONArrayBuilder[E any](buf IOBuffer, nestedFields ...string) *JSONArray
 
 func (b *JSONArrayBuilder[E]) Add(event E) {
 	if b.started {
-		if _, err := b.buf.Write([]byte("\n\t,")); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reporting/batch_jsonarray.go
```diff
@@ -51,7 +51,7 @@ func NewJSONArrayBuilder[E any](buf IOBuffer, nestedFields ...string) *JSONArray
 
 func (b *JSONArrayBuilder[E]) Add(event E) {
 	if b.started {
-		if _, err := b.buf.Write([]byte("\n\t,")); err != nil {
+		if _, err := b.buf.Write([]byte(",")); err != nil {
```
mikhail-sakhnov, line 54: nit: should we have a unittest for that or e2e, may be not now (to speed up fixing) but later?

Omrigan (approved):

PR #1325, opened 2025-03-21
kernel-config: Update x86-64 config to gcc 13.3

cloneable:
Ubuntu 24.04 now ships with gcc toolchain 13.3 instead of 13.2.
Update the kernel config before making further changes.
Skip the arm64 kernel config because it's in a weird state.


sharnoff:
> Skip the arm64 kernel config because it's in a weird state.

IIUC this would mean that we get more drift between x86 and ARM, right? If so, let's fix ARM first

petuhovskiy:
+1 to Em, I think we should apply the same patch to the arm config to update gcc there as well.

mikhail-sakhnov:
+1 to the previous comments

cloneable:
> > Skip the arm64 kernel config because it's in a weird state.
> 
> IIUC this would mean that we get more drift between x86 and ARM, right? If so, let's fix ARM first

I'm all for it as long as it's done soon. I need some kernel changes to debug an issue and would like this in prod next week.

Issue #1326, opened 2025-03-21
make arm kernel config being autogenerated


mikhail-sakhnov:
Currently we have manually changed amd config to work on arm, let's make it generated by the kconfig tooling

related #1093

PR #1327, opened 2025-03-21
neonvm-kernel: run make olddefconfig over aarch kernel config

mikhail-sakhnov:
Results of applying `make olddefconfig`. Kernel validated via e2e tests and manually via running in local tilt environment. 

#1326


neonvm-kernel/linux-config-aarch64-6.6.64
```diff
@@ -294,191 +267,341 @@ CONFIG_TRACEPOINTS=y
...
-CONFIG_CC_HAS_SANE_STACKPROTECTOR=y
-
-#
-# Processor type and features
-#
 CONFIG_SMP=y
-# CONFIG_GOLDFISH is not set
-# CONFIG_IOSF_MBI is not set
-# CONFIG_SCHED_OMIT_FRAME_POINTER is not set
-CONFIG_HYPERVISOR_GUEST=y
```
petuhovskiy, line 335: `CONFIG_HYPERVISOR_GUEST` looks like a valid config option that should be set to `y`. And the same probably can be applied to other config options on lines below.

I wonder why `olddefconfig` decided to remove it?

petuhovskiy (approved):

neonvm-kernel/linux-config-aarch64-6.6.64
mikhail-sakhnov: https://cateee.net/lkddb/web-lkddb/HYPERVISOR_GUEST.html

it looks like it is defined as x86 specific.

petuhovskiy: Makes sense then! I was looking at this config on another website and it didn't say that this option was x86 only

PR #1337, opened 2025-04-08
plugin/state: Include "committed" resources in metrics

sharnoff:
Follow-up to #1289, increasing observability.
Part of neondatabase/cloud#25458.

This commit adds a new field to the plugin's `state.NodeResources` to store the amount of resources "committed" — i.e. the amount before applying any discounts from overcommit factors.

Then, surfacing it in the metrics is as simple as adding the field to the list from `(NodeResources[T]).Fields()`.


mikhail-sakhnov (approved):
looks reasonable

pkg/plugin/state/node.go
```diff
@@ -77,7 +77,12 @@ type NodeResources[T constraints.Unsigned] struct {
 	// * Other pods are scheduled without going through our scheduler plugin
 	Reserved T
 
-	// Migrating is the amount of T that we expect will be removed by ongoing live migration.
+	// Committed is the sum of all Pods' <resource>.Reserved values, BEFORE applying any overcommit
+	// factors.
+	Committed T
```
Omrigan, line 82: Why is it called `Commited`, what's the intuition?

Maybe something like `ReservedNominal` or `Effective` 🤔

sharnoff: Reasoning for `Committed` was to match the naming for "overcommit", which allows "committing" more resources than actually exist on the node.

Omrigan: Fair enough, makes sense.

Omrigan (approved):

PR #1341, opened 2025-04-09
vxlan-controller: enable learning for neon-vxlan0

Omrigan:
Unfortunately, it seems it is impossible to enable learning on an existing interface, so we would have to use -delete for this to have effect.


neonvm-vxlan-controller/cmd/main.go
```diff
@@ -164,9 +164,10 @@ func createVxlanInterface(name string, vxlanID int, ownIP string, bridgeName str
 		LinkAttrs: netlink.LinkAttrs{
 			Name: name,
 		},
-		VxlanId: vxlanID,
-		SrcAddr: net.ParseIP(ownIP),
-		Port:    4789,
+		VxlanId:  vxlanID,
+		SrcAddr:  net.ParseIP(ownIP),
+		Port:     4789,
+		Learning: true,
```
mikhail-sakhnov, line 170: (question, not-blocking): Should it be passed as an cli argument to the binary? That way only a redeployment would need to be done if we need to rollback the change.

mikhail-sakhnov (approved):

sharnoff:
> we would have to use -delete for this to have effect

What would rollout look like here?

neonvm-vxlan-controller/cmd/main.go
Omrigan: I don't think there is a need to rollback this change quickly, so I'll revert this PR if needed.

Omrigan:
> > > we would have to use -delete for this to have effect
> > 
> > 
> > What would rollout look like here?
> 
> I'll do the `-delete` rollout myself later this week on staging.
> 
> To not forget prod, I've created [neondatabase/cloud#27215](https://github.com/neondatabase/cloud/issues/27215)

Actually, it is currently recreated on stop, see https://github.com/neondatabase/autoscaling/pull/1349.

PR #1342, opened 2025-04-09
neonvm-runner: Use HTTP 502,504 for neonvm-daemon issues

### LINKED ISSUE #1334: neonvm-runner should return 504 on neonvm-daemon timeout

Currently, neonvm-runner returns HTTP 500 when a request to neonvm-daemon times out.

AFAICT, a more correct error code would be 504 "Gateway Timeout", which would also help disambiguate.

*Some context in this thread: https://neondb.slack.com/archives/C03TN5G758R/p1743716781911049*

## Implementation ideas

We "just" need to investigate the error type instead of blanket returning HTTP 500.

sharnoff:
Currently neonvm-runner returns HTTP 500 on any error with getting or setting CPU inside the VM. This could be more precise.

This PR adopts the following conventions:

1. HTTP 504 "Gateway Timeout" when the request from neonvm-runner to neonvm-daemon fails due to timeout.
2. HTTP 502 "Bad Gateway" for any issue that is plausibly due to an issue with neonvm-daemon. For example:
    1. Any HTTP status from neonvm-daemon other than 200 "OK"
    2. Any error reading the response body after we receive a 200 "OK"
    3. Any errors parsing the resonse from neonvm-daemon
3. HTTP 500 "Internal Server Error" for everything else

For more, see: https://www.rfc-editor.org/rfc/rfc9110.html#section-15.6

Closes #1334, which is about the timeout case.

### DIFF
--- a/neonvm-runner/cmd/httpserver.go
+++ b/neonvm-runner/cmd/httpserver.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/samber/lo"
 	"go.uber.org/zap"
 
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
@@ -19,8 +20,8 @@ import (
 )
 
 type cpuServerCallbacks struct {
-	get   func(*zap.Logger) (*vmv1.MilliCPU, error)
-	set   func(*zap.Logger, vmv1.MilliCPU) error
+	get   func(*zap.Logger) (*vmv1.MilliCPU, *int, error)
+	set   func(*zap.Logger, vmv1.MilliCPU) (*int, error)
 	ready func(*zap.Logger) bool
 }
 
@@ -87,7 +88,7 @@ func handleCPUChange(
 	logger *zap.Logger,
 	w http.ResponseWriter,
 	r *http.Request,
-	set func(*zap.Logger, vmv1.MilliCPU) error,
+	set func(*zap.Logger, vmv1.MilliCPU) (*int, error),
 ) {
 	if r.Method != "POST" {
 		logger.Error("unexpected method", zap.String("method", r.Method))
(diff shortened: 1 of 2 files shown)


neonvm-runner/cmd/httpserver.go
```diff
@@ -87,7 +88,7 @@ func handleCPUChange(
 	logger *zap.Logger,
 	w http.ResponseWriter,
 	r *http.Request,
-	set func(*zap.Logger, vmv1.MilliCPU) error,
+	set func(*zap.Logger, vmv1.MilliCPU) (*int, error),
```
mikhail-sakhnov, line 91: what is the reason to use pointer for a error code?

sharnoff: The intent was to help avoid needing to write the HTTP status in more places (so if `err != nil` then `nil` is converted to HTTP 500, and if `err != nil` then the status wouldn't be used anyways)

I'm not totally sure it's worth it though TBH

mikhail-sakhnov: >The intent was to help avoid needing to write the HTTP status in more places

Not sure I follow how it helps :( 

What can't be done with int but can be done with *int in that scenario, assuming we use default value 0 as a logical nil?

sharnoff: It can be done either way, I just thought the semantics of having zero as a special value was a little odd :sweat_smile:

mikhail-sakhnov: In general in go it is a common practice to use default values (I guess mostly to avoid putting more pressure on GC and mem fragmentation, both are not related to our scenario) so at least to me that code was really confusing to read. In general I'd rather stick with common practices even if I don't like them, just to make cognitive load during reading less. Anyway, no hard feelings here.

sharnoff: Makes sense - implemented in 4c35c85f898fc68b17f7c539c272b2ffe0f3d2bb

neonvm-runner/cmd/httpserver.go
```diff
@@ -12,15 +12,16 @@ import (
...
 	vmv1 "github.com/neondatabase/autoscaling/neonvm/apis/neonvm/v1"
 	"github.com/neondatabase/autoscaling/pkg/api"
 )
 
 type cpuServerCallbacks struct {
-	get   func(*zap.Logger) (*vmv1.MilliCPU, error)
-	set   func(*zap.Logger, vmv1.MilliCPU) error
+	get   func(*zap.Logger) (*vmv1.MilliCPU, *int, error)
+	set   func(*zap.Logger, vmv1.MilliCPU) (*int, error)
 	ready func(*zap.Logger) bool
```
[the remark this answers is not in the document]
sharnoff: Makes sense - made the change in d1e027940913b29c73baac10444df57f2c756426

mikhail-sakhnov (approved):

PR #1344, opened 2025-04-09
neonvm-controller: Don't emit k8s events

sharnoff:
On larger clusters with a lot of churn, we see substantial amounts of space in etcd taken up by events from neonvm-controller, even with a relatively short TTL.

So, let's completely remove event reporting in the short term to mitigate risks of running out of space in EKS.

In the meantime, all the relevant information should still be available in the neonvm-controller logs, and we can re-add select events if we think it's worthwhile.

Closes neondatabase/cloud#27262.
Ref https://neondb.slack.com/archives/C08N5D2CF7A/p1744224210018349
Ref https://neondb.slack.com/archives/C03TN5G758R/p1744225420614349

---

**Notes for review:** I haven't examined the logs yet, will do that with the output from e2e tests on CI. I think there's also a lot of cleanup we should to make our log lines more structured, but better to leave that for a follow-up IMO.

### DIFF
--- a/neonvm-controller/cmd/main.go
+++ b/neonvm-controller/cmd/main.go
@@ -219,12 +219,11 @@ func main() {
 	defer ipam.Close()
 
 	vmReconciler := &controllers.VMReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		Recorder: mgr.GetEventRecorderFor("virtualmachine-controller"),
-		Config:   rc,
-		Metrics:  reconcilerMetrics,
-		IPAM:     ipam,
+		Client:  mgr.GetClient(),
+		Scheme:  mgr.GetScheme(),
+		Config:  rc,
+		Metrics: reconcilerMetrics,
+		IPAM:    ipam,
 	}
 	vmReconcilerMetrics, err := vmReconciler.SetupWithManager(mgr)
 	if err != nil {
@@ -241,11 +240,10 @@ func main() {
 	}
 
 	migrationReconciler := &controllers.VirtualMachineMigrationReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		Recorder: mgr.GetEventRecorderFor("virtualmachinemigration-controller"),
-		Config:   rc,
-		Metrics:  reconcilerMetrics,
+		Client:  mgr.GetClient(),
+		Scheme:  mgr.GetScheme(),
+		Config:  rc,
+		Metrics: reconcilerMetrics,
 	}
 	migrationReconcilerMetrics, err := migrationReconciler.SetupWithManager(mgr)
 	if err != nil {
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -50,7 +50,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apiserver/pkg/storage/names"
-	"k8s.io/client-go/tools/record"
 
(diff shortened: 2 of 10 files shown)


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -239,20 +235,15 @@ func (r *VMReconciler) doFinalizerOperationsForVirtualMachine(ctx context.Contex
 	log := log.FromContext(ctx)
 
 	// The following implementation will raise an event
-	r.Recorder.Event(vm, "Warning", "Deleting",
-		fmt.Sprintf("Custom Resource %s is being deleted from the namespace %s",
-			vm.Name,
-			vm.Namespace))
+	log.Info("Deleting VirtualMachine")
```
mickael-carl, line 238: Do we really need that logline? Realistically it conveys almost 0 info. I'd suggest adding either which machine gets deleted or just remove this altogether

sharnoff: The VM object it refers to is already included in the log context IIRC

pkg/neonvm/controllers/vm_controller.go
```diff
@@ -239,20 +235,15 @@ func (r *VMReconciler) doFinalizerOperationsForVirtualMachine(ctx context.Contex
...
 	if vm.Spec.ExtraNetwork != nil {
 		ip, err := r.IPAM.ReleaseIP(ctx, types.NamespacedName{Name: vm.Name, Namespace: vm.Namespace})
 		if err != nil {
-			return fmt.Errorf("fail to release IP: %w", err)
+			return fmt.Errorf("fail to release overlay IP: %w", err)
 		}
-		message := fmt.Sprintf("Released IP %s", ip.String())
-		log.Info(message)
-		r.Recorder.Event(vm, "Normal", "OverlayNet", message)
+		log.Info(fmt.Sprintf("Released overlay IP %s", ip.String()))
```
mickael-carl, line 246: note: the idiomatic way to log with this lib is probably more like:
```suggestion
		log.Info("Released overlay IP", ip, ip.String())
```

mickael-carl: https://pkg.go.dev/github.com/go-logr/logr#Logger.Info has more examples

sharnoff: +1; see "notes for review" in the PR description:

> I think there's also a lot of cleanup we should to make our log lines more structured, but better to leave that for a follow-up IMO.

mickael-carl:
On verbosity: a lot of the existing logs should realistically be debug-level (V>5? maybe?)

neonvm-controller/cmd/main.go
```diff
@@ -219,12 +219,11 @@ func main() {
 	defer ipam.Close()
 
 	vmReconciler := &controllers.VMReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		Recorder: mgr.GetEventRecorderFor("virtualmachine-controller"),
```
petuhovskiy, line 224: Maybe it'd be better to just implement a noop `record.EventRecorder` and pass it here?

Omrigan (approved):

neonvm-controller/cmd/main.go
petuhovskiy: Yes, my thinking is that having a shim would be less moving parts and would be less risky for a hotfix. But if we know for sure that we want to move away from recording k8s events, current diff is better for this.

petuhovskiy (approved):

neonvm-controller/cmd/main.go
sharnoff: Yeah, I think it's unlikely we're able to support writing to k8s events in the medium term unfortunately :/

PR #1356, opened 2025-04-14
neonvm-controller: remove nodelete build tag

Omrigan:
Closes https://github.com/neondatabase/cloud/issues/18828


Omrigan:
Given the absence of CI failures, it seems we didn't even rely on that in e2e 🤔

sharnoff:
Is this the route we want to go here? Without the equivalent of this behavior, we lose the ability to see logs from VMs in failed CI runs.

What's your thinking?

Omrigan:
> Is this the route we want to go here? Without the equivalent of this behavior, we lose the ability to see logs from VMs in failed CI runs.
> 
> What's your thinking?

Let me think about mounting log folder from the node as a replacement.

Omrigan:
Potential solution: https://github.com/neondatabase/autoscaling/pull/1373

mikhail-sakhnov:
>Is this the route we want to go here? Without the equivalent of this behavior, we lose the ability to see logs from VMs in failed CI runs.

I am more concerned not about losing logs but about losing the state of the failed pod, it is great to force pods not being deleted to inspect the events, metadata and other stuff by kubectl. 
No hard opinion, but I'd prefer to have env variable or cli argument to force the same behavior as we have with build tag. 

Speaking the mentioned PR with fluentbit and log collections, regardless of which way we are going to take in this PR, I think collecting logs via fluentbit is a nice improval and we should go for it.

PR #1365, opened 2025-04-22
vm-deploy.yaml: fix TLS fields in the example

mikhail-sakhnov:
Without the change the `vm-deploy.yaml` fails with 

```
Error from server (BadRequest): error when creating "vm-deploy.yaml": VirtualMachine in version "v1" cannot be handled as a VirtualMachine: strict decoding error: unknown field "spec.tlsCertificateIssuer", unknown field "spec.tlsExpiration", unknown field "spec.tlsRenewal"
```

### DIFF
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -14,9 +14,11 @@ spec:
   schedulerName: autoscale-scheduler
   enableNetworkMonitoring: true
   enableSSH: true
-  tlsCertificateIssuer: "neon-ca-issuer"
-  tlsExpiration: 5m
-  tlsRenewal: 3m
+  tls:
+    certificateIssuer: "neon-ca-issuer"
+    expireAfter: 5m
+    renewBefore: 3m
+    serverName: "example.local"
   guest:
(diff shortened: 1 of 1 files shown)


vm-deploy.yaml
```diff
@@ -14,9 +14,6 @@ spec:
   schedulerName: autoscale-scheduler
   enableNetworkMonitoring: true
   enableSSH: true
-  tlsCertificateIssuer: "neon-ca-issuer"
-  tlsExpiration: 5m
-  tlsRenewal: 3m
```
conradludgate, line 19: Ah, I see. we later changed the format, I guess I missed the example.

We could fix it to
```yaml
tls:
  certificateIssuer: "neon-ca-issuer"
  expireAfter: 5m
  renewBefore: 3m
  serverName: "example.local"
```

but I don't care much whether we just get rid of it either

conradludgate (approved):

vm-deploy.yaml
mikhail-sakhnov: Thanks, I fixed the example to reflect the correct format

PR #1366, opened 2025-04-22
neonvm: use virtio console if use-virtio-console arg is true

mikhail-sakhnov:
The PR introduces VM field based on the value of which qemu uses either pty device or virtio-serial for /dev/console.

The way I tested the rate was dead simple (thanks to @Omrigan for the idea), by using cat inside the VM to direct 16mb file to the /dev/console.

Comment https://github.com/neondatabase/autoscaling/pull/1366#issuecomment-2837885515 has more results explained.

Both pods run on the same host x86 machine.

```
# pty pod

## preparation
vm-postgres16-disk-test-serial-bv8jd:~# dd if=/dev/urandom bs=1M count=16 | base64
 > random_data.txt
16+0 records in
16+0 records out
16777216 bytes (17 MB, 16 MiB) copied, 0.0733734 s, 229 MB/s

## run 0
vm-postgres16-disk-test-serial-bv8jd:~# time cat random_data.txt > /dev/console
real	2m 1.70s
user	0m 0.00s
sys	2m 1.11s

## run 1
vm-postgres16-disk-test-serial-bv8jd:~# time cat random_data.txt > /dev/ttyS1
real	2m 3.12s
user	0m 0.00s
sys	2m 2.53s



# virtio pod
vm-postgres16-disk-test-w28vh:~# dd if=/dev/urandom bs=1M count=16 | base64 > random_data.txt
16+0 records in
16+0 records out
16777216 bytes (17 MB, 16 MiB) copied, 0.0536018 s, 313 MB/s

## run 0
vm-postgres16-disk-test-w28vh:~# time cat random_data.txt > /dev/console
real	0m 4.62s
user	0m 0.00s
sys	0m 4.60s

## run 1
vm-postgres16-disk-test-w28vh:~# time cat random_data.txt > /dev/hvc0
real	0m 4.61s
user	0m 0.00s
sys	0m 4.59s
```

virtio console start is delayed compare to a native pty, which means we loose first 0.1 seconds of kernel logs (not the user-space logs though).

To mitigate the issue with delayed virtio console init output of the uname -a is added to the init via https://github.com/neondatabase/autoscaling/pull/1366/commits/2853ddfec38db2c460131e58cb00a44f743ca2c3, since kernel version was the most useful information we lost.

### DIFF
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -96,6 +96,10 @@ type VirtualMachineSpec struct {
 	// +optional
 	RunnerPort int32 `json:"runnerPort,omitempty"`
 
+	// +kubebuilder:default:=true
+	// +optional
+	VirtioConsole bool `json:"virtioConsole,omitempty"`
+
 	// +kubebuilder:default:=5
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds"`
--- a/neonvm/apis/neonvm/v1/virtualmachine_types.go
+++ b/neonvm/apis/neonvm/v1/virtualmachine_types.go
@@ -96,10 +96,6 @@ type VirtualMachineSpec struct {
 	// +optional
 	RunnerPort int32 `json:"runnerPort,omitempty"`
 
-	// +kubebuilder:default:=true
-	// +optional
-	VirtioConsole bool `json:"virtioConsole,omitempty"`
-
 	// +kubebuilder:default:=5
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds"`
--- a/vm-builder/files/inittab
+++ b/vm-builder/files/inittab
@@ -1,7 +1,7 @@
 ::sysinit:/neonvm/bin/vminit
 ::once:/neonvm/bin/touch /neonvm/vmstart.allowed
 # we might be using hvc0 console, so we need to print the kernel version in user space, after virtio console is initialized
-::once:/neonvm/bin/uname -a
+::once:cat /proc/version
 ::respawn:/neonvm/bin/udhcpc -t 1 -T 1 -A 1 -f -i eth0 -O 121 -O 119 -s /neonvm/bin/udhcpc.script
 ::respawn:/neonvm/bin/udevd
 ::wait:/neonvm/bin/udev-init.sh
--- a/vm-builder/files/inittab
+++ b/vm-builder/files/inittab
@@ -1,5 +1,7 @@
 ::sysinit:/neonvm/bin/vminit
 ::once:/neonvm/bin/touch /neonvm/vmstart.allowed
+# we might be using hvc0 console, so we need to print the kernel version in user space, after virtio console is initialized
+::once:/neonvm/bin/uname -a
 ::respawn:/neonvm/bin/udhcpc -t 1 -T 1 -A 1 -f -i eth0 -O 121 -O 119 -s /neonvm/bin/udhcpc.script
 ::respawn:/neonvm/bin/udevd
 ::wait:/neonvm/bin/udev-init.sh
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -19,6 +19,7 @@ spec:
     expireAfter: 5m
     renewBefore: 3m
     serverName: "example.local"
+  virtioConsole: true
   guest:
     cpus: { min: 0.25, use: 0.25, max: 1.25 }
     memorySlotSize: 1Gi
--- a/vm-deploy.yaml
+++ b/vm-deploy.yaml
@@ -19,7 +19,6 @@ spec:
     expireAfter: 5m
     renewBefore: 3m
     serverName: "example.local"
-  virtioConsole: true
   guest:
     cpus: { min: 0.25, use: 0.25, max: 1.25 }
     memorySlotSize: 1Gi
--- a/neonvm-controller/cmd/main.go
+++ b/neonvm-controller/cmd/main.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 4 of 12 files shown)


vm-deploy.yaml
```diff
@@ -14,9 +14,9 @@ spec:
   schedulerName: autoscale-scheduler
   enableNetworkMonitoring: true
   enableSSH: true
-  tlsCertificateIssuer: "neon-ca-issuer"
-  tlsExpiration: 5m
-  tlsRenewal: 3m
+  virtioConsole: true
+  extraNetwork:
```
mikhail-sakhnov, line 18: not related, was debugging issues with kind.

Omrigan:
Couple more thoughts about throughput measurement:
1. It makes sense to measure different sizes of file, to verify there is a linear dependency with constant equal to 0.
2. Better to express the throughput in MB/s.

mikhail-sakhnov:
> Couple more thoughts about throughput measurement:
> 
> 1. It makes sense to measure different sizes of file, to verify there is a linear dependency with constant equal to 0.
> 2. Better to express the throughput in MB/s.

here is the commit with benchmarks https://github.com/neondatabase/autoscaling/commit/0fea4521bb8926bb250ef99830fb19b923dbedb7

here ist he commit with results https://github.com/neondatabase/autoscaling/commit/77c960c4d4dfbb26ed41f8a502940b822fc52922

For each step generator generates X kbs for 15 seconds, generator should be ran in the vm.

Files with results have collectors output.

steps.0 uses steps `1000 2000 5000 10000 15000 25000 50000 100000`
steps.1 uses steps `1000 2000 5000 10000 15000 25000 50000 100000 200000 300000 600000`

neonvm/apis/neonvm/v1/virtualmachine_types.go
```diff
@@ -96,6 +96,10 @@ type VirtualMachineSpec struct {
 	// +optional
 	RunnerPort int32 `json:"runnerPort,omitempty"`
 
+	// +kubebuilder:default:=true
+	// +optional
+	VirtioConsole bool `json:"virtioConsole,omitempty"`
```
petuhovskiy, line 101: Our end goal here is to enable it everywhere, and after that it's unlikely that we will want to switch it back.

I think it might be easier (for rollout) to add this config to neonvm controller and runner, without modifying the CRD. Then we will be able to control the rollout without modifying control plane.

WDYT?

Found some previous feature that was implemented like this:
https://github.com/neondatabase/autoscaling/pull/1034/files#diff-47c5f6211bd0862ce74d260f1d133abd65980eaeb0416e960b72da66db426158R599

vm-builder/files/inittab
```diff
@@ -1,5 +1,7 @@
 ::sysinit:/neonvm/bin/vminit
 ::once:/neonvm/bin/touch /neonvm/vmstart.allowed
+# we might be using hvc0 console, so we need to print the kernel version in user space, after virtio console is initialized
+::once:/neonvm/bin/uname -a
```
petuhovskiy, line 4: For comparison:
```
root:~# cat /proc/version
Linux version 6.12.24 (root@buildkitsandbox) (x86_64-linux-gnu-gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42) #1 SMP PREEMPT_DYNAMIC Wed Apr 23 17:31:20 UTC 2025

root:~# uname -a
Linux vm-compute-msf-unreal-xxdpm 6.12.24 #1 SMP PREEMPT_DYNAMIC Wed Apr 23 17:31:20 UTC 2025 x86_64 GNU/Linux
```

Not a blocker, but I slightly prefer the first option.

petuhovskiy (commented):
Mostly LGTM

neonvm/apis/neonvm/v1/virtualmachine_types.go
mikhail-sakhnov: Sounds fair, though I am in general not sure how granular we want to go with that, is it going to be something like "gradually roll out per region" or "rollout to staging and validate, then roll out to production"!

>Our end goal here is to enable it everywhere, and after that it's unlikely that we will want to switch it back.

I was thinking about enabling it by default everywhere but leaving a way to switch off for a particular VM if we found issues with their logs. Later we can drop the 

>I think it might be easier (for rollout) to add this config to neonvm controller and runner, without modifying the CRD. Then we will be able to control the rollout without modifying control plane.

Do we actually need to envolve control plane into that? Can we, in case we want to rollback, revert the PR and cut of a release?

To sum up, I am open to change the code to control target console via controller/runner arguments, but that will leave no option to change for a particular VM, and with that set up it feels almost the same as just use it by default everywhere (which is not something I am against).

vm-builder/files/inittab
mikhail-sakhnov: Changed in [79041126e09d871f68da63694055f530891d906a](https://github.com/neondatabase/autoscaling/pull/1366/commits/79041126e09d871f68da63694055f530891d906a)

mikhail-sakhnov:
```
[    0.304055] EXT4-fs (vda): resizing filesystem from 262144 to 2097152 blocks
[    0.307157] EXT4-fs (vda): resized filesystem to 2097152
The filesystem on /dev/vda is now 2097152 (4k) blocks long.

Linux version 6.6.64 (root@buildkitsandbox) (x86_64-linux-gnu-gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42) #1 SMP PREEMPT_DYNAMIC Mon Feb  3 10:05:38 UTC 2025
udhcpc: started, v1.35.0
starting up with netlink and the input layer
```

neonvm/apis/neonvm/v1/virtualmachine_types.go
petuhovskiy: I think global on/off is acceptable here and to me it looks like the simplest way to rollout. Avoiding cplane modifications is the main motivation here.

> Can we, in case we want to rollback, revert the PR and cut of a release?

If we modify the CRD, I think it's almost impossible to rollback.

> but that will leave no option to change for a particular VM

fine with me

mikhail-sakhnov: >If we modify the CRD, I think it's almost impossible to rollback.

I am not sure I understand where is that coming from, but I see the concernt/opinion, let's make it via controller args

mikhail-sakhnov: changed in https://github.com/neondatabase/autoscaling/pull/1366/commits/3b9a6999791ce44b6591846324c314930f547ee3

petuhovskiy (approved):

sharnoff (approved):
LGTM, nice :)

One question -- what's the plan for rollout? Wait for a compute release with the vm-builder changes, and then enable `--use-virtio-console` on neonvm-controller?

mikhail-sakhnov:
>LGTM, nice :)

> One question -- what's the plan for rollout? Wait for a compute release with the vm-builder changes, and then enable --use-virtio-console on neonvm-controller?

We can do it either way, no personal preference:
- wait for new compute release
- deploy without waiting for a new release
 
The only change inside the vm is the cat /proc/version during VM boot, which, if missing for a limited span of the time, doesn't change a lot of stuff I think, unless we have metrics/alerts set up for the log line "Linux version". I have no hard opinion on that and since it affects the ongloing @petuhovskiy's kernel related investigations, I'd leave the decision up to @petuhovskiy

petuhovskiy:
I'm fine with not having kernel version in the logs for a few days – if it means that we will rollout virtio-console to prod faster.

PR #1373, opened 2025-04-29
local: add fluent-bit log collector

Omrigan:
This solves the problem of getting logs while debugging locally. Something I've been hacking on in the background, not sure we want to go this route.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,6 +392,9 @@ deploy: check-local-context docker-build load-images render-manifests kubectl ##
 	$(KUBECTL) -n kube-system rollout status deployment autoscale-scheduler
 	$(KUBECTL) apply -f $(RENDERED)/autoscaler-agent.yaml
 	$(KUBECTL) -n kube-system rollout status daemonset autoscaler-agent
+	# Apply Fluent Bit logging agent
+	@echo "Applying Fluent Bit manifests..."
+	$(KUBECTL) apply -f tests/fluent-bit/
 
 .PHONY: load-images
 load-images: check-local-context kubectl kind k3d ## Push docker images to the local kind/k3d cluster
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -392,6 +392,9 @@ deploy: check-local-context docker-build load-images render-manifests kubectl ##
 	$(KUBECTL) -n kube-system rollout status deployment autoscale-scheduler
 	$(KUBECTL) apply -f $(RENDERED)/autoscaler-agent.yaml
 	$(KUBECTL) -n kube-system rollout status daemonset autoscaler-agent
+	# Apply Fluent Bit logging agent
+	@echo "Applying Fluent Bit manifests..."
```
mikhail-sakhnov, line 396: do we need this echo? any other manifest is applied "silently"

mikhail-sakhnov:
looks good to me, great stuff!

Couple of thoughts a loud:

- if I recall correctly, there is initiative to improve getting logs from live compute units, and it's going to use rsyslogd, may be once it is ready, we might want to change configs from this PR to have more unified settings with live images.

- will the fluentbit also be deployed during CI runs? If so, I'd suggest to archive logs collected by fluentbit as run artifacts, would make life so much easier while debugging e2e failures.

sharnoff (approved):
Broadly lgtm. I have no clue about the configuration — if it works it works :smile:

Beyond that, a few notes:

1. If you adapted the config files from somewhere, can you add a link at the top so we can re-discover it later if needed?
2. IMO CI improvements are ok to do as a follow-up, but +1 that we should do what @mikhail-sakhnov suggested
3. Could you make sure that the logs directory is gitignore'd?

Omrigan:
@mikhail-sakhnov:  
>  if I recall correctly, there is initiative to improve getting logs from live compute units, and it's going to use rsyslogd, may be once it is ready, we might want to change configs from this PR to have more unified settings with live images.

Maybe, but AFAIK, even with rsyslogd we are still aiming for collecting logs from pod's stdout.

But yes, let's aim to keep this setup consistent with production setup.

> will the fluentbit also be deployed during CI runs? If so, I'd suggest to archive logs collected by fluentbit as run artifacts, would make life so much easier while debugging e2e failures.

Created https://github.com/neondatabase/autoscaling/issues/1394

@sharnoff:
> If you adapted the config files from somewhere, can you add a link at the top so we can re-discover it later if needed?
    
I adapted it from AI 😨 

>Could you make sure that the logs directory is gitignore'd?

Done 👍

mikhail-sakhnov (approved):

PR #1381, opened 2025-05-14
neonvm: Remove neonvm-runner version [2/2]

sharnoff:
We no longer use it anywhere, so we should probably remove it to get rid of unnecessary code.

This PR is the second of two changes, removing the remaining places where we set the neonvm-runner version label. It must be a separate change so that neonvm-controller can first be updated to ignore whether the label is present, for rollback safety.

See #1380 for more.

---

**NOTE: This PR must not be merged until #1380 has been released!**

### DIFF
--- a/pkg/neonvm/controllers/vm_controller.go
+++ b/pkg/neonvm/controllers/vm_controller.go
@@ -876,14 +876,10 @@ func updatePodMetadataIfNecessary(ctx context.Context, c client.Client, vm *vmv1
 		ignoreExtra map[string]bool // use bool here so `if ignoreExtra[key] { ... }` works
 	}{
 		{
-			metaField: "labels",
-			expected:  labelsForVirtualMachine(vm, nil), // don't include runner version
-			actual:    runnerPod.Labels,
-			ignoreExtra: map[string]bool{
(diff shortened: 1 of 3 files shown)


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -876,14 +876,10 @@ func updatePodMetadataIfNecessary(ctx context.Context, c client.Client, vm *vmv1
...
-			metaField: "labels",
-			expected:  labelsForVirtualMachine(vm, nil), // don't include runner version
-			actual:    runnerPod.Labels,
-			ignoreExtra: map[string]bool{
-				// Don't override the runner pod version - we need to keep it around without
-				// changing it; otherwise it's not useful!
-				vmv1.RunnerPodVersionLabel: true,
-			},
+			metaField:   "labels",
+			expected:    labelsForVirtualMachine(vm), // don't include runner version
```
mikhail-sakhnov, line 880: comment is outdated now, with the changes we don't have runner version argument at all

sharnoff: good catch, thanks! Fixed in a6d39ee32c7171aff2a70efbd865bc183526289e

PR #1382, opened 2025-05-15
neonvm-controller: Move CLI flags to separate file

sharnoff:
I was looking at adding a few additional CLI flags, figured it's worth refactoring main beforehand, so it's at least a bit more manageable.


mikhail-sakhnov (approved):
keeping in mind that e2e tests are green, lgtm.

PR #1384, opened 2025-05-15
CI: set environment for Azure to push images on tags

bayandin:
Azure OICD is configured for GitHub environments to make it possible to push docker images on tag creation (otherwise we have to list all tags which is not possible to do — Azure doesn't support wildcards https://github.com/Azure/azure-workload-identity/issues/373).

`environment` was lost from the workflow in https://github.com/neondatabase/autoscaling/pull/1283

This PR adds `environment` back to `_push-to-container-registry.yml` workflow.


sharnoff (approved):
lgtm, thanks!

PR #1385, opened 2025-05-15
neonvm-controller: use RequeueAfter=1s only when scaling or migrating

Omrigan:
Previous code also matched e.g. VmSucceeded.

Found here: https://neondb.slack.com/archives/C03TN5G758R/p1747326017592529


pkg/neonvm/controllers/vm_controller.go
```diff
@@ -214,11 +214,11 @@ func (r *VMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		}
 	}
 
+	requeueAfter := 15 * time.Second
 	// Only quickly requeue if we're scaling or migrating. Otherwise, we aren't expecting any
 	// changes from QEMU, and it's wasteful to repeatedly check.
-	requeueAfter := time.Second
-	if vm.Status.Phase == vmv1.VmPending || vm.Status.Phase == vmv1.VmRunning {
-		requeueAfter = 15 * time.Second
+	if vm.Status.Phase == vmv1.VmScaling || vm.Status.Phase == vmv1.VmMigrating {
```
sharnoff, line 220: Should we also include `PreMigrating` ? Entirely up to you, just flagging

sharnoff (approved):
Nice, good catch :)

pkg/neonvm/controllers/vm_controller.go
Omrigan: Fair point, thanks.

Although maybe with migration logic evolving in future, maybe we can get rid of all of that.

PR #1387, opened 2025-05-22
scheduler: Fix randomization in NormalizeScores

petuhovskiy:
There was an issue with incorrect returning of randomized scores. This commit fixes it and adds a test for randomness


sharnoff (approved):
Oof — good catch!

Issue #1391, opened 2025-06-03
Bug: kernel null pointer dereference on virtio-mem hotplug


sharnoff:
## Environment

Production

## Context

We recently upgraded from kernel version 6.6.64 to 6.12.26 in production (ref #1376).

Since then, we've seen a _very small_ (~1 in a million) rate of our VMs hitting a bug that looks roughly like:

1. We try to hotplug more memory with virtio-mem
2. There's an allocation failure while setting up the physical pagetables
3. That allocation failure isn't handled, and we end up dereferencing the null pointer

See this slack thread for more: https://neondb.slack.com/archives/C0807C9SSJ2/p1748363834798349

## Impact

1. The rate of occurrence is very small
2. When VMs hit this, the kworker responsible for handling virtio-mem operations exits _and does not restart_ — i.e. 

So overall impact is pretty small: the VMs continue operating normally, with memory scaling broken.
(notably in contrast to the kcompactd issue, where the VM will eventually fall over)

## Example stack trace(s)

Here's an example of what we saw for a particular VM:

<details>
<summary>Allocation failure</summary>

```
[ 1259.521867] virtio_mem virtio1: plugged size: 0x0
[ 1259.521939] virtio_mem virtio1: requested size: 0x40000000
[ 1259.534610] kworker/0:2: page allocation failure: order:0, mode:0x920(GFP_ATOMIC|__GFP_ZERO), nodemask=(null),cpuset=/,mems_allowed=0
[ 1259.534738] CPU: 0 UID: 0 PID: 140 Comm: kworker/0:2 Not tainted 6.12.26 #1
[ 1259.534740] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS rel-1.16.2-0-gea1b7a073390-prebuilt.qemu.org 04/01/2014
[ 1259.534742] Workqueue: events_freezable virtio_mem_run_wq
[ 1259.534748] Call Trace:
[ 1259.534750]  <TASK>
[ 1259.534751]  dump_stack_lvl+0x5b/0x70
[ 1259.534755]  dump_stack+0x10/0x20
[ 1259.534756]  warn_alloc+0x103/0x180
[ 1259.534760]  __alloc_pages_slowpath.constprop.0+0x738/0xf30
[ 1259.534763]  __alloc_pages_noprof+0x1e9/0x340
[ 1259.534765]  alloc_pages_mpol_noprof+0x47/0x100
[...]

PR #1392, opened 2025-06-03
neonvm-kernel: Patch null ptr deref in x86 pagetable setup

sharnoff:
ref https://neondb.slack.com/archives/C0807C9SSJ2/p1748363834798349
ref https://neondb.slack.com/archives/C03TN5G758R/p1748944236243709
ref #1391

Broader discussion in the slack threads, but at a high level:

1. We started seeing null pointer dereferences after failed allocations in `phys_pud_init()` and friends
2. Those functions don't handle allocation failure (because normally they're called during boot, where the allocation cannot fail)
3. The allocation uses the `GFP_ATOMIC` flag, which *can* fail
4. This has been the case for a while, so whatever reason we're starting to see allocation failures, this code probably does need to be able to handle them.

This change consists of several patches, matching [this thread on LKML](https://lore.kernel.org/all/7d0d307d-71eb-4913-8023-bccc7a8a4a3d@neon.tech/).

The main idea of the change is to propagate allocation failures by changing the `phys_*_init()` functions to return `-ENOMEM` (which in turn requires changing their return type).

This patch was tested by injecting failure in `phys_pud_init()` for a single call to `alloc_low_page()` after boot, confirming that:

* an allocation failure will no longer cause a null pointer dereference
* virtio-mem hotplug succeeds after the initial failure, if made to retry

Check the slack threads for more detail on that testing.


stradig (commented):
I looked at the C code, looks good to me modulo some comments

neonvm-kernel/.gitignore
```diff
@@ -1 +1,6 @@
 vmlinuz
+
+# When working on kernel patches locally, it's sometimes helpful to download the source to this
+# directory. If it's not excluded from git, certain operations become unexpectedly VERY expensive.
```
Omrigan, line 4:
```suggestion
# Exclude files for local kernel development.
# Otherwise, certain git operations become unexpectedly VERY expensive.
```

nit: a bit more concise IMO

.dockerignore
```diff
@@ -4,3 +4,8 @@ bin/
 testbin/
 
 rendered_manifests
+
+# When working on kernel patches locally, don't try to include the entire kernel source in the
+# docker context. See neonvm-kernel/.gitignore for more.
```
Omrigan, line 9:
```suggestion
# Exclude files for local kernel development. Those are not needed inside docker context.
```

nit: a bit more concise IMO

neonvm-kernel/patches/00-phys_pgtable_nullptr.patch
Omrigan, line 1: Can you add description to this patch, suitable with how it would be sent upstream?

neonvm-kernel/.gitignore
Omrigan: One more thing: consider adding makefile commands for 
1. Cloning the right kernel version for local development 
2. Applying patches on top of the local tree

This might simplify dev workflow when we need to change/review those patches. WDYT?

Omrigan (commented):
Couple high level comments, didn't look at the patch itself yet.

neonvm-kernel/patches/00-phys_pgtable_nullptr.patch
```diff
@@ -0,0 +1,148 @@
...
++			return 0; // allocation failed
+ 		paddr_last = phys_p4d_init(p4d, __pa(vaddr), __pa(vaddr_end),
+ 					   page_size_mask, prot, init);
+ 
+@@ -771,7 +792,12 @@
+ 					  (pud_t *) p4d, init);
+ 
+ 		spin_unlock(&init_mm.page_table_lock);
++
++		if (!paddr_last)
```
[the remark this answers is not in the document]
sharnoff: (resolving in favor of the thread above)

neonvm-kernel/echo-version.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+#
+# Helper script to print the current kernel version in use.
+#
+# Example usage:
+#
+#   echo "Using version $(./get-version.sh)"
```
petuhovskiy, line 7: Looks like the file was renamed before a commit?

```suggestion
#   echo "Using version $(./echo-version.sh)"
```

sharnoff: Oh, yup - thanks!

neonvm-kernel/patches/00-phys_pgtable_nullptr.patch
sharnoff: Switched to using `git format-patch` for the patches after discussing offline - should be good here I think

neonvm-kernel/patches/0001-x86-mm-Update-mapped-addresses-in-phys_-pmd-pud-_ini.patch
```diff
@@ -0,0 +1,379 @@
+From 11e5209ef9e94792580d376b6cf1e4305e2dae4c Mon Sep 17 00:00:00 2001
+From: Em Sharnoff <sharnoff@neon.tech>
+Date: Thu, 12 Jun 2025 10:36:21 +0100
+Subject: [PATCH 1/5] x86/mm: Update mapped addresses in phys_{pmd,pud}_init()
```
Omrigan, line 4: Where is the 5th patch?

Makefile
```diff
@@ -305,6 +304,37 @@ kernel: ## Build linux kernel.
 	docker cp $$id:/vmlinuz neonvm-kernel/vmlinuz; \
 	docker rm -f $$id
 
+.PHONY: kernel-source
+kernel-source: ## Set up git repo for the current kernel version and apply existing patches
+	kernel_version="$$(neonvm-kernel/echo-version.sh)"; \
```
Omrigan, line 309: Consider extracting those make targets into helper scripts as well, would help with `; \`

Omrigan (approved):

Makefile
sharnoff: Discussed separately, planning to leave this for now - may revisit later & change to something easier to maintain.

PR #1393, opened 2025-06-04
go: add syntax for error handling to Neon Go

olegbbtr:
After a recent announcement [1] by Go team, saying they won't proceed
with adding error handling syntax to Go, it is evident we have to
implement a solution downstream.

We already build Linux kernel with custom patches (see #1386, #1392), it
is logical that as a next step we should have our own fork of Go, called
Neon Go.

The first feature is a syntax to handle errors. It supports the
following syntax:

  ... := ... handle err {
    ...
  }

Which is equivalent to:

  ..., err := ...
  if err != nil {
    ...
  }

Advantages:
1. Easy to read.
2. Easy to write (no need to look for the ! key).

Timeline to implement:
5 June - merge to main
6 June - deploy to staging
7 June - deploy to production

[1]: https://go.dev/blog/error-syntax

Also, disable parts of the CI, as it doesn't have good support for Neon Go (yet).

NOTE: this is, of course, a joke. I don't have any strong opinions about
error handling in Go.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,16 +56,8 @@ jobs:
       archs: '["amd64","arm64"]'
     secrets: inherit
 
-  e2e:
-    needs: [ get-tag, build-images, build-test-vm ]
-    uses: ./.github/workflows/e2e-test.yaml
-    with:
-      tag: ${{ needs.get-tag.outputs.tag }}
-      push-yamls: true
-    secrets: inherit
-
   release:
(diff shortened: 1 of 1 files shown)


mikhail-sakhnov:
>NOTE: this is, of course, a joke. I don't have any strong opinions about
error handling in Go.

Should be first line of the PR 😆

mikhail-sakhnov (approved):
I didn't read, but looks good to me 🚀

PR #1395, opened 2025-06-05
billing: Switch billing to ndjson format

dsavelev:
Switch billing to ndjson format without the container (one billing event per line)

Relates: https://github.com/neondatabase/cloud/issues/29993


pkg/agent/billing/clients.go
```diff
@@ -84,7 +85,7 @@ func createClients(ctx context.Context, logger *zap.Logger, cfg ClientsConfig) (
 			Name:            "s3",
 			Base:            client,
 			BaseConfig:      c.BaseClientConfig,
-			NewBatchBuilder: jsonArrayBatch(reporting.NewGZIPBuffer),
+			NewBatchBuilder: jsonLinesBatch(reporting.NewGZIPBuffer),
```
mikhail-sakhnov, line 88: I think, we already once had incident when the format was not exactly like we expected it to be.

May i ask you to add a simple unit/integration test, to double validate that produced data format matches the expectations?

dsavelev: I've fixed e2e tests and added unit tests.

pkg/agent/billing/clients.go
```diff
@@ -69,7 +70,7 @@ func createClients(ctx context.Context, logger *zap.Logger, cfg ClientsConfig) (
 			Name:            "azureblob",
 			Base:            client,
 			BaseConfig:      c.BaseClientConfig,
-			NewBatchBuilder: jsonArrayBatch(reporting.NewGZIPBuffer),
+			NewBatchBuilder: jsonLinesBatch(reporting.NewGZIPBuffer),
```
Omrigan, line 73: Is there value in preserving old behaviour through adding a configuration option to select the format? 

This would be helpful for the rollout.

tests/e2e/s3/utils.py
```diff
@@ -201,15 +200,6 @@ def s3_check_file(local_endpoint: str):
     do_assert(now - stop_time < datetime.timedelta(seconds=10))
 
 
-def validate_metrics_schema(metrics: dict):
-    billing_schema = yaml.safe_load(
-        open("billing-v1.yaml", "r").read()
```
Omrigan, line 206: billing-v1.yaml is no longer used?

pkg/agent/billing/billing.go
```diff
@@ -264,11 +264,11 @@ func (h *vmMetricsHistory) finalizeCurrentTimeSlice() {
 // tryMerge attempts to merge s and next (assuming that next is after s), returning true only if
 // that merging was successful.
 //
-// Merging may fail if s.endTime != next.startTime or s.metrics != next.metrics.
-func (s *metricsTimeSlice) tryMerge(next metricsTimeSlice) bool {
-	merged := s.endTime == next.startTime && s.metrics == next.metrics
+// Merging may fail if m.endTime != next.startTime or m.metrics != next.metrics.
+func (m *metricsTimeSlice) tryMerge(next metricsTimeSlice) bool {
```
Omrigan, line 268: Ideally, should have those non-functional changes split-off to a separate commit and merge-with-rebase.

pkg/agent/config.go
```diff
@@ -201,7 +201,6 @@ func (c *Config) validate() error {
 	if c.Billing.Clients.S3 != nil {
 		validateBaseReportingConfig(&c.Billing.Clients.S3.BaseClientConfig, "billing.clients.s3")
 		validateS3ReportingConfig(&c.Billing.Clients.S3.S3ClientConfig, ".billing.clients.s3")
-		erc.Whenf(ec, c.Billing.Clients.S3.PrefixInBucket == "", emptyTmpl, ".billing.clients.s3.prefixInBucket")
```
Omrigan, line 204: This should also ideally be a separate commit.

pkg/agent/billing/clients.go
```diff
@@ -108,9 +115,13 @@ func newBlobStorageKeyGenerator(prefix string) func() string {
...
 
-		return fmt.Sprintf("%s/year=%d/month=%02d/day=%02d/%s_%s.ndjson.gz",
+		if prefix != "" {
+			prefix = strings.TrimRight(prefix, "/") + "/"
+		}
+
+		return fmt.Sprintf("%syear=%d/month=%02d/day=%02d/hour=%02d/%s_%s.ndjson.gz",
 			prefix,
-			now.Year(), now.Month(), now.Day(),
+			now.Year(), now.Month(), now.Day(), now.Hour(),
```
Omrigan, line 124: Same question about preserving backward compatibility.

Omrigan (commented):
Broadly looks good, couple of questions.

pkg/agent/billing/clients.go
dsavelev: We don't actually use current data from S3 anywhere. It exists only as a backup. There is no need for a configuration option, imo. No one will ever switch it back after this update.

pkg/agent/billing/clients.go
dsavelev: We don't really need it. Nothing depends on it.

pkg/agent/billing/billing.go
dsavelev: I was trying to make the code a bit more consistent. Ok, let's revert

Omrigan (approved):

PR #1399, opened 2025-06-18
neonvm: do not use targetArchitecture in affinity, if not set

mikhail-sakhnov:
Do not add affinity if targetArchitecture is nil.

Should imrove quality of live with local development not on x86 and should simplify import jobs spec.

Issue https://github.com/neondatabase/cloud/issues/30328


Omrigan (approved):

sharnoff (approved):
lgtm!

PR #1403, opened 2025-06-24
billing: Apply changes to S3 bucket key to the proper events

dsavelev:
By mistake, changes to the S3 bucket's key were applied to scaling events. This PR reverses the previous changes and applies them to the billing events.

This reverts commit 8f29f3a. (https://github.com/neondatabase/autoscaling/pull/1402)

### DIFF
--- a/pkg/agent/billing/clients.go
+++ b/pkg/agent/billing/clients.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


dsavelev:
> PR title is super long, but not sure if there's a more succinct way to capture "undo X and try it again". Would you be up for separating the revert into its own PR?

I changed the PR title to make it shorter. Hope it's better.

PR #1406, opened 2025-07-02
vm-builder: Suport docker contexts

sharnoff:
Colima on MacOS uses a docker context to redirect to the socket at `~/.colima/default/docker.sock`.

If you look in `~/.docker/contexts/meta/.../meta.json`, the content of the context looks something like:

```json
{"Name":"colima","Metadata":{"Description":"colima"},"Endpoints":{"docker":{"Host":"unix:///<HOME>/.colima/default/docker.sock","SkipTLSVerify":false}}}
```

So, in order to allow vm-builder to work with colima, use docker contexts.

### DIFF
--- a/vm-builder/main.go
+++ b/vm-builder/main.go
@@ -18,6 +18,8 @@ import (
 	"github.com/alessio/shellescape"
 	"github.com/distribution/reference"
 	cliconfig "github.com/docker/cli/cli/config"
+	dockercontext "github.com/docker/cli/cli/context"
+	contextstore "github.com/docker/cli/cli/context/store"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/registry"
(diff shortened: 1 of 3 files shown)


vm-builder/main.go
```diff
@@ -200,10 +202,39 @@ func main() {
 		authConfigs[key] = registry.AuthConfig(value)
 	}
 
+	dockerClientOpts := []client.Opt{
+		client.FromEnv,
+		client.WithAPIVersionNegotiation(),
+	}
+
+	if dockerConfig.CurrentContext != "" {
```
mikhail-sakhnov, line 210: does it always use the current context? eg. the same as in the `docker context show`?

mikhail-sakhnov: For example, I have multiple docker contexts:
```
do%                                                                                                                                                                                                                             ➜  ~ docker context ls
NAME            DESCRIPTION                               DOCKER ENDPOINT                                           ERROR
colima *        colima                                    unix:///Users/misha.sakhnov/.colima/default/docker.sock
default         Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux   Docker Desktop                            unix:///Users/misha.sakhnov/.docker/run/docker.sock
```

The selected one is in use if I don't specify --context argument.
```
➜  ~ docker --context default ps
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
➜  ~ docker --context colima ps
CONTAINER ID   IMAGE                            COMMAND                  CREATED          STATUS          PORTS                                         NAMES
81183bcb26d9   yuzutech/kroki                   "/bin/sh -c 'exec ja…"   22 seconds ago   Up 21 seconds   0.0.0.0:8000->8000/tcp, [::]:8000->8000/tcp   lakebase-docsold-kroki-1

# the same!
➜  ~ docker ps
CONTAINER ID   IMAGE                            COMMAND                  CREATED          STATUS          PORTS                                         NAMES
81183bcb26d9   yuzutech/kroki                   "/bin/sh -c 'exec ja…"   52 seconds ago   Up 51 seconds   0.0.0.0:8000->8000/tcp, [::]:8000->8000/tcp   lakebase-docsold-kroki-1
```


How does the value of dockerConfig.CurrentContext is filled? is it the same way as in using docker cli command?

petuhovskiy (approved):

PR #1407, opened 2025-07-02
Support multi-arch image SHAs

sharnoff:
When building both x86 and ARM images on the same machine, docker can get confused if you end up building the same image twice, giving an error like:

```
cannot overwrite digest sha256:e5d0aea7f7d2954678a9a6269ca2d06e06591881161961ea59e974dff3f12377
```

This happens because pulling a multi-platform image using the SHA of the *manifest* will resolve to different underlying images.

To fix this, we need to use the platform-specific digests, which we can get from the 'docker manifest inspect' output of the multi-platform digest.

---

To test on apple silicon with rosetta enabled, you can run:

```bash
make HOST_ARCH=arm64 kernel docker-build
make HOST_ARCH=amd64 kernel docker-build
```

(i.e. with both architectures)


Makefile
```diff
@@ -21,15 +21,84 @@ GOOS ?= $(shell go env GOOS)
...
+# For each of these images, you can fetch the platform-specific SHAs with:
+#
+#   docker manifest inspect <image>:<tag>@sha256:<SHA> \
+#     | jq '.manifests[] | select(.platform.os == "linux") | select(.platform.architecture == "amd64" or .platform.architecture == "arm64")'
+#
+# If you get "manifest verification failed", it might be that the tag no longer refers to that SHA.
+# You can try again with 'docker manifest inspect <image>@sha256:<SHA>' instead (dropping the ':<tag>')
+
+# alpine images for @sha256:e5d0aea7f7d2954678a9a6269ca2d06e06591881161961ea59e974dff3f12377
+ALPINE_IMG_TAG ?= 3.19.7
```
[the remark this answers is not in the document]
sharnoff: Yeah, I think that makes a lot of sense :)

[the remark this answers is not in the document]
sharnoff: added a `versions.env` in ef55c20df1a1e8298b2fccca92c1dbc5f955aeaf so that it can be used from github workflows as well

.github/workflows/vm-kernel.yaml
```diff
@@ -258,11 +258,33 @@ jobs:
...
+              'amd64' | 'arm64')
+                  ;;
+              *)
+                  echo "Unexpected architecture $MATRIX_ARCH"
+                  exit 1
+                  ;;
+          esac
+
+          arch="${MATRIX_ARCH@U}" # upper-case
+          grep -v '^#' versions.env | grep -E "^[A-Z_]+(_TAG|_SHA_$arch)=" | sed -e "s/_$arch=/=/g" >> $GITHUB_ENV
```
mikhail-sakhnov, line 275: We have that command repeated couple of times and the pattern is pretty complicated. 

Wdyt about extracting it to a helper shell script, to use it like `source $(export_versions.sh)` ?

sharnoff: Yeah, makes sense. During development I had some issues because this entire step is repeated for a few workflows -- I'll see if I can extract it into a common step shared between workflows.

Omrigan (approved):

.github/workflows/vm-kernel.yaml
sharnoff: Implemented in 964f72157459aec7a8d6c3aa6985b177425ed21c, 33d5de187efc590d23a1a528ff5f2102b69e77f6, 4397b075df566f2aca19cee96bbe2ce7d8d48894

PR #1409, opened 2025-07-09
fluent-bit: use better logs format

olegbbtr:
1. Prepend file name with pod creation timestamp.
2. Bring timestamp and log line to the start of the line.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -224,8 +224,9 @@ jobs:
         uses: docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4 # v6.15.0
         id: build-go-dependencies-image
         with:
-          provenance: false
-          sbom: false
+          attests: |
+            type=provenance,mode=max
+            type=sbom,generator=docker.io/docker/buildkit-syft-scanner:1
           context: .
           push: true
           file: go-base.Dockerfile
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -536,15 +536,28 @@ k3d-destroy-cluster:
...
+
+.PHONY: apply-fluent-bit
+apply-fluent-bit: kubectl
+	$(KUBECTL) apply -f tests/fluent-bit/
+
+.PHONY: restart-fluent-bit
+restart-fluent-bit: kubectl
+	$(KUBECTL) -n kube-system rollout restart daemonset/fluent-bit
+
+.PHONY: status-fluent-bit
```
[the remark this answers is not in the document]
Omrigan: Meh

Makefile
```diff
@@ -536,15 +536,28 @@ k3d-destroy-cluster:
...
-deploy-fluent-bit: kubectl
-	$(KUBECTL) apply -f tests/fluent-bit/
-	$(KUBECTL) -n kube-system rollout status daemonset fluent-bit
-
 .PHONY: logs-dir-cleanup
 logs-dir-cleanup:
 	@rm -rf tests/logs/*
 
+.PHONY: deploy-fluent-bit
+deploy-fluent-bit: kubectl apply-fluent-bit status-fluent-bit
```
sharnoff, line 544: As a follow-up: Should we just make this part of `make {k3d,kind}-setup`? (wondering what utility there is in keeping makefile targets for fluent-bit)

sharnoff (approved):
lgtm, nice!

Makefile
Omrigan: It is already part of `deploy`. 

Separate targets are mostly for testing, and in general, "more targets" is better than "less targets" - same logic as for functions in normal code, more flexibility when working with repo locally. WDYT?

PR #1410, opened 2025-07-09
Makefile: fix bugs and add run-e2e target

Omrigan:


Makefile
```diff
@@ -477,14 +477,17 @@ pg16-disk-test: docker-build-pg16-disk-test load-pg16-disk-test
 
 ##@ End-to-End tests
 
-.PHONE: e2e-tools
```
mikhail-sakhnov, line 480: 📞

mikhail-sakhnov (approved):

Makefile
Omrigan: Yeah, we either need a linter or that thing @sharnoff suggests (what is it's name?), I imagine it doesn't have such problems.

PR #1417, opened 2025-07-21
agent,plugin: Cancel orca ctx before wait

sharnoff:
In [LKB-2414](https://databricks.atlassian.net/browse/LKB-2414), we found that the autoscaler-agent can deadlock after a failure during startup. This is because the deferred call to `srv.GetOrchestrator().Wait()` is *before* the canceling of the context, which means that it just blocks forever.

To fix the immediate issue, move the `cancel()` into the same deferred function.

---

I reproduced the issue locally - and confirmed that this fixes it - by modifying the autoscaler-agent's main function to always panic.

### DIFF
--- a/autoscale-scheduler/cmd/main.go
+++ b/autoscale-scheduler/cmd/main.go
@@ -53,12 +53,14 @@ func runProgram(logger *zap.Logger) (err error) {
 	// services and be able to more coherently wait for shutdown
 	// without needing a sleep.
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM)
-	defer cancel()
 	ctx = srv.SetShutdownSignal(ctx)
 	ctx = srv.WithOrchestrator(ctx)
 	ctx = srv.SetBaseContext(ctx)
 	orca := srv.GetOrchestrator(ctx)
-	defer func() { err = orca.Service().Wait() }()
+	defer func() {
+		cancel()
+		err = orca.Service().Wait()
+	}()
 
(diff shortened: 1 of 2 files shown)


petuhovskiy (approved):

autoscale-scheduler/cmd/main.go
```diff
@@ -53,12 +53,14 @@ func runProgram(logger *zap.Logger) (err error) {
...
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM)
-	defer cancel()
 	ctx = srv.SetShutdownSignal(ctx)
 	ctx = srv.WithOrchestrator(ctx)
 	ctx = srv.SetBaseContext(ctx)
 	orca := srv.GetOrchestrator(ctx)
-	defer func() { err = orca.Service().Wait() }()
+	defer func() {
+		cancel()
+		err = orca.Service().Wait()
```
mikhail-sakhnov, line 62: I am struggling to understand how is it supposed to work.

Looking into the implementation of `orca.Service().Wait()` I see it never uses the context passed via `srv.GetOrchestrator(ctx)`, do I read it wrong? In Wait (and in other places of the orca lib) it uses internal.BackgroundContext which is defined as context.Background and from the top of my head I don't see internal.BackgroundContext be overriden?

mikhail-sakhnov: I don't see nothing like that in the code, I think. 

It does the opposite: creates a context with orchestrator in the context available by key `orchestratorCtxKey{}` (basically, orca variable is the object).

mikhail-sakhnov (approved):

PR #1418, opened 2025-07-21
vm-builder: allow to set `alpine` and `busybox` from custom sources

bayandin:
In some cases, we need pull base images for `vm-builder` from different registries.
This PR adds `alpine-image` and `busybox-image` flags to make it possible to specify that.

Note: These new flags are a bit inconsistent with `daemon-image`, which allows a full image string (including tag/SHA). The new ones only take registry + image. We can unify this later


sharnoff (approved):
lgtm! The inconsistency isn't ideal, but I think it's ok enough for now

PR #1421, opened 2025-07-30
ci: include sbom and provenance attestations in container images

jcgruenhage:
As part of LKB-2641, we're adding sbom and provenance attestations to all container images, to make them more auditable.


fedordikarev (approved):
LGTM, but please add description for the PR

PR #1427, opened 2025-08-28
neonvm-controller: fix a bug of migration state machine

NaroZeol:
Hello Neon team,

I really like the idea behind the autoscaling implementation, and I’ve been experimenting with it myself.

However, I noticed that the state machine management in neonvm-controller might have an issue. **In the `PreMigrating` state, the phase is always transitioned directly into `Migrating`, regardless of the actual result of the QMPrequest.**

Most of the time this isn’t a problem, but in some corner cases where the target QEMU startup is slower than expected and the QMP port becomes available later, this can cause a request failure.

When that happens, neonvm-controller doesn’t perform any retries—it simply moves into the `Migrating` phase, which causes the migration flow to get stuck. After applying the change in this PR, that issue is resolved.

I’m happy to have a chance to contribute to the autoscaling project. I’d really appreciate any feedback if there are issues with my changes.

: )

### DIFF
--- a/pkg/neonvm/controllers/vmmigration_controller.go
+++ b/pkg/neonvm/controllers/vmmigration_controller.go
@@ -327,15 +327,15 @@ func (r *VirtualMachineMigrationReconciler) Reconcile(ctx context.Context, req c
 			}
 			// Migrate only running VMs to target with plugged devices
 			if vm.Status.Phase == vmv1.VmPreMigrating {
+				// trigger migration
+				if err := QmpStartMigration(vm, migration); err != nil {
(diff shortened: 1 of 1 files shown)


NaroZeol:
I’ve been looking into a few serverless solutions, and Neon’s autoscaling is my favorite one since it feels truly serverless. While digging into the code implementation, I noticed this obvious small bug.

By the way, will Neon turn to close source after being acquired? The update frequency in recent months is much slower than ever before.

olegbbtr:
> By the way, will Neon turn to close source after being acquired?

We are continuing to accept contributions in this and other repositories.

Please send more!

